PR #2, opened 2019-09-26
chore,fix(various): Get Vega Strike to build and run on Ubuntu 18.04 LTS

stephengtuggy:


src/physics.h
```diff
@@ -27,7 +27,7 @@
 static const float oocc = (float) 0.0000000000000000111265005605; //1/c^2
 static const float c    = (float) 299792458.0;
 static const float co10 = (float) 29979245.8;
-float copysign( float x, float y );
+// float copysign( float x, float y );
```
klaussfreire: I'm guessing you did this because it conflicts with the stdlib.

In those cases commenting the code isn't a good solution. A simple `#ifndef` may work, and/or a configure test to check whether it's available would be the proper approach.

Or just deleting it. AFAIK this was introduced in C99, which should be widely supported by now, so we may no need to keep compatibility with older compilers that lack C99 support. In any case you shouldn't leave commented code everywher (I know we did this already, but it's wrong).

I'm not sure about windows (MSVC) support, but I'm guessing it should be there.

src/ship_commands.cpp
```diff
@@ -8,14 +8,14 @@
...
-}
-static inline float fmax( float a, float b )
-{
-    return (a > b) ? a : b;
-}
+// static inline float fmin( float a, float b )
+// {
+//     return (a < b) ? a : b;
+// }
+// static inline float fmax( float a, float b )
```
klaussfreire: Similarly here. Either conditionally based on a configure check, or remove if widely supported already.

Some consideration must be given to windows, though. MSVC lacked fmin/fmax a while ago, I'm not sure whether it includes it now in its math templates.

src/vsfilesystem.cpp
```diff
@@ -602,23 +602,31 @@ void InitDataDirectory()
...
+        // data_paths.push_back( vegastrike_cwd+"/../../data4.x" );
+        // data_paths.push_back( vegastrike_cwd+"/data4.x" );
         data_paths.push_back( vegastrike_cwd+"/../../data" );
-        data_paths.push_back( vegastrike_cwd+"/data4.x" );
         data_paths.push_back( vegastrike_cwd+"/data" );
         data_paths.push_back( vegastrike_cwd+"/../data" );
         data_paths.push_back( vegastrike_cwd+"/../Resources" );
+        data_paths.push_back( vegastrike_cwd+"/Assets-Production");
+        data_paths.push_back( vegastrike_cwd+"/../Assets-Production");
+        data_paths.push_back( vegastrike_cwd+"/../../Assets-Production");
```
klaussfreire: Why did you add these?

src/vsfilesystem.cpp
```diff
@@ -602,23 +602,31 @@ void InitDataDirectory()
...
     }
     data_paths.push_back( "." );
     data_paths.push_back( ".." );
-    //data_paths.push_back( "../data4.x" ); DELETE 4.x no longer used since .5 release obsolete, removal required.
+    // DELETE 4.x no longer used since .5 release obsolete, removal required.
+    //data_paths.push_back( "../data4.x" );
```
klaussfreire: Just remove all 4.x data entries and the comment

boost/1_53/CMakeLists.txt
```diff
@@ -1,34 +0,0 @@
-add_definitions(-DBOOST_PYTHON_NO_PY_SIGNATURES -DBOOST_PYTHON_STATIC_LIB)
-include_directories("${vsUTCS_SOURCE_DIR}/boost/1_53")
-add_library(boost_python
```
klaussfreire: This is from Boost, you shouldn't be changing it.

Maybe you have to update boost. You can build using system boost libraries, instead of the included ones. I would recommend you try.

setup/src/include/central.cpp
```diff
@@ -20,6 +20,8 @@ struct catagory CATS;
 struct group GROUPS;
 struct global_settings CONFIG;
 
+static char NULL_CHAR = '\0';
```
klaussfreire: This is a weird way of passing an empty string.

Why not

``` C++
static char EMPTY_STR[] = "";
```

src/physics.h
BenjamenMeyer: For MSVC compat, I would recommend making a header file that can be used to explicitly provide compatibility; sometimes functions just need to be renamed; sometimes they need wrappers to ensure the same interface. Fortunately, MS has gotten better about compatibility with the standards - at least providing the functions; but they've gotten worse about implementation details (f.e snprintf vs _snprintf - there's differences in behavior that need to be accounted for, and MS favors s_snprintf which is also different); not sure how much you'll run into that with copysign, hopefully not at all.

$0.02 FWIW.

stephengtuggy: Yes, I did this because it was already declared, evidently in stdlib. If this was introduced in C99, I think it would be fine to take out these declarations altogether. (That standard is now 20 years old. ;-) )

src/vsfilesystem.cpp
stephengtuggy: If I understand correctly, [vegastrike/Assets-Production](https://github.com/vegastrike/Assets-Production) is the GitHub equivalent to the `data`/`hqtextures` folder on SourceForge. When I clone it to my local machine, it ends up in a local folder that is also called `Assets-Production` by default.

I figured others would encounter the same behavior. Not only that, but when I tried configuring the data directory using `ccmake`, it didn't like it **at all**. I tried several different ways of specifying that path, but none of them worked.

src/vsfilesystem.cpp
stephengtuggy: OK, good. I was erring on the side of caution. I'm glad I have permission to remove these entirely.

src/ship_commands.cpp
stephengtuggy: I would expect that these are widely supported by this point. I think I will go ahead and remove them.

boost/1_53/CMakeLists.txt
stephengtuggy: This file is one of only two files checked into GitHub under the boost directory. It didn't come with the boost source code when I downloaded that.

With this file in place, the entire project failed to build. Without this file, the project built and ran fine. So I decided just to remove this file.

If there is an alternative, I would be happy to hear about it.

setup/src/include/central.cpp
stephengtuggy: Good point. I'll fix that.

boost/1_53/CMakeLists.txt
stephengtuggy: Oh, I forgot to mention: I did try building using the system boost libraries also. That didn't work either.

src/vsfilesystem.cpp
stephengtuggy: Done.

src/physics.h
stephengtuggy: Removed altogether.

src/ship_commands.cpp
stephengtuggy: Removed.

Loki1950:
Hi guys will be checking in everyday as I'm retired now and have lost's of time

Loki1950:
Would love to Stephen but my C and C++ is nonexistent :( and I just get by in Python after all I started coding before FORTRAN

stephengtuggy:
> Would love to Stephen but my C and C++ is nonexistent :( and I just get by in Python after all I started coding before FORTRAN

No worries!

src/physics.h
BenjamenMeyer: MSVC is kind of its own standard... depending on what version you're using. MS has gotten better about adhering to ISO standard C/C++ (starting with VS2008 which finally introduced `stdint.h` and slowly getting better each iteration since) but it can still be hit-or-miss on some things last I knew...but I haven't used VS in quite a while now so may be it's no longer an issue...

boost/1_53/CMakeLists.txt
stephengtuggy: UPDATE: I have since figured out how to build using the system boost libraries.

src/physics.h
stephengtuggy: > The Universal C Runtime library (UCRT) provides many integral and floating-point math library functions, including all of those required by ISO C99.

[https://docs.microsoft.com/en-us/cpp/c-runtime-library/floating-point-support?view=vs-2019](https://docs.microsoft.com/en-us/cpp/c-runtime-library/floating-point-support?view=vs-2019)

boost/1_53/CMakeLists.txt
klaussfreire: These boost directories should be git submodules. I think this is a migration error carried over since the migration from SVN. I'll look into that.

In the meanwhile, I think it's ok to delete these.

src/cmd/unit_generic.cpp
```diff
@@ -655,14 +655,6 @@ void Unit::DeactivateJumpDrive()
         jump.drive = -1;
 }
 
-float copysign( float x, float y )
```
klaussfreire: Check out where these come from (the header file), you'll have to include the proper header. I think it's `<math.h>` in this case. The include would go in the cpp.

src/physics.h
stephengtuggy: So it sounds like Visual Studio 2019, at least, supports `copysign` as required by the standard.

src/ship_commands.cpp
```diff
@@ -8,15 +8,6 @@
 
 
 
-static inline float fmin( float a, float b )
```
klaussfreire: Same here, you have to include them from <math.h>

src/physics.h
BenjamenMeyer: Unfortunately they dropped docs for VS2013 and older and stop listing what version introduced something, so it's hard to tell where support began. My guess is they're only supporting VS2015 and newer (right now); that said, many folks may still have access to older compilers...one can only hope folks use a newer compiler. Certainly would be good to say "using this version and newer" if it doesn't already.

src/cmd/unit_generic.cpp
stephengtuggy: I believe `math.h` is already included. Line 7 of `unit_generic.cpp` includes `"unit_generic.h"`, which includes `"vegastrike.h"` on line 40. In turn, `"vegastrike.h"`, line 38, includes `"vs_math.h"`. And that file includes `<math.h>`, on line 27 and again on line 45.

Do your findings match what I'm seeing?

src/ship_commands.cpp
stephengtuggy: `ship_commands.cpp` includes `"cmd/unit_generic.h"`. So it will automatically have `<math.h>` available, via much the same dependency chain as `unit_generic.cpp`.

stephengtuggy: I have built and run this entire project on Linux. Everything works, more or less. I want to build Vega Strike on Windows as well, but I'm running into some difficulties on that front, configuring CMake.

src/cmd/unit_generic.cpp
stephengtuggy: How's this?

src/ship_commands.cpp
stephengtuggy: `<math.h` included explicitly, as requested.

klaussfreire: Thanks

klaussfreire:
If there's no objection I'll merge

Loki1950:
Pouring myself a dollop of single malt congrats on our first merge guys
Enjoy the Choice

PR #7, opened 2019-10-03
Fix/debug output

### LINKED ISSUE #3: Some debug output statements missing EOL
I have noticed that some of the debugging statements in VS that output to the console, are missing the terminating newline. So multiple output lines get jammed together on one line at times.

One of the most common culprits is this message:
```
cannot execute suborders as Linear Velocity Matcher
```

### LINKED ISSUE #4: Some debug output statements seem to be crashing vegastrike
I haven't been able to pin down exactly what is causing these crashes. But I suspect buggy debugging output statements. (Ironic, right?) I think it's a good idea to make each output statement as safe as reasonably possible. This will probably help reduce the frequency of crashes.

To achieve this, I propose using C++-style stream I/O, rather than C-style `printf` type statements. This will automatically improve type safety, memory safety, etc.

Calls to `VSFileSystem::vs_fprintf()` can be refactored relatively easily. Calls to `VSFileSystem::vs_dprintf()` will be a little trickier, because of the different debug verbosity levels, but I have an idea for this too.

I am in the process of replacing `vs_dprintf()` with a function called `vs_dbg()`:

```C++
static boost::iostreams::stream< boost::iostreams::null_sink > nullOstream( ( boost::iostreams::null_sink() ) );

std::ostream& vs_dbg( char level )
{
    if (use_volumes || (level > g_game.vsdebug)) {
        return nullOstream;
    } else {
        return std::cerr;
    }
}
```

You can write to `vs_dbg` using the `<<` operator, just like `cout` or `cerr`. If the current user settings mean that a given output line should be suppressed, `vs_dbg()` will return a boost `null_sink` stream, which just throws away any data sent to it. (See [https://stackoverflow.com/a/8244321/5067822](https://stackoverflow.com/a/8244321/5067822).)

Example usage:

```C++
VSFileSystem::vs_dbg(1) << "Not deleting mission twice: " << this->mission_name << std::endl;
```

instead of

```C++
VSFileSystem::vs_dprintf( 1, "Not deleting mission twice: %s\n", this->mission_name.c_str() );
```

Already, my changes seem to be improving Vega Strike's stability somewhat.

Let me know what you guys think.

stephengtuggy:
Thanks for merging the first pull request, guys. Here comes the second! :-)

I went through all the debugging output statements in the code that used `vs_dprintf()`, and converted them to use a new method `vs_dbg()` instead, plus `boost::format()` where applicable.

There were lots of buggy debug statements. (Ironic, right?) Among other things:

- Many debug stmts lacked a newline at the end
- Many such stmts assumed that `size_t` equated to `int` (`%d`), which it doesn't always
- Some debug output stmts used the wrong type or precision of format specifier
- In at least one case, the format string had a `%u` format specifier, but no arguments!
- `va_end` didn't seem to be called anywhere, which probably would have led to significant memory leaks

Vega Strike appears to be running much more smoothly with these changes, crashing a lot less often.

Sorry about the size of this pull request. Let me know if you have any questions. You may want to ignore whitespace changes when viewing the diff on GitHub. My editor cleaned up some trailing whitespace in the files I edited.

Closes #3 
Closes #4


CMakeLists.txt
```diff
@@ -553,7 +553,7 @@ SET(DATADIR "" CACHE STRING "Default vegastrike data directory")
 
 # Hide our temporary variable and initialize it to empty
 SET(CPU_OPTS  )
-SET(DEFINES " -DNV_CUBE_MAP -DBOOST_PYTHON_NO_PY_SIGNATURES ")
+SET(DEFINES " -D NV_CUBE_MAP=1 -D BOOST_PYTHON_NO_PY_SIGNATURES=1 ")
```
BenjamenMeyer: It would be good if the values specified mapped to some configuration variables elsewhere to make them easy to change and document. $0.02 FWIW

src/vsfilesystem.cpp
```diff
@@ -479,28 +481,6 @@ int vs_fprintf( FILE *fp, const char *format, ... )
...
 }
 
-void vs_dprintf( char level, const char *format, ... )
-{
-    if (!use_volumes && level <= g_game.vsdebug) {
-        va_list ap;
-        va_start( ap, format );
-        vfprintf( stderr, format, ap );
-    }
-}
```
stephengtuggy: Don't need this anymore!

src/vsfilesystem.cpp
```diff
@@ -479,28 +481,6 @@ int vs_fprintf( FILE *fp, const char *format, ... )
...
-{
-    if (!use_volumes) {
-        va_list arglist;
-        va_start( arglist, format );
-        //return _input(fp,(unsigned char*)format,arglist);
-        return vfscanf( fp, format, arglist );
-    } else {}
-    return 0;
-}
-#endif
```
stephengtuggy: This was effectively commented out anyway

src/vsfilesystem.h
```diff
@@ -146,15 +150,7 @@ bool vs_feof( FILE *fp );
...
 int vs_fprintf( FILE *fp, const char *format, ... );
-/* Debug level is a character, 1, 2, or 3.   1 is most important warnings only. 2 is warnings that are important but not hazardous to gameplay
- *  3 is for developer type messages, more akin to informational output but not relevant to anything an end user would care about.
- *  example usage.   vegastrike --debug=3       leaving --debug empty as an argument is the same as --debug=1
- *  all output is sent to stderr
- */
-void vs_dprintf( char level, const char *format, ... );
-#if 0
-int vs_fscanf( FILE *fp, const char *format, ... );
-#endif
```
stephengtuggy: This was effectively commented out, so I removed it entirely

src/vsfilesystem.h
```diff
@@ -146,15 +150,7 @@ bool vs_feof( FILE *fp );
 long vs_getsize( FILE *fp );
 
 int vs_fprintf( FILE *fp, const char *format, ... );
-/* Debug level is a character, 1, 2, or 3.   1 is most important warnings only. 2 is warnings that are important but not hazardous to gameplay
- *  3 is for developer type messages, more akin to informational output but not relevant to anything an end user would care about.
- *  example usage.   vegastrike --debug=3       leaving --debug empty as an argument is the same as --debug=1
- *  all output is sent to stderr
- */
-void vs_dprintf( char level, const char *format, ... );
```
stephengtuggy: Don't need this anymore, as far as I can tell

src/gfx/vid_file.cpp
```diff
@@ -104,10 +104,8 @@ class VidFileImpl
                         pCodecCtx, pNextFrameYUV, &frameFinished,
                         packetBuffer, packetBufferSize );
                     #endif
-                    VSFileSystem::vs_dprintf(3, "dts %lld: Decoded %d bytes %s\n", 
-                        int64_t(packet.dts),
-                        int(bytesDecoded),
-                        (frameFinished ? "Got frame" : "")
```
stephengtuggy: vid_file.cpp as a whole doesn't compile with current versions of FFMPEG et al, so I've had to disable FFMPEG. As a result, this file hasn't actually been compiled or tested. However, my IDE (Visual Studio Code) did have syntax highlighting enabled for this file, and I think it would have caught most common types of errors.

stephengtuggy:
@klaussfreire I think this PR is ready for you to review again, whenever you have the chance.

CMakeLists.txt
stephengtuggy: I am now working on CMake stuff in a separate branch, although I'm having a LOT of issues building the project, on both Linux and Windows.

BenjamenMeyer: @stephengtuggy cool - let me know if you need help. I haven't successfully built it yet myself, but am certainly willing to help out where I can.

pyramid3d:
Merged the commits. Thanks for the changes.

Issue #8, opened 2019-10-03
Use C++11 std::unique_ptr, std::shared_ptr, std::weak_ptr where applicable?


stephengtuggy:
Personally, I think the C++11 standard is old enough (8 years old!) that we can begin using it. And it comes with some great added features! Like the smart pointer types `std::unique_ptr`, `std::shared_ptr`, and `std::weak_ptr`: [https://en.cppreference.com/w/cpp/memory](https://en.cppreference.com/w/cpp/memory).

If we start using these smart pointer types wherever possible, a lot of Vega Strike's memory management issues could probably be solved.

What do you guys think?

klaussfreire:
I agree with the sentiment, but a lot of distros still come with gcc versions that don't enable c++11 support by default. No big deal, just a heads-up.

FTR, we're already using a few features from c++11 through tr1 and boost, so it's not like we're missing on them.

stephengtuggy:
Looks like boost includes a `boost::shared_ptr` type, which internally maps to c++11 `std::shared_ptr` when that is available, or to a different implementation otherwise. I would be happy to use `boost::shared_ptr` instead.

stephengtuggy:
Perhaps even more suited to our needs: [`boost::intrusive_ptr`](https://www.boost.org/doc/libs/1_71_0/libs/smart_ptr/doc/html/smart_ptr.html#intrusive_ptr). The name sounds a little weird, but the concept actually fits what Vega Strike already does: keep an internal ref count within each object instance (of `Unit`, for example).

Potential advantages of using `boost::intrusive_ptr` over what we already have:
- It's written and reviewed by domain experts, and thus very likely to be correct
- It's an established pattern as part of a well-known and respected library that we already use to some degree
- It is (or can be) thread-safe, which the current implementation... uh... isn't entirely ;-)

LifWirser:
In theroy I agree
now look at what else could be afected
- Vegastrike shares data referenced from pointers back and forth between c++, Python> GL(whichever), sound and various network functions.
 As I am new to contributing I want to know that all sides are looking at what they are supposed to
- from my perspective a through "Audit" of th VS codebase needs to be performed. Basically for this topic all pointers need to be idenified bot of origin(s) and users, along what tools are needed to test/verify proposed changes

Making this a Goal sounds like a result of properly auditing the codebase.

BenjamenMeyer:
See #15 for the audit proposal by @LifWirser

[...]

PR #12, opened 2019-10-17
Fix reloading of guns independently of mount volume.

crts-crome:
The proposed changes fix the reload mechanism for mounted guns. Currently most vessels cannot be reloaded. Some vessels that can reload their mounted gun slots do so in an inconsistent manner. I made a lengthy post in the vegastrike forum which contains screenshot and a detailed explanation why I think that the reload mechanism was intended to work this way.

### DIFF
--- a/src/cmd/unit_generic.cpp
+++ b/src/cmd/unit_generic.cpp
@@ -6452,7 +6452,7 @@ bool Unit::UpgradeMounts( const Unit *up,
                                         std::string weaponname = ammoname.substr(0, ammopos);
 
                                         /* Do NOT delete this Unit because it will be either fetched
-                                        * fetched from a cache or - if it has to be created - it will
+                                        * from a cache or - if it has to be created - it will
                                         * be automatically put in a cache.
                                         * Deletion will corrupt the cache!
                                         */
--- a/src/cmd/unit_generic.cpp
+++ b/src/cmd/unit_generic.cpp
@@ -6430,21 +6430,48 @@ bool Unit::UpgradeMounts( const Unit *up,
                                 int tmpammo = mounts[jmod].ammo;
                                 if (mounts[jmod].ammo != -1 && up->mounts[i].ammo != -1) {
                                     tmpammo += up->mounts[i].ammo;
-                                    if (templ) {
-                                        if (templ->GetNumMounts() > jmod) {
-                                            if (templ->mounts[jmod].volume != -1) {
-                                                if (templ->mounts[jmod].volume < mounts[jmod].type->volume*tmpammo) {
-                                                    tmpammo =
-                                                        (int) floor( .125
-                                                                    +( (0
-                                                                        +templ->mounts[jmod].volume)
-                                                                      /mounts[jmod].type->volume ) );
+                                    if (ismissiletype) {
+                                        if (templ) {
+                                            if (templ->GetNumMounts() > jmod) {
+                                                if (templ->mounts[jmod].volume != -1) {
+                                                    if (templ->mounts[jmod].volume < mounts[jmod].type->volume*tmpammo) {
+                                                        tmpammo =
+                                                            (int) floor( .125
+                                                                        +( (0
+                                                                            +templ->mounts[jmod].volume)
+                                                                          /mounts[jmod].type->volume ) );
+                                                    }
                                                 }
                                             }
                                         }
+                                        if (tmpammo*mounts[jmod].type->volume > mounts[jmod].volume)
+                                            tmpammo = (int) floor( .125+( (0+mounts[jmod].volume)/mounts[jmod].type->volume ) );
+                                    } else {
+                                        std::string ammoname = up->name.get();
+                                        std::size_t ammopos = ammoname.find("_ammo");
+                                        std::string weaponname = ammoname.substr(0, ammopos);
+
+                                        /* Do NOT delete this Unit because it will be either fetched
+                                        * fetched from a cache or - if it has to be created - it will
+                                        * be automatically put in a cache.
+                                        * Deletion will corrupt the cache!
+                                        */
+                                        const Unit * weapon = getUnitFromUpgradeName(weaponname);
+
+                                        if (weapon == NULL || weapon->name == LOAD_FAILED) {
+                                            // this should not happen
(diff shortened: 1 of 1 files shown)


crts-crome:
Here is the link to the forum thread with the corresponding screenshots: https://forums.vega-strike.org/viewtopic.php?f=2&t=49183

stephengtuggy:
Thanks for the PR! I will take a look when I get a chance.

crts-crome:
Please also have a look at the thread I linked. It has a more detailed explanation why I think the issue should be fixed in the proposed way.
To summarize:
- Better gameplay
- Consistent ammo amounts
- Better maintainability in the future

src/cmd/unit_generic.cpp
```diff
@@ -6430,21 +6430,48 @@ bool Unit::UpgradeMounts( const Unit *up,
...
                                         }
+                                        if (tmpammo*mounts[jmod].type->volume > mounts[jmod].volume)
+                                            tmpammo = (int) floor( .125+( (0+mounts[jmod].volume)/mounts[jmod].type->volume ) );
+                                    } else {
+                                        std::string ammoname = up->name.get();
+                                        std::size_t ammopos = ammoname.find("_ammo");
+                                        std::string weaponname = ammoname.substr(0, ammopos);
+
+                                        /* Do NOT delete this Unit because it will be either fetched
+                                        * fetched from a cache or - if it has to be created - it will
```
stephengtuggy: The word `fetched` is repeated. Omit the second occurrence.

src/cmd/unit_generic.cpp
```diff
@@ -6430,21 +6430,48 @@ bool Unit::UpgradeMounts( const Unit *up,
...
+                                        * be automatically put in a cache.
+                                        * Deletion will corrupt the cache!
+                                        */
```
stephengtuggy: Also: Will it introduce a memory leak if this unit never gets deleted? That wouldn't be good either.

I am working on a major code overhaul that would use smart pointers wherever possible, so that we wouldn't have to worry about when to manually delete objects. See issue #8 for further discussion. (The code itself isn't ready for review yet.)

stephengtuggy:
Mostly, this code change looks good, although I haven't tested the functionality yet.

klaussfreire:
I seem to recall I did a fix for a similar issue a while ago. Not sure whether they're related, but it certainly sounds related. I'll have to look into it a bit, because if it's the same issue, it means not all commits have been migrated from SVN.

src/cmd/unit_generic.cpp
crts-crome: Done in commit 38a3a0b.

stephengtuggy: :+1:

src/cmd/unit_generic.cpp
crts-crome: No, I do not think it will. Here is the code for getUnitFromUpgradeName() (file: src/cmd/unit_functions_generic.cpp):

```
const Unit * getUnitFromUpgradeName( const string &upgradeName, int myUnitFaction = 0 )
{
    const char *name     = upgradeName.c_str();
    const Unit *partUnit = UnitConstCache::getCachedConst( StringIntKey( name, FactionUtil::GetUpgradeFaction() ) );
    if (!partUnit) {
        partUnit = UnitConstCache::setCachedConst( StringIntKey( name,
                                                                FactionUtil::GetUpgradeFaction() ),
                                                  UnitFactory::createUnit( name, true, FactionUtil::GetUpgradeFaction() ) );
    }
    if (partUnit->name == "LOAD_FAILED") {
        partUnit = UnitConstCache::getCachedConst( StringIntKey( name, myUnitFaction ) );
        if (!partUnit)
            partUnit = UnitConstCache::setCachedConst( StringIntKey( name, myUnitFaction ),
                                                      UnitFactory::createUnit( name, true, myUnitFaction ) );
    }
    return partUnit;
}
```

As you can see, it always tries to fetch the requested unit from the cache before it creates it. I verified by inserting logging statements when the function gets called. It does not "forget" the previously created unit and does not start filling the memory with redundant copies. In fact, when I tested it, it never created a new unit, all units were already in the cache. I also checked other usages of the method in the code. They all just call "getUnitFromUpgradeName()". No further postprocessing is required.

crts-crome:
@klaussfreire Thanks for the feedback. I think someone in the forums mentioned that the GITHUB version is not as stable as the SVN version. I do not recall the thread, though. This would further indicate that some changes from SVN did not make it into GITHUB.
If you have the time, could you also check if you fixed something as described in issue #13 ?
This would be the next issue that I wanted to chase down, however, if it turns out that the port from SVN to GIT was in fact incomplete then I might be doing a Sisyphean task. I would very much like to avoid that :-)

src/cmd/unit_generic.cpp
stephengtuggy: Good to know. It sounds, then, like there is confusion between different parts of the code over what constitutes a "Unit": Is it a unit **type**, with associated Mesh and other attributes, as loaded from a data file? Or is it an **instance** of said unit type?

It sounds like your code change here is correct, though.

stephengtuggy:
> Mostly, this code change looks good, although I haven't tested the functionality yet.

I tried to test the functionality of this PR the other day. Unfortunately, I didn't get very far. The game crashed after only about 15-20 minutes.

To be fair, so does the main code branch, on my machine. Not sure what's going on. Anyone else having these issues?

pyramid3d:
> I seem to recall I did a fix for a similar issue a while ago. Not sure whether they're related, but it certainly sounds related. I'll have to look into it a bit, because if it's the same issue, it means not all commits have been migrated from SVN.
It seems the code was moved from svn to github in 2015, while I saw still commits to svn from you klauss on in 2017: https://sourceforge.net/p/vegastrike/code/13731/log/). Ideally those changes between 2015 and 2017 would be merged to github too.
See also my more generic posts on contributor friendly github handling:
https://forums.vega-strike.org/viewtopic.php?f=53&t=26104&view=unread#p169815
https://openteq.wordpress.com/2019/04/25/open-source-invites-abandonment-culture/

pyramid3d:
Merged since review has been concluded.

PR #29, opened 2020-04-01
Remove compilation warnings

nabaco:
I want to cleanup the compilation log from all the unnecessary warnings
Additionally, I started adapting for Ogre 1.9, to fix compilation with it.

### DIFF
--- a/engine/launcher/general.cpp
+++ b/engine/launcher/general.cpp
@@ -126,7 +126,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
 	current= new char [LENGTH+1];
 	length = strlen(line);
 	strcpy(current, line);
-	while ((location = strstr(current, search)) > 0) {
+	while ((location = strstr(current, search)) != NULL) {
 		chr_new[0] = '\0';
 		calc = strlen(current) - strlen(search) + strlen(replace);
 		if (calc > LENGTH) { strcpy(line, current); return line; }
--- a/engine/objconv/asteroidgen.cpp
+++ b/engine/objconv/asteroidgen.cpp
@@ -48,7 +48,7 @@ public:
     }
     void Yaw( float rad ) //only works with unit vector
     {
-        float theta;
+        float theta = 0.0f;
         float m = Mag();
         if (i > 0)
             theta = (float) atan( k/i );
@@ -65,7 +65,7 @@ public:
 
     void Roll( float rad )
     {
-        float theta;
+        float theta = 0.0f;
         float m = Mag();
         if (i > 0)
             theta = (float) atan( j/i );
@@ -82,7 +82,7 @@ public:
 
     void Pitch( float rad )
     {
-        float theta;
+        float theta = 0.0f;
         float m = Mag();
         if (k > 0)
             theta = (float) atan( j/k );
@@ -144,7 +144,7 @@ public:
 struct asteroid
 {
     Vector center;
-    float  radius;
+    float  radius = 0.0f;
     Vector YawPitchRoll;
     int    num_polys;
     vector< Vector >points;
--- a/engine/objconv/mesher/to_OgreMesh.cpp
+++ b/engine/objconv/mesher/to_OgreMesh.cpp
@@ -4,7 +4,7 @@
 
 #include <Ogre.h>
 #include "to_OgreMesh.h"
-
+#include <OGRE/OgrePrerequisites.h>
 #include "OgreDefaultHardwareBufferManager.h"
 
 
@@ -150,7 +150,6 @@ static string getMaterialHash( const GFXMaterial &mat,
 static string getMaterialHash( const Mesh_vec3f &v )
 {
     static const char hexdigs[] = "0123456789ABCDEF";
-    static char tmp[16];
     string ret;
     for (size_t i = 0; i < sizeof (v); i++) {
         ret += hexdigs[( ( (char*) &v )[i]>>4 )];
@@ -189,7 +188,7 @@ static string getMaterialHash( const XML &memfile )
                              memfile.alphatest );
     hash += "-";
 
-    sprintf( tmp, "-%d-", memfile.textures.size() );
+    sprintf( tmp, "-%lu-", memfile.textures.size() );
     hash += tmp;
     for (i = 0; i < memfile.textures.size(); i++) {
         sprintf( tmp, "%d,%d,", memfile.textures[i].index, memfile.textures[i].type );
@@ -197,7 +196,7 @@ static string getMaterialHash( const XML &memfile )
         hash += memfile.textures[i].name;
         hash += '#';
     }
-    sprintf( tmp, "-%d-", memfile.detailplanes.size() );
+    sprintf( tmp, "-%lu-", memfile.detailplanes.size() );
     hash += tmp;
     sprintf( tmp, "%d,%d,", memfile.detailtexture.index, memfile.detailtexture.type );
     hash += tmp;
@@ -443,7 +442,7 @@ static void SetupColorTplVars( map< string, string > &vars,
     bool   eqg   = (fabs( defg-g ) < 0.001);
     bool   eqb   = (fabs( defb-b ) < 0.001);
     bool   eqa   = (fabs( defa-a ) < 0.001);
-    bool   eqrgb = eqr && eqb && eqa;
+    bool   eqrgb = eqr && eqg && eqb && eqa; // Nabaco: Added eqg back
 
     string rgbs  = "$("+rvar+") $("+gvar+") $("+bvar+")";
     string rgbas = "$("+rvar+") $("+gvar+") $("+bvar+") $("+avar+")";
@@ -1074,10 +1073,16 @@ void Add( void *outputcontext, const XML &memfile )
 
 static void AutoOrganiseBuffers( Ogre::VertexData *data, Ogre::MeshPtr mesh )
 {
+// FIXME: Fix compilation for newer versions of OGRE
 #if (OGRE_VERSION_MAJOR == 1) && (OGRE_VERSION_MINOR < 1)
     Ogre::VertexDeclaration *newDcl =
         data->vertexDeclaration->getAutoOrganisedDeclaration(
             mesh->hasSkeleton() );
+#elif (OGRE_VERSION_MINOR == 1) && (OGRE_VERSION_MINOR >= 9)
+    Ogre::VertexDeclaration *newDcl =
+        data->vertexDeclaration->getAutoOrganisedDeclaration(
+            mesh->hasSkeleton(), mesh->hasVertexAnimation() || (mesh->getPoseCount() > 0), false);
+    // We don't include normals animations, but should we?
 #else
     Ogre::VertexDeclaration *newDcl =
         data->vertexDeclaration->getAutoOrganisedDeclaration(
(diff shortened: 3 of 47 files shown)


nabaco:
Regarding Ogre 1.9
ProgressiveMesh was switched to ProgressiveMeshGenerator.
I found some technical details here: https://forums.ogre3d.org/viewtopic.php?t=76309

I still need to figure out how to use this new class.

engine/objconv/mesher/to_OgreMesh.cpp
```diff
@@ -440,7 +440,7 @@ static void SetupColorTplVars( map< string, string > &vars,
                                bool force_rgba = false )
 {
     bool   eqr   = (fabs( defr-r ) < 0.001);
-    bool   eqg   = (fabs( defg-g ) < 0.001);
+    //bool   eqg   = (fabs( defg-g ) < 0.001);
```
BenjamenMeyer, line 443: The real bug here seems to be the lack of including `eqg` on line 446 as this effectively removes green from the RGBA calculation. Either that or a comment needs to be provided as to why it's being dropped.
@LifWirser  thoughts? Should we take it commented out? Or include it in 446 with a note about history in case a bug is found?

engine/src/mmorpgclient.cpp
```diff
@@ -139,7 +139,9 @@ void mmoc::ParseRemoteInput( char *buf )
     //parse commands to be executed <FOO=BAR>
     for (counter = 0; counter < end; counter++) {
         if (open) {
-            if (start[counter] == '=') {hackedup.append( "\" " ); } else if (start[counter] == '>') {
+            if (start[counter] == '=') {
+				hackedup.append( "\" " );
+			} else if (start[counter] == '>') {
```
BenjamenMeyer, line 144: Indentation

nabaco:
Following the different links on the [wiki](https://wiki.vega-strike.org/Development:Ogre), I understand that VegaStrike's integration with Ogre is not complete, so I'll remove the changes related to Ogre from this PR. I guess will have to deal with this separately.

engine/objconv/mesher/to_OgreMesh.cpp
```diff
@@ -139,9 +140,9 @@ static string getMaterialHash( const GFXMaterial &mat,
     ret += tmp;
     sprintf( tmp, "%06X", (unsigned int) blend_dst );
     ret += tmp;
-    sprintf( tmp, "%06X", *(unsigned int*) &polygon_offset );
+    sprintf( tmp, "%06X", *(unsigned int*) &polygon_offset ); // Why do we need this weird casting?
```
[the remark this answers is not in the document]
nabaco: From my local experiments, it seems like a trick to turn a floating-point number into a hex. Since this seems to work consistently, and it is for generating a hash, I think we should keep it as is.

engine/objconv/mesher/to_OgreMesh.cpp
nabaco: I added it in 446 with a note

nabaco:
Build should pass now. I rebased on the latest changes on master.

I'll be glad for a final review and merge.
@pyramid3d @LifWirser @Loki1950 @royfalk @BenjamenMeyer

engine/launcher/general.cpp
```diff
@@ -126,7 +126,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
 	current= new char [LENGTH+1];
 	length = strlen(line);
 	strcpy(current, line);
-	while ((location = strstr(current, search)) > 0) {
+	while ((location = strstr(current, search)) != NULL) {
```
BenjamenMeyer, line 129: we may want to use `std::nullptr` (https://en.cppreference.com/w/cpp/language/nullptr) instead of NULL. Not sure if that's good for this PR or not, so not blocking but something to start thinking about.

engine/objconv/asteroidgen.cpp
```diff
@@ -48,7 +48,7 @@ class Vector
     }
     void Yaw( float rad ) //only works with unit vector
     {
-        float theta;
+        float theta = 0.0;
```
BenjamenMeyer, line 51: `0.0f` for `float` otherwise it's a `double`

engine/objconv/mesher/to_OgreMesh.cpp
```diff
@@ -189,15 +188,15 @@ static string getMaterialHash( const XML &memfile )
                              memfile.alphatest );
     hash += "-";
 
-    sprintf( tmp, "-%d-", memfile.textures.size() );
+    sprintf( tmp, "-%lu-", memfile.textures.size() );
```
BenjamenMeyer, line 191: Please use `snprintf`:
```
snprintf(tmp, 64*sizeof(char), "-%lu-", memfile.texturse.size());
```

BenjamenMeyer: helps prevent buffer overflows ;)

engine/src/audio/Listener.cpp
```diff
@@ -10,13 +10,13 @@
...
 
     Listener::Listener() throw() :
-        cosAngleRange(-1,-1),
         position(0,0,0),
         atDirection(0,0,-1),
         upDirection(0,1,0),
         velocity(0,0,0),
         radius(1),
         gain(1),
+        cosAngleRange(-1,-1),
```
BenjamenMeyer, line 19: any info on this change? Seems a little arbitrary

engine/src/cmd/ai/hard_coded_scripts.cpp
```diff
@@ -888,7 +888,7 @@ void Takeoff( Order *aisc, Unit *un )
     ord = ( new Orders::FaceTargetITTS( 0, 3 ) );
     AddOrd( aisc, un, ord );
     un->SelectAllWeapon( false );
-    un->Fire( (weapon_info::LIGHT || weapon_info::MEDIUM), false );
+    un->Fire( (weapon_info::LIGHT | weapon_info::MEDIUM), false );
```
BenjamenMeyer, line 891: bit mapping, right?

engine/src/cmd/base_interface.cpp
```diff
@@ -1238,7 +1238,7 @@ void BaseInterface::Room::Launch::Click( BaseInterface *base, float x, float y,
                 playa->getAIState()->Communicate( c );
             abletodock( 5 );
             if (playa->name == "return_to_cockpit")
-                if (playa->faction == playa->faction)
+                //if (playa->faction == playa->faction) // This will always return true :/
```
BenjamenMeyer, line 1241: perhaps this was suppose to be `playa->faction == bas->faction`?

engine/src/cmd/collide2/Ice/IcePoint.h
```diff
@@ -207,14 +207,17 @@
...
-									return *this;
-								}
+                                    if(x<min) x=min;
+                                    else if(x>max) x=max;
+
+                                    if(y<min) y=min;
+                                    else if(y>max) y=max;
+
+                                    if(z<min) z=min;
+                                    else if(z>max) z=max;
```
BenjamenMeyer, line 217: These probably need to be in a min-max function:
x = min <= x <= max

So:
if x < min: x = min
if x > max: x = max
or something like:
```
x = std::min(x, std::max(x, max))
y = std::min(y, std::max(y, max))
z = std::min(z, std::max(z, max))
```

Can probably define a `minmax()` function to make it more obvious and less typing.

BenjamenMeyer: Or see https://en.cppreference.com/w/cpp/algorithm/minmax

royfalk:
We probably want a separate PR for replace all NULL for null_ptr.
At least on modern compiler, this generates a ton of warnings.
Not sure what negative side effects this could have.

engine/objconv/mesher/to_OgreMesh.cpp
nabaco: I want to do that in a separate PR, as part of #32.

engine/src/audio/Listener.cpp
nabaco: The compiler complains when members are not initialized in the same order as they were declared.

engine/src/cmd/ai/hard_coded_scripts.cpp
nabaco: yes, bitwise operations

engine/launcher/general.cpp
nabaco: Due to @royfalk's comment, I think that such a change should be brought up for discussion, and should be done in a separate PR after some testing.

engine/src/cmd/base_interface.cpp
nabaco: Changed with a note, for history's sake

engine/src/cmd/collide2/Ice/IcePoint.h
nabaco: Hi @BenjamenMeyer, I see such things in a lot of places in the code. I want to replace all of those places with a generic function that will be defined once for the entire codebase. I'll do that in a separate PR.

nabaco:
Hi @BenjamenMeyer, I have done some minor changes due to your comments.
If you have no objections for the rest of my comments, I'll merge the change.

nabaco:
I have to add that in this PR I have enabled Clang compilation, and it's no more an allowed failure. So if the tests have passed, then both GCC and Clang compilation passed.

nabaco:
> 
> 
> A few changes. Please check your editor and make sure you're using the same formatting. I think you might be using larger tab sizes or inserting tabs instead of spaces which is throwing some of the indentations off.

@BenjamenMeyer This is fixed too. Anything else left?

stephengtuggy:
Also: Last time I tried compiling the code, there were literally thousands of warnings. Have all of these been cleaned up? I haven't had time to check yet.

Thanks guys!

stephengtuggy:
> We probably want a separate PR for replace all NULL for null_ptr. At least on modern compiler, this generates a ton of warnings. Not sure what negative side effects this could have.

I definitely agree. Trying to put too much in one PR can be problematic for sure. Why don't we focus on one type of warning at a time, at least for now?

nabaco:
Hi @stephengtuggy, sorry for the squashing of the commits, that's the way I'm used to work from my workplace.

Codewise, is there anything specific that bother you? Or you just prefer me to break the PR down to several smaller PRs?

nabaco:
Regarding the compilation warnings, after merging this, there should be 3 types:
- warning: ‘template<class> class std::auto_ptr’ is deprecated [-Wdeprecated-declarations]
- warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]
-  warning: ‘...’ may be used uninitialized in this function [-Wmaybe-uninitialized]

The first two are related to switching to C11. I don't have enough knowledge for dealing with them, and either way, they should be treated in a separate PR.
The third one is challenging me, so I thought to deal with it separately too.

stephengtuggy:
> Hi @stephengtuggy, sorry for the squashing of the commits, that's the way I'm used to work from my workplace.

Hi @nabaco . Thanks, and no worries. Squashing commits isn't necessarily a problem; nor do we necessarily have to set a hard-and-fast rule on this. I just get nervous whenever code is changing and I can't see every single line that changed, when it changed. (I've gotten burned a few times before with bad merges and the like.)

> 
> Codewise, is there anything specific that bother you? Or you just prefer me to break the PR down to several smaller PRs?

Yes, if you could break the PR down into several smaller PRs, that would be great. Thanks.

stephengtuggy:
One more thing: I think when I built the code last year on Ubuntu 18.04, I was using gcc 8.4 (which wasn't the default compiler on that distro/release). I was also turning on ALL the warnings. Because, in my experience, warnings usually point to an actual problem in the code.

But yeah. It was a LOT to deal with all at once. Perhaps we should turn on warning types one at a time or something, going forward.

PR #35, opened 2020-04-04
Refactor particle.cpp

royfalk:
* particle.cpp
  - Remove the need for the overly complicated vectorize.h. Replace with several explicit loops in functions UpdateColor and UpdateAlpha
  - Replace custom min functions with stl
  - Replace NULL with nullptr to remove warning
* gfxlib_struct - add operator for substracting const and non-const GFXColor
* vectorize - remove
* CMakeLists - add C++11 support to suppress warnings about new style of C++. This allows us to use for(auto:) without warnings for example. 
* gitignore - Suppress build files in build-engine-Desktop-Default. Assuming it's the same for everyone, it'll save others the need to suppress a whole bunch of files after the first build.


royfalk:
I just took a look at the travis build logs and merge request. clang is broken as expected but so is gcc on Ubuntu 16.4. It's complaining of `/home/travis/build/vegastrike/Vega-Strike-Engine-Source/engine/src/vs_math.h:49:32: error: ‘isnan’ was not declared in this scope`. I didn't go anywhere near that, so I'd appreciate if anyone can shed light on this error. Did we actually compile successfully on this platform?

BenjamenMeyer:
Interesting, I'm building on 19.04 and don't have that issue - I've managed some game play. Probably missing something in the dependencies is my guess, but don't know what.

BenjamenMeyer:
Looking closer at https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/vs_math.h (the file in question) I see a few things:

1. We double include `math.h` - once outright and once after checking if configure found it. Seems like a bug right there; perhaps the second one was suppose to include `cmath` instead?
2. Based on https://en.cppreference.com/w/cpp/header/cmath we probably want to be includeing the `cmath` header and convert to using std::isnan and std::isfinite... those seem to be C++11 things; so now that that's enabled it'd probably be good to move over as well. :money_with_wings:

stephengtuggy:
> Looking closer at https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/vs_math.h (the file in question) I see a few things:
> 
> 1. We double include `math.h` - once outright and once after checking if configure found it. Seems like a bug right there; perhaps the second one was suppose to include `cmath` instead?
> 2. Based on https://en.cppreference.com/w/cpp/header/cmath we probably want to be includeing the `cmath` header and convert to using std::isnan and std::isfinite... those seem to be C++11 things; so now that that's enabled it'd probably be good to move over as well. money_with_wings

Yes. 100% agree with the above.

And yes, I can't tell you how excited I am to see this community finally making so much forward progress... together! It's truly awesome!

Issue #42, opened 2020-04-05
Travis-CI: Enhancements


BenjamenMeyer:
Travis-CI has been integrated to do PR gate checks. It presently only supports a few of the Linux distros, Ubuntu versions at that. Update the Travis-CI configuration to support the following:
- [ ] Ubuntu 18.04
- [ ] Ubuntu 20.04
- [ ] CentOS/Fedora Version ????
- [ ] Mac OS 10.15 (Catalina)
- [ ] Windows 10

Additional Possible Versions:
- [ ] Mac OS 10.14 (Mojave)

If this is not possible with Travis-CI, then is there an open-source project friendly and no cost CI system that can support these platforms? If not, we may need some community members to do periodic testing on supported platforms that can't be PR gated.

BenjamenMeyer:
Looks like Mac OS X is supported by Travis-CI.
- https://docs.travis-ci.com/user/reference/osx/
- https://docs.travis-ci.com/user/multi-os/
- Limited to 10.14 right now; no support for 10.15. That's probably okay for now.

Looks like Windows support is still experimental by Travis-CI:
- https://docs.travis-ci.com/user/reference/windows/

I suggest we start adding both the Windows and MacOS support to what we can, and allow them to fail for the time being. As Travis-CI matures and as we mature the config we can move to requiring them to pass.

BenjamenMeyer:
I've figured out OSX. I can match all our dependencies to brew installations so we should be okay; will see once I get the PR up.

Windows support is going to be harder. The package manager Travis is using for Windows (chocolatey.org) only has the following packages that match up with our dependencies:
- python2
- openal, openalsdk

We're probably going to have to just vendor stuff for Windows or add support to Chocolotey if we wanted to use that. The Windows support is still very very lacking; enough that it's not worth adding yet. Errors when trying to test using the https://config.travis-ci.com/explore:
- Doesn't recognize `windows` as an OS
- Doesn't recognize `choco` or `chocolatey` as an addon

It'll be a lot of custom work to support Windows right now. Let me get the Mac OS X support up and then we can see about Windows in a later PR.

BenjamenMeyer:
Added #43 which adds some more OS configurations, namely Ubuntu 18.04 and OS X.

OS X builds are allowed to fail right now, which is good b/c they are. More work is needed to get it fully working but that PR at least builds the foundation.

PR #43, opened 2020-04-05
Enhancement: Testing more OS configurations

BenjamenMeyer:
- Added support for OS X 10.14 Mojave
- Added support for Ubuntu 18.04

Like Clang, OSX builds are allowed to fail for now

Contributes To: #42


BenjamenMeyer:
ea50c3a is going to fail b/c the "allowed failures" section wasn't right.
2bf28e8 should succeed though.

It's interesting that we can use xcode on Linux; that's not an allowed failure yet, so if 2bf28e8 fails that probably why.

.travis.yml
```diff
@@ -26,8 +28,19 @@ jobs:
...
           compiler: clang
           name: "Ubuntu 16.04 LTS - Clang"
+        - os: linux
+          dist: bionic
+          compiler: gcc
+          name: "Ubuntu 18.04 LTS - GCC"
+        - os: linux
+          dist: bionic
+          compiler: clang
+          name: "Ubuntu 18.04 LTS - Clang"
```
nabaco, line 38: Those are unnecessary as we included `dist: bionic`. so it is built automatically for both compilers.

.travis.yml
```diff
@@ -1,13 +1,15 @@
...
+ - linux
+ - osx
 
 dist: bionic
 
 compiler:
     - gcc
     - clang
-    
+    - xcode
```
nabaco, line 11: AFAIK xcode is not a separate compiler, but xcode == clang.
meaning that it's enough to add `os: osx`, and it'll compile.

BenjamenMeyer: yeah, it's weird in the matrix. Listing `xcode` here does add a Linux+XCode build which passes. So I'm not sure if it fully matches `clang` or not. The builds on OS X do fail though, which I'm more interested in trying to get fixed.

nabaco: It seems that there is a conflict
 between python2 and python3

.travis.yml
BenjamenMeyer: perhaps...but it'd probably be good to keep so that we can make it easy to roll forward with new editions, such as when Travis-CI supports 20.04.

BenjamenMeyer:
Build of 2bf28e8 failed due to the 16.04 build failing, which is related to `vs_math.h` having problems. This is a known error for the time being. Not looking to fix that in this PR unless folks request it be added to the `accept failure` list. Otherwise, any feedback on fixing the OS X builds so they pass if there's something easy to do would be greatly appreciated; I'll look as I can. I personally only have Linux systems available to test on.

.travis.yml
nabaco: The issue is that we're running those tests twice 😅 
I'd suggest to add a comment next to `dist: bionic` that once updated, the old dist needs to be added manually to the `jobs` section.
I'll see if I can file a feature request with Travis-CI to add distros to the matrix like you add OSes.

.travis.yml
nabaco: On the same topic, looking at the OSX build log:

```sh
GCC version

Apple LLVM version 9.1.0 (clang-902.0.39.2)

Target: x86_64-apple-darwin17.7.0

Thread model: posix

InstalledDir: /Applications/Xcode-9.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

LLVM version

Apple LLVM version 9.1.0 (clang-902.0.39.2)

Target: x86_64-apple-darwin17.7.0

Thread model: posix

InstalledDir: /Applications/Xcode-9.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

Both GCC and Clang point at Xcode, which in turn is just Clang.

Look at the Linux Xcode build log:
```sh
$ export CXX=${CXX:-g++}

$ export CXX_FOR_BUILD=${CXX_FOR_BUILD:-g++}

$ export CC=${CC:-gcc}

$ export CC_FOR_BUILD=${CC_FOR_BUILD:-gcc}

$ xcode --version

Command 'xcode' not found, did you mean:

  command 'code' from snap code (0ba0ca52)

  command 'tcode' from deb emboss

See 'snap info <snapname>' for additional versions.
```
Xcode doesn't exist there, and it just compiles with GCC.

I think we can remove Xcode from the `compiler` list.

royfalk:
I’m confused about what we’re trying to do here.
Xcode isn’t a compiler. It’s an IDE. It’s running LLVM/clang underneath.
I do have a Mac (work machine), so I can try and compile it and fix errors, though it may take a while.

nabaco:
> 
> I’m confused about what we’re trying to do here. Xcode isn’t a compiler. It’s an IDE. It’s running LLVM/clang underneath. I do have a Mac (work machine), so I can try and compile it and fix errors, though it may take a while.

I didn't know it's an IDE, but you seem to just prove my point - no need to add is a compiler :)

Current issues with MacOS compilation:
- Python2 vs. Python3 conflict
- libjpeg didn't install
- Cmake looks for GTK+2, though GTK+3 was installed

BenjamenMeyer:
Just trying to get an OSX env up. If I'm reading the Travis config right we have to specify xcode for the environment we need.
Per Gtk...I couldn't find a Gtk2 install in brew and was hoping the Gtk+3 formula would get it; the doc there was a little confusing as to the actual Gtk version.
I'll get to some changes you all have said this evening, though maintainers can push against it too if y'all want to try stuff in the meantime.
Any and all guidance for OSX appreciated.

Sent from Yahoo Mail on Android 
 
  On Sun, Apr 5, 2020 at 2:43 AM, Nachum Barcohen<notifications@github.com> wrote:   



I’m confused about what we’re trying to do here. Xcode isn’t a compiler. It’s an IDE. It’s running LLVM/clang underneath. I do have a Mac (work machine), so I can try and compile it and fix errors, though it may take a while.


I didn't know it's an IDE, but you seem to just prove my point - no need to add is a compiler :)

Current issues with MacOS compilation:
   
   - Python2 vs. Python3 conflict
   - libjpeg didn't install
   - Cmake looks for GTK+2, though GTK+3 was installed

nabaco:
> Per Gtk...I couldn't find a Gtk2 install in brew and was hoping the Gtk+3 formula would get it; the doc there was a little confusing as to the actual Gtk version.

The package is gtk+. I have tested it locally, and CMake still fails, looking for gtk+2-x11. I'm unsure how much it's relevant for MacOS, so maybe we need to adapt it somehow.

> I'll get to some changes you all have said this evening, though maintainers can push against it too if y'all want to try stuff in the meantime. Any and all guidance for OSX appreciated.

I'm not sure what you mean. Since it's your branch, I don't think I can push changes there.

BenjamenMeyer:
So apparently the Homebrew folks dropped the X11 support for Quartz (see https://github.com/Homebrew/homebrew-core/issues/21826); MacPorts seems to still has X11 support (https://github.com/macports/macports-ports/tree/master/gnome/gtk2); might have to ese if I can mix the two in Travis; though we probably want to move to Quartz support on Mac in the long run any way.

Loki1950:
May have to move away from OpenGL as well and move toward Vulkan which appears to be the replacement on Windows and OSX

BenjamenMeyer:
```
Skipping install of jpeg formula. It is already up-to-date.
Using jpeg
Skipping install of libpng formula. It is already up-to-date.
Using libpng
...
-- Found JPEG: /usr/local/lib/libjpeg.dylib (found version "90") 
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'libpng'
--   Found libpng, version 1.6.37
-- Found PNG: /usr/local/lib/libpng.dylib
```
Seems libjpeg and libpng are already installed.

SDL2 installed but we need SDL1. Will have a few fixes up shortly.

BenjamenMeyer:
@Loki1950  I think we just need to figure out how to detect Gtk2-Quartz in `engine/FindGTK2.cmake`; but that's better for a separate PR.

BenjamenMeyer:
NOTE: Build failure is just due to the known issue around 16.04, which #44 should resolve.

nabaco:
> 
> 
> @Loki1950 I think we just need to figure out how to detect Gtk2-Quartz in `engine/FindGTK2.cmake`; but that's better for a separate PR. Quartz still seems to support OpenGL. It should be doable to just swap the Gtk backend and have it work unless we tie into something too closely, but since we support Windows I doubt that's the case.
> 
> Filed #46 regarding the Vulkan side of things.

I think it's enough to add in engine/FindGTK2.cmake L199 `gtk-quartz` or `gtk+-quartz`. I'll check whether this works later tonight.

BenjamenMeyer:
merged master; that should resolve the build failure.

stephengtuggy:
I would like to request some time to review this in more detail, please. I have some catching up to do haha.

I will say, though, that I really appreciate that this project finally (finally!) has Continuous Integration. I can't tell you how excited I am about that. I've been wanting to do this for, like, a year or more. CI is going to make our lives so much easier.

stephengtuggy:
So. Do we have `gcc` 8.4/8.x yet? That might be another good option to add.

Also `clang` on Windows. Visual Studio, of course -- probably VS2017 to start with. Maybe `LLVM` as well?

stephengtuggy:
More discussion of the Windows side of things at #53

stephengtuggy:
Maybe `LLVM` across the board would be nice. (I do have that compiler name correct, right?)

stephengtuggy:
And I would be totally okay with including Ubuntu 19.10. Along with Ubuntu 20.04 when it comes out. But that doesn't all have to happen right away.

royfalk:
I'm not sure what there is to review here. It's one file outside the source code and it seems OK. We're allowing new additions to fail anyway. You've got my vote.

stephengtuggy:
Fair enough. Looks good to me as well. Thanks, guys.

PR #44, opened 2020-04-05
Enhancement: std::(isfinite, isnan)

BenjamenMeyer:
Convert to using the provided versions of the functions via the standard library as of C++11.
Hopefully this will fix the build error on Ubuntu 16.04 too.

### DIFF
--- a/engine/src/vs_math.h
+++ b/engine/src/vs_math.h
@@ -26,6 +26,7 @@
 #include "macosx_math.h"
 #include <math.h>
 #include <float.h>
+#include <cmath>
 
 #ifdef HAVE_IEEEFP_H
  #include <ieeefp.h>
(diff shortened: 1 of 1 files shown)


stephengtuggy:
:tada:

engine/src/vs_math.h
```diff
@@ -26,6 +26,7 @@
 #include "macosx_math.h"
 #include <math.h>
 #include <float.h>
+#include <cmath>
```
stephengtuggy, line 29: Perfect! Based on the documentation I've read of c++11 and later standards, this does indeed look like the correct form/correct syntax of the header to include.

engine/src/vs_math.h
```diff
@@ -26,6 +26,7 @@
 #include "macosx_math.h"
 #include <math.h>
```
stephengtuggy, line 27: Overall, this looks excellent. Absolutely outstanding! :tada:

Just one, relatively minor point: Now that we're including <cmath>, we probably don't need <math.h> any more. :sunglasses: 

Loving all the progress, you guys!

Issue #45, opened 2020-04-05
C11 - Remove compilation warnings


nabaco:
Since moving to C11 standard in #35, there is a flood of compilation warnings, where are either of the following two:

- [x] `warning: ‘template<class> class std::auto_ptr’ is deprecated [-Wdeprecated-declarations]`
- [x] `warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]`

It would be nice to remove the warnings and align our codebase with that standard.

nabaco:
@royfalk Would you be willing to take this, as you seem to be more proficient with C11?

BenjamenMeyer:
The `std::auto_ptr` just needs to be updated to a new smart pointer type; see https://en.wikipedia.org/wiki/Auto_ptr for some info; `std::unique_ptr` is the primary replacement, and `std::shared_ptr` might be useful too but we'll have to identify which ones need it. We can probably start by swapping them to `std::unique_ptr`.

Not sure about the other one.

LifWirser:
Congradulations somebody managed to make an issue out of something that was  SOLVED.


Not only were the "auto_ptr" issue already dealt with, but the attempts to solve it AGAIN corrupted the 0.5.3 branch.
NICE JOB

BenjamenMeyer:
@LifWirser I'm not seeing any issue on the 0.5.3 branch. On master, this hasn't been solved. If there is a change that needs to be pushed to master to solve it then that should get pushed to master IMHO.

Also, we're seeing it on master now b/c of the addition of support for C++11 and C++11 deprecated `std::auto_ptr`.

nabaco:
Currently both branches are corrupted due to a change that was merged, without CI checks or a PR review (df7f7d420dec684d0af775a8a0a1d6e9bb59bc0a).
My PR should fix the issue, #29.

The issue with the auto_ptr appeared due to a decision to switch fully to C11. In my opinion, the decision was taken a bit to quickly, and without proper discussion, but I think I'm too new in the community to affect such decisions. Either way, if we don't want to revert the change, we need to adapt the code to it. It's your call.

BenjamenMeyer:
@nabaco not sure a PR gate would have caught that without unit tests in place

royfalk:
For the record, I’m the one who pushed the change.
I actually debated whether to add this or not, but figured this would come up in the review.
It was clearly marked as one of the changes in the PR.

I only merged the change after:

[...]

PR #47, opened 2020-04-06
WIP - Refactor codebase to comply with C11

nabaco:
Replaced all auto_ptr with unique_ptr to comply with C11.
Relevant to #45

### DIFF
--- a/engine/src/gfx/radar/radar.cpp
+++ b/engine/src/gfx/radar/radar.cpp
@@ -12,21 +12,21 @@
 namespace Radar
 {
 
-std::auto_ptr<Display> Factory(Type::Value type)
+std::unique_ptr<Display> Factory(Type::Value type)
 {
     switch (type)
     {
     case Type::NullDisplay:
-        return std::auto_ptr<Display>(new NullDisplay);
+        return std::unique_ptr<Display>(new NullDisplay);
 
     case Type::SphereDisplay:
-        return std::auto_ptr<Display>(new SphereDisplay);
+        return std::unique_ptr<Display>(new SphereDisplay);
 
     case Type::BubbleDisplay:
-        return std::auto_ptr<Display>(new BubbleDisplay);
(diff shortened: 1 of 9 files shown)


engine/src/gfx/radar/radar.cpp
```diff
@@ -12,21 +12,21 @@
 namespace Radar
 {
 
-std::auto_ptr<Display> Factory(Type::Value type)
+std::unique_ptr<Display> Factory(Type::Value type)
```
BenjamenMeyer, line 15: These might need to be `std::shared_ptr`; not sure. Does it run well with `std::unique_ptr`?

LifWirser:
I just tried to clone and compile 0.5.3 only to get "bunches of "auto_ptr " errors when I have had none WTF happened. 0.5.3 had NONE before. did somebody know what they did, OR were they trying to impress somebody?

nabaco:
> 
> 
> I just tried to clone and compile 0.5.3 only to get "bunches of "auto_ptr " errors when I have had none WTF happened. 0.5.3 had NONE before. did somebody know what they did, OR were they trying to impress somebody?

I think @royfalk can comment on this. To get rid of the warnings requires a single change in engine/CMakeList.txt, to remove the C11 standard from there.

royfalk:
It sounds right. I just replied to a more recent email with a more full explanation.

nabaco:
> 
> 
> So yeah. I think this PR was well-intentioned, but a little too simplistic. It's going to take a LOT more work to figure out which c++11/boost ptr types to use in each case. I started down this road myself, but would really like more input.
> 
> See, for example, #8 for further discussion.

Point taken. I thought there might be a quick way to solve the issue and remove all the warnings, but I guess it'll take some more time and effort. I'll close the PR.

stephengtuggy:
Cool. Thanks, @nabaco .

PR #51, opened 2020-04-07
Bug Fixes

BenjamenMeyer:
- 16.04 seems to stil be breaking; we don't need `math.h` in
  vs_math.h since `cmath` is included, so just use the one as the
  math functions (std::isnan, std::isfinite) come from cmath.
- Typo in IcePoint - need a space between `else` and `if`

### DIFF
--- a/engine/src/cmd/collide2/Ice/IcePoint.h
+++ b/engine/src/cmd/collide2/Ice/IcePoint.h
@@ -212,7 +212,7 @@
 									if(y<min)	y=min;
 			                                                    else if(y>max)	y=max;
 									if(z<min)	z=min;
-			                                                     elseif(z>max)	z=max;
+			                                                     else if(z>max)	z=max;
 									return *this;
 								}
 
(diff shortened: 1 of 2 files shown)


engine/src/cmd/collide2/Ice/IcePoint.h
```diff
@@ -212,7 +212,7 @@
 									if(y<min)	y=min;
 			                                                    else if(y>max)	y=max;
 									if(z<min)	z=min;
-			                                                     elseif(z>max)	z=max;
+			                                                     else if(z>max)	z=max;
```
nabaco, line 215: This is fixed in my PR too, so it'll cause a conflict to whoever merges second. I won't block on this.

BenjamenMeyer: @nabaco since this is focused, can we merge this to master then update your PR?

nabaco: Sure, the effort to rebase my PR will be minimal.

PR #57, opened 2020-04-08
Refactor (Q)Vector

royfalk:
Vector and QVector are two, nearly identical classes based on float and double. Current implementation uses ifdef to generate these two files by applying #define macro to xvector.h and generating the two classes in vec.h. 
This commit replaces it with tvector, a template based implementation that uses typedef to replicate the two classes.


royfalk:
It builds but it doesn't run. It prints an error of:
`16:07:21: /home/roy/vegastrike/Vega-Strike-Engine-Source/build-engine-Desktop-Default/objconv/replace exited with code 255`

When I first tested it, I think I ran the stable repository by accident, thinking everything's OK. Closing until I figure out what went wrong. If you have an idea of what this error means, feel free to comment.

BenjamenMeyer:
@royfalk  perhaps some unit testing would help show how the functionality differs?

royfalk:
I'm reopening my pull request provisionally. For some reason, the IDE started running _objconv/replace_ utility without any parameters (which returns with -1) instead of _vegastrike_. It runs for me but I would like someone to actually run it as well before approving.
I can work on other stuff in the meanwhile, so no rush.

nabaco:
I have tested the changes locally, including docking on a planet, SPEC flight, fighting, and crashing into a station. Everything seems OK, so I'm giving my approve.

nabaco:
@danielrh @BenjamenMeyer @stephengtuggy Could you guys please review the PR?

engine/src/gfx/tvector.h
```diff
@@ -0,0 +1,180 @@
...
+    };
+    union
+    {
+        S k;
+        S z;
+    };
+
+// Constructors
+    TVector() : i(0),j(0),k(0){}
+    TVector( const TVector<T,S> &a );
```
stephengtuggy, line 47: Question: Why is this `T,S` and not `S,T`?

stephengtuggy, line 47: Question: Why is this `T,S` and not `S,T`?

I have the same question about a few other lines below, also.

nabaco: @royfalk ?

royfalk: Quick disclaimer - still a beginner C++ dev, as I learned today to my sorrow. This is not a default copy constructor. That one is covered by the rule of 0. This is a constructor to create QVector from Vector and vice-versa. I worked off the original xvector.h and vec.h. If you look there, you'll find my reference constructors.

stephengtuggy: @royfalk Thanks for the reply.

Perhaps the default copy constructor is implied. But based on my logging, the program was hitting this one quite a lot, until I defined the default copy constructor `<S,T>` explicitly, alongside this one.

Will it hurt anything to do this?

BenjamenMeyer: @stephengtuggy that might mean we could simply do:
```
typedef TVector<float, double> Vector
typedef Vector QVector
```

Not sure, but it shouldn't hurt anything. Might help if we put a high level verbose logs in each so we can trace which is getting called too (e.g logs we have to explicitly enable for dev/diag purposes but would normally be turned off - probably a compile-time option).

stephengtuggy: @BenjamenMeyer Hmm. Not sure. Wouldn't that mean that both `Vector` and `QVector` would be `float` precision, rather than `double`?

I do have the logging set up much the way you described. It's at the highest verbosity level -- `trace`, in Boost.Log terms -- which only gets enabled when you pass `--debug=3` to vegastrike on the command line. (As opposed to `--debug=2`, `--debug=1`, or `--debug` by itself.)

This whole thing is turning out to be quite the rabbit hole. I think I could use more eyes on it.

BenjamenMeyer: @stephentuggy yeah...not sure. There's certainly some complications. Honestly we should push it to be all doubles and convert to float only when necessary.

stephengtuggy: @BenjamenMeyer yeah, that was my thought as well. Ideally it would be all doubles. However, there are explicit `float`s all over the code. It would be a massive change.

BenjamenMeyer: @stephentuggy. Do we want to wait until 0.9.x to do that?

royfalk:
Welcome to my rabbit warren 😊
I actually search/replaced floats everywhere a while back and it didn’t work.
I found some code where the game parses a double into two floats (I think in save game).

I think the key to mapping the important places in the code is the cast() method.
It converts between Vector and QVector.

I would totally recommend waiting till 9.0. Hopefully other code refactors would simplify this as well.

BenjamenMeyer:
@royfalk yes it'll have to be a mix of moving to doubles and casting as a number of external APIs use floats in their functions.

stephengtuggy:
If we wait until 0.9.x to move to doubles, does that mean that we will be rolling back this PR for the time being? On master, or else just on the 0.7.x and 0.8.x branches?

BenjamenMeyer:
@stephengtuggy it means we'll do the same fix on the branches for 0.7.x and 0.8.x that we did on 0.6.x; not going to do that on master though unless we decide to abandon these changes entirely.

stephengtuggy:
I am not happy with having master in an unplayable state for that long.

Hopefully there is some easier solution that we're just not seeing.

BenjamenMeyer:
@stephengtuggy I'm hoping we can push through those releases relatively quickly once we get 0.6.x out the door - most of what is holding up 0.6.x right now isn't code, but all the other things that need to be done.

PR #65, opened 2020-04-09
(refactor) Fix clang compilation

nabaco:
Did some minor refactoring to let compilation pass with the Clang compiler.
Removed Clang from allowed failures in Travis-CI.
Tested on Archlinux with Clang-9.0.1.
From a minor test, game-play seems unaffected.

### DIFF
--- a/engine/setup/src/include/file.cpp
+++ b/engine/setup/src/include/file.cpp
@@ -167,7 +167,7 @@ void LoadConfig( void )
         n_parm = next_parm( parm );                     //next_parm is a line splitter included with general.c
         if (strcmp( "groups", parm ) == 0) {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
                 G_CURRENT->name = NewString( parm );
                 G_NEXT = (struct group*) malloc( sizeof (struct group) );
                 if (G_NEXT == 0) {
@@ -188,7 +188,7 @@ void LoadConfig( void )
             n_parm = next_parm( parm );
             group  = NewString( parm );
             parm   = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
                 C_CURRENT->name = NewString( parm );
                 C_NEXT = (struct catagory*) malloc( sizeof (struct catagory) );
                 if (C_NEXT == 0) {
@@ -317,7 +317,7 @@ void Modconfig( int setting, const char *name, const char *group ) {
         parm++;
         if (strcmp( name, parm ) == 0) {commenting = setting; } else {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
                 if (parm[0] == '<') break;
                 if (strcmp( name, parm ) == 0) {
                     commenting = setting;
--- a/engine/setup/src/include/file.cpp
+++ b/engine/setup/src/include/file.cpp
@@ -167,7 +167,7 @@ void LoadConfig( void )
         n_parm = next_parm( parm );                     //next_parm is a line splitter included with general.c
         if (strcmp( "groups", parm ) == 0) {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
+            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
                 G_CURRENT->name = NewString( parm );
                 G_NEXT = (struct group*) malloc( sizeof (struct group) );
                 if (G_NEXT == 0) {
@@ -188,7 +188,7 @@ void LoadConfig( void )
             n_parm = next_parm( parm );
             group  = NewString( parm );
             parm   = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
+            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
                 C_CURRENT->name = NewString( parm );
                 C_NEXT = (struct catagory*) malloc( sizeof (struct catagory) );
                 if (C_NEXT == 0) {
@@ -317,7 +317,7 @@ void Modconfig( int setting, const char *name, const char *group ) {
         parm++;
         if (strcmp( name, parm ) == 0) {commenting = setting; } else {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
+            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
                 if (parm[0] == '<') break;
                 if (strcmp( name, parm ) == 0) {
                     commenting = setting;
--- a/engine/setup/src/include/general.cpp
+++ b/engine/setup/src/include/general.cpp
@@ -120,7 +120,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
 	calc = strlen(line) - strlen(search) + strlen(replace);
 	if (calc > LENGTH) { return line; }
 	strcpy(current, line);
-	while ((location = strstr(current, search)) > 0) {
+	while ((location = strstr(current, search)) != NULL) {
 		chr_new[0] = '\0';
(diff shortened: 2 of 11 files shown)


nabaco:
This is a breakdown from #29

engine/setup/src/include/file.cpp
```diff
@@ -167,7 +167,7 @@ void LoadConfig( void )
         n_parm = next_parm( parm );                     //next_parm is a line splitter included with general.c
         if (strcmp( "groups", parm ) == 0) {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
```
BenjamenMeyer, line 170: Please verify this is actually doing pointer comparison against zero and not getting evaluated as something else first.

engine/src/gfx/nav/navitemstodraw.h
```diff
@@ -25,14 +25,14 @@ void NavigationSystem::DrawCursor( float x, float y, float wid, float hei, const
...
-            x+0.35*wid, y-0.6*hei,  0,
-            x+wid,      y-0.75*hei, 0,
+            x,                              y,                              0,
+            x,                              static_cast<float>(y-hei),      0,
+            x,                              y,                              0,
+            static_cast<float>(x+wid),      static_cast<float>(y-0.75*hei), 0,
+            x,                              y-hei,                          0,
+            static_cast<float>(x+0.35*wid), static_cast<float>(y-0.6*hei),  0,
+            static_cast<float>(x+0.35*wid), static_cast<float>(y-0.6*hei),  0,
+            static_cast<float>(x+wid),      static_cast<float>(y-0.75*hei), 0,
```
BenjamenMeyer, line 35: Is this all space aligned or tab aligned?
Let's make sure it's space aligned.

nabaco:
> Some of the comparisons in the while loops look wrong in the change since the original was `> 0` instead of `!= 0`; these need to be verified as pointer comparisons and ensure that they are not getting morphed via operators into comparing another value.

@BenjamenMeyer I would say even further, **all** the replacements are from `> 0` to `!= NULL`. Please notice that I have replaced in all the places that the Clang failed due to an error such as this:
```sh
/home/travis/build/vegastrike/Vega-Strike-Engine-Source/engine/setup/src/include/central.cpp:39:37: error: 

      ordered comparison between pointer and zero ('struct group *' and 'int')

        } while ((CURRENT = CURRENT->next) > 0);

                 ~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
```

engine/src/gfx/nav/navitemstodraw.h
nabaco: Everything is space aligned, I have rechecked just to make sure.

engine/setup/src/include/file.cpp
nabaco: See response below

BenjamenMeyer:
@nabaco that's why I generally - in my own code - don't favor doing tricks like that. Writing code that is easier to understand is more important, and the compiler will generally get the optimizations needed. So I'd personally do:
```c++
do {
    ...
    CURRENT = CURRENT->next;
} while (CURRENT > 0);
```
Making it explicit what is being shown and not relying on assignments within an evaluation to occur. That's generally considered better coding practice too; though a lot of older code did that kind of thing.

Now, I'm not saying re-write it. Just noting my personal coding style and approach.
It'd be interesting to see if the clang still generates an error in such a case though - if not, then another evaluation mechanism may be coming into play too - likely not; it's highly likely to just be the pointer comparison, and checking the data structure definition should show that too as well as looping for any `operator>()` definitions that take the structure.

nabaco:
Hi @BenjamenMeyer, I'm having a hard time understand what is the bottom line. Do you suggest to refactor all of those places as you do? Or it's good enough as it is now?
Currently I just wanted to fix Clang compilation, and as you can see in the Travis CI checks, it works. Compilation passes on all versions of Clang.

BenjamenMeyer:
@nabaco if you want to test a change to see, that's fine - but no I'm not saying we should do that refactor now.

Bottom line is I'd like to hear from @LifWirser or @stephengtuggy  and others that have more familiarity with the code right now to ensure there's nothing non-obvious happening.

nabaco:
> I gotta be honest: I kind of hate pointer math. :-/ These kinds of older code constructs are too easy to get wrong, in my experience.
> 
> I would rather leave this as a known `clang` failure for now. Then come back and fix it later, by doing a larger refactor to use C++ STL standard algorithms or similar. Stuff that is more likely to be correct; properly bounds-checked; and so forth.
> 
> Yes, it will be a lot of work. I think it could be worth it though.

Hi @stephengtuggy,
I feel like I'm missing something here. The changes that I have made do not touch any pointer arithmetic at all. All I'm doing is changing the style of NULL pointer tests. I have to say it's quite standard in C/C++. In these cases you see a pointer assigned to a different pointer, and then tested whether it's a NULL pointer or not.
I believe that we will benefit from enabling the Clang compilation, as it catches more warnings/errors, and has more friendly messages. And though you're right, there is not hurry, but I don't see a reason to postpone it, since we'll only accumulate issues with it. Especially considering the simplicity of the fix.
It's your call.

engine/setup/src/include/file.cpp
royfalk: I didn't actually see a response from Nachum, but it sure looks like it's pointer comparison against zero.
n_parm is of type char *. next_parm gets the next parameter from the string, sorted by a space, sticks a \0 at the space and returns a pointer to the right side. I'm actually refactoring in this area and may well replace everything here with a line or two from boost.

engine/setup/src/include/file.cpp
```diff
@@ -167,7 +167,7 @@ void LoadConfig( void )
...
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
```
royfalk, line 170:
```suggestion
            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
```

engine/setup/src/include/file.cpp
```diff
@@ -188,7 +188,7 @@ void LoadConfig( void )
             n_parm = next_parm( parm );
             group  = NewString( parm );
             parm   = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
```
royfalk, line 191:
```suggestion
            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
```

engine/setup/src/include/file.cpp
```diff
@@ -317,7 +317,7 @@ void Modconfig( int setting, const char *name, const char *group ) {
         parm++;
         if (strcmp( name, parm ) == 0) {commenting = setting; } else {
             parm = n_parm;
-            while ( ( n_parm = next_parm( parm ) ) > 0 ) {
+            while ( ( n_parm = next_parm( parm ) ) != NULL ) {
```
royfalk, line 320:
```suggestion
            while ( ( n_parm = next_parm( parm ) ) != nullptr ) {
```

engine/setup/src/include/general.cpp
```diff
@@ -120,7 +120,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
 	calc = strlen(line) - strlen(search) + strlen(replace);
 	if (calc > LENGTH) { return line; }
 	strcpy(current, line);
-	while ((location = strstr(current, search)) > 0) {
+	while ((location = strstr(current, search)) != NULL) {
```
royfalk, line 123:
```suggestion
	while ((location = strstr(current, search)) != nullptr) {
```

BenjamenMeyer:
Q: Does this pass our PR Validation?
https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

nabaco:
> Q: Does this pass our PR Validation?
> https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Yes, the full suit, no abnormalities found.

BenjamenMeyer:
Ok. Please merge master I'll remove my block once I'm on a computer. Can't figure out how through the GitHub app.

stephengtuggy:
With the switch to comparing against `nullptr` instead of `NULL`, I am good with what you guys have done here. Thanks.

PR #77, opened 2020-04-21
Remove deprecated throw definition

nabaco:
Partially solves #45.
Since we moved C++11, we're receiving a plethora of compilation warnings. This PR addresses one specific warning that is printed a lot:
`warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]`

Be warned, the PR is big, so review it when you have some free time.

Gameplay tests are yet to be done.


nabaco:
> 
> 
> Q: What is ` engine/src/cmd/ai/sedxjaQZy` for? If there's no reason it should get removed.
> @bmorel

Seems like a glitch from the command I used. Deleted.

nabaco:
Tested according to test plan. All works fine.

PR #80, opened 2020-04-27
Fixed -Wcatch-value warnings

nabaco:
The fix is based on [this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#e15-catch-exceptions-from-a-hierarchy-by-reference) and [this](https://github.com/leela-zero/leela-zero/pull/2134/files)

Didn't test gameplay yet.

### DIFF
--- a/engine/src/audio/RenderableListener.cpp
+++ b/engine/src/audio/RenderableListener.cpp
@@ -22,7 +22,7 @@ namespace Audio {
     {
         try {
             updateImpl(flags);
-        } catch(Exception e) {}
+        } catch(const Exception& e) {}
     }
 
 };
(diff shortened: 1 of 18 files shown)


nabaco:
@royfalk @stephengtuggy @BenjamenMeyer 
Any chance you guys can have a look at this?

nabaco:
I have tested the code on the latest master. All seems fine, except a glitch with the flight speed that seems to be jumping.
I have tested it on previous versions, and it seems to happen there too sometimes. So I'm merging this PR, but we'll have to see when did we introduce the issue.
I'll probably open a separate ticket for it.

BenjamenMeyer:
@nabaco yeah - I can confirm I've seen the flight speed bounce back and forth; typically seems to be when going from normal speed (300kps) to light speeds as you're moving away from a planet/etc and speeding up.

Issue #83, opened 2020-05-01
Git repository history might contain bloat


BottledByte:
Currently, size of this repo is around 150 MB and size of Assets-Production repo is around 1.4 GB. However, when git `clone`-ing those repos, full history included (~1.5 GB without it), *approx. 3.4 GB is downloaded*. ~800 MB of this is VS's engine history and another ~1.1 GB is from the asset repo's history.

Vega Strike has many years of development behind it, so it can be said that the size of engine's history is okay. However, assets repo might be a problem, as it contains all the models, textures, etc. Git treats those files badly (and those files tend to be quite big), so if it is not a problem now, it might become one in the future (if those assets would get updated several times).

Also, this issue somewhat relates to #38. From those 150 MB of actual engine repo size... ~90% is Boost. If I understood it correctly, Boost was not originally kept in the SVN repo, only as a reference. But Git makes snapshots of everything, so it is probable that it keeps snapshots of entire Boost in history and if the Boost would be upgraded (and vendored the same way as is now), another snapshot would be created (creating a lot of bloat).

@nabaco suggested that a notice is added to readme that the size of history is somewhat big and maybe suggest there to use shallow git `clone`.

Feel free to close this if you do not consider it an issue. (and rather a Git feature :) )

BenjamenMeyer:
Yes, asset repo is large and will grow that can't be helped too much.
Third party libs like boost need a better vendoring solution than being simply included in the repository.

evertvorster:
I just built two packages for Arch Linux AUR for vegastrike. Both of them pull the full git repo before building the package. One for the engine, one for the data. 

Download size for the engine was around 700mbm which is a bit vulgar for something that boils down to 7mb package once built. 

The data one can understand that it is big, and that is around a 1gb download.

The way the AUR works is that anyone that wants to install this package would have to download both git repos as well, so a little de-bloating here would be much appreciated.

BenjamenMeyer:
@evertvorster our goal is that they won't have to download the git, but can have two packages: (1) engine (small) and (2) data, with the data package being used to support multiple games - we already know of 2 separate games (Under the Coldest Sea, provided by our Asset Production repo) and the Privateer WCU Parallel Universe (provided by the PWCU GH Org). Let's continue that discussion in the #131 issue.

stephengtuggy:
[...]

Issue #53, opened 2020-04-07
Figure out and fix Windows compilation


stephengtuggy:
So guys. Last year, I literally spent **2 months** trying to build Vega Strike on Windows 10! One of the most frustrating experiences of my developer life.

One thing I CAN tell you, is that Visual Studio 2019 is a no-go as a development environment. The required third-party dependencies just hadn't been updated to support VS2019. VS2017 is probably the latest version that will work.

We will also (probably) want to try clang as an alternate compiler on Windows. Perhaps llvm as well?

And finally, TBH, CMake may just not be up to the task any more. It is a venerable tool, and has served us well for years. But now, it might be time to look at other options? What do you guys think? I have a couple of suggestions for things we could try, if you guys want to hear them.

stephengtuggy:
Relates to #46

stephengtuggy:
Relates to #52

BenjamenMeyer:
I don't think there's a good cross-platform replacement for CMake. AutoTools are a no-go on Windows.  QMake is nice but Qt is looking to drop it in favor of CMake; even their qbs is being dropped in favor of CMake.

Starting with VS2017 sounds good as far as Windows goes.

As to changing compilers - I'm more in favor of supporting multiple compilers - so supporting GCC, Clang, LLVM, and VC at the same time. It tends to help C++ code as it clears out bugs through the different behaviors. Right now, we just need to focus on one at a time, which looks like GCC. Clang will be a follow-on; LLVM can be proposed but won't likely happen until after CLang. VC will come with the Windows support; but first we have to have a CI system that will support Windows so that we can keep from breaking it as we move forward.

royfalk:
From Wikipedia: The Clang project provides a language front-end and tooling infrastructure for languages in the C language family (C, C++, Objective C/C++, OpenCL, CUDA, and RenderScript) for the LLVM<https://www.llvm.org/> project.
So that's one thing less to worry about.
________________________________

stephengtuggy:
> I don't think there's a good cross-platform replacement for CMake. AutoTools are a no-go on Windows. QMake is nice but Qt is looking to drop it in favor of CMake; even their qbs is being dropped in favor of CMake.

Fair enough. Mozilla `mach` might be an option though. Maybe. I still need to find out a little more about it.

[...]

PR #84, opened 2020-05-01
Removed libboost from our source

### LINKED ISSUE #38: Drop vendoring of libboost
Per https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/33#issuecomment-609089044
VS provides a vendored copy of libboost. Is this still necessary? Or can we assume it is available?

nabaco:
Closes #38, related to #83.

I removed the boost source code from our repo, made the compilation with system boost default, and added a note in the README in case issues arise.

This PR is a proposition and not a decision.
In my opinion, if the build passes CI cleanly, we can merge it, but I'll leave it for the community to decide.


nabaco:
Well darn, the build is failing for Ubuntu 14.04 GCC. Oddly enough, for Clang it passes :/

I'm think to enhance the CMake file to download libboost by itself, unpack it and compile it, if the system boost is not good enough.

Anyone knows what is the minimal Boost version we can use?

Loki1950:
I will hold on that review till it complies with gcc

BenjamenMeyer:
@nabaco since we have a limited amount of GitHub Package storage (500Mb), we might just want to setup a repository for developer assets where we can just drop a tarball (Unix/Linux) and Zip file (Windows) to support this. Then CMake could be setup to just download the repository into a specific folder (external-assets) and pull and extract whatever versions from there. That would also give us a bit better delta comparisons for any 3rd party library we need to track that way. $0.02 FWIW

ermo:
Is Ubuntu 14.04 *seriously* a compile target?! There's been *3* LTS releases since then!

nabaco:
> @nabaco since we have a limited amount of GitHub Package storage (500Mb), we might just want to setup a repository for developer assets where we can just drop a tarball (Unix/Linux) and Zip file (Windows) to support this. Then CMake could be setup to just download the repository into a specific folder (external-assets) and pull and extract whatever versions from there. That would also give us a bit better delta comparisons for any 3rd party library we need to track that way. $0.02 FWIW

What do we achieve by that? If we're going in the direction, might as well pull the sources from the origin, without duplicating them.

BenjamenMeyer:
@nabaco we achieve having a copy that we can rely on in case something happens and the origin URL changes, goes off-line, etc. Gives us just a little more ability to reproduce throughout the years ahead. Otherwise, I fully agree just pull from the origin directly.

royfalk:
I think we need to consider boost just like any other library. Do we keep a tarball for openAL? If not, why should we for boost? 
**I do agree that we should use the system boost** and expand it's usage. I have commits in progress that use boost filesystem for file operations instead of the old hacks currently used, for example.

Loki1950:
The openAL tarball contains the windows .dll IIRC needed for building a windows installer creation

royfalk:
It's going to be a bit hard to review, as github chokes on the number of files. 
Can we assume you only edited readme and cmakelist and deleted only stuff in the attached boost library?

danielrh:
Do we have a designated person to track windows build/compat? When doing a lot of these refactors, making sure our Visual Studio builds go swimmingly could be an advantage-- with different memory allocators and better debugging mechanisms, Visual Studio has prevented Vega Strike from a number of pretty serious bugs

nabaco:
> It's going to be a bit hard to review, as github chokes on the number of files.
> Can we assume you only edited readme and cmakelist and deleted only stuff in the attached boost library?

And `sh/vsbuild.sh`

nabaco:
> Do we have a designated person to track windows build/compat? When doing a lot of these refactors, making sure our Visual Studio builds go swimmingly could be an advantage-- with different memory allocators and better debugging mechanisms, Visual Studio has prevented Vega Strike from a number of pretty serious bugs

This is still in progress, see #53 for more info. If you have some light to shed on the matter, we'll be glad to hear.

BenjamenMeyer:
@danielrh no, we don't have the Windows build working yet. Once we do, we'll try to figure out how to add it to the PR gate check

nabaco:
Conflict resolved, and since Ubuntu 14.04 LTS was removed, compilation passes.

BenjamenMeyer:
@nabaco given there are 14,000+ files changed; which ones do I need to pay attention to for a final review? README and CMakeLists.txt?

nabaco:
Yes, that's the two files that I have changed.

BenjamenMeyer:
@nabaco thanks - assuming builds work; it's approved

PR #88, opened 2020-05-02
Add base name and type to base computer header

ermo:
In Priv:WCU, it is very convenient to be able to check the name and type
of the current base when trading via the base computer, since the
computer tracks the prices seen at other in-system bases.

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -1771,11 +1771,15 @@ void BaseComputer::recalcTitle()
     if (baseUnit) {
         if (baseUnit->isUnit() == PLANETPTR) {
             string temp = ( (Planet*) baseUnit )->getHumanReadablePlanetType()+" Planet";
-            baseName = temp;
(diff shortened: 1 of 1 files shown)


stephengtuggy:
Hmm. Why the force-push, @ermo ? I am generally wary when I see one of those.

ermo:
> Hmm. Why the force-push, @ermo ? I am generally wary when I see one of those.

Typo in commit message header ('type' instead of 'typo'). That's the only difference.  I figured that since it's a short-lived branch, force-pushing would be ok.

ermo:
> Looks good to me. Can you supple a couple of screen shots how it actually looks in the game?

https://imgur.com/a/mS2RLqC (don't know how long the imgur link will work, but there you go)

nabaco:
Have you tested it with the UtCS data-set?

ermo:
> Have you tested it with the UtCS data-set?

The top image is from the UtCS data-set, so yes -- but only a cursory test where I load up the game and verify that it looks nice enough on the starting planet.

Issue #40, opened 2020-04-04
Decision on compiler/platform support


BenjamenMeyer:
Several questions raised in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/33#issuecomment-609089044:

> I'm a cutting edge kind of guy, and it got me into trouble before. I vote we ditch older environments for building but not for playing. We require new if not newest compilers so we can clean the code as much as possible but release new version binaries for older systems as well. We're an open source project, it should be fun to contribute. No one likes to maintain legacy code. If we wanted to do that, we'd spend another hour at our day job. As an aside, the code is filled with ifdefs for MS Visual C 6 (circa mid-90's!) and Solaris. I think we can drop support for these.

1. What compilers do we want to support?
- [x] gcc
- [x] clang
- [x] ~VC6~ (dropped per https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/40#issuecomment-625288625)
- [x] ~VC2020~ VS2017

2. What legacy platforms do we want to support?
- [x] ~Ubuntu 14.04~ (dropped per https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/40#issuecomment-623743970)
- [x] Ubuntu 16.04
- [x] Ubuntu 18.04 (current LTS for Ubuntu until 20.04 is released in April 2020)

3. How long should a platform be supported?

Loki1950:
First we have to get Windows build at all which will be not fun and where is the Mac OSX version yeah we did support that as well

BenjamenMeyer:
Windows support for Travis-CI is still experimental; according to the documentation they're currently working on VS2017 support. Propose we move to VS2017 as the official compiler until that matures.

NOTE: Even once the gate is in place it'll still be allowed to fail for a while as we figure it out and as Travis-CI matures it's Windows target support.

Loki1950:
Good reasoning @BenjamenMeyer IIRC VS2017 is supported by M$'s community version of VS

BenjamenMeyer:
@Loki1950  updated the description per VS 2017.

nabaco:
I suggest that for compiling, we support officially maintained distro's and officially maintained compilers.

So for Ubuntu, 14.04, 16.04, 18.04 with the GCC/Clang version that is supplied with them.
Once 14.04's support is fully dropped, we can drop the support for compilation on it too.
[...]

PR #90, opened 2020-05-04
Removed support of Ubuntu 14.04 LTS

nabaco:
Removed Ubuntu 14.04 LTS from Travis-CI, thus officially dropping support of **compilation** on that distro.

Currently this PR is a preparation/offer and will be merged after finishing the [vote on the matter](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/40#issuecomment-623743970)

@vegastrike/core-reviewers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,14 +14,6 @@ compiler:
 jobs:
     include:
         - os: linux
-          dist: trusty
-          compiler: gcc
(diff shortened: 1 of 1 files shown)


nabaco:
The vote results: 7 - for, 0 - against. I'm merging this. Worst case scenario, if we see a reason to, we can revert.

PR #102, opened 2020-05-09
README.md: -DUSE_SYSTEM_BOOST=ON is the default now

ermo:
Besides, -DSYSTEM_BOOST=ON was wrong. Two birds with one stone.

[skip ci]


nabaco:
@BenjamenMeyer we might have an issue here, since [skip ci] tag is set, but GH is waiting for the CI to complete.
We need to change either the settings in GH to allow such merges, or the setting in Travis-CI to not run on README (or any .md) files.

BenjamenMeyer:
@nabaco there is a related bug...need to review this more as I was having issues with doing a build when solely following the build instructions in the readme, though vsbuild.sh works fine.

As to CI...will have to investigate more. Travis CI might wait for all potential builds to finish before sending the pass/fail to GH, even if some are marked as ignore/skip.

BenjamenMeyer:
ahh...reading more.. yeah seems GitHub doesn't recognize the `skip ci` tag in the commit history.
Personally I'd probably just let it go through the CI process regardless.

@nabaco well, we have a choice: either request the skip ci tag be removed, or do an administrative merge. We'll have to decide on a policy for that.

@ermo can you please update against master? That might in itself resolve the issue.

Issue #97, opened 2020-05-08
Linux Desktop Integration


BenjamenMeyer:
We should provide Desktop Integration via the Free Desktop Entry Specification. This is basically a text file of the name `VegaStrike.desktop` (or similar) that we could ultimately drop in `/usr/share/applications` which would then allow Vega Strike appear on the launcher menus of the various desktop environments (KDE, GNOME, etc).

For reference see https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

nabaco:
Agreed. But I think we should start dealing with that once we figure out the packaging side of things.

pyramid3d:
https://github.com/vegastrike/Assets-Production/blob/f3309d0d63218341c0bce5bf966522b66a8866d6/vegastrike.desktop

BenjamenMeyer:
@pyramid3d thanks; we probably should have a default one with the Core Engine repo too at least as reference; asset repos can then make one specific for the games they for.

Loki1950:
Thx for putting this in my face ;) missed it's creation and as @nabaco mentioned it is something we need to do for packaging even on the way back machine having to add VS to the menu manually was a bit of a chore

BenjamenMeyer:
We now have icons and short cuts for VS and the configuration tool as part of VS:UtCS. I think we can close this out for now.

Issue #95, opened 2020-05-07
Proposal: Use boost::logs to capture log data


BenjamenMeyer:
Instead of always dumping data to the console, let's capture it to a file properly. Boost provides `boost::logs` which can fulfill this in a platform independent way.

UPDATE: As a general rule everything should go to our log file. The exceptions:
- getting the log going
- fatal errors where we crash out, in which case if we can it should go to the log but should also go to the console/std::error.

We should also force a limit on log file size and do file rotations; these probably should be part of the configuration.

nabaco:
That's sound like some of the changes that @stephengtuggy has done in the past

BenjamenMeyer:
@nabaco yes, there's probably something related to that. The current gitter discussion with trying to capture data to a log file got me thinking of this, so just wanted to capture it for the moment. If we already have something that can be re-enabled that'd be great and then we can pull this into one of the current milestones as a bug fix.

stephengtuggy:
Yeah, basically I think my `BOOST_LOG_TRIVIAL` stuff just needs to be re-enabled. That is how Vega Strike does its debug output now (as opposed to `stdout` or `stderr` output, which are separate in the current version of VS).

Perhaps a Phase 2 of this issue would be redirecting **everything** to `BOOST_LOG_TRIVIAL`. No more `stdout`/`cout`/`stderr`/`cerr`.

nabaco:
Sounds good. @stephengtuggy are you willing to undertake this endeavor?

BenjamenMeyer:
@stephengtuggy yes, I'd like to see everything go to file the boost logs and have only minimal output on stdout/cout/stderr/cerr - namely output before the logs can get configured and after they're shutdown, perhaps with info on where the logs are being stored too so people can easily find them.

stephengtuggy:
@nabaco I'll think about it.

stephengtuggy:
Related to #5

BenjamenMeyer:
@stephengtuggy #5 will have some impact certainly - though we'll have to ensure that log messages are readable enough by both users and us to be useful for actually debugging stuff too.

royfalk:
I was actually looking into BOOST_LOG_TRIVIAL as well and did some work on it before discovering everyone else is also thinking/working on it. @stephengtuggy since you started already, it's yours if you want it. Just be sure to assign the issue to yourself so we'll know.

stephengtuggy:
[...]

PR #87, opened 2020-05-01
Try to make savegame memory allocation slightly more robust

ermo:
Fixes #85 -- will likely need to use something else than vs_dprintf(), but I'll leave that to people with more C++ fu.

Looks like this fix was never merged in from the VS-taose codebase, yet its commit message is present in the ChangeLog?  What's that all about?

PR #108, opened 2020-05-10
Re-enable boost logging

### LINKED ISSUE #107: The boost::log stuff I put in got turned off
Need to turn it back on again, after figuring out how to get it to compile properly with the latest Boost. This is essential for proper VS debugging output.

Relates to #95

stephengtuggy:
So the issue with the code not compiling on the latest versions of Boost, was a real issue. I followed the guidance found [here](https://stackoverflow.com/a/55689686) to get past this.

Re-enabled the various places in the code that use `BOOST_LOG_TRIVIAL`. About 30 files in all. Re-enabled the `initLogging` function. Made a few other associated changes.

Testing steps completed:
Bought and sold cargo
Bought and sold ship upgrades
Flew between Atlantis and Serenity, docking at each multiple times
Picked a fight (and lost)
Jumped to Oldziey; landed on a rocky planet

I feel like I'm missing someone in terms of reviewers. Feel free to add anyone else who should be a reviewer on this PR.

Closes #107


Loki1950:
Will let the others review the code @stephengtuggy first but I do think it's a good idea it should let us the fix the various errors in the vs logs with more info

stephengtuggy:
Well, crap. Travis errored out. I will troubleshoot.

stephengtuggy:
Looks like Travis is fixed.

ermo:
@stephengtuggy 

I added a vs_dprintf in PR #87 -- if you want to change that to boost logging format, you're more than welcome to do so (I tried to do it myself, but it didn't compile back then so I figured I'd better leave it to the pros).

BenjamenMeyer:
Please merge master

stephengtuggy:
> BTW there is a way to increase the verbosity of the logs --verbose # where the num character is 1-5 not sure on this

I believe it is --debug=# where the num character is 1-3.

stephengtuggy:
> Hi @stephengtuggy, shouldn't we just delete all `VSFileSystem::vs_dprintf` calls?

I would agree that that would be a good next step.

> 
> Additionally, can you show the results? Where will we find the log files? How will the look?

This is what the results look like on the console. Not sure yet how to redirect to log files, sadly.

Vega Strike boost log console output screenshot 2020-05-11

stephengtuggy:
> Please merge master

Will do. It may take a little time though; there are several conflicting files now.

Loki1950:
@stephengtuggy `./vegastrike > output.log 2>error.log
`

stephengtuggy:
So I can add to the list of tests run, that I have now tested jumping as well.

When redirecting output to output.log and error.log, using a command such as the following:

```bash
./vegastrike --debug=2 >output.log 2>error.log
```

all the Boost::Log output goes to output.log.

stephengtuggy:
Sweet! Thanks for the merge, @nabaco

nabaco:
I think we should add a separate issue regarding the redirection of the logs into a standard file.

stephengtuggy:
[error.log](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/4611344/error.log)
[output.log](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/4611345/output.log)

stephengtuggy:
> I think we should add a separate issue regarding the redirection of the logs into a standard file.

Can you do that, @nabaco ?

Let me see if I understand correctly. Do we want the console output and error output **always** to go to a certain file location, instead of to the terminal? If so, would that be part of #95 ?

BenjamenMeyer:
@stephengtuggy yes, it should only go to a file location, and yes that would be part of #95.

Loki1950:
Expatiation of the earlier comment currently we create .vegastrike directly under Home but current practice is to put that folder in .config also the **save sectors serialized_xml sounds textures** should be under .local/share/.vegastrike extra work for us I know but it does seem to current practice to keep hidden files in those classifications

nabaco:
Agreedm it's according the XDG specification. This is part of #97.

ermo:
Guys, fatal errors really ought to be printed (or at least also copied) to stderr?

nabaco:
> 
> 
> > > Let me see if I understand correctly. Do we want the console output and error output **always** to go to a certain file location, instead of to the terminal? If so, would that be part of #95 ?
> > 
> > 
> > yes, it should only go to a file location, and yes that would be part of #95.
> 
> Guys, fatal errors **_really_** ought to be printed (or at least also copied) to stderr?
> 
> Not doing so tends to break the principle of least astonishment.
> 
> Bonus points for posting a short text blurb after VS has exited with an error code, where the blurb points to the latest log file.

Agreed. Let's do this under a separate issue.

BenjamenMeyer:
> > > Let me see if I understand correctly. Do we want the console output and error output **always** to go to a certain file location, instead of to the terminal? If so, would that be part of #95 ?
> > 
> > 
> > yes, it should only go to a file location, and yes that would be part of #95.
> 
> Guys, fatal errors **_really_** ought to be printed (or at least also copied) to stderr?
> 
> Not doing so tends to break the principle of least astonishment.

I would agree that would be in the exception to the rule portion. Let's update #95 for that.

Issue #50, opened 2020-04-07
Enable -Werror compilation


nabaco:
This is related to #29 and #45.

To enable this, we need to remove all the existing compilation warnings. Once we enable it, the build will fail on any new compilation warnings.
Thus, developers will deal with compilation warnings in real-time, and we'll see the failures in the CI.

ermo:
Speaking from experience, adding this by default for Release builds (which is what the distributions and platforms will build) is _no bueno_.

I would personally prefer if this was only enabled for the Debug target but left off for the other targets.

Then require have CI (and thus VS Devs, since CI builds will now gate changes) to build with `-DCMAKE_BUILD_TYPE=Debug` by default?

BenjamenMeyer:
@ermo agree - the `-Werror` is more for us; and good points as far as getting integration into the distros

I do want to _eventually_ get to where there are no errors/warnings, but I don't think we have to necessarily make all warnings errors.

Setting it for *our* CI pipeline once we get there is a probably a good compromise.

Question: do we want this for the 0.6.x release? Or make it as part of the long term goals?

ermo:
> Question: do we want this for the 0.6.x release? Or make it as part of the long term goals?

My take is "both": Make it a long term goal with a milestone deadline (e.g. 0.9) and then make it a target to get "substantial" work on it done for each preceding milestone (i.e. don't release a milestone unless it contains a subset of this work).

This should get the ball rolling while maintaining forward momentum I should think.

BenjamenMeyer:
@ermo sounds like a plan. I've been holding off on the 0.9.x target as I've wanted to get at least 0.6.x out first and hadn't identified must specific beyond 0.8.x. This might become the first issue for it. We'll have to think about some related stories to break down the larger work for how to include it into each previous release (0.7.x and 0.8.x)... it might be beyond 0.9.x depending on how much work there is but that'd be okay as long as we're progressing.

nabaco:
There actually not that much work. Currently the main warnings are:
- Unused variables/functions
- Unsafe string operations
- Class/stracture direct memory access (memset/memcopy)

BenjamenMeyer:
@nabaco agreed; though some of those might be better broken into smaller parts.

The DMA one at least is probably going to require more dedicated investigation/testing to verify things don't break. IIRC some of those warnings are due to polymorphic behaviors.

The other two should be pretty easy to clean up, as well as any required type casting, etc.

[...]

PR #112, opened 2020-05-12
Set -fno-PIE by default as a Ubuntu/Mint workaround

### LINKED ISSUE #94: Building on Mint 19.3
Having great difficulty building do not get an executable get
vegastrike: ELF 64-bit LSB shared object, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/l, for GNU/Linux 3.2.0, BuildID[sha1]=7037dafcdf95535245ab216b3e22191299d30af8, not stripped

ermo:
Older versions of Ubuntu and its derivates have an issue where the
`file` utility will interpret Position Independent Executables as
shared libraries:

https://bugs.launchpad.net/ubuntu/+source/file/+bug/1747711

As a workaround, allow setting the option `ENABLE_PIE`
(defaults to `OFF`) and add a suitable message for maintainers in both
the CMake code, during configuration runs and in ccmake.

Update the build instructions in `README.md` to reflect this change.

Additionally, switch sh/vsbuild.sh to the Debug target (see #50) and enable verbose compilation by default to make it easier to spot errors in CI.

Closes #94

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -580,6 +580,46 @@ IF(DATADIR)
 ENDIF(DATADIR)
 
 
+# On some Ubuntu versions and derivatives, a bug exists whereby enabling
+# PIE compilation (Position Independent Executables) results in the
+# `file` utility incorrectly recognising the compiled vegastrike binary
+# as a shared library instead of a position independent shared executable
+# object.
+#
+# The effect of the bug is that vegastrike can still be started from the
+# command line but that it will not be recognised as an executable by GUI
+# file managers such as Nautilus and Dolphin.
+#
+# To avoid this scenario, turn off this flag by default and let packagers
+# on other distributions turn this on if their OS is able to correctly deal
+# with Position Independent Executables.
+
+# For more info, see:
+# - https://bugs.launchpad.net/ubuntu/+source/file/+bug/1747711
+# - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/94
+#
+UNSET(CMAKE_POSITION_INDEPENDENT_CODE)
+UNSET(ENV_CFLAGS)
+UNSET(ENV_CXXFLAGS)
+SET(ENV_CFLAGS "${CMAKE_C_FLAGS}")
+SET(ENV_CXXFLAGS "${CMAKE_CXX_FLAGS}")
+OPTION(COMPILE_WITH_FPIE
+  "Enable Position Independent Executables/Shared Libraries (NOT RECOMMENDED on Ubuntu/Mint)"
+  OFF)
+IF(COMPILE_WITH_FPIE)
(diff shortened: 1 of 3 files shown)


Loki1950:
@ermo you might add Arch to the distros that enable `COMPILE_WITH_FPIE` they where the first to do so I think

engine/CMakeLists.txt
```diff
@@ -650,6 +650,36 @@ ELSE("${CMAKE_BUILD_TYPE}" STREQUAL "Profiler")
 ENDIF("${CMAKE_BUILD_TYPE}" STREQUAL "Profiler")
 
 
+# On some Ubuntu versions and derivatives, a bug exists whereby enabling
+# PIE compilation (Position Independent Executables) results in the
+# `file` utility incorrectly recognising the compiled vegastrike binary
+# as a shared library instead of a position independent shared executable
+# object.
```
BenjamenMeyer, line 657: Might add text about it effecting operation from GUI interfaces, but accessible from command line.

Not going to block on it.

BenjamenMeyer:
not sure what is up with the builds; seems GCC/G++ doesn't like something but Clang is happy.

engine/CMakeLists.txt
```diff
@@ -650,6 +650,42 @@ ELSE("${CMAKE_BUILD_TYPE}" STREQUAL "Profiler")
...
+  message("!! Enabling Position Independent Executables/Shared Libraries (NOT RECOMMENDED on Ubuntu/Mint) !!")
+  #SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+  SET(CMAKE_C_FLAGS "-fPIC -fPIE" ${CMAKE_C_FLAGS})
+  SET(CMAKE_CXX_FLAGS "-fPIC -fPIE" ${CMAKE_CXX_FLAGS})
+ELSE(COMPILE_WITH_FPIE)
+  message("++ Disabling Position Independent Executables/Shared Libraries (Recommended on Ubuntu/Mint)")
+  #add_compile_options(-fno-PIE)
+  #add_compile_options(-fno-PIC)
+  SET(CMAKE_C_FLAGS "-fno-PIC -fno-PIE" ${CMAKE_C_FLAGS})
+  SET(CMAKE_CXX_FLAGS "-fno-PIC -fno-PIE" ${CMAKE_CXX_FLAGS})
```
BenjamenMeyer, line 685: You probably have the quote in the wrong place

```
 SET(CMAKE_C_FLAGS "-fno-PIC -fno-PIE ${CMAKE_C_FLAGS}")
 SET(CMAKE_CXX_FLAGS "-fno-PIC -fno-PIE ${CMAKE_CXX_FLAGS}")
```

BenjamenMeyer: These should get set on any call to the compiler (e.g GCC, G++, Clang); https://cmake.org/cmake/help/v3.10/envvar/LDFLAGS.html should get set on the linker calls.

ermo:
PR message updated -- ready for merging on my end.

PR #119, opened 2020-05-21
Remove redundant definitions from gnuhash.h

nabaco:
While working on windows compilations I saw that we have some redundant code in `gnuhash.h`, as the code is defind int `<functional>`

At first I thought to just delete the file and replace all place that include it with `#include <functional>` but it didn't work due to defines and `hash_compare`. So I'm doing this as a midstep, but I think we should consider merging this with some other header.

### DIFF
--- a/engine/src/gnuhash.h
+++ b/engine/src/gnuhash.h
@@ -2,63 +2,22 @@
 #define _GNUHASH_H_
 #include "config.h"
 #include <unordered_map>
+#include <functional>
 
 #define vsUMap     std::unordered_map
 #define vsHashComp std::hash_compare
 #define vsHash     std::hash
-class Unit;
 
 namespace std
 {
-template < >
-class hash< void* >
-{
-    hash< size_t >a;
-public:
-    size_t operator()( const void *key ) const
-    {
-        return a( (size_t) key );
-    }
-};
-template < >
-class hash< const void* >
-{
-    hash< size_t >a;
-public:
-    size_t operator()( const void*const &key ) const
-    {
-        return a( (size_t) key );
-    }
-};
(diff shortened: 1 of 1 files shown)


engine/src/gnuhash.h
```diff
@@ -2,63 +2,22 @@
...
-template < >
-class hash< std::pair< Unit*, Unit* > >
-{
-    hash< size_t >a;
-public:
-    size_t operator()( const std::pair< Unit*, Unit* > &key ) const
-    {
-        return (size_t) (size_t) ( a( (int) ( ( (size_t) key.first )>>4 ) )
-                                  ^a( (int) ( ( (size_t) key.second )>>4 ) ) );
-    }
```
BenjamenMeyer, line 52: Since this is operating on the Unit this should probably be kept, though perhaps defined in Unit.

stephengtuggy:
Hmm. I don't much like the idea of changing our hash algorithms. Might break something unforeseen.

engine/src/gnuhash.h
```diff
@@ -2,63 +2,22 @@
...
 
 #define vsUMap     std::unordered_map
 #define vsHashComp std::hash_compare
 #define vsHash     std::hash
-class Unit;
 
 namespace std
 {
-template < >
-class hash< void* >
```
stephengtuggy, line 14: Based on some warnings I'm seeing in our build logs, it looks like maybe `hash` is supposed to be a `struct`, not a class. I wonder if that would fix the problem?

stephengtuggy:
@nabaco Is this PR still needed? If not, could it perhaps be closed? Thanks

PR #120, opened 2020-05-21
Removed config.h includes due to redundancy

nabaco:
Since we have -include config.h as a compiler parameter, the compiler adds automatically the header into each file it compiles.
The presence of this includes breaks windows compilation.


nabaco:
Got it working on windows. The MSVC options is [/FI](https://docs.microsoft.com/en-us/cpp/build/reference/fi-name-forced-include-file?view=vs-2019) - Force Include

Issue #14, opened 2020-02-15
Missing Contributors in Master branch


crts-crome:
I have mentioned this issue in the forum previously, not sure whether it has been noticed or not. After the merge of the SVN repository the new Master branch does not show the Contributors any more. The contributed code changes are still present but the files themselves show 0 Contributors when selected.
Is this going to be fixed at some point?

crts-crome:
It is not just my contributions, by the way, e.g., the changes from #7 are also no longer attributed to the contributor in the Master branch.

pyramid3d:
Answered in the forum

BenjamenMeyer:
@pyramid3d can you link the answer here?

crts-crome:
Here is the link to [the answer](https://forums.vega-strike.org/viewtopic.php?f=53&t=26104&p=169930#p169930).

However, the explanation is not entirely correct, specifically this part:

> ... Dunno why they were not included when I merged the code again into master ...

My changes were, in fact, merged back into the master branch and they are still there. I am just not listed as the contributor of those changes. Apparently, during the great merge the repository got squashed, i.e., its entire history is gone. 
This is what the [current history](https://github.com/vegastrike/Vega-Strike-Engine-Source/commits/master/engine/src/cmd/unit_generic.cpp) of the affected file looks like.

I still have the version [before the great merge](https://github.com/crts-crome/Vega-Strike-Engine-Source/commits/master/src/cmd/unit_generic.cpp) in my repository.

The history of all other files has also been cut short.

Another excerpt from the answer states

> ... klauss mentioned that the bug was already fixed by him ...

Klauss said, that he remembers fixing [something similar](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/12#issuecomment-544969625). I went through the file's (unit_generic.cpp) history in the SVN repository. I did not find any changes that might have fixed that bug.

BenjamenMeyer:
> @BenjamenMeyer Here is the link to [the answer](https://forums.vega-strike.org/viewtopic.php?f=53&t=26104&p=169930#p169930).
> 

Thanks I'll give it a read.

> However, the explanation is not entirely correct, specifically this part:
> 
> > ... Dunno why they were not included when I merged the code again into master ...
> 
[...]

PR #125, opened 2020-05-24
Enhancement: Auto-generating the AUTHORS file

BenjamenMeyer:
Functionality to auto-generate the AUTHORS file per #14. I'll make another for 0.6.x as the AUTHORS file as generated against current `master`.


nabaco:
I think we can integrate that into our CD once we set it up

PR #129, opened 2020-05-25
Pull Request Template

### LINKED ISSUE #68: Define validation procedure for Pull Requests
I suggest that we define a validation procedure for Pull Requests, that will present a basic acceptance criteria for changes based on tests that the CI can't make.
I'm still new to the VS universe to say all the details, but I'd that it'd be something like:
- Start with ship A in location XYZ. Fly to N.
- Start with ship B in location ___. Make a fight.
- Fly into a nebula.
- Etc.

It should be a set of short tests that will test different aspects of the game, widely enough to make sure that a change hasn't broken anything drastic.
I would even say, maybe we can create a set of savefiles, to make the process easier.

If you think that doing that per PR is too much, we can still use that per release.

BenjamenMeyer:
- an initial pull request template

Closes #68


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.
+
+Please answer the following:
+
+Code Changes:
+- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
```
nabaco, line 6: Maybe we should list some of the things from the Wiki here?

Loki1950: go for it

BenjamenMeyer: I'd rather them not get out of sync which is why I just listed the wiki article as a reference.

Issue #91, opened 2020-05-07
Strange velocity 'bouncing' behaviour with SPEC


Calendis:
Platform: Linux
Using the latest master (vssetup identifies as 0.6.0)

Sometimes (about 50% of the time? It's hard to say), when using Auto-pilot to approach or leave a major object, the ship's velocity will not behave properly. It will rapidly 'bounce' between seemingly random velocities from the hundreds range to the tens of thousands. During this, the positions of major objects also seem to bounce, but maybe this is only an illusion due to the rapid velocities involved?

This will continue to happen even after auto is turned off, making it essentially impossible to control your vessel. The effect can sometimes be calmed by switching to another target, but it will come back. This is while auto pilot is off, so targeting different objects shouldn't affect my control. I have tested using "default", "realistic", and "classic" physics, and it can occur on all three modes.

Flying close to a major object using autopilot and then switching targets seems especially likely to trigger this. I can provide additional information if needed, or record a video

nabaco:
Hi Calendis,
Thank you for the report. We have seen this issue too. Could you please attach logs to the issue, after issuing the following command?
`vegastrike > output.log 2>error.log`

A few questions
1. Does this happen while targeting any object? I have seen it happen when flying to a jump point sarrounded by asteroids, and thought it might be related due to the complicated maneuver the autopilot tries to do.
2. Did you try to do a manual SPEC flight? I think I had the same issue flying manually too, so it might not be related to autopilot.
3. In case it happens in manual SPEC flight too, does it happen in long regular flight?

Meanwhile, we're investigating the issue too.

Loki1950:
Deja voodoo very old behaviour any gravity well affects SPEC and the behaviour can seem random since I have been unable to build an executable on Mint 19.3 I can not give a informed option but it has been a comment about SPEC that is long standing

stephengtuggy:
I have seen this same bug. Was planning to report it; just hadn't gotten my thoughts together yet. Thanks, @Calendis .

Loki1950:
Well i am seeing the expected behaviour used auto from starting planet orbit Atlantis set ship to proper orientation on Serenity base toggled auto SPECed straight to the base dropping out of auto 2500 m from base hit "Home" to match velocity asked for docking clearance docked with no issues saved game now to hack the save file for mega cash and continue testing

Loki1950:
[...]

PR #150, opened 2020-06-23
Revert "Refactor (Q)Vector"

BenjamenMeyer:
This reverts a commit identified by @nabaco that is linked to issue #91 in some form.

Verified:
- [x] it compiles
- [x] game is playable
- [x] the issue is at least reduced to more historic levels

Game Play Verification:
- launched using a saved game that put me on an uninhabited planet in the origin system
- jumped to the Everett System
- Tried to land on an astroid mine base.

Commits reverted:
- 04f99ab64535fc0146a90ac9e226d71cbbac0095

Previously I've done this and could never even get close to the base.
More verification is needed; request the reviewers also do a game play test to verify.

For the 0.6.x release it is acceptable to revert the commits; however, we want to find a proper fix in master.


royfalk:
I can confirm the issue is only present after my QVector refactor. I'm looking into it. Meanwhile, I suggest we revert.

PR #157, opened 2020-07-03
Fix resume crash on 0.6.x

stephengtuggy:
Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [ ] This is a documentation change only

Issues:
Closes #28 

Purpose:
- What is this pull request trying to do? **Make it so that users can pause and resume the game freely, without crashes.**
- What release is this for? **0.6.0**
- Is there a project or milestone we should apply this to? **0.6.x**

PR #161, opened 2020-07-06
Fix resume crash on master

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Fix the equivalent issue to #157 , only on the `master` branch.**

Purpose:
- What is this pull request trying to do? **Make Pause and Resume work properly on the `master` branch**
- What release is this for? **0.7.x and following**
- Is there a project or milestone we should apply this to? **0.7.x?**

### DIFF
--- a/engine/src/lin_time.cpp
+++ b/engine/src/lin_time.cpp
@@ -113,15 +113,22 @@ void setTimeCompression( float tc )
 bool toggle_pause()
 {
     static bool paused = false;
+    fprintf(stderr, "toggle_pause() called in lin_time.cpp\n");
     if (paused)
     {
+        fprintf(stderr, "toggle_pause() in lin_time.cpp: Resuming (unpausing)\n");
         setTimeCompression(1);
         paused = false;
     }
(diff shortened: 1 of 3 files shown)


engine/src/lin_time.cpp
```diff
@@ -113,15 +113,22 @@ void setTimeCompression( float tc )
 bool toggle_pause()
 {
     static bool paused = false;
+    fprintf(stderr, "toggle_pause() called in lin_time.cpp\n");
```
BenjamenMeyer, line 116: fine with me; do we wnat to send the to the Boost Log?

stephengtuggy: @BenjamenMeyer Yes. Duh. All I had to do to make that work was `#include "vsfiilesystem.h"`. I was forgetting.

Do you think we need to run the play tests again?

BenjamenMeyer: Wouldn't hurt, but probably doesn't need to be as extensive

Issue #142, opened 2020-06-07
Python 3 test fails on some platforms


nabaco:
On some platforms CMake configuration fails due to Python dependency.

When running normaly, it expects python2, but for some reason it links against python3 and compilation fails.

On `-DUSE_PYTHON_3=ON` it fails, claiming to not find python. To overcome this, I have to replace in a single place from `python3` to `python`.

I will post logs tomorrow.

nabaco:
@ermo will you be willing to have a look at this?

BenjamenMeyer:
Linux distros are moving over to Py3 by default. Arch and a few others jumped early; guessing that might be part of what you're running into (based on gitter chat).

FWIW - So CMake use to have a mailing list for support, which was always very helpful. Apparently they transitioned to solely using discourse - https://discourse.cmake.org/. Haven't logged in there yet, but they're devs were always helpful.

BenjamenMeyer:
A few questions:
- [ ] what all platforms are we seeing this on?
- [ ] is Py2 installed on those platforms?
- [ ] what is the default Python version on those platforms?

nabaco:
I'm seeing this on Arch and MacOS. For the latter you can see here:
https://travis-ci.org/github/vegastrike/Vega-Strike-Engine-Source/jobs/695768054

For Arch I still need to collect the logs and upload.

In the case of Arch, Python2 is installed alongside Python3. Regarding default, I'm not sure, need to check.

BenjamenMeyer:
@nabaco Yeha, I think Arch is Py3 by default. `/usr/bin/python` should point to the default Python on any platform. If you want a Py2 or Py3 then use `python2` or `python3` respectively.

nabaco:
OK, agreed about Arch. But what about MacOS?

BenjamenMeyer:
Keep in mind were only continuing to support Py2 for a few more months. We need to get 0.6.x and 0.7.x out and then hit dropping Py2 hard to quickly follow with 0.8.x.

Until then I'd suggest we do a best effort approach and work as quickly as we can to get the tasks done to get through those releases.

evertvorster:
Hi there... 
I am actually in the process of writing the AUR for vegastrike, and I'll give you 1 guess as to why I am here. 
XD

[...]

PR #163, opened 2020-07-07
Correct boost::python component specification wrt boost 1.67+

s0600204:
#### Purpose
As of `boost` `1.67`, the naming of `python` components now have version-specific suffixes (e.g. `python27`, `python34`), instead of distribution-specific prefixes (e.g. `python3`, `python`).

See:
 * https://www.boost.org/doc/libs/1_73_0/libs/python/doc/html/rn.html
 * https://cmake.org/cmake/help/latest/module/FindBoost.html
 * https://repology.org/badge/vertical-allrepos/boost.svg?minversion=1.67
   - (All repos listed that are not bright red are affected by this)

This change should be backward compatible with earlier versions of `boost`.

#### Code Changes
- [x] Build-system modification; no game-play alteration.

Tested on ArchLinux with `python` (`2.7`, `3.8`), `boost` (`1.72`)

- [x] Compiles and runs; `boost` >= `1.67` with `python` `2.7`
- [x] Compiles and runs; `boost` >= `1.67` with `python` `3`+
- [ ] Compiles and runs; `boost` < `1.67` (Travis should confirm it compiles. Hopefully.)

#### Related Issues
- #142

#### Suggested Meta
- *Release*: 0.6.x+
- *Project*: "Coming Up To Snuff"
- *Milestone*: "0.6.x Release"


engine/CMakeLists.txt
```diff
@@ -710,30 +710,34 @@ ENDIF (PYTHONLIBS_FOUND OR PYTHON_LIBRARIES AND PYTHON_INCLUDE_DIRS)
...
-        FIND_PACKAGE(Boost COMPONENTS python log log_setup REQUIRED)
-    ENDIF (USE_PYTHON_3)
-
+    FIND_PACKAGE(Boost)
     IF (Boost_FOUND)
         MESSAGE("++ Found System Boost version : ${Boost_VERSION}")
+
+        # With boost version 1.67 and above, python components are specified with a version suffix
+        # (`python27`, `python38`) instead of their invocation name (`python`, `python3`)
+        if (Boost_VERSION VERSION_GREATER_EQUAL 1.67) # >= 1.67
```
nabaco, line 719: Capitalize `if` as per our CMAKE coding conventions

nabaco:
After merging, we'll need to cherry-pick this into 0.6.x branch

s0600204:
I might be wrong, but I think the reason why the MacOS builds fail can be found at line 926 of that build log:
> This formula provides Boost.Python for Python 2. Due to a
> collision with boost-python3, the CMake Config files are not
> available. Please use -DBoost_NO_BOOST_CMAKE=ON when building
> with CMake or switch to Python 3.

In other words, CMake complains (at line 963) that it can't find any suitable configuration files... because they aren't being installed. :shrug:

nabaco:
Ah, good! So we could add something like
```IF (XCODE)
    SET(Boost_NO_BOOST_CMAKE ON)
ENDIF(XCODE)```

Either way, we can do that in a different PR.

**Comment by BenjamenMeyer** [comment]
@nabaco adding a py3 build that's allowed to fail is fine with me

File: engine/CMakeLists.txt [hunk 2]
```diff
@@ -18,7 +18,7 @@
 # > cmake -DCMAKE_BUILD_TYPE=Debug <path to source tree>
 #====================================
 
-CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.8)
```
BenjamenMeyer, line 21: Is this supported on all our platforms? I think we kept it at 2.6 for Ubuntu 14.04.

s0600204: I bumped it up because `PYTHONLIBS_VERSION_STRING` (which was already in use prior to this PR) requires at least 2.8.8, according to the relevant CMake [documentation page](https://cmake.org/cmake/help/latest/module/FindPythonLibs.html).

A general overview of CMake versions across various linux/bsd distributions (that have it in their repositories) can be seen [here](https://repology.org/badge/vertical-allrepos/cmake.svg?minversion=2.8.8) (all bright red are below 2.8.8). As to MacOS or Windows... well that's more up to the user.

As to Ubuntu 14.04, specifically:
* According to [repology](https://repology.org/project/cmake/versions), 14.04 "Trusty" has two versions available: 2.8.12 and 3.5.1 (trusty/main and trusty-updates respectively)
  - (The official Ubuntu packages site no longer lists Trusty)
  - (Note that 12.04 - the previous LTS - has 2.8.7)
* I was under the impression that this project had officially dropped support for 14.04 in #90?

Loki1950: Nice to see you here @s0600204 glad you have the time to help.

BenjamenMeyer: @s0600204 I was noting the history. I'd love to bump it up to 3 even, just have to be aware of supporting platforms. If 2.8.8 or later is on all our supported platforms then great, let's bump it up.

Thanks!!!

evertvorster:
This may clear up the issue for Arch Linux as well, as I have been using something very similar in my patches of the CMakefile.txt to get it to work on Arch. Rejoice!

ermo:
Really happy to see someone actually fixing my half-arsed attempt at getting py2 and py3 working!

Issue #82, opened 2020-04-30
Python2 is deprecated -- ensure that VS can (optionally) be built against Python3 out of the box


ermo:
Just tried building VS against python3 using both the included boost-1.72 and the system boost (1.66 in my case on Solus).

For the included boost-1.72 I get the following error:
```
[ 50%] Linking CXX executable vegaserver
/usr/bin/ld: cannot find -lboost_python
```

For the system boost-1.66 I get the following errors:
```
[ 63%] Linking CXX executable vegaserver
/usr/bin/ld: CMakeFiles/vegaserver.dir/src/cmd/script/director_server.cpp.o: in function `PyInit_Base':
director_server.cpp:(.text+0x1e): undefined reference to `boost::python::detail::init_module(PyModuleDef&, void (*)())'
/usr/bin/ld: director_server.cpp:(.text+0x49): undefined reference to `boost::python::detail::init_module(PyModuleDef&, void (*)())'
/usr/bin/ld: CMakeFiles/vegaserver.dir/src/cmd/script/director_server.cpp.o: in function `PyInit_Briefing':
director_server.cpp:(.text+0xa6): undefined reference to `boost::python::detail::init_module(PyModuleDef&, void (*)())'
/usr/bin/ld: director_server.cpp:(.text+0xd1): undefined reference to `boost::python::detail::init_module(PyModuleDef&, void (*)())'
/usr/bin/ld: libengine_com.a(unit_wrapper.cpp.o): in function `PyInit_VS':
unit_wrapper.cpp:(.text+0x86): undefined reference to `boost::python::detail::init_module(PyModuleDef&, void (*)())'
/usr/bin/ld: libengine_com.a(unit_wrapper.cpp.o):unit_wrapper.cpp:(.text+0xb1): more undefined references to `boost::python::detail::init_module(PyModuleDef&, void (*)())' follow
/usr/bin/ld: /usr/lib/libboost_python.so: undefined reference to `PyString_Size'
/usr/bin/ld: /usr/lib/libboost_python.so: undefined reference to `PyUnicodeUCS4_FromEncodedObject'
/usr/bin/ld: /usr/lib/libboost_python.so: undefined reference to `PyFile_FromString'
/usr/bin/ld: /usr/lib/libboost_python.so: undefined reference to `PyString_Type'
/usr/bin/ld: /usr/lib/libboost_python.so: undefined reference to `PyInt_Type'
[...]

Issue #78, opened 2020-04-21
Figure out MacOS compilation


nabaco:
Continuation of #42.

BenjamenMeyer:
0.6.0 release not dependent on this; would be nice but it can be 0.6.1 if needed.

s0600204:
Hope no-one objects, but after playing around with Travis I feel like I've made a little progress, so here's my findings thus far...

Firstly, Travis's `XCode 9.4`[^1] image seems to be a no-go... at least with regards to `python 3`:

The image appears to come with the homebrew packages `python@2` (providing `python 2.7.7`[^2]) and `python` (providing `python 3.7.5`) preinstalled. During the early stages of our build process, the brew `python@3.8` gets downloaded and installed. Whilst the installation itself doesn't fail, Homebrew complains the "`brew link` step failed" due to a conflict with the `python@2` package. Thus, later on, when CMake is looking for a `3.x` version of `python`, it finds the `3.7` version, but not the `3.8`. Which causes a problem because `boost-python3` needs `3.8`.

`XCode 9.4` with `python 2` works a little better, although there's a lot of [repeated complaints](https://travis-ci.com/github/s0600204/Vega-Strike-Engine-Source/jobs/359105169#L1319) about boost itself being newer than it expects.

I also tried `XCode 11.3` and `XCode 12`[^3], and these worked even better. Just. I did encounter the same problem as mentioned [here](https://github.com/mikepurvis/ros-install-osx/issues/73), which fortunately also provided a workaround.

The next hurdle to overcome appears to be [`error: typedef redefinition`](https://travis-ci.com/github/s0600204/Vega-Strike-Engine-Source/jobs/359083043#L580).

----

(OSX-only build file: [.travis.yml](https://github.com/s0600204/Vega-Strike-Engine-Source/blob/osx/.travis.yml))

[^1] - `XCode 9.4` is the default `osx_image` Travis uses if you don't specify one.
[^2] - Oddly enough, `OSX 10.13` supposedly [comes with](https://opensource.apple.com/release/macos-1013.html) `python 2.7.10`.
[^3] - Whilst the `XCode 9.4` image is based on `OSX 10.13`, the `XCode 11.3` and `12` images are based on `OSX 10.14` and `10.15` respectively.

nabaco:
Good progress @s0600204! Thanks for having a look at this

s0600204:
You're welcome.

[...]

PR #166, opened 2020-07-08
Enhanced .travis.yml

nabaco:
Code Changes:
- CI Only changes

Issues:
- #78 
- #82 
- #163 

Purpose:
- What is this pull request trying to do? **Extend CI coverage**
- What release is this for? **None right now**

* Notify Gitter room only about builds in the official repo
* Add Python3 build
* Further attempt to progress MacOS build

MacOS related changes should be further enhanced when #163 will be merged.


.travis.yml
```diff
@@ -11,6 +10,10 @@ compiler:
     - gcc
     - clang
 
+env:
+    - {}
+    - FLAGS='-DUSE_PYTHON_3=ON'
```
BenjamenMeyer, line 15: This would seem to either:
- add Python3 to all builds
- add a second set of builds that are Py3 but still required to pass for the PR to pass the gate.

I think the better solution right now would be to add a single build, perhaps Ubuntu 18.04 or 20.04, to check the Py3 functionality and allow it to fail.

.travis.yml
```diff
@@ -23,6 +26,10 @@ jobs:
           name: "Ubuntu 16.04 LTS - Clang"
         - os: osx
           osx_image: xcode11.3
+          env: FLAGS='-DUSE_PYTHON_3=OFF -DBoost_NO_BOOST_CMAKE=ON'
+        - os: osx
+          osx_image: xcode11.3
+          env: FLAGS='-DUSE_PYTHON_3=ON'
```
BenjamenMeyer, line 32: This is okay since it's OSX and all OSX builds are allowed to fail.

.travis.yml
nabaco: This expands the matrix so that each build (in the matrix only) either compiles regularly or with Python3.
result is:
- Bionic + gcc
- Bionic + clang
- Bionic + gcc + py3
- Bionic + clang + p3
- Xenial + gcc
- Xenial + clang

s0600204:
(The MacOS Py3 build will most likely need the `boost-python3` package.)

.travis.yml
BenjamenMeyer: image

Yes, we should have Xenial Py3 too at minimum; not sure why it only expanded the Bionic ones. We probably should add Focal (20.04) too.

nabaco:
@s0600204 thanks. I'll investigate MacOS compilation further once we'll merge this and your PR

.travis.yml
Loki1950: I tend to follow Ben on this

BenjamenMeyer: Note: Focal isn't currently available, so ignore that part.

.travis.yml
nabaco: So what's the bottom line for this PR? Add Xenial compilation with + py3 and we're good to go?

.travis.yml
nabaco: So what's the bottom line for this PR? Add Xenial + Py3 and we're good to go?

nabaco:
@s0600204 you were right. But I'd like to note that with py2 on MacOS we managed to pass configuration stage and proceed to compilation, which is good progress

.travis.yml
BenjamenMeyer: @nabaco sorry for the delay - xenial + py3 as an optional build that doesn't need to pass yet, just like we do for Mac.

nabaco: Done. Let's see the CI to make sure I haven't messed up anything

PR #172, opened 2020-07-09
Boost log to file

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Further implements #95** 

Purpose:
- What is this pull request trying to do? **Have Boost logging output go to a file; convert more stdout, stderr, and logging messages to go through Boost.Log; log only fatal errors to the console. (I was going to try to finish converting the entire project, but that appears impractical to do within this PR.)**
- What release is this for? **0.7.0?**
- Is there a project or milestone we should apply this to? **0.7.x?**


engine/src/cmd/basecomputer.cpp
```diff
@@ -4113,43 +4103,36 @@ void trackPrice(int whichplayer, const Cargo &item, float price, const string &s
...
         {
             for (size_t i = 0; i < recordedLowestPrices.size(); ++i) {
+                                                                    // POSIX-printf style
                 BOOST_LOG_TRIVIAL(info) << boost::format("    %1% : %2$.2f") % i % recordedLowestPrices[i];
-                // VSFileSystem::vs_dprintf(1, "    %d : %.2f\n", i, recordedLowestPrices[i]);
             }
         }
 
-        fflush(stderr);
+        // TODO: Find a way to flush the Boost logs? Console log specifically?
```
BenjamenMeyer, line 4135: part of the boost log configuration IIRC

stephengtuggy: @BenjamenMeyer yes, you can set the log sinks to auto-flush after every line of output. I have done that with the console logger, but my thinking is that that will be too much overhead with the file logger; it will slow down the game too much. So we will need a way to manually flush the file log sink when needed.

Luckily, I now have a way to do that:

```C++
VSFileSystem::pFileLogSink->flush();
```

I just haven't updated this file yet to use it.

stephengtuggy: Thanks for the review, by the way.

stephengtuggy:
OK, I think this one is ready for review. I was going to try to finish converting the entire project. But other people are working on the same files, of course, and so much is shifting around that that goal seems pretty impractical at this point.

I did get the Boost logging to go to a file by default -- once it has reached the stage of game startup where it is able to. And I believe I converted quite a few of the most common and annoying messages, so they won't show up on the console anymore.

What do you guys think of the log location I chose? `$HOME/.vegastrike/logs/`. Since this isn't a system program, that seemed like a good location. But maybe I'm wrong.

The file logs include a timestamp in the filename. They rotate at 10 MiB, or at midnight. Logging will stop if free space on the volume drops to 1 GiB.

Any other questions?

evertvorster:
@stephengtuggy Thanks for the hard work that you put in.
If you don't mind, I'll also use the .vegastrike/logs directory to keep console logs of the stuff that is not in boost. Currently I am using files in .vegastrike. As you mentioned, this is not system software, so /var/log seems a bit far away. 

I am a little concerned about the size of the logs. If you have a rather large /home, this would mean that the files will never be deleted. Is this something you turn on, or is the boost logging on by default? 

For the console logs, a backup of the last log is made, and then the new log is written in it's place. This gives us a log rotation of 2, keeping the size of the thing under control. How far back in history do you want to go? It may be worth considering a log rotation scheme of 10, and compressing the backups.

stephengtuggy:
Sorry, one more commit is on the way. Need to make sure to flush the logs before exit, wherever possible.

stephengtuggy:
> @stephengtuggy Thanks for the hard work that you put in.

You're welcome. :-)

> If you don't mind, I'll also use the .vegastrike/logs directory to keep console logs of the stuff that is not in boost. Currently I am using files in .vegastrike. As you mentioned, this is not system software, so /var/log seems a bit far away.
> 

Sure, that's fine.

> I am a little concerned about the size of the logs. If you have a rather large /home, this would mean that the files will never be deleted. Is this something you turn on, or is the boost logging on by default?
> 

The boost logging is on by default, but by default, it only logs warning, error, and fatal messages. This means that during some runs of the game, no log file will be created at all. In others, only one or two messages will be logged.

> For the console logs, a backup of the last log is made, and then the new log is written in it's place. This gives us a log rotation of 2, keeping the size of the thing under control. How far back in history do you want to go? It may be worth considering a log rotation scheme of 10, and compressing the backups.

That is worth considering. However, I don't know how to do that with Boost::Log. I'm not actually sure if their system supports that without some customization. Maybe a future PR can deal with that.

stephengtuggy:
There. NOW it's ready.

stephengtuggy:
> Hi @stephengtuggy, good job. I think we can merge it, and add some documentation on the week how to use this infrastructure (maybe as art of our coding style) so future developers will know how to use it.

Thanks.

Usage instructions remain the same as before. There is at least a preliminary usage guide on the wiki [here](https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/How-to-convert-a-VSFileSystem::vs_dprintf()-call-to-the-Boost-logging-way).

stephengtuggy:
> we can merge once a play test has been confirmed

I have play tested it. Do you mean you want someone else to as well?

BenjamenMeyer:
@stephengtuggy it would be nice if we could have a secondary person do play tests...but that's a lot of extra work so as long as we have someone willing to take ownership of the play test and affirm it I'm good.

evertvorster:
I'll be playing it now, just rebuilding my game engine. :)

PR #179, opened 2020-07-16
Fixes the empty name in the dynamically generated universe xml that i…

evertvorster:
…s saved in the home directory

Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [X] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [ ] This is a documentation change only

Issues:
- Please list any related issues
The dynamically generated universe is saved in the home directory under sectors. 
In some solar systems, there is units with an empty name. This fix fills that empty variable with a random name. 

There is still Unit naming weirdness going on that is unrelated to this fix, as the Python scripts also add units to the universe, but they are not saved in the sectors sub directory. I am still on the hunt for those, but as they reside in Assets-Production, I'll make a PR on that repository when I have them sorted.

Purpose:
- What is this pull request trying to do?
Fix empty names for some installations in the universe.
- What release is this for?
Works on all of them. 
- Is there a project or milestone we should apply this to?
Nope.

PR #182, opened 2020-07-19
Fixes the starbase names in a slightly better way

evertvorster:
This one replaces #179

With #179 WriteUnit was modified to just write a random name into the name field. 
Technically, this is not correct, as it takes a name as a variable, and should use it. So, leaving this one alone. 

MakeSmallUnit used to call WriteUnit with no name, and used the "nam" variable to get the unit type. 
MakeSmallUnit now calls WriteUnit with a name and type. 

The whole reason that I updated my fix is that there might be another function that calls WriteUnit, and expects it to handle the "nam" variable correctly

Please answer the following:

Code Changes:

Issues:
This is a better fix for the same problem that #179 was made for. Just to be clear, I don't want #179 to be applied anymore.

Purpose:
It fixes the names for installations that galaxy_gen.cpp makes. 

- What release is this for?
Any, sooner is better than later, I suppose. 
Nope.


stephengtuggy:
I will want to test this, but it looks good to me so far.

evertvorster:
After all the approvals & checks, what's the process for getting this one merged?

stephengtuggy:
Merged.

Issue #183, opened 2020-07-20
Nav computer does not display path.


evertvorster:
The navigation computer does not display the set path. 

As far as I can tell, it has three modes:
Path Off: In this mode it should display all the systems & paths that it knows about, but not highlight a path. 
Path On: In this mode it should display everything from the previous mode, but highlight the path. 
Path Only: In this mode it should only display the path, and not the rest of the universe. 

When I make a path with the path computer, and switch between the modes, I get the following three screens:
One screen with the known universe. 
One screen with the known universe, with the path not visible
One screen with nothing. 

This leads me to believe that the path display function is not making this path visible.
The module that runs this screen is NavigationSystem::DrawGalaxy() in engine/src/gfx/nav/drawgalaxy.cpp

So far I've been picking at this one for about a day, but can't keep this juicy bug all to myself.

PR #184, opened 2020-07-20
Fixes the missing colors from the Navigation Computer, and Galaxy view

evertvorster:

Found the core issue is that the pathcolor was not initialized. I picked brilliant white, with no alpha blending to make it really stand out.

Please answer the following:

Code Changes:

Issues:
Issue #183 

To see it in action, "Shift+M" for your nav computer. Click on some other system that you are not in, and select "select target" Now cycle through: Path off/on/only to see the path. 

I am still trying to wrangle the navigation computer extended functions to make sense to me. In theory, you should be able to make a path with it as well, and it should show up in the Galaxy display. However, this it outside the scope of this small bugfix.

Purpose:
Just make it easier to navigate the universe. 
- What release is this for?
It compiles in any branch. 
ASAP, it adds a little polish to the game.


evertvorster:
Looking at this one a little more, the fix works, but it probably is not the proper fix.

The variable "pathcol" is defined in navscreen.cpp, and referenced in navscreen.h. 
navscreen.h is in the includes of drawgalaxy.cpp. 

The reason I am mentioning this is that there are other parts of the navigation computer that also seems to suffer from un-initialised color variables. 

What is missing, why does this not work like it used to? Is there a more elegant fix?

BenjamenMeyer:
@evertvorster I assume you are referring to the colors at https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/nav/navscreen.h#L200? If so, then proper place to initialize them to values other than the default set by `GFXColor` is to initialize them in https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/nav/navscreen.cpp#L59.

evertvorster:
Hi there, Benjamin. 
https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/14c5dc0cbcb082f415232c65e07db722e2e3f8b0/engine/src/gfx/nav/navscreen.cpp#L343
Is where the colors variables are defined & initialized. 

In https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/14c5dc0cbcb082f415232c65e07db722e2e3f8b0/engine/src/gfx/nav/drawgalaxy.cpp#L949 is the first time that variable is getting used, but it's not initialized. Why is that? Do I need to initialize them again for drawgalaxy.cpp?

evertvorster:
With this one I am now initializing the variables in each of navscreen.cpp and drawgalaxy.cpp
It seems the initial issue was that it is trying to read these variables from a config file, and they simply are not in the config file. So, initializing them to their backup values. 

Even when initializing the colors in navscreen.cpp, the colors still did not show until I moved the initialization of the colors in navscreen.cpp to the function they are being used in.
Lastly, removed the initialization of pathcol from navscreen.cpp, as my compiler complained about initializing a variable that is not being used.

BenjamenMeyer:
@evertvorster https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/nav/navscreen.cpp#L59 is what is called the Class Constructor. It's for initializing any class variables and doing anything required for setting up the class. So yes, it's where they should be initialized.

Now if it's still giving an issue, then check the config function. It might be that the config function is setting a value without checking if the value is in the config; so then it's overwriting it with the default value from the object. Instead it should leave the value alone.

Basically in pseudo-code:
```
if config.has_option(option_name) then:
   option = config.get_option_value(option_name)
end if
```

Python does this easily if the config is stored/presented as a dictionary:
```python
# value =  dict.get(name, default_value)
option = config.get(option_name, option)
```

I'm not sure yet where it's loaded; but hopefully that will give you some pointers.

evertvorster:
@BenjamenMeyer What I don't understand is that when I initialized the variables in that class constructor, it still did not work. 

It's only when I moved the initialization of the variables to the functions that they are used in that the colors came back in the nav computer. What would cause those values to not be propogated?

evertvorster:
@BenjamenMeyer So, I did a few more tests. 
---test 1---
I made the variables in the config file, and recompiled the engine without any of my changes. 
The colors were still messed up in the nav computer. 

---test 2---
Then, I directly assigned the colors to the variables in the function that they are in now, and recompiled the game. 
Still no change, colors are all messed up. 

---test 3---
I even moved the block of color definitions to the function that starts at line 59 in navscreen.cpp like you suggested, and still no change. The colors are still all messed up. 

Is it really such a bad thing to have the colors defined directly, in two places? The fix works, and the default colors are OK.

BenjamenMeyer:
@evertvorster the tests you did are quite interesting in their result. There has to be something else in play that we're not seeing.

> Is it really such a bad thing to have the colors defined directly, in two places? The as it stands now, this PR works, and the default colors are OK.

Generally you want to limit where things are changed/defined. In this case:
- the GFXColor should provide a default in case something isn't set
- the class constructor should define different default colors for each instance
- the configuration should override the values of the class constructor

We shouldn't really have anything else changing those values unless we provide the user an interface for setting them (which we don't). That said, as a temporary fix - then go ahead and do it; just make a comment about a bigger fix being needed as I'm sure we'll track it down as we refactor stuff. Perhaps wrap that value in a config option so we can disable and test during refactors.

engine/src/gfx/nav/navscreen.cpp
```diff
@@ -340,15 +340,7 @@ void NavigationSystem::Setup()
...
 
-    //Get special colors from the config
-    static GFXColor currentcol = vs_config->getColor( "nav", "current_system",
-                                           GFXColor( 1, 0.3, 0.3, 1.0 ) );
-    static GFXColor destinationcol = vs_config->getColor( "nav", "destination_system",
-                                           GFXColor( 1, 0.77, 0.3, 1.0 ) );
-    static GFXColor selectcol = vs_config->getColor( "nav", "selection_system",
-                                           GFXColor( 0.3, 1, 0.3, 1.0 ) );
-    static GFXColor pathcol = vs_config->getColor( "nav", "path_system",
-                                           GFXColor( 1, 0.3, 0.3, 1.0 ) );
```
BenjamenMeyer, line 351: these should probably be here, but not have the `static GFXColor` portion; that basically creates a new variable that then gets dropped when it leaves scope.

BenjamenMeyer: @evertvorster the above issue might be why you were having issues with the tests you did - not sure, just a guess

evertvorster: Let me test your theory.... if it works, then I know where to put them in the config file as well.

evertvorster:
Just checked it out, and it works beautifully. Thanks for the help, @BenjamenMeyer

engine/src/gfx/nav/navscreen.cpp
```diff
@@ -41,6 +41,7 @@
 #include "universe_util.h"
 #include "in_mouse.h"
 #include "gui/glut_support.h"
+#include "networking/netclient.h"
```
BenjamenMeyer, line 44: I believe this file isn't in master any more; it might still be in 0.6.x

evertvorster: I'm removing it now. I was building against 0.6.0, as that is what is playable for me...

PR #186, opened 2020-07-22
vsfilesystem.cpp: Fixes potential path buffer overflows

### LINKED ISSUE #180: Two potential buffer overflows in VSFileSystem::InitDataDirectory()
`char` array `tmppath` is size `16384`. So the maximum length string it can store is 16383 characters. (Leaving room for the null terminator.)

Thus, each of the two calls below to `getcwd()` should be:

```C++
getcwd( tmppath, 16383 )
```

... or better yet, the equivalent using a constant for the buffer size.

Instead, both instances use `16384`. An off-by-one error, it seems to me.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Fixes #180** 

Purpose:
- What is this pull request trying to do? **Fix the potential buffer overflows in vsfilesystem.cpp**
- What release is this for?
**0.7.0, and possibly 0.6.0 as well.**

### DIFF
--- a/engine/src/vsfilesystem.cpp
+++ b/engine/src/vsfilesystem.cpp
@@ -52,7 +52,8 @@ int VSFS_DEBUG()
     return 0;
 }
 char *CONFIGFILE;
-char  pwd[65536];
+const size_t VS_PATH_BUF_SIZE = 65536;
+char  pwd[VS_PATH_BUF_SIZE];
 VSVolumeType isin_bigvolumes = VSFSNone;
 string curmodpath = "";
 
@@ -114,10 +115,10 @@ std::string vegastrike_cwd;
 void ChangeToProgramDirectory( char *argv0 )
 {
     {
-        char pwd[8192];
+        char pwd[VS_PATH_BUF_SIZE];
         pwd[0]    = '\0';
-        if (NULL != getcwd( pwd, 8191 )) {
-            pwd[8191] = '\0';
+        if (NULL != getcwd( pwd, VS_PATH_BUF_SIZE - 1 )) {
+            pwd[VS_PATH_BUF_SIZE - 1] = '\0';
             vegastrike_cwd = pwd;
         } else {
             VSFileSystem::vs_fprintf( stderr, "Cannot change to program directory: path too long");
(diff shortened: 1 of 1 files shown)


stephengtuggy:
Should I submit another PR to fix the same issue on the 0.6.x branch?

stephengtuggy:
Note: During the fight test, towards the end when my ship was extremely damaged, the sky was intermittently filled with white light, fading to blue, then normal again. That was strange. But I don't think it was introduced with this pull request.

Perhaps just an artifact that can happen during fights under certain circumstances, or something.

BenjamenMeyer:
@stephengtuggy since it's a bug, you can if you want. I do want to start limiting what we accept for 0.6.x to only what we need to make the release and any release related infrastructure.

stephengtuggy:
@BenjamenMeyer Sure thing.

nabaco:
Good one @stephengtuggy, fixed a bug and removed all those magic numbers.

I would like to question whether 65536 is the correct number to use. See the following article for more detail.

We might want to consider adding `<linux/limits.h>` instead, but only under a Linux only #ifdef, as to not break Windows compilation further.

https://www.systutorials.com/maximum-allowed-file-path-length-for-c-programming-on-linux/

nabaco:
Regarding windows, it seems that <windows.h> supplies the same define:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamea

nabaco:
Regarding windows, it seems that <windows.h> supplies the same define:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamea

stephengtuggy:
@nabaco Hmm. 65536 (64K) isn't necessarily the right number. However, in theory at least, Vega Strike is supposed to support other operating systems too. Mac OS X for sure, and I have also seen references to IRIX and BeOS in the code.

nabaco:
I'm fine with merging this, and issuing a ticket to make all such places in the code based on platform specific defines, rather than magic numbers, if that's preferred by you guys

stephengtuggy:
We could use one of, or a combination of, the POSIX-compliant solutions listed here: https://stackoverflow.com/questions/9449241/where-is-path-max-defined-in-linux

Or we could just compromise on a smaller value, hardcoding, say, 4096 for example.

stephengtuggy:
> I'm fine with merging this, and issuing a ticket to make all such places in the code based on platform specific defines, rather than magic numbers, if that's preferred by you guys

Okay, sure. Let's do that.

PR #193, opened 2020-07-23
Fixed Travis CD

nabaco:


nabaco:
After we make sure it works correctly :)

PR #194, opened 2020-07-23
Fix/bouncing bug

### LINKED ISSUE #91: Strange velocity 'bouncing' behaviour with SPEC

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
Fixes #91 

Purpose:
- What is this pull request trying to do? Fix the bouncing bug!
- What release is this for? 0.7.0

### DIFF
--- a/engine/src/cmd/unit.cpp
+++ b/engine/src/cmd/unit.cpp
@@ -381,7 +381,7 @@ void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &par
     if (!cam_setup_phase) {
         // Following stuff is only needed in actual drawing phase
         if (this->cloaking > this->cloakmin) {
-            cloak = (int) (this->cloaking-interpolation_blend_factor*this->pImage->cloakrate * SIMULATION_ATOM /*simulation_atom_var*/ );
+            cloak = (int) (this->cloaking-interpolation_blend_factor*this->pImage->cloakrate * simulation_atom_var );
             cloak = cloakVal( cloak, this->cloakmin, this->pImage->cloakrate, this->pImage->cloakglass );
         }
         if (this->hull < this->maxhull) {
@@ -398,7 +398,7 @@ void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &par
             QVector camerapos = camera->GetPosition();
 
             float minmeshradius =
-                ( camera->GetVelocity().Magnitude()+this->Velocity.Magnitude() ) * SIMULATION_ATOM; //simulation_atom_var;
+                ( camera->GetVelocity().Magnitude()+this->Velocity.Magnitude() ) * simulation_atom_var;
 
             unsigned int numKeyFrames = this->graphicOptions.NumAnimationPoints;
             for (i = 0, n = this->nummesh(); i <= n; i++) {
@@ -470,19 +470,22 @@ void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &par
             double backup = interpolation_blend_factor;
             int    cur_sim_frame = _Universe->activeStarSystem()->getCurrentSimFrame();
             for (un_iter iter = this->getSubUnits(); (un = *iter); ++iter) {
-                //float sim_atom_backup = simulation_atom_var;
-                /*if (this->sim_atom_multiplier && un->sim_atom_multiplier) {
+                float sim_atom_backup = simulation_atom_var;
+                if (sim_atom_backup != 1) {
+                    BOOST_LOG_TRIVIAL(debug) << boost::format("void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &parentMatrix ): sim_atom as backed up != 1: %1%") % sim_atom_backup;
+                }
+                if (this->sim_atom_multiplier && un->sim_atom_multiplier) {
                     //BOOST_LOG_TRIVIAL(trace) << boost::format("void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &parentMatrix ): simulation_atom_var as backed up  = %1%") % simulation_atom_var;
                     simulation_atom_var = simulation_atom_var * un->sim_atom_multiplier / this->sim_atom_multiplier;
                     //BOOST_LOG_TRIVIAL(trace) << boost::format("void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &parentMatrix ): simulation_atom_var as multiplied = %1%") % simulation_atom_var;
-                }*/
+                }
                 interpolation_blend_factor = calc_blend_factor( saved_interpolation_blend_factor,
                                                                 un->sim_atom_multiplier,
                                                                 un->cur_sim_queue_slot,
                                                                 cur_sim_frame );
                 (un)->Draw( *ct, *ctm );
 
-                //simulation_atom_var = sim_atom_backup;
+                simulation_atom_var = sim_atom_backup;
             }
             interpolation_blend_factor = backup;
         }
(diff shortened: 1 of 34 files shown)


engine/src/gfx/tvector.cpp
```diff
@@ -39,7 +39,7 @@ float _sqrt(float f) {
 
 template <>
 double _sqrt(double d) {
-  return sqrtf(d);
```
BenjamenMeyer, line 42: That would make a difference

stephengtuggy: Yep

stephengtuggy:
This code may be messy, but it does seem to fix the bouncing bug. What do you guys think? Should I clean it up a little bit more, or shall we merge it in like this?

More testing **is** still needed, to see if too many physics calculations pile up on top of each other with the "Current computer" setting. Especially in the "Sol" system. If so, we might need to run the game at a slightly lower setting.

That reminds me. When you change the "Current computer" setting to a lower setting in vegasettings, it doesn't seem to comment out the corresponding simulation_atom setting correctly in the config file. So I am attaching my config file, in case it helps you test.

stephengtuggy:
[vegastrike_config_SGT_2020-07-24.zip](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/4974676/vegastrike_config_SGT_2020-07-24.zip)

stephengtuggy:
During my play testing, the camera zoomed in and out suddenly on the background a few times, but that could have been explainable by passing ships interfering with SPEC/the autopilot. None of the objects or other units bounced back and forth that I noticed.

engine/src/cmd/unit_util_generic.cpp
```diff
@@ -78,8 +78,8 @@ bool hasDockingUnits( const Unit *my_unit )
 
 int getPhysicsPriority( Unit *un )
 {
-    static const bool FORCE_TOP_PRIORITY = XMLSupport::parse_bool(
-        vs_config->getVariable( "physics", "priorities", "force_top_priority", "false" ) );
+    static const bool FORCE_TOP_PRIORITY = true; //XMLSupport::parse_bool(
+        //vs_config->getVariable( "physics", "priorities", "force_top_priority", "false" ) );
```
BenjamenMeyer, line 82: This could probably just `bool FORCE_TOP_PRIORITY` or `const bool FORCE_TOP_PRIORITY` for its type; it's only used in this function. `static` is unnecessary.

Not a blocker, just noting it.

engine/src/cmd/unit.cpp
```diff
@@ -381,7 +381,7 @@ void GameUnit< UnitType >::Draw( const Transformation &parent, const Matrix &par
     if (!cam_setup_phase) {
         // Following stuff is only needed in actual drawing phase
         if (this->cloaking > this->cloakmin) {
-            cloak = (int) (this->cloaking-interpolation_blend_factor*this->pImage->cloakrate*simulation_atom_var);
+            cloak = (int) (this->cloaking-interpolation_blend_factor*this->pImage->cloakrate * SIMULATION_ATOM /*simulation_atom_var*/ );
```
BenjamenMeyer, line 384: I assume all the `/*simulation_atom_var*/` is just documentation about `SIMULATION_ATOM`

stephengtuggy: No, it's an alternative variable with the same initial value, but one that looks like maybe it's meant to be changed -- unlike the one in ALL_CAPS.

engine/src/cmd/unit_util_generic.cpp
stephengtuggy: If it was just `bool FORCE_TOP_PRIORITY`, it would probably read the setting correctly, after vs_config is initialized properly with the values from the file. Maybe I should just do that.

engine/src/cmd/unit.cpp
stephengtuggy: These are places where I was using the lowercase one in an intermediate commit.

BenjamenMeyer: @stephengtuggy okay - so then let's either:
- wrap these in a #define so to switch between them, or 
- if not needed clean them up

so we don't confuse people

engine/src/lin_time.cpp
```diff
@@ -149,7 +150,8 @@ void micro_sleep( unsigned int n )
 
 #elif defined (IRIX)
 
-#include <unistd.h>
+// Now included above, for all operating systems except Windows
+// #include <unistd.h>
```
BenjamenMeyer, line 154: yes we need to clean a lot of this up; feel free to just drop it out entirely.

NOTE: FTR Only time a header should be included lower in a file is if there's a define that changes it as a result, even then we should try to keep as much of that at the top as possible.

evertvorster:
I have done some play testing on this PR, too. 
I'd say this one cures the bouncing bug, and puts my play testing on the master branch. I will of course do more play testing, but I have visited a handful of bases all over the Universe, including Sol/Sol, and the game performs beautifully.

stephengtuggy:
> I have done some play testing on this PR, too.
> I'd say this one cures the bouncing bug, and puts my play testing on the master branch. I will of course do more play testing, but I have visited a handful of bases all over the Universe, including Sol/Sol, and the game performs beautifully.

Awesome.

What simulation_atom setting did you use in your tests? Did you use the vegastrike.config I provided, with the lower/older setting, or did you use the "Current computer" setting, as vegasettings refers to it?

stephengtuggy:
So I just tested test/9k_idle_1k_fighting.mission, with simulation_atom 0.01 ("Current Computer"), resolution 3800x2000 (near 4K), and --debug=3, no less. Everything seemed to work fine.

I think this PR is ready to merge.

stephengtuggy:
Hold on, I take that back.

Loki1950:
Glad you still looking for the underling reason.

stephengtuggy:
Well, guys. This latest patch is looking pretty promising. It passes the "bouncing" test; it passes the 9k_idle_1k_fighting test; and it appears to pass soltest as well.

Let me know what you think.

engine/src/lin_time.cpp
stephengtuggy: Dropped entirely.

engine/src/cmd/unit.cpp
stephengtuggy: I believe `simulation_atom_var` is the correct one to use in most cases, so I have made that change.

engine/src/cmd/unit_util_generic.cpp
stephengtuggy: Took out the `const` clause. Now it reads its value correctly from the config file.

evertvorster:
I have played with it some as well, and I like how smooth everything is. Have not noticed anything untoward while approaching quite a few bases.

stephengtuggy:
The major changes are:

- In lin_time.cpp, uses `clock_gettime` to get the system time, where this function is available
- Corrects a couple of minor potential precision issues with the (Q)Vector refactor
- Switches to using `simulation_atom_var` (variable naming convention) instead of `SIMULATION_ATOM` (constant naming convention) in most places
- Disables randomization of physics priority (per-unit sim_atom), controlled by a compile-time `#define`
- Gives all system installations a fixed physics priority, defaulting to 3 (planets, asteroids, bases, and such)
- Changes various `static const` declarations of config file variables to just `static`, so that they will actually pick up their values from the vegastrike.config file. One of these is force_top_priority, which forces all units/objects to run at priority 1 when set.
- Makes a few other changes that mean that individual units will change priority less often

PR #202, opened 2020-07-27
Fix/Bug 154

stephengtuggy:
Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [x] This is a **build and** documentation change only

Issues:
- Please list any related issues
Fixes #154 

Purpose:
- What is this pull request trying to do? Fix the manual build process, as well as the instructions for performing said manual build.
- What release is this for? 0.7.x
- Is there a project or milestone we should apply this to? 0.7.x

PR #207, opened 2020-07-30
Fix version numbers for boost

evertvorster:
This PR is purely to fix the boost detection so that the package compiles on Arch Linux. 

The version numbers for boost does not exist as it was defined, this would then cause CMake to look for python3_boost, which does not exist on Arch Linux systems, and fail compilation. 
Python3_boost does exist on other systems, and it would give the appearance of working properly on other distros when in fact it was failing to choose the proper path and falling back to "python3"

This python2/3 thing has been the source of many headaches, as implementations differes across distros, with some updating "python" to mean python3, and others staying with "python" meaning "python2" for compatibility reasons. 

The way forward for specifying which boost version to use is to use both the mayor and minor python version on the system in conjunction with boost, as this is the same across all distros. 

More about it here:
https://cmake.org/cmake/help/v3.15/module/FindBoost.html


evertvorster:
I'm trying a few different edits here on the CMakeLists.txt file to see if I can make it pass compilation on all the targets

evertvorster:
Sorry for the many commits, I'm just trying to make it compile on all the targets that we care about. However, the last commit is a bit of an ugly hack.

evertvorster:
OK, I have banged my head on the table enough on this one. 
The builds we care about passes with this last edit. 

There is even a build that passes now that did not when the changes were put in. (LTS Xenial with Python 3)

However, none of the MacOS builds pass, and they have not passed when these changes were put in either. 

I can see on the Python2 branch of the MacOS that python & boost are detected and configured properly. It then fails on something else. 
The Python 3 branch fails to find the python3 boost, as it looks like the mayor and minor version numbers were not found.

s0600204:
Hmm, my apologies... the CMake documentation does not state (on the relevant page) that `VERSION_GREATER_EQUAL` requires CMake v3.7 or greater; if it had, I most likely would not have used it in #163.

Anyway:
* The `VERSION_*` comparators expect the version numbers to be in the form `X.Y.Z` ([link](https://cmake.org/cmake/help/latest/command/if.html)).
* Until CMake v3.15, `Boost_VERSION` is in the same form as it appears in the `boost/version.hpp` file (e.g. `106700`) ([link](https://cmake.org/cmake/help/v3.14/module/FindBoost.html?highlight=boost_version)).

As such, I'd like to suggest the following alternate approach to this issue:
```diff
diff --git a/engine/CMakeLists.txt b/engine/CMakeLists.txt
index a6ac62634..bf8659877 100644
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -715,18 +715,23 @@ IF (USE_SYSTEM_BOOST)
     IF (Boost_FOUND)
         MESSAGE("++ Found System Boost version : ${Boost_VERSION}")
 
+        # VERSION_EQUAL / VERSION_GREATER requires a version in the form X.Y.Z
+        # Until CMake v3.15, Boost_VERSION is in the form as written in boost/version.hpp (e.g. 106700)
+        # Thus, build the version from parts, which should work across all supported CMake versions
+        SET(BOOST_DERIVED_VERSION "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
+
         # With boost version 1.67 and above, python components are specified with a version suffix
         # (`python27`, `python38`) instead of their invocation name (`python`, `python3`)
-        IF ((Boost_VERSION VERSION_EQUAL 106700) OR (Boost_VERSION VERSION_GREATER 106700))
+        IF ((BOOST_DERIVED_VERSION VERSION_EQUAL 1.67) OR (BOOST_DERIVED_VERSION VERSION_GREATER 1.67))
             STRING(REPLACE "." ";" PYTHONLIBS_VERSION_LIST ${PYTHONLIBS_VERSION_STRING})
             LIST(GET PYTHONLIBS_VERSION_LIST 0 PYTHON_VERSION_MAJOR)
             LIST(GET PYTHONLIBS_VERSION_LIST 1 PYTHON_VERSION_MINOR)
             SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
         ELSEIF (USE_PYTHON_3)
                 SET(BOOST_PYTHON_COMPONENT "python3")
-        ELSE ((Boost_VERSION VERSION_EQUAL 106700) OR (Boost_VERSION VERSION_GREATER 106700))
+        ELSE ((BOOST_DERIVED_VERSION VERSION_EQUAL 1.67) OR (BOOST_DERIVED_VERSION VERSION_GREATER 1.67))
                 SET(BOOST_PYTHON_COMPONENT "python")
-        ENDIF ((Boost_VERSION VERSION_EQUAL 106700) OR (Boost_VERSION VERSION_GREATER 106700))
+        ENDIF ((BOOST_DERIVED_VERSION VERSION_EQUAL 1.67) OR (BOOST_DERIVED_VERSION VERSION_GREATER 1.67))
 
         UNSET(Boost_${BOOST_PYTHON_COMPONENT}_FOUND)
         MESSAGE("++ Looking for System Boost::python")
```

----

The OSX py3 build fails where it does because the `boost-python3` package is not installed.

stephengtuggy:
@s0600204 I took **still** another approach here: https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/202 . I used [CMake Policy 0093](https://cmake.org/cmake/help/latest/policy/CMP0093.html#policy:CMP0093) to make newer versions of CMake report the Boost version in 106700 format, the way the older CMake versions did. Then I compare the resulting Boost version to 106700.

I believe It works because the minor and patch versions are optional. (They're listed in square brackets `[]` in the [link](https://cmake.org/cmake/help/latest/command/if.html) you referenced.)

s0600204:
Polite reminder (from *your* referenced link):
> **Note:** The `OLD` behavior of a policy is [`deprecated by definition`](https://cmake.org/cmake/help/latest/manual/cmake-policies.7.html#manual:cmake-policies(7)) and may be removed in a future version of CMake.

stephengtuggy:
@s0600204 True. Your way probably _is_ better.

Do you mind if I incorporate it into #202 ?

s0600204:
Sure, I've no objection.

stephengtuggy:
@s0600204 I incorporated a slightly adapted version of your fix into #202 . It's working great so far. Thank you very much.

PR #210, opened 2020-07-31
In basecomputer.cpp, convert all instances of sprintf to use boost::format instead

### LINKED ISSUE #168: basecomputer.cpp uses memory-unsafe sprintf, in multiple places
See https://stackoverflow.com/questions/3662899/understanding-the-dangers-of-sprintf .

Alternatives include:
- `snprintf`
- `boost::format` (which I think I would prefer)

### LINKED ISSUE #216: Use-after-free getting stardate in basecomputer.cpp
In basecomputer.cpp, there's a line that gets the current stardate, as follows:

```C++
const char *stardate = _Universe->current_stardate.GetFullTrekDate().c_str();
```

Unfortunately, by the time the computer finishes evaluating the full expression on the right, it throws away the result of c_str(). So we are left with `stardate` as a dangling pointer. Which may or may not hold the same value when it is used further down to set the title in the Trade Cargo and other in-game computer screens.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Fixes #168 .
Fixes #216 .**

Purpose:
- What is this pull request trying to do? **Use best practices for memory safety when formatting strings.**
- What release is this for? **0.7.x?**
- Is there a project or milestone we should apply this to? **@BenjamenMeyer ?**

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -1778,8 +1778,7 @@ void BaseComputer::configureCargoCommitControls( const Cargo &item, TransactionT
 
         //Total price display.
         const double   totalPrice   = item.price*maxQuantity;
-        char tempString[2048];
-        sprintf( tempString, "Total: #b#%.2f#-b", totalPrice );
+        std::string tempString = (boost::format("Total: #b#%1$.2f#-b") % totalPrice).str();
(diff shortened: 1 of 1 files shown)


BenjamenMeyer:
Dont see anything obvious from my phone

PR #211, opened 2020-08-01
Correct boost_python3 component on non-Debian systems

s0600204:
### Purpose
In #202, the following change was made:

```diff
         ELSEIF (USE_PYTHON_3)
-            SET(BOOST_PYTHON_COMPONENT "python3")
+            SET(BOOST_PYTHON_COMPONENT "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")   # This may be specific to Ubuntu xenial
```

Whilst this works fine for Debian and Debian-based distros (e.g. Ubuntu, Mint) - at least those that still use a version of `Boost` below `1.67` - this will have broken `python 3` builds for most, if not all, distros that are *not* based on Debian (and are *also* running a `Boost` version lower than `1.67`).

The attached changeset resolves this, putting in a specific conditional for the one distro (Xenial) that actually needs the `boost-python` specifying in such a manner for the `python 3` build to work, and restoring the old component specification for everyone else.

- **What release is this for?** Any
- **Is there a project or milestone we should apply this to?** master / 0.6.x

### Code Changes
- [x] Build System change only, no gameplay alteration
- [ ] Passes Travis CI
    - [x] [Fork](https://travis-ci.com/github/s0600204/Vega-Strike-Engine-Source/builds/178026856)
    - [ ] [PR](https://travis-ci.org/github/vegastrike/Vega-Strike-Engine-Source/builds/713908487)

### Related Issues
- #142

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -764,7 +764,21 @@ IF (USE_SYSTEM_BOOST)
         IF (Boost_1_67_Or_Later_Result)
             SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
         ELSEIF (USE_PYTHON_3)
-            SET(BOOST_PYTHON_COMPONENT "python-py${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")   # This may be specific to Ubuntu xenial
+            SET(BOOST_PYTHON_COMPONENT "python3")
+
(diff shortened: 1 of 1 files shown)


stephengtuggy:
Well, so far so good. This doesn't break anything on Ubuntu 16.04 xenial; Ubuntu 18.04 bionic; Ubuntu 20.04 focal; or Debian 10 buster. @evertvorster can you try it out on your system?

evertvorster:
Hi there. 
This PR has no effect on the Arch based systems, as we have new enough python and boost to avoid that ELSEIF statement. That is, when the libraries are detected correctly. However, this is a good fallback. 

So, Looks Good To Me.

PR #213, opened 2020-08-02
CPack: Rename vsUTCS => Vega-Strike

stephengtuggy:

Please answer the following:

Code Changes:
- [x] **Packaging change only**

Issues:

Purpose:
- What is this pull request trying to do? Name the .deb packages appropriately
- What release is this for? 0.6.0
- Is there a project or milestone we should apply this to? 0.6.0 Release

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1,14 +1,6 @@
 #====================================
 # @file   : CMakeLists.txt
-# @version: 2020-07-27
-# @created: 2008-06-08
-# @author : safemode
-# @author : Anth0rx
-# @author : pyramid
-# @author : vegastrike development team
 # @brief  : cmake configuration file for VegaStrike
-# @usage  : cmake dir-with-this-file
-# @param  : none
 #====================================
 # create a build directory somewhere outside of this source tree
 # > cd .. && mkdir build && cd build
@@ -1132,6 +1124,7 @@ MARK_AS_ADVANCED(FORCE
 )
 INSTALL(TARGETS vegastrike DESTINATION bin)
 
+SET(CPACK_PACKAGE_NAME "Vega-Strike")
 SET(CPACK_PACKAGE_DESCRIPTION_SUMMARY "3D OpenGL spaceflight simulator")
 SET(CPACK_PACKAGE_VENDOR "Vega Strike")
 SET(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/../README.md")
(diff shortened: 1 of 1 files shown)


engine/CMakeLists.txt
```diff
@@ -1132,6 +1124,7 @@ MARK_AS_ADVANCED(FORCE
 )
 INSTALL(TARGETS vegastrike DESTINATION bin)
 
+SET(CPACK_PACKAGE_NAME "Vega-Strike")
```
BenjamenMeyer, line 1127: Not sure if we want to do this on 0.6.x...but we'll certainly want to on master....
Then again....since it's a new release we might as well go ahead and do so, no?

evertvorster: The package name change does not bother Arch Linux at all as this part of the code is not used by the packaging tools. Once we've settled down on a name, then I can rename the packages. Since it's a bit of a pain, I would like to do it only once, so I hope no one minds if the Arch packages are still called vegastrike-git and vegastrike-engine-git for the time being?

The vegasettings/vssetup below... have you changed the name of the setup executable that gets installed, too? The vssettings script calls vegasettings... I wanted to make some small adjustments to the scripts in any case, like moving the log directory... I might go for name changes as well. "vs" apparently, has been grabbed by Visual Studio on MS. Since we're the little fish here we'll think of something else. :)

stephengtuggy: > Not sure if we want to do this on 0.6.x...but we'll certainly want to on master....
> Then again....since it's a new release we might as well go ahead and do so, no?

Yeah, my thinking is that if we wait to do the name change, that will only cause unnecessary confusion for users when upgrading from 0.6.x to a later release.

stephengtuggy: > The vegasettings/vssetup below... have you changed the name of the setup executable that gets installed, too? 

@evertvorster vegasettings is still called vssetup on the 0.6.x branch.

BenjamenMeyer: @evertvorster agreed for Arch. Certainly use `vegastrike` for the engine; and `vegastrike-utcs` (upon the coldest sea) or something similar for our Asset Production package. No need for you to wait on that at least for the `.git` related packages.

@stephengtuggy  yeah, I had the same thought which is part of why I approved it.

On the one hand, 0.6.x should be a simple update of 0.5.x with minimal changes so the package name should probably be the same; on the other hand, it's been years since 0.5.x was released and we're going to try to move at a faster pace, especially this year so confusion is more likely to abound.

so let's just go with it and rename it.

BenjamenMeyer: oh, and I also see that @Loki1950 approved it.

BenjamenMeyer:
@stephengtuggy please make sure this goes to master as well, and go ahead and release another alpha with the updated package names.

stephengtuggy:
Updated alpha released. Pull request to master in progress.

PR #217, opened 2020-08-02
vsfilesystem.cpp: Fix potential path buffer overflows on 0.6.x

### LINKED ISSUE #180: Two potential buffer overflows in VSFileSystem::InitDataDirectory()

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Fixes #180 on the 0.6.x branch**

Purpose:
- What is this pull request trying to do? **Fix the two path buffer overflows that are (theoretically) possible on the 0.6.x branch**
- What release is this for? **0.6.0**
- Is there a project or milestone we should apply this to? **0.6.x**


stephengtuggy:
What do you think, @BenjamenMeyer ? Is it worth it to backport this fix to the 0.6.x branch? Or should we focus our efforts on master at this point?

BenjamenMeyer:
@stephengtuggy ahh...I knew I recognized this one... to answer your question...I could go either way. We're only doing alpha releases right now so we could push it in. If it fixes something, then certainly do so. Otherwise we'll hopefully have a release of 0.7.x out soon enough it's not a big deal if we don't.

stephengtuggy:
I think I'll go ahead and push it in.

Issue #226, opened 2020-08-05
Compiler (gcc 10.1.0) warning in cmd/ai/hard_coded_scripts.cpp


evertvorster:
[  3%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/ai/hard_coded_scripts.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/ai/hard_coded_scripts.cpp: In function 'void Takeoff(Order*, Unit*)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/ai/hard_coded_scripts.cpp:892:51: warning: enum constant in boolean context [-Wint-in-bool-context]
  892 |     un->Fire( (weapon_info::LIGHT || weapon_info::MEDIUM), false );
      |                                                   ^~~~~~
[  3%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/ai/ikarus.cpp.o

Looks like it's complaining about the boolean operator being used on enums.

stephengtuggy:
Hmm. I'm not sure what this one is supposed to be.

Issue #233, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/cmd/drawable.cpp


evertvorster:
[  1%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/drawable.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp: In member function 'bool Drawable::DrawableInit(const char*, int, Flightgroup*, const char*)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:30:5: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
   30 |     if(animationExt)
      |     ^~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:33:2: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
   33 |  std::vector< Mesh* > *meshes = new vector<Mesh *>();
      |  ^~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp: In member function 'void Drawable::AnimationStep()':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:120:9: warning: this 'if' clause does not guard... [-Wmisleading-indentation]
  120 |         if(nextactiveMesh >= vecAnimations.at(activeAnimation)->size())
      |         ^~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:123:13: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the 'if'
  123 |             if(loopCount > 0)
      |             ^~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:103:13: warning: unused variable 'numvold' [-Wunused-variable]
  103 |         int numvold = 0;
      |             ^~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:104:13: warning: unused variable 'numvertices' [-Wunused-variable]
  104 |         int numvertices = 0;
      |             ^~~~~~~~~~~
[  1%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/movable.cpp.o

Hmm... this one looks fairly easy. :)

[...]

PR #265, opened 2020-08-06
Fixes the two enums being compared to booleans

evertvorster:

Please answer the following:

Code Changes:

Issues:
#226  - enum constant in boolean context
#233 - enum constant in boolean context

Purpose:
Clear two compiler errors
- What release is this for?
Any
Nope

### DIFF
--- a/engine/src/cmd/ai/hard_coded_scripts.cpp
+++ b/engine/src/cmd/ai/hard_coded_scripts.cpp
@@ -889,7 +889,8 @@ void Takeoff( Order *aisc, Unit *un )
     ord = ( new Orders::FaceTargetITTS( 0, 3 ) );
     AddOrd( aisc, un, ord );
     un->SelectAllWeapon( false );
-    un->Fire( (weapon_info::LIGHT || weapon_info::MEDIUM), false );
+    un->Fire( (weapon_info::LIGHT), false );
+    un->Fire( (weapon_info::MEDIUM), false );
     TurnTowards( aisc, un );
 }
 
--- a/engine/src/cmd/ai/hard_coded_scripts.cpp
+++ b/engine/src/cmd/ai/hard_coded_scripts.cpp
@@ -889,8 +889,7 @@ void Takeoff( Order *aisc, Unit *un )
     ord = ( new Orders::FaceTargetITTS( 0, 3 ) );
     AddOrd( aisc, un, ord );
     un->SelectAllWeapon( false );
-    un->Fire( (weapon_info::LIGHT), false );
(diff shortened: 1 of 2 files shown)


engine/src/cmd/ai/hard_coded_scripts.cpp
```diff
@@ -889,7 +889,8 @@ void Takeoff( Order *aisc, Unit *un )
     ord = ( new Orders::FaceTargetITTS( 0, 3 ) );
     AddOrd( aisc, un, ord );
     un->SelectAllWeapon( false );
-    un->Fire( (weapon_info::LIGHT || weapon_info::MEDIUM), false );
```
stephengtuggy, line 892: I think this is supposed to be the bitwise OR operator `|`, rather than the logical OR operator `||`. I.e.,

```C++
un->Fire( (weapon_info::LIGHT | weapon_info::MEDIUM), false );
```

engine/src/cmd/unit_generic.cpp
```diff
@@ -5930,7 +5930,7 @@ Vector Unit::MountPercentOperational( int whichmount )
     return Vector( mounts[whichmount].functionality,
                   mounts[whichmount].maxfunctionality,
                   ( (mounts[whichmount].status == Mount::ACTIVE || mounts[whichmount].status
-                     == Mount::INACTIVE) ? 0.0 : (Mount::UNCHOSEN ? 2.0 : 1.0) ) );
+                     == Mount::INACTIVE) ? 0.0 : (mounts[whichmount].status == Mount::UNCHOSEN ? 2.0 : 1.0) ) );
```
stephengtuggy, line 5933: This looks correct. Good job.

engine/src/cmd/ai/hard_coded_scripts.cpp
evertvorster: You might be right, of course. My reasoning here was that this un->Fire thing would be done if either of those weapon_info's evaluated to true, so why not run them after another? I will update to your suggestion, and see if that makes the compile error go away as well. 
I'm pretty sure your suggestion is closer to the original intent.

engine/src/cmd/unit_generic.cpp
evertvorster: Heh, it was quite the adventure getting there. I first had to figure out what the "?" was. Then, I pulled it apart, and made a proper if statement from it. When I was still getting the compiler error, it finally dawned on me that one of the three "boolean" comparisons were not like the other. 
So, this inline if is just one step short of a switch statement?

stephengtuggy:
Fixes #226

Issue #273, opened 2020-08-08
Two unused variables in src/cmd/drawable.cpp


stephengtuggy:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:103:13: warning: unused variable 'numvold' [-Wunused-variable]
  103 |         int numvold = 0;
      |             ^~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.cpp:104:13: warning: unused variable 'numvertices' [-Wunused-variable]
  104 |         int numvertices = 0;
      |             ^~~~~~~~~~~

Successor to #233 . (As far as I know, these two unused variable warnings remain unfixed.)

evertvorster:
Moved the two variable initialization to the ifdef block where they are actually used. 
It seems strange to me that the preprocessor chops them out so completely when they are used in an ifdef.

evertvorster:
This issue has been solved with PR #275

Issue #260, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: asteroidgen.dir/asteroidgen.cpp


evertvorster:
[ 99%] Building CXX object objconv/CMakeFiles/asteroidgen.dir/asteroidgen.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/asteroidgen.cpp: In function 'int main(int, char**)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/asteroidgen.cpp:144:8: warning: '<anonymous>.asteroid::radius' may be used uninitialized in this function [-Wmaybe-uninitialized]
  144 | struct asteroid
      |        ^~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/asteroidgen.cpp:144:8: warning: '<anonymous>.asteroid::num_polys' may be used uninitialized in this function [-Wmaybe-uninitialized]
[ 99%] Linking CXX executable asteroidgen

evertvorster:
Hopefully the fix here is to simply initialize them when they are declared.

Issue #252, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/cmd/faction_xml.cpp


evertvorster:
[ 44%] Building CXX object CMakeFiles/vegastrike.dir/src/cmd/faction_xml.cpp.o
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/universe_generic.h:29,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/vs_globals.h:5,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/matrix.h:26,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/quaternion.h:5,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/drawable.h:4,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_generic.h:33,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/ai/communication.h:3,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/faction_xml.cpp:8:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/faction_generic.h: In static member function 'static void Faction::beginElement(void*, const XML_Char*, const XML_Char**)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/faction_generic.h:25:12: warning: '<anonymous>.Faction::faction_stuff::stats' may be used uninitialized in this function [-Wmaybe-uninitialized]
   25 |     struct faction_stuff
      |            ^~~~~~~~~~~~~
[ 45%] Building CXX object CMakeFiles/vegastrike.dir/src/cmd/pilot.cpp.o

evertvorster:
This one will be fixed when #235 is fixed, as it references the same error

Issue #251, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/cmd/music.cpp


evertvorster:
[ 44%] Building CXX object CMakeFiles/vegastrike.dir/src/cmd/music.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/music.cpp: In member function 'void Music::_SkipRandSong(int, int)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/music.cpp:509:5: warning: 'nonnull' argument 'this' compared to NULL [-Wnonnull-compare]
  509 |     if (this != NULL) {
      |     ^~
[ 44%] Building CXX object CMakeFiles/vegastrike.dir/src/cmd/nebula.cpp.o

stephengtuggy:
This appears to have been fixed in #275 . Closing.

PR #275, opened 2020-08-09
Compiler fixes 2

evertvorster:

Please answer the following:

Code Changes:

Issues:
#235 - faction_generic.cpp - un-initialized function
#239 - start_system_xml.cpp - defined boolean that is not used
#245 - particle.cpp - comparison of integers with different signedness
#251 - music.cpp - non-null compared to NULL
#252 - faction_xml.cpp - maybe uninitialized variable
#260 - asteroidgen.cpp - uninitialized variable
#273 - Fixes the uninitialized variables

Purpose:
clear up some compiler warnings
- What release is this for?
Any
No


engine/src/faction_generic.cpp
```diff
@@ -18,14 +22,16 @@ void Faction::ParseAllAllies()
 void Faction::ParseAllies( unsigned int thisfaction )
 {
     unsigned int i, j;
+    std::vector< faction_stuff >faction;
```
stephengtuggy, line 25: OK, so I think what's happening here is that you're redeclaring this variable. I'm not sure that is the correct fix.

engine/src/faction_generic.cpp
```diff
@@ -37,6 +43,7 @@ void Faction::ParseAllies( unsigned int thisfaction )
         tempvec[faction[i].stats.index].stats = tmp;
     }
     faction.swap( tempvec );
+    }
```
stephengtuggy, line 46: I think this `}` should go after line 34.

engine/src/faction_generic.cpp
```diff
@@ -1,14 +1,18 @@
...
+
 
 using namespace FactionUtil;
 
 vector< boost::shared_ptr<Faction> >factions;  //the factions
 
 void Faction::ParseAllAllies()
 {
-    //MSVC has a bug where it won't let you reuse variables intitliezed in the paramater list of the for loop
+    //MSVC has a bug where it won't let you reuse variables initialized in the paramater list of the for loop
```
stephengtuggy, line 13: I'm pretty sure MSVC doesn't do that anymore. 😄

engine/src/faction_generic.cpp
evertvorster: I fully agree with you. Why would the compiler not see the previous declaration, though?

stephengtuggy: I think the compiler is seeing the previous **declaration**, just not a previous **initialization with a value.** I think I'll take a look at the code and see if the compiler is right about that.

engine/src/faction_generic.cpp
evertvorster: Which } are you talking about?

engine/src/faction_generic.cpp
stephengtuggy: Looks like gcc 10.x is complaining that [this field](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/7934960a2fa32a1ec1311c5fb6490c250df10d4f/engine/src/faction_generic.h#L33) is not initialized in [this constructor](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/7934960a2fa32a1ec1311c5fb6490c250df10d4f/engine/src/faction_generic.h#L37). Oddly enough, I don't get this warning in either gcc 9.3.0 or clang 10.0.0.

Try changing faction_generic.h, line 37, to the following:

```C++
        faction_stuff() : relationship(0.0)
        {
            stats.index = 0;
        }
```

See if that takes care of it.

engine/src/faction_generic.cpp
stephengtuggy: The one after `faction.swap( tempvec );`

engine/src/cmd/music.cpp
```diff
@@ -506,17 +506,15 @@ void Music::_SkipRandSong( int whichlist, int layer )
...
-                GotoSong( whichlist, random ? randInt( playlist[whichlist].size() ) : playlist[whichlist].counter++
-                          %playlist[whichlist].size(), true, layer );
-            else
-                fprintf( stderr, "Error no songs in playlist %d\n", whichlist );
-            return;
-        }
+    if ( whichlist != NOLIST && whichlist >= 0 && whichlist < (int) playlist.size() ) {
+        lastlist = whichlist;
+        static bool random = XMLSupport::parse_bool( vs_config->getVariable( "audio", "shuffle_songs", "true" ) );
+        if ( playlist[whichlist].size() )
```
nabaco, line 512: if where at it, let's put all the if..else statements in { }

engine/src/faction_generic.cpp
nabaco: who knows, we'll see once we figure out the MSVC compilation

engine/src/faction_generic.cpp
nabaco: Agreed, unless we have a bug hiding here due to missing `{` :)

engine/src/gfx/jpeg_memory.h
```diff
@@ -1,3 +1,5 @@
+#pragma once
```
nabaco, line 1: The generally accepted method in the project is to use guards instead of this pragma:
```
#ifndef _FILENAME_H
#define _FILENAME_H

// File content

#endif
```

engine/src/gfx/jpeg_memory.h
```diff
@@ -24,7 +26,7 @@
 #endif //tells VCC not to generate min/max macros
 #include <windows.h>
 #endif
-#include "vsfilesystem.h"
+#include "../vsfilesystem.h"
```
nabaco, line 29: Are you sure about this? relative path in include is not the best practice... do you see any issues in compilation due to this?

engine/src/star_system_xml.cpp
```diff
@@ -121,36 +121,37 @@ static bool EvalCondition( const char *op, const T &left, const T &right )
...
-        string sval = vs_config->getVariable( "graphics", varname, "0.0" );
-        rv = EvalCondition< string > ( op, sval, varval );
-    } else {
-        float fval = XMLSupport::parse_floatf( vs_config->getVariable( "graphics", varname, "0.0" ) );
-        float fref = XMLSupport::parse_floatf( varval );
-        rv = EvalCondition< float > ( op, fval, fref );
-    }
-    return rv;
-}
+// Commented out 2020-08, issue #239
```
nabaco, line 124: Can you elaborate on this? why are we commenting a whole function out?

stephengtuggy: @nabaco because it's not used anywhere, except in code that is also commented out. See #239 for further details.

stephengtuggy: Turns out that `ConfigCondition()` _was_ used prior to #181 . Hmm. We might want to look at this a little more closely.

engine/src/gfx/jpeg_memory.h
evertvorster: @nabaco , I'm not sure of anything at this point. My compiler does not complain about missing headers, it was just my IDE. However, with a little help from Stephen Tuggy I may have sorted out the IDE as well.

engine/src/faction_generic.cpp
evertvorster: Indeed it did. Learn something new every day!

engine/src/gfx/jpeg_memory.h
evertvorster: Ah, this was my IDE being crazy again. I've removed it now.

engine/src/star_system_xml.cpp
evertvorster: Hmm. The string "ConfigCondition" occurs exactly twice throughout the entire codebase. 
Once here, where it is defined, and once in system_factory.cpp, where it is commented out. Where was it used?

stephengtuggy: @evertvorster I put more details in #239 .

engine/src/faction_generic.cpp
evertvorster: I fixed that one up, as suggested

engine/src/cmd/music.cpp
evertvorster: Do you mean all the if ... else statements in the file, or just the ones in this function?

engine/src/star_system_xml.cpp
evertvorster: @stephengtuggy true, it was used in a few places pre #181. Not so much afterwards.
@royfalk what do you think about this function. Is it used anywhere, or can we finally rip it out?

stephengtuggy: I think we can leave it commented out for now, but not delete it entirely. More details in #239 discussion.

engine/src/cmd/music.cpp
```diff
@@ -475,12 +475,13 @@ void Music::GotoSong( int whichlist, int whichsong, bool skip, int layer )
...
-            if ( (layer < 0) && (muzak_count >= 2) )
+            if ( (layer < 0) && (muzak_count >= 2) ){
                 muzak[0].lastlist = muzak[1].lastlist = whichlist;
-
-            else
+            }
+            else{
                 lastlist = whichlist;
             GotoSong( playlist[whichlist][whichsong], layer );
+            }
```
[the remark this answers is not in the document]
evertvorster: I thought C++ is indentation agnostic. So, I was putting the braces at the end of the function, and ignoring the indentation. If you would rather have me respect the indentation, I can start again on these files, of course. I am just afraid that I will introduce changes to how the program operates.

engine/src/cmd/music.cpp
evertvorster: I have wrapped all the control statements in curly brackets in this file. It's nice exercise.

engine/src/cmd/music.cpp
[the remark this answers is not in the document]
evertvorster: I have moved the brace as requested. Also, I have put all the ifs elses and whiles in curly brackets.

engine/objconv/asteroidgen.cpp
```diff
@@ -144,9 +144,9 @@ class Tri
 struct asteroid
 {
     Vector center;
-    float  radius;
+    float  radius = 0; //Initializing to a value so the compiler shuts up about it
```
BenjamenMeyer, line 147: should use a float (`0.0f`) to initialize a float; compiler will cast it any way so not a big deal.

engine/src/cmd/drawable.cpp
```diff
@@ -99,13 +99,12 @@ void Drawable::AnimationStep()
...
-
-        int numvold = 0;
-        int numvertices = 0;
         //copy reference to data
         meshdata.at(0) = vecAnimations.at(activeAnimation)->at(activeMesh);
 
 #ifdef DEBUG_MESH_ANI
+        int numvold = 0;
+        int numvertices = 0;        
         std::cerr << "vertices changed from: " << numvold << " to: " << numvertices << std::endl;
```
BenjamenMeyer, line 108: does this actually do anything? If it's just printout out `0` and `0` then why not remove it?

engine/src/cmd/music.cpp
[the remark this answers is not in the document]
BenjamenMeyer: @evertvorster yes, C++ and C are indentation agnostic. However, for control structures - like `else` and `if`  if it's not wrapped in a block only takes the next line. So:
```
if (X) do_something();
```
and
```
if (X)
    do_something();
```
and
```
if (X) {
    do_something();
}
```
are equivalent.

However the last one clarifies for readers the difference between:
```
if (X)
   do_something();   // only run if X is true
   do_something_else();   // run regardless of whether X is true
```
and it's more clear form:
```
if (X) {
   do_something();
}
do_something_else();
```

engine/src/cmd/music.cpp
```diff
@@ -504,17 +547,17 @@ void Music::SkipRandSong( int whichlist, int layer )
...
+    }
+    if ( whichlist != NOLIST && whichlist >= 0 && whichlist < (int) playlist.size() ) {
+        lastlist = whichlist;
+        static bool random = XMLSupport::parse_bool( vs_config->getVariable( "audio", "shuffle_songs", "true" ) );
+        if ( playlist[whichlist].size() ){
+            GotoSong( whichlist, random ? randInt( playlist[whichlist].size() ) : playlist[whichlist].counter++
+                        %playlist[whichlist].size(), true, layer );
+        } else {
+            fprintf( stderr, "Error no songs in playlist %d\n", whichlist );
             return;
```
BenjamenMeyer, line 561: The return can probably go away. There doesn't seem to be any loop here. Line 508/551 makes sense as it's short cutting the function; but this one on 518/561 doesn't make so much sense as if it's missed the function will naturally return any way, just like when going through the other side of the if-block.

The addition of the braces makes it more clear that it's an unnecessary statement that can be removed.

engine/src/gldrv/gl_globals.h
```diff
@@ -315,8 +315,8 @@ typedef struct
     int  max_texture_dimension;
     int  max_movie_dimension;
     int  max_rect_dimension;
-    int  max_array_indices;
-    int  max_array_vertices;
+    unsigned int  max_array_indices;
+    unsigned int  max_array_vertices;
```
BenjamenMeyer, line 319: should these be unsigned? It might be useful to use a `-1` for certain cases (like an invalid reference).
Please verify.

evertvorster: This is to solve #245, where max_array_indices and max_array_vertices are compared to unsigned ints vertices_per_particle and vertices_per_particle in lines 26 to 49 of gfx/particle.cpp
They are only checked to be larger than 0, then used.

engine/objconv/asteroidgen.cpp
evertvorster: Done!

engine/src/cmd/drawable.cpp
evertvorster: I've chopped out this ifdef, as requested

engine/src/cmd/music.cpp
evertvorster: Chopped out the return function

engine/src/gfx/jpeg_memory.h
BenjamenMeyer: Yeah, there's two methods of header guards:
```
#pragma once
```
where the compiler does all the work
and
```
#ifndef __SOME_CONSTANT
#define __SOME_CONSTANT
...
#endif //__SOME_CONSTANT
```
where the programmer does more of the work.

Microsoft and some communities prefer the `pragma once` method; however, I've found in practice that the `ifndef/define/endif` method is far easier to trace, debug, etc when things go wrong. So I agree with @nabaco that that should be our preferred method. We can update the style-guide if needed.

evertvorster:
@nabaco, I have implemented the changes you have requested. Can I merge now?

engine/src/cmd/music.cpp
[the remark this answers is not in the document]
evertvorster: Thanks for the clarifications. 
As mentioned before, I have put the curly braces as requested.

engine/src/cmd/music.cpp
```diff
@@ -491,11 +533,12 @@ void Music::SkipRandSong( int whichlist, int layer )
 {
     if (muzak) {
         if (layer < 0) {
-            if (muzak_count >= 2)
+            if (muzak_count >= 2){
                 muzak[muzak_cross_index]._SkipRandSong( whichlist );
-
-            else
+            }
```
nabaco, line 538: In all other places you have `} else {` on the same line and here on a different. not a blocker, just noting.

nabaco (approved):

nabaco:
Let's get this merged after updating the branch and solving the merge conflicts

PR #278, opened 2020-08-12
Packaging for openSUSE Leap 15.2 and Fedora 32; etc.

### LINKED ISSUE #220: Add openSUSE packaging support

stephengtuggy:

Please answer the following:

Code Changes:
- [x] **This is a build system change only**

Issues:
**Closes #220 
Closes #278**

Purpose:
- What is this pull request trying to do? **Put in place the RPM dependencies for openSUSE Leap 15.2 and Fedora 32; distinguish between py2 and py3 builds in the package filename; refactor and fix CMake/CPack bugs as necessary along the way**
- What release is this for? **0.7.x, probably**
- Is there a project or milestone we should apply this to? **0.7.x, probably**


engine/CMakeLists.txt
```diff
@@ -1265,8 +1308,70 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
+            ENDIF (USE_PYTHON_3)
+            SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL, libglvnd, expat, boost-log, boost-system, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic")
+        ELSEIF (LINUX_ID STREQUAL fedora AND LINUX_VERSION_ID STREQUAL "31")    # Not yet tested
+            IF (USE_PYTHON_3)
+                SET(CPACK_RPM_PACKAGE_REQUIRES "python38, boost-python3")
+            ELSE (USE_PYTHON_3)
+                SET(CPACK_RPM_PACKAGE_REQUIRES "python26, boost-python2")
+            ENDIF (USE_PYTHON_3)
+            SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL, libglvnd, expat, boost-log, boost-system, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic")
+        ELSEIF (LINUX_ID STREQUAL fedora AND LINUX_VERSION_ID STREQUAL "30")    # Not yet tested
```
BenjamenMeyer, line 1345: Q: is there some way to combine the Fedora sections? Are the package names different between each release? I find that hard to be the case. It would be better if we could do:
```
if fedora:
   if version >= X:
   else Y =< version < X:
   else version <Y:
...
```

engine/CMakeLists.txt
```diff
@@ -1189,6 +1210,7 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
     IF(HAS_APT)
         MESSAGE("-- Configuring Debian Packaging")
         # See https://cmake.org/cmake/help/v3.3/module/CPackDeb.html
+        SET(CPACK_DEBIAN_PACKAGE_NAME "Vega-Strike")
```
BenjamenMeyer, line 1213: Did we miss this in master in the recent changes?

engine/CMakeLists.txt
stephengtuggy: The Python versions are different. The other package names are probably the same.

I'll try Fedora 30 and 31 as well, then get back to you.

engine/CMakeLists.txt
stephengtuggy: So we were already setting `CPACK_PACKAGE_NAME`. And `CPACK_DEBIAN_PACKAGE_NAME` was supposed to inherit its default setting from the former. But that didn't seem to work, at least on one of the distros I tried. That part of the filename was missing. So I added this line in to set `CPACK_DEBIAN_PACKAGE_NAME` explicitly.

BenjamenMeyer: Thanks!

engine/CMakeLists.txt
stephengtuggy: OK. Tested and working on Fedora 30, 31, and 32. Let me know what you think.

PR #282, opened 2020-08-15
Refactor universe

royfalk:
This is part of the merging of X and X_generic. This one merges star_system. I've tried to merge universe as well, but it turned out to be too difficult and a large project.

Code Changes:
- [Yes] Have the PR Validation Tests been run? 

Issues:
- Auto-pilot has difficulty reaching Atraxia. I suspect this is not related to this PR.

Purpose:
- What is this pull request trying to do? See above
- What release is this for? 0.7?


vincele:
FWIW, I think I remember also having problems reaching Ataraxia without this patchset

evertvorster:
When you say autopilot has difficulty reaching Atraxia... 
Does it get close, then veer off course repeatedly?
I have noticed behavior like this, and it is worse with fast craft.

royfalk:
> would appreciate changes for better read-ability and adding braces (coding standards), but otherwise looks good to go

I'm concentrating on rough changes/refactoring only and generating code without warnings on new files only. If I do a deep dive, I could spend a month on a single file and that wouldn't be productive at all. Unless you mean that I added new code with the wrong standard. That sometimes happen when I switch from work code, which has a different one.
In any case, we should consider using a tool like what golang has for doing this.

> When you say autopilot has difficulty reaching Atraxia...
> Does it get close, then veer off course repeatedly?
Yes. Exactly.

BenjamenMeyer:
> > would appreciate changes for better read-ability and adding braces (coding standards), but otherwise looks good to go
> 
> I'm concentrating on rough changes/refactoring only and generating code without warnings on new files only. If I do a deep dive, I could spend a month on a single file and that wouldn't be productive at all. Unless you mean that I added new code with the wrong standard. That sometimes happen when I switch from work code, which has a different one.
> In any case, we should consider using a tool like what golang has for doing this.

Unfortunately there really isn't a good code formatter for C/C++. I've known several (gnu indent,etc); each has its own set of bugs and formatting issues.

I'm not going to block on moving code around for formatting; but I will for obviously new code.

PR #283, opened 2020-08-17
Joystick cleanups

vincele:
Code Changes:

Issues:
- N/A

Purpose:
small cleanups in joystick handling code

- What release is this for?
next one

next one

Additional infos:
compiles cleanly & runs OK (without any joystick, mouse only)

### DIFF
--- a/engine/src/in_joystick.cpp
+++ b/engine/src/in_joystick.cpp
@@ -113,15 +113,6 @@ void myGlutJoystickCallback( unsigned int buttonmask, int x, int y, int z )
     }
 }
 
-JoyStick::JoyStick()
-{
-    for (int j = 0; j < MAX_AXES; ++j) {
-        axis_axis[j]    = -1;
-        axis_inverse[j] = false;
-        joy_axis[j]     = axis_axis[j] = 0;
-    }
-    joy_buttons = 0;
-}
 int JoystickPollingRate()
 {
     return game_options.polling_rate;
--- a/engine/src/in_joystick.cpp
+++ b/engine/src/in_joystick.cpp
@@ -134,7 +134,8 @@ void InitJoystick()
 #ifdef HAVE_SDL
     num_joysticks = SDL_NumJoysticks();
     printf( "%i joysticks were found.\n\n", num_joysticks );
-    printf( "The names of the joysticks are:\n" );
+    if (num_joysticks > 0)
+        printf( "The names of the joysticks are:\n" );
 #else
     //use glut
     if (glutDeviceGet( GLUT_HAS_JOYSTICK ) || game_options.force_use_of_joystick) {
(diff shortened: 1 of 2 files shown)


engine/src/in_joystick.cpp
```diff
@@ -113,15 +113,6 @@ void myGlutJoystickCallback( unsigned int buttonmask, int x, int y, int z )
...
 
-JoyStick::JoyStick()
-{
-    for (int j = 0; j < MAX_AXES; ++j) {
-        axis_axis[j]    = -1;
-        axis_inverse[j] = false;
-        joy_axis[j]     = axis_axis[j] = 0;
-    }
-    joy_buttons = 0;
-}
```
stephengtuggy, line 124: Hmm. I don't think getting rid of the JoyStick constructor is the right approach. Let's try something else.

I take it that you are trying to clean up the output from the program when no joystick is present?

engine/src/in_joystick.cpp
```diff
@@ -143,7 +134,8 @@ void InitJoystick()
 #ifdef HAVE_SDL
     num_joysticks = SDL_NumJoysticks();
     printf( "%i joysticks were found.\n\n", num_joysticks );
-    printf( "The names of the joysticks are:\n" );
+    if (num_joysticks > 0)
```
stephengtuggy, line 137: Good idea, to use an `if` here.

engine/src/in_joystick.cpp
vincele: This PR does a few independent things / cleanups:
- getting rid of an apparently unused constructor
- getting rid of a useless console message
- getting rid of a duplicated extern definition

I separated each in its own commit

This constructor looks unused, the only one used is the one taking a `JoyStick::JoyStick(int)` parameter.

I may be missing something, though...

vincele:
@BenjamenMeyer I cannot test that joysticks still work as I don't have one available ATM.

engine/src/in_joystick.cpp
stephengtuggy: I would say that if a default constructor has already been implemented, it's generally a good idea to leave it, even if it does not appear to be used anywhere at the moment. Future-proofing, you might call it.

BenjamenMeyer:
@vincele that's fine - just partner with someone that can;  please reach out in our gitter channel to find someone.

Loki1950:
Both @stephengtuggy and I have the same model joystick Logitech Extreme 3D Pro so I would not mind testing this when it's ready.

BenjamenMeyer:
@vincele as it has been nearly a year since any real activity and there are conflicts I'm going to close this for the time being. Please feel free to re-open once you have updated it or file a new PR which might be easier with many of the changes going on.

Issue #176, opened 2020-07-12
Enhancement: Use GitHub's normalized script pattern?


stephengtuggy:
I suggest that we use GitHub's normalized script pattern for open source projects, found [here](https://github.com/github/scripts-to-rule-them-all). This could make it easier for new contributors to get started with VS.

Not all of the scripts would apply -- like `server`, for instance -- but many would. I.e., the `sh` directory would become `script`. `clean` would become `setup`. And perhaps `vsbuild.sh` would just become `build`?

What do you guys think?

nabaco:
Sounds good to me. Let's open a PR with the changes. We could use this occasion to clean out old irrelevant scripts.

BenjamenMeyer:
Certainly work looking into.

PR #284, opened 2020-08-17
Add Ubuntu 20.04 Focal support to Travis-CI

nabaco:
CI changes only, to increase coverage. Shouldn't affect functionality in any way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ addons:
   apt:
     packages:
       - python-dev
-      - python2-dev
+#      - python2-dev
       - libgl1-mesa-glx
       - freeglut3-dev
       - libopenal-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,7 @@ addons:
   apt:
     packages:
       - python-dev
+      - python2-dev
       - libgl1-mesa-glx
       - freeglut3-dev
       - libopenal-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: cpp
-os:
- - linux
+os: linux
+distro: focal
 
 # When updating be sure to add a section
 # to the matrix to cover the distro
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -45,6 +60,7 @@ addons:
   apt:
     packages:
       - python-dev
+#      - python2-dev
```
stephengtuggy, line 63: Now python2 support is going to be broken on all 3 Ubuntu versions. We need separate `apt` sections for 20.04 vs. the others, if that's possible.

Python2 support **is** possible on Ubuntu 20.04, but only if you specifically install the 1.67.0 versions of the Boost libraries. Which is what I have done.

stephengtuggy: ...Or maybe python2 support won't be broken on Xenial and Bionic. Never mind that part then.

I would still like to get py2 working on Travis Focal though. Perhaps this would be a good time to implement #176 , and include a `bootstrap` script that installs any necessary development dependencies on that specific distro. I'm working on that right now, actually.

nabaco: What's the bottom line here though? can we merge this?

.travis.yml
```diff
@@ -11,11 +11,26 @@ compiler:
     - clang
 
 env:
-    - {}
+# Seems like Python2 doesn't work well on Focal.
+# Should be uncommented if someone fixes that
+#    - {}
```
BenjamenMeyer, line 16: Focal I think is Py3 by default; so not likely going to get fixed very well. Probably have to add the extra Py2 specific packages.

We may want to keep the primary/base-OS at bionic until we move to Py3 support being required. $0.02

stephengtuggy: There are other distros that do not have python2 support. Fedora 32, for example. It doesn't seem to support python2 **at all**.

stephengtuggy: In other words, it looks like 0.6.x will need to be limited to those distros that do support py2, and 0.7.x will need to be truly a mixed py2-py3 release. At least the way I see it.

nabaco: Agreed on that, and that is exactly what I have done here. Focal compiles Py3 only, and Bionic compiles both

.travis.yml
BenjamenMeyer: Let's make a decision about https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/284/files#r471870159 and then this also, and we can decide.

1. We can't break Py2 support.
2. Question: do we want to start requiring Py3 builds to pass on master? (We need to make sure they're not required for 0.6.x.)

Depending on our answer to '2' we might make a decision specifically about Focal too for practical purposes as far as our provided packages go.

nabaco: Py3 shows to be very stable (both in buildtime, and runtime according to @evertvorster), and since we do won't to convert to it, I think we should start requiring it as to make sure no one breaks what we already have.
If there is some doubts about this, we can still add this to the allowed failures list, though I think that would be wrong considering where we are standing now

stephengtuggy: @nabaco I agree. I have also switched to using Py3 by default. There are a few runtime errors, but the game seems to work fine. And it compiles very reliably.

nabaco: waiting for your approval to get this merged :)

Issue #246, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/gfx/star.cpp


evertvorster:
[ 45%] Building CXX object CMakeFiles/vegastrike.dir/src/gfx/star.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/star.cpp: In function 'GFXColorVertex* AllocVerticesForSystem(std::string, float, int*, int)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/star.cpp:163:64: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct GFXColorVertex'; use assignment or value-initialization instead [-Wclass-memaccess]
  163 |     memset( tmpvertex, 0, sizeof (GFXVertex)*(*num)*repetition );
      |                                                                ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfxlib.h:47,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/star.h:3,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/star.cpp:1:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfxlib_struct.h:234:8: note: 'struct GFXColorVertex' declared here
  234 | struct GFXColorVertex
      |        ^~~~~~~~~~~~~~
[ 45%] Building CXX object CMakeFiles/vegastrike.dir/src/gfx/stream_texture.cpp.o

Issue #229, opened 2020-08-05
Compiler (gcc 10.1.0) error in: src/cmd/unit_csv.cpp


evertvorster:
[  1%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/unit_csv.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_csv.cpp: In member function 'void Unit::LoadRow(CSVRow&, std::string, std::string*)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_csv.cpp:1058:38: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct Shield'; use assignment or value-initialization instead [-Wclass-memaccess]
 1058 |     memset( &two, 0, sizeof (Shield) );
      |                                      ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/damageable.h:4,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_generic.h:32,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_csv.cpp:3:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_armorshield.h:38:8: note: 'struct Shield' declared here
   38 | struct Shield
      |        ^~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_csv.cpp:1059:39: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct Shield'; use assignment or value-initialization instead [-Wclass-memaccess]
 1059 |     memset( &four, 0, sizeof (Shield) );
      |                                       ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/damageable.h:4,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_generic.h:32,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_csv.cpp:3:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_armorshield.h:38:8: note: 'struct Shield' declared here
   38 | struct Shield
      |        ^~~~~~
[...]

Issue #228, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/cmd/planet_generic.cpp


evertvorster:
[  2%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/planet_generic.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.cpp: In constructor 'Planet::Planet()':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.cpp:223:55: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct Shield'; use assignment or value-initialization instead [-Wclass-memaccess]
  223 |     memset( &(this->shield), 0, sizeof (Unit::shield) );
      |                                                       ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/damageable.h:4,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_generic.h:32,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.h:4,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.cpp:1:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/unit_armorshield.h:38:8: note: 'struct Shield' declared here
   38 | struct Shield
      |        ^~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.cpp: In constructor 'Planet::Planet(QVector, QVector, float, const Vector&, float, float, float, const string&, const string&, const string&, const std::vector<std::__cxx11::basic_string<char> >&, const QVector&, Unit*, int, std::string, bool, unsigned int)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/planet_generic.cpp:369:55: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct Shield'; use assignment or value-initialization instead [-Wclass-memaccess]
  369 |     memset( &(this->shield), 0, sizeof (Unit::shield) );
      |                                                       ^
[...]

PR #288, opened 2020-08-18
Compiler fixes 3

evertvorster:

Please answer the following:

Code Changes:

Issues:
#228 
#229 
#246 

Purpose:
Clear up some compiler errors
- What release is this for?
Any
Nope


engine/src/cmd/planet_generic.cpp
```diff
@@ -222,8 +222,13 @@ Planet::Planet() :
     terraintrans = NULL;
     atmospheric  = false;
     //Force shields to 0
-    memset( &(this->shield), 0, sizeof (Unit::shield) );
-    this->shield.number = 2;
+//    memset( &(this->shield), 0, sizeof (Unit::shield) ); //Let's not use memset on non-trivial stuff
```
BenjamenMeyer, line 225: if we know the structure is okay to do so, then it's fine to do.

However, for non-trivial stuff it'd be better to make it a formal `class` (as opposed to a `struct`) and add a method that can be called to do this; or use a constructor to do it (for initialization).

evertvorster: Thanks for the review. I'm sure to return to this code in the future, but for now it just makes the compiler happy.

Issue #287, opened 2020-08-18
Lots of error messages to the console with Python3


evertvorster:
When running VS from the console, I get a lot of error messages like this:
................
ERROR: NULL Unit used in Python script; returning default value...
ERROR: NULL Unit used in Python script; returning default value...
.................
This error message originates from engine/src/python/unit_wrapper_class.h, line 3. 
It looks like anything in this wrapper throws this error. We'll have to divide and conquer. ;)

BenjamenMeyer:
@evertvorster  is this still an issue?

evertvorster:
Nope, console nice and clean now.

:)


Evert Vorster
Awesome Chapters Tours
http://www.awesomechapters.com
Tel: +264 (0) 811477690

BenjamenMeyer:
Thanks. Closing

PR #289, opened 2020-08-19
Refactor universe - now actually refactoring the universe classes

royfalk:
This PR is part of a refactoring of the code to remove server parts and merge *_generic with *. In this PR I'm dealing with the universe and universe_generic files.

Code Changes:
- [Yes ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Auto-pilot landed me really close to facilities. I did change destinations, so it's probably an existing issue

Purpose:
- What is this pull request trying to do? see above


evertvorster (approved):
It plays wonderfully. I have not noticed anything out of the ordinary. 

Actually, this PR greatly alleviates #287, for some reason. It's not a complete cure, but definitely a step in the right direction. I have done the full flight test, buying and selling cargo and ship parts, did Janus' mission and shot at Oswald.
Auto pilot still has issues closing when I am in a fast ship, especially when the vessel speed is more than it's GCNT:Maneuver limits.

engine/src/universe.cpp
```diff
@@ -46,116 +46,45 @@
...
-    GFXSetMaterial( tmp, mat );
-    GFXSelectMaterial( tmp );
-    int ligh;
-    GFXCreateLightContext( ligh );
-    GFXSetLightContext( ligh );
-    GFXLightContextAmbient( GFXColor( 0, 0, 0, 1 ) );
-    GFXEndScene();
-}
+// Local Variables
+vector< StarSystem* >deleteQueue;
```
BenjamenMeyer, line 84: Just FYI - another good refactor would be to take things like `vector< StarSystem*>` and make them types:
```
typedef vector<StarSystem*> StarSystemList;
```
and then use that new type everywhere instead of the more expansive form. This will also help eliminate bugs, but also make typing easier.

Not asking for this round since it's copy/paste moves; but noting as something to do for another round at some point.

BenjamenMeyer: https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/290 will track this kind of thing.

royfalk:
@nabaco, it looks like one build just stalled or something. Is there a way to retry that one (or all of them) without another commit?

stephengtuggy:
@royfalk I just restarted the stalled job. 👍🏼

PR #298, opened 2020-08-20
Remove a function that does nothing

royfalk:
It changes from the original directory to the same one.

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Because this is such an initial part of the game and it does nothing according to the debugger, I checked the game runs properly. I suspect that if it fails, it won't be on my machine.

Issues:
- none identified.

Purpose:


stephengtuggy:
I think this function was supposed to change to the directory in which the program is located. It doesn't currently work. But it could probably be made to work -- e.g. by using argv[0].

royfalk:
> I think this function was supposed to change to the directory in which the program is located. It doesn't currently work. But it could probably be made to work -- e.g. by using argv[0].
Yep. That was what I was missing. It's not that it doesn't work. It is simply already running in this folder under normal conditions. They meant this to fix a crash when you do something such as _./build-folder/vegastrike_.

> If we need it, we can redo it via the Boost::FileSystem calls instead.
That's exactly what I did. I'll add this code in main, since it should be a one or two liner.

evertvorster (approved):
if in doubt, rip it out.
This function seems overly long, just to change a directory & and it's never called. As was mentioned in the comments, should we need this functionality we can probably implement it in a much better way.

engine/src/main.cpp
```diff
@@ -301,9 +302,13 @@ void initLoggingPart2(char debugLevel)
 
 int main( int argc, char *argv[] )
 {
-    initLoggingPart1();
+    // Change to program directory if not already
+    std::string argv0_string(argv[0]);
+    string program_directory = argv0_string.substr(0, argv0_string.length() - 11);
+    boost::filesystem::path program_directory_path(program_directory);
+    boost::filesystem::current_path(program_directory);
```
stephengtuggy, line 309: Perhaps this would be even more elegant:

```C++
    std::string program_path_str(argv[0]);
    boost::filesystem::path program_path(program_path_str);
    boost::filesystem::path program_directory_path = program_path.parent_path();
    boost::filesystem::current_path(program_directory_path);
```

stephengtuggy:
This change is very functional -- the new code changes to the program directory just like it's supposed to.

Unfortunately, this breaks the neat little trick I was using to load @evertvorster 's galaxy backgrounds without having to manually specify the path to them as a command-line argument to vegastrike. I'm still not sure **why** it breaks it, but it does.

My trick was to place a symbolic link in the `bin` directory, named `data`, linking to the `Assets-Production` directory. It worked for a while, but no longer. Oh well.

royfalk:
>This looks like it will do what it's supposed to. And in a much more elegant way, too.
>The number 11 is a "magic number", and it could be improved, but not a blocker, IMO.

Replaced with your code. Strangely enough, this code is what I had in my stashed version from a few months ago.

>Looks good to me. Has this now been play-tested?

Yes, the whole suite. Realistically though, if the game loads properly, it's probably good enough in this case.  

>Unfortunately, this breaks the neat little trick I was using to load @evertvorster 's galaxy backgrounds without having to manually specify the path to them as a command-line argument to vegastrike. I'm still not sure why it breaks it, but it does.

I'm sorry and I'm not sure why. Right now we're moving from one directory to the same one. There really should be no change. Please open an issue and we'll see what we can do. Maybe a config file for developers that would load extra stuff?

PR #314, opened 2020-08-24
WIP: Remove nebula and some other refactors

royfalk:
I've decided to remove nebula code. According to @Loki1950, it's not used at the moment and it's littering the entire code base. We can revisit it once we have the code stabilized.
Also move some mount functions to mount.cpp

Code Changes:
- [not yet] Have the PR Validation Tests been run? 

Issues:
- None identified

Purpose:
- What is this pull request trying to do? Code cleanup. 


evertvorster:
While I am not against the removal of this nebula code, as it really, really does not look like a nebula, what it does look like is some serious space anomaly. There is some place for this sort of thing, but I suppose we can implement it better. Good job on cleaning up the code base!

evertvorster:
Hi there. 
I just built and played this branch. 

There is a positional instability that feels different from the previous one. 

In this one it actually feels like it's the player's position that is unstable. I captured a video clip, which will shortly available here: https://youtu.be/T_bfUeF4Igg (Don't look at my landing, I'm still not sure how I survived that.)

In the beginning, if you look closely, it looks like the planet or the player is bouncing. However, no matter where I was in space, the speed meter kept bouncing as well. This is different from the previous "SPEC bouncy bug"

Secretly I am happy that this bug is coming to the fore, as now we can see it's behavior from a different angle, and learn more about it. Unfortunately this is a show stopper to this PR getting approved.

evertvorster (changes requested):
There is a positional instability in the craft that is unique to this branch at this point. 
There is a clip that shows the speed counter bouncing. This affects SPEC travel, as well as the auto-pilot when it is climbing out of a gravity well. This unfortunately has not been captured in this clip. 

Unfortunately we'll have to leave the nebula code in until we fix this bug.

royfalk:
I think we'll freeze this for now. I'll get bak to this when I'm done with the coming merge and I'll see what I've missed.

royfalk:
@evertvorster  Can you share the save game with me please. I'd start looking at this issue. I've seen it before too.

evertvorster:
[Evert_Mission.zip](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/5223565/Evert_Mission.zip)
It's not exactly where the video was shot, but the issue is clear in that file, too. 

Since I had docked with a transient, you'll be thrown into space straight away as you load the file. You'll see your speed counter bounce around. Fly over to the fighter barracks, and you'll see it bounce around in relation to you. I suspect it's the vessel. that is doing the bouncing. 

I started a new game, and did not see the issue with the Llama.

royfalk:
OK. I just downloaded and played the saved game with multiple variations. My current dev branch (not this), showed some initial speed for a second and then set to zero. The game worked fine. Did not find a fighter's barrack but did navigate to a research facility.
I then switched to this branch and did not see anything at all. 
Finally, in this branch, switched the settings to realistic physics and saw some jittering in spec near the star but, again, the flight to the research facility was fine.
@BenjamenMeyer - what are your thoughts about this? There are obviously some bugs in the flight system. We're seeing them manifest in all sorts of places, but it's not necessarily reproducible in a consistent manner.

BenjamenMeyer:
@royfalk I think we'll figure out the issues in the flight as we refactor stuff and clean things up so I'm not too concerned there right now. Do we have a regression with this PR in that aspect? That might help point some of the issue too.

royfalk:
@BenjamenMeyer  As I noted before, I was unable to see anything substantial with Evert's saved game but he did. Not really sure how to proceed here - do we merge anyway? close the PR? wait?

BenjamenMeyer:
@royfalk @evertvorster let's see if someone else can repro what @evertvorster saw

evertvorster:
Hi there!
Looks like the instability settles down after a while. Anyone would like to hazard a guess as to why? It also does not happen for me on a new game, only when I load that saved game. 
I made a new screen recording of the issue:

https://youtu.be/S75_movQtyM

Enjoy!

royfalk:
Saved games suffer from a variety of bugs, some dating 15 years back. Worse, reloading a saved game after dying or from a base computer is a recipe for bugs. I suspect this has to do with memory leaks.

BenjamenMeyer:
@royfalk would you mind if we moved this target to 0.9.x? I don't think it's blocking 0.8.x and from conversation on gitter I'm thinking time is going to prevent it from 0.7.x.  Thoughts?

royfalk:
I would say that in general, releases should not be delayed for specific issues unless these are critical. Feel free to bump issues in progress to later releases.

BenjamenMeyer:
@royfalk thanks. Just wanted to confirm. It's mixed right now until I get the 0.9.x project board up.

royfalk:
The code changed significantly since then. Closing.

Issue #318, opened 2020-08-28
Crash on program startup.


evertvorster:
I traced this one to commit: 2e0bf753985b7497f4c0878e01da65d237e836e9

Basically, what is happening on my system is that for whatever reason the starting directory initializes to an empty string. 
Bad things happen after that. 

I put in a pr that will fix this

stephengtuggy:
I suspect that this comes from #316 combined with #209 . VS changes directory just fine, but then it can't find the data files, so it crashes.

evertvorster:
I've merged my quick hack into master now, closing this issue. However, on my system vegastrike still crashes if not run from /usr/share/vegastrike, but does happily accept the "-d" parameter and then runs fine if you tell it where the data is. 
So, back to it's old behavior.

PR #319, opened 2020-08-28
Fixes a crash on my system

evertvorster:

Please answer the following:

Code Changes:

Issues:
#318 

Purpose:
This fixes a core dump on startup on my system
- What release is this for?
Any
Not really

### DIFF
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -306,7 +306,9 @@ int main( int argc, char *argv[] )
     std::string program_path_str(argv[0]);
     boost::filesystem::path program_path(program_path_str);
     boost::filesystem::path program_directory_path = program_path.parent_path();
-    boost::filesystem::current_path(program_directory_path);
+    if ( ! program_directory_path.empty())                  // Changing to an empty path does bad things
+    {    boost::filesystem::current_path(program_directory_path);
+    }
 
     initLoggingPart1();
 
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -307,7 +307,8 @@ int main( int argc, char *argv[] )
     boost::filesystem::path program_path(program_path_str);
     boost::filesystem::path program_directory_path = program_path.parent_path();
     if ( ! program_directory_path.empty())                  // Changing to an empty path does bad things
-    {    boost::filesystem::current_path(program_directory_path);
+    {    
+        boost::filesystem::current_path(program_directory_path);
     }
 
     initLoggingPart1();


engine/src/main.cpp
```diff
@@ -306,7 +306,9 @@ int main( int argc, char *argv[] )
     std::string program_path_str(argv[0]);
     boost::filesystem::path program_path(program_path_str);
     boost::filesystem::path program_directory_path = program_path.parent_path();
-    boost::filesystem::current_path(program_directory_path);
+    if ( ! program_directory_path.empty())                  // Changing to an empty path does bad things
+    {    boost::filesystem::current_path(program_directory_path);
+    }
```
BenjamenMeyer, line 311: Probably should generate a warning if it doesn't change the path.

Also, not sure why GitHub is formatting the code line to be on the same line as the first brace...
Should be:
```
if (...)
{
    ...
}
```
Please verify

evertvorster: Ah, it was not Github's fault. 
In my coding studies I have picked up the habit of putting the first statement after an opening brace on the same line as the brace, rather than on it's own line. So, completely my fault, and is corrected now.

engine/src/main.cpp
```diff
@@ -306,7 +306,10 @@ int main( int argc, char *argv[] )
     std::string program_path_str(argv[0]);
     boost::filesystem::path program_path(program_path_str);
```
stephengtuggy, line 307: What if we were to add in here something like:

```C++
boost::filesystem::path abs_program_path = boost::filesystem::absolute(program_path);
```

And then the next line (308) would become:

```C++
boost::filesystem::path program_directory_path = abs_program_path.parent_path();
```

Then we could be reasonably sure that we had a non-empty path to cd to.

BenjamenMeyer: We also probably only want to change directory there if some of the required data files exist at that path too.
So it'd be good to have a function that did the following:
1. Take a target path (absolute path, f.e `boost::filesystem::absolute()`)
2. Verify that the path contains the required data files at that location
3. Warn if no data files are present at that location, and return an error to the caller
4. Caller can then decide to try another path or error out entirely

engine/src/main.cpp
BenjamenMeyer: @evertvorster fully understand, my personal preferred C++ style is:
```
block statement
   {
   // block code
   }
```

I find that much more readable than putting the braces elsewhere; it seems to be unique enough it's very hard to get code generators/formaters to do that style; so I've had to learn to adapt to what others use.

BenjamenMeyer:
It would be nice to incorporate the changes mentioned in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/319#discussion_r479797736, but as this prevents a basic crash on at least one platform we can merge and do a follow-up if you want; but please be quick with a follow-up PR in that case.

engine/src/main.cpp
evertvorster: I've merged my small change. I really don't know enough about this boost stuff to properly implement Benjamin's suggestions. As far as I know, Vega Strike used to look for Version.txt, and if it did not exist, would error out.

Obviously the best way forward would be for one to be able to just type "vegastrike" anywhere in the file system, and the game to read your config file, figure out where the data is supposed to be and changing to that directory. 
There are still some messages about a game file directory supposed to be in the config file, but I have not seen the variable and do not know what it is supposed to look like. 

One other thing to think about is that vegastrike is a game engine that can run more than one game, that could conceivably be installed concurrently.

BenjamenMeyer: @evertvorster agree - it really should be in the config file. There's a number of ways that could be tackled:

- by default it would look at one config file but for alternate games a different config file could should be specified, or
- get the executable base name (e.g vegastrike) and look for the config at a given path (`~/.local/<executable base name>/...); alternate games would simply symlink the binary to rename it to change the path.

Basically we do something like:
```c++
boost::filesystem::path program_path(program_path_str);
boost::filesystem::path version_file = progrgram_path + "/" + "Version.txt"
if (!version_file.exists()) // or `false == version_file.exists()`
{
   // generate error
}
// allow change directory to happen
```
See https://www.boost.org/doc/libs/1_74_0/libs/filesystem/doc/reference.html#exists

PR #320, opened 2020-08-31
Refactor: get rid of opcodegarray macro

### LINKED ISSUE #225: Compiler (gcc 10.1.0) warning in: src/cmd/collide2/CSopcodecollider.cpp
I'm poking through the compiler errors in the compile log, to see if we can tidy things up a little.
This is the very first one in the log, and from the looks of it, it is not so trivial:

[1st_error.docx](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/5027049/1st_error.docx)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Closes #225**

Purpose:
- What is this pull request trying to do? **Get rid of a big, ugly preprocessor macro that uses realloc() where it shouldn't, and instead use an adaptation of std::vector called vs_vector.**
- What release is this for? **0.7.x**
- Is there a project or milestone we should apply this to? **0.7.x**

**I started testing this code, but IIRC, I ran into some weird issues when leaving Serenity. My velocity, as listed on the HUD, steadily increased until it was at 40C. Yet in reality, I was barely moving at all -- relative to Serenity or anything else. It was very strange.**

**Update: #338 filed for the above issue.**


BenjamenMeyer:
@stephengtuggy given current speed, I can see this going into 0.7.x easily enough.

engine/src/cmd/collide2/CSopcodecollider.cpp
```diff
@@ -33,25 +33,26 @@
...
-	vertholder = 0;
-	pairs.IncRef();
-	TreeCollider.SetFirstContact(true);
-	TreeCollider.SetFullBoxBoxTest(false);
-	TreeCollider.SetTemporalCoherence(false);
-	opcMeshInt.SetCallback (&MeshCallback, this);
-	GeometryInitialize (polygons);
+    m_pCollisionModel = nullptr;
+    vertholder = nullptr;
+    //pairs.IncRef();
```
BenjamenMeyer, line 50: I'd be a little caution about removing reference counting

engine/src/cmd/collide2/Stdafx.h
```diff
@@ -6,6 +6,8 @@
  */
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+#pragma once
```
BenjamenMeyer, line 9: 1. I'd rather remove `StdAfx.h`...it's a windows oriented files namely for pre-compiled headers which are themselves issues....that aside...
2. Please use the other header guard method:
```
#ifndef SOME_CONSTANT_H__
#define SOME_CONSTANT_H__
...
#endif // SOME_CONSTANT_H__ 
```
It's far more reliable and easier to debug than `#pragma once`.

engine/src/vs_collections.hpp
```diff
@@ -0,0 +1,46 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
+#pragma once
```
BenjamenMeyer, line 23: Same as https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/320/files?diff=split&w=1#r479895086

stephengtuggy: Done

engine/src/cmd/collide2/Stdafx.h
stephengtuggy: #2 Done

engine/src/cmd/collide2/CSopcodecollider.cpp
stephengtuggy: My understanding was that this was a custom implementation of `std::shared_ptr`, before there was such a thing. So I just used that instead. But perhaps `boost::intrusive_ptr` would be a better fit, I don't know.

Note also that the `pairs` variable is static anyway.

stephengtuggy:
So yeah, this code doesn't work right. It will need more fine-tuning before it does what it's supposed to do.

engine/src/cmd/collide2/CSopcodecollider.cpp
stephengtuggy: I ended up using a static `vector` directly, without a `shared_ptr` around it. The code seems to be working well now.

BenjamenMeyer (approved):

BenjamenMeyer:
Let's get some play testing on this

evertvorster (approved):
Gave this PR a good play test. There are no new issues that I picked up on.

Issue #258, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: vegasettings.dir/src/include/general.cpp


evertvorster:
[ 93%] Building CXX object setup/CMakeFiles/vegasettings.dir/src/include/general.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/setup/src/include/general.cpp: In function 'char* ptr_copy(char*)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/setup/src/include/general.cpp:83:9: warning: 'char* strncpy(char*, const char*, size_t)' specified bound depends on the length of the source argument [-Wstringop-overflow=]
   83 |  strncpy(alloc, string, strlen(string));
      |  ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/setup/src/include/general.cpp: In function 'char* replace(char*, char*, char*, int)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/setup/src/include/general.cpp:129:10: warning: 'char* strncat(char*, const char*, size_t)' specified bound depends on the length of the source argument [-Wstringop-overflow=]
  129 |   strncat(chr_new, replace, strlen(replace));
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/setup/src/include/general.cpp:131:10: warning: 'char* strncat(char*, const char*, size_t)' specified bound depends on the length of the source argument [-Wstringop-overflow=]
  131 |   strncat(chr_new, ptr_new, strlen(ptr_new));
      |   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[ 93%] Building CXX object setup/CMakeFiles/vegasettings.dir/__/src/common/common.cpp.o

evertvorster:
I see on stackoverflow that using memcpy instead of strncpy would solve the first bit of this warning, but when I tried it, the second and third errors did not go away. 

Here is the link:
https://stackoverflow.com/questions/56782248/gcc-specified-bound-depends-on-the-length-of-the-source-argument

I'll come back to this one.

PR #322, opened 2020-08-31
Fixes a compiler warning in vegasettings/src/include/general.cpp

evertvorster:

Please answer the following:

Code Changes:

Issues:
#258 
Purpose:
Fixes a compiler warning for vegasettings. 
The compiler was complaining that strncpy plus the full length of the source is the same as strcpy without the length check. 

- What release is this for?
any
nope

### DIFF
--- a/engine/setup/src/include/general.cpp
+++ b/engine/setup/src/include/general.cpp
@@ -87,18 +87,28 @@ char *split_words(char *string, int max_words) {
 
 // Better yet, use glib
 
-char *ptr_copy(char *string) 
+char *ptr_copy(const char *string)
 {
-	char *alloc;
-	alloc = (char *)malloc(strlen(string) + 1);
-	if (alloc == 0) 
-        {
-            fprintf(stderr, "Out of memory\n"); 
-            exit(-1);
-        }
-	alloc = string;
-	alloc[strlen(string)] = '\0';
-	return alloc;
+#if _XOPEN_SOURCE >= 500 || _POSIX_C_SOURCE >= 200809L
+    return strdup(string);
+#else
+    size_t buf_size = strlen(string) + 1;
+    char *alloc;
+    alloc = (char *)malloc(buf_size);
+    if (alloc == nullptr)
+    {
+        fprintf(stderr, "Out of memory\n");
+        fflush(stderr);
+        exit(-1);
+    }
+#ifdef __STDC_LIB_EXT1__
+    strcpy_s(alloc, buf_size, string);
+#else
+    strncpy(alloc, string, buf_size);
+#endif
+    alloc[buf_size - 1] = '\0';
+    return alloc;
+#endif
 }
 
 /* chomp
--- a/engine/setup/src/include/general.cpp
+++ b/engine/setup/src/include/general.cpp
@@ -310,7 +310,7 @@ int pwr(int start, int end)
 {
 	int current, val_return;
 	val_return = 1;
-	for (current = 2; current <= end; current++) 
+	for (current = 2; current <= end; ++current) 
         {
 		val_return *= start;
 	}
--- a/engine/setup/src/include/general.cpp
+++ b/engine/setup/src/include/general.cpp
@@ -96,7 +96,7 @@ char *ptr_copy(char *string)
             fprintf(stderr, "Out of memory\n"); 
             exit(-1);
         }
-	strcpy (alloc, string);
+	alloc = string;
 	alloc[strlen(string)] = '\0';
 	return alloc;
 }
@@ -131,47 +131,47 @@ char *pre_chomp(char *line)
  * This function searches the first parameter for the second parameter
  * If the second parameter is found, it is replaced with the third parameter
  */
-
-char *replace(char *line, char *search, char *replace, int LENGTH) 
-{
-	int  dif, calc;
-	char *ptr_new, *location;
-
-	char *chr_new,*current;
-	chr_new = (char *) malloc (sizeof (char)*LENGTH);
-	current = (char *) malloc (sizeof (char)*LENGTH);
-	calc = strlen(line) - strlen(search) + strlen(replace);
-    if (calc > LENGTH) 
-    {
-        free(chr_new);
-        free(current);
-        return line;
-    }
-	strcpy(current, line);
-	while ((location = strstr(current, search)) != nullptr) 
-        {
-		chr_new[0] = '\0';
-		calc = strlen(current) - strlen(search) + strlen(replace);
-                if (calc > LENGTH) 
-                {
-                    strcpy(line, current);
-                    free(current);
-                    free(chr_new);
-                    return line;
-                }
-		dif = location - current;
-		strncat(chr_new, current, dif);
-		strcat(chr_new, replace);
-		ptr_new = current + dif + strlen(search);
-		strcat(chr_new, ptr_new);
-		strcpy(current, chr_new);
-	}
-	strcpy(line, current);
-	free(chr_new);
-	free(current);
-	return line;
-}
-
+/*    --- Seems this function is not used at all.
+*char *replace(char *line, char *search, char *replace, int LENGTH) 
+*{
+*	int  dif, calc;
+*	char *ptr_new, *location;
+*
+*	char *chr_new,*current;
+*	chr_new = (char *) malloc (sizeof (char)*LENGTH);
+*	current = (char *) malloc (sizeof (char)*LENGTH);
+*	calc = strlen(line) - strlen(search) + strlen(replace);
+*    if (calc > LENGTH) 
+*    {
+*        free(chr_new);
+*        free(current);
+*        return line;
+*    }
+*	strcpy(current, line);
+*	while ((location = strstr(current, search)) != nullptr) 
+*        {
+*		chr_new[0] = '\0';
+*		calc = strlen(current) - strlen(search) + strlen(replace);
+*                if (calc > LENGTH) 
+*                {
+*                    strcpy(line, current);
+*                    free(current);
+*                    free(chr_new);
+*                    return line;
+*                }
+*		dif = location - current;
+*		strncat(chr_new, current, dif);
+*		strcat(chr_new, replace);
+*		ptr_new = current + dif + strlen(search);
+*		strcat(chr_new, ptr_new);
+*		strcpy(current, chr_new);
+*	}
+*	strcpy(line, current);
+*	free(chr_new);
+*	free(current);
+*	return line;
+*}
+*/
 char *strmov(char *to, char *from) 
 {
 	char *end;
--- a/engine/setup/src/include/general.cpp
+++ b/engine/setup/src/include/general.cpp
@@ -28,14 +28,22 @@ int RANDOMIZED = 0;
 // Gets the next parameter from the string, sorted by a space
 // Sticks a \0 at the space and returns a pointer to the right side.
 
-char *next_parm(char *string) {
+char *next_parm(char *string) 
+{
 	char *next;
-	if (string[0] == '\0') { return 0; }
+	if (string[0] == '\0') 
+        { 
+            return 0; 
+        }
 	next = string;
-	while (next[0] != ' ' && next[0] != '\0') {
+	while (next[0] != ' ' && next[0] != '\0') 
+        {
 		next++;
 	}
-	if (next[0] == '\0') { return next; }
+	if (next[0] == '\0') 
+        { 
+            return next; 
+        }
 	next[0] = '\0';
 	next++;
 	return next;
@@ -79,11 +87,16 @@ char *split_words(char *string, int max_words) {
 
 // Better yet, use glib
 
-char *ptr_copy(char *string) {
+char *ptr_copy(char *string) 
+{
 	char *alloc;
-	alloc = (char *)malloc(strlen(string)+1);
-	if (alloc == 0) { fprintf(stderr, "Out of memory\n"); exit(-1); }
-	strncpy(alloc, string, strlen(string));
+	alloc = (char *)malloc(strlen(string) + 1);
+	if (alloc == 0) 
+        {
+            fprintf(stderr, "Out of memory\n"); 
+            exit(-1);
+        }
+	strcpy (alloc, string);
 	alloc[strlen(string)] = '\0';
 	return alloc;
 }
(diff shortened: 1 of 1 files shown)


evertvorster:
I fixed up the formatting to be more to what Benjamin has recommended while I was at it. 
That's why the whole file has had my grubby hands all over it.

engine/setup/src/include/general.cpp
```diff
@@ -113,34 +132,38 @@ char *pre_chomp(char *line) {
...
+                {
+                    strcpy(line, current);
+                    free(current);
+                    free(chr_new);
+                    return line;
+                }
 		dif = location - current;
 		strncat(chr_new, current, dif);
-		strncat(chr_new, replace, strlen(replace));
+		strcat(chr_new, replace);
```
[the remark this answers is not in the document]
evertvorster: @stephengtuggy, why is it a step backwards? 
As I am learning this stuff, when you make a statement like that, you have to qualify it.

According to the documentation, the only difference between strcat & strncat is that one copies the whole string while the other expects a length of the string to be supplied as well. In this case, since we are copying the entire length of the source, we are effectively copying the whole string, so why not use the tool that is designed to do it?
 
I agree on the refactor, though. The code reads incredibly clunky.

evertvorster:
Most of these functions is to deal with the text formatting of the config file. Surely there is a better way to set game options?

To provide a more smooth and better experience, the setup needs to be incorporated into the game engine, and the config file retired. That would greatly simplify this bit, but I have a suspicion that this might be sometime off in the future.

engine/setup/src/include/general.cpp
```diff
@@ -79,21 +87,28 @@ char *split_words(char *string, int max_words) {
...
-	alloc = (char *)malloc(strlen(string)+1);
-	if (alloc == 0) { fprintf(stderr, "Out of memory\n"); exit(-1); }
-	strncpy(alloc, string, strlen(string));
+	alloc = (char *)malloc(strlen(string) + 1);
+	if (alloc == 0) 
+        {
+            fprintf(stderr, "Out of memory\n"); 
+            exit(-1);
+        }
+	strcpy (alloc, string);
```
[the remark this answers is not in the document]
evertvorster: Again, step backwards? What's the goal that you have in mind? My goal here was to make as small a change as possible to clear the compiler warning and not change the functionality of the code. 
In my searches on the difference between strncpy and strcpy on the web, there was no clear preference for one over the other. 
The compiler obviously thinks that in this situation the strcpy is the better tool, as it warns you when you use strncpy to copy an entire string.

engine/setup/src/include/general.cpp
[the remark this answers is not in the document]
BenjamenMeyer: @evertvorster always use the `strn*` functions when possible. They're better form since you have to specify how much data to copy, not leave it open ended.

So to be more specific in this case (though this is true of all the related functions):
- `strcat` will copy data in the provided buffer until it reaches a null terminator character (`\0'). If the buffer doesn't have one for some reason, then it will continue copying until something bad happens and the program crashes or overwrites something it shouldn't. This kind of thing often results in buffer-overflow attacks.
- `strncat` takes a parameter that specifies the maximum amount of data to copy; this prevents the buffer overflow attack; must marke sure the `size_t` parameter is `type * length`; another 

Similar methods:
- vsprintf vs vsnprintf
- sprintf vs snprintf
-

engine/setup/src/include/general.cpp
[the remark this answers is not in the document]
BenjamenMeyer: @evertvorster Microsoft VS will warn you if you use `strncpy` as they want you to use their special `s_strncpy` variant - which only works for Microsoft platforms.

BenjamenMeyer:
@evertvorster so in 0.9.x I'd like to revisit the game config and look at moving to using JSON or YAML format instead. But right now we need to use safer functions like strncpy/strncat/strnlen/etc as they perform a boundary check.

engine/setup/src/include/general.cpp
```diff
@@ -149,7 +172,8 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
 	return line;
 }
 
-char *strmov(char *to, char *from) {
+char *strmov(char *to, char *from)
```
BenjamenMeyer, line 175: this should take a boundary parameter 9 `strmov(char*, char*, size_t)` same format as `strnlen/strncat/strncpy`...probably should call it `strnmov` too for consistency

engine/setup/src/include/general.cpp
```diff
@@ -160,61 +184,94 @@ char *strmov(char *to, char *from) {
...
+        {
+		if (line[current] >= 65 && line[current] <= 90) 
+                {
 			line[current] += 32;
 		}
 	}
 }
 
-void strappend(char *dest, char *source, int length) {
+void strappend(char *dest, char *source, int length)
```
BenjamenMeyer, line 199: we should use `size_t` instead of `int` for the length parameter

engine/setup/src/include/general.cpp
```diff
@@ -299,10 +382,14 @@ void SetString(GString **ptr, char *line) {
 
 #endif
 
-char *NewString(const char *line) {
+char *NewString(const char *line)
```
BenjamenMeyer, line 385: We should check the usages and add a size_t length parameter too. Ideally though this would be a C++ class.

engine/setup/src/include/general.cpp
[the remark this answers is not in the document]
stephengtuggy: As I understand it, the advantage of strncat is that it will not overflow the destination character buffer. At least, not when used properly. This is important to avoid security flaws (buffer overflows).

This code, as it was, may have been technically correct. But the compiler couldn't tell that with its limited logic. It looked to the compiler like we were limiting the number of chars copied by the length of the **source** string, not the capacity of the **destination** buffer. In fact, this was not the case.

Long-term, our best bet is probably to switch to using `std::string`s (C++ style instead of C style), and use [`boost::algorithm::replace_all()`](https://www.boost.org/doc/libs/1_74_0/doc/html/boost/algorithm/replace_all.html) in place of this function.

@BenjamenMeyer @nabaco @royfalk do you guys have any input as far as what the best short-term fix would be? Or should we just skip straight to the long-term ("proper") solution?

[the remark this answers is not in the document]
evertvorster: @BenjamenMeyer & @stephengtuggy : Thanks for the explanation, it makes sense now.

As a learning exercise, I suppose it makes sense to implement C++ functions, as those are the ones that I would like to learn. The boost libraries seem like a whole new layer of complexity that I am afraid that I am not currently ready for. This function seems to be part of an xml interpreter. I don't know how much there is to gain from re-writing it, though. 

I'm pretty certain that we'll end up throwing out the entire vegasettings code base before long in favor for a better engineered solution, so I'll treat it as a way of getting more familiar with the C++ way of dealing with strings, unless you have strong feelings on the subject?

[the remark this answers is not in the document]
BenjamenMeyer: @stephengtuggy agree long term we should move to std::string/std::wstring (0.9.x target); for the short term the the strn* functions are fine but we should be using them.

@evertvorster yes, as we get more familiar with the code and apply more modern designs we'll probably get rid of chunks of code. We may also introduce new chunks of code.  We need to make sure to maintain performance targets while doing so. I wouldn't go after many of those big things quite yet; I'd like to wait for those until 0.9.x. Trying to limit 0.7.x/0.8.x as far as large things being done. 0.8.x will be very limited actually; hoping we'll only have a month between releases to get to 0.9.x out once we get 0.6.x fully out since it's mostly backend infrastructure that is holding up 0.6.x now.

evertvorster:
This problem changes depending on the level that you look at it. 
From the highest level, the settings should not be a separate program to the game, and all of it needs to be replaced. 

Slightly down from that high level, the configuration file is xml, and the settings program is an xml interpreter and modifier. There are excellent boost libraries for messing with xml with the minimum fuss, and it would be the easiest option to re-implement the settings program if we wanted to start from scratch today. 

At a lower level than that, all of this code is more C than C++. 
C++ has better simpler string handling capabilities, and a re-write of the settings in to proper C++ would probably not be too terrible to attempt. It just seems like a lot of wasted effort. It would take a few days of dedicated effort, and as there are libraries for handling this stuff better it would be a lot of effort with nothing to show for it at the end.

At the lowest level, it's C. The functions are unsafe, and if the game crashes because someone has messed with their own config file to the point that they manage to get a non-null terminated string in there no one is going to die. It would probably be an easy fix... so is the couple of fixes I threw in there to make the compiler shut up for now really such a bad idea?

My personal goal here was to get a nice clean compile log, and all of the options here apart from the last one is quite a bit of work.

BenjamenMeyer:
@evertvorster in the long run run I'd like to see:
- move to using libraries instead of implementing our own (Boost XML/JSON/etc)
- more C++ as it makes sense (FWIW, my C++ is still very C-ish; I mostly use C++ for objects and associating functions to data structures).
- moving the config file from XML to a more modern, easier to work with format such as JSON or YAML.

I think it makes sense for our configuration editor to be outside the game as it could then be used with multiple games; if we do it correctly we could even provide hooks for games to add their own config specs with default values.

But that's the long term, and we'll certainly have to do some design work to figure out how all of this will work, what we want, etc.

For the purposes of this PR, let's clean things up and keep things in working order; make sure we're using the correct functions with proper checking, etc. I could even see us moving to Boost XML parsing in the short term too as a good first step would be consolidating the configuration code and functionality; for instance there should be one global Game Config object instead of everything loading the configuration and using static vars.

evertvorster:
Here is an interesting wrinkle in this saga:

[evert@Vorster engine]$ find -name general.cpp 
./mission/include/general.cpp
./launcher/general.cpp
./setup/src/include/general.cpp


Of the three general.cpp in the Vega Strike code, this one is the only one with a compile warning. 
Why would we have three versions of the same thing?

Would it not be better to include the configuration editor in the engine? That way we only need one xml reader/interpreter.

BenjamenMeyer:
@evertvorster don't know... will have to see which ones are actually used. Would be good to rename them to be more applicable; but that can be a different PR.

BenjamenMeyer:
@evertvorster the compiler is probably not complaining about the function but because it thinks the function is doing something it shouldn't - like going past a buffer, or providing the wrong length value. If you point out in the code using a comment I'll be able to review that easier.

evertvorster:
The commented out function, as well as the other "fix" are here:
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/322/commits/c46fa990bf617fb97e163622b15caeeeda5dc873

I did a string search for "replace(" through all our sources, and really nothing calls that function. There are a couple of calls to class members called replace, but they are using a different set of parameters. Kdevelop agrees with me, and I will test this thing thoroughly tonight. 

I don't really know if the fix where I replace "strcpy (alloc, string);" with "alloc = string;" is really the right approach.


The book I am using to study C++ with is an O'Reilly's Practical C++ Programming. It was released in 1997, so it's arguably older than Vega Strike, but it allows me to read the code of the game better the more I study it. I'm about halfway through, so it won't be long until I can give Vega Strike more of my attention again. 
I will probably learn more modern C++ while refactoring older bits of the game.

engine/setup/src/include/general.cpp
```diff
@@ -79,21 +87,28 @@ char *split_words(char *string, int max_words) {
...
-	alloc = (char *)malloc(strlen(string)+1);
-	if (alloc == 0) { fprintf(stderr, "Out of memory\n"); exit(-1); }
-	strncpy(alloc, string, strlen(string));
+	alloc = (char *)malloc(strlen(string) + 1);
+	if (alloc == 0) 
+        {
+            fprintf(stderr, "Out of memory\n"); 
+            exit(-1);
+        }
+	alloc = string;
```
stephengtuggy, line 99: This isn't quite correct either, since it will simply end up returning the original string, rather than a copy of the string like it's supposed to.

This function was bugging me, so I spent some time on it today, and came up with a rather over-engineered solution. LOL.

Here it is. Function `ptr_copy` becomes:

```C++
char *ptr_copy(const char *string)
{
#if _XOPEN_SOURCE >= 500 || _POSIX_C_SOURCE >= 200809L
    return strdup(string);
#else
    size_t buf_size = strlen(string) + 1;
    char *alloc;
    alloc = (char *)malloc(buf_size);
    if (alloc == nullptr)
    {
        fprintf(stderr, "Out of memory\n");
        fflush(stderr);
        exit(-1);
    }
#ifdef __STDC_LIB_EXT1__
    strcpy_s(alloc, buf_size, string);
#else
    strncpy(alloc, string, buf_size);
#endif
    alloc[buf_size - 1] = '\0';
    return alloc;
#endif
}
```

Then in general.h in the same directory, the following line gets added before line 29:

```C++
#define __STDC_WANT_LIB_EXT1__ 1
```

stephengtuggy: The `strcpy_s` should work on Windows, and the `strdup` should work pretty much everywhere else. Theoretically, at least.

evertvorster: Pushed that change in on the PR. Tested it as well, the vegasettings still works as expected.

stephengtuggy (approved):
I'm happy with this PR as it stands. We might want to make some additional changes in later PR(s). @BenjamenMeyer , what do you think?

engine/setup/src/include/general.cpp
```diff
@@ -228,38 +295,45 @@ void vs_itoa(char *line, int number, int length) {
...
 	return val_return;
 }
 
-int pwr(int start, int end) {
+int pwr(int start, int end) 
+{
 	int current, val_return;
 	val_return = 1;
-	for (current = 2; current <= end; current++) {
+	for (current = 2; current <= end; current++)
```
BenjamenMeyer, line 313: Just FYI - `++current` is more efficient for this usage

BenjamenMeyer (approved):
It'd be nice if you could update the `current++` to `++current` as that's more efficient.

`current++` - save a copy, increment, return the old copy
`++current` - increments, and returns the new value

The overall result is the same in these loops; but the `++current` saves the a couple steps.

Good enough for now, but we'll want to make that change.

engine/setup/src/include/general.cpp
evertvorster: I have added in the requested changes, and tested the changes with modifying the vegastrilke.config file.

stephengtuggy:
Good job, @evertvorster . Thanks for your patience with this process.

Issue #73, opened 2020-04-20
Refactor Unit


nabaco:
As discussed in the last community meeting, we want to refactor the Unit classes, merging them together and simplifying the codebase.
I'm still not very familiar with that codebase, so maybe @danielrh will be able to offer some more technical details on how this can be done.

nabaco:
@BottledByte will you be willing to take this task upon yourself? Feel free to reject if you're dealing with other things right now.

BottledByte:
I see that @danielrh is assigned to this task already, so I am not sure if that means that he is going to refactor it. If he is only gathering some info about it, I will gladly take this task myself, but some ideas on how to do this would be indeed useful. But I will try my best anyway :)

Loki1950:
I do not think that danielrh can be very active ATM

nabaco:
I have assigned this to Danielrh only for him to suggest some information on the implementation.

Loki1950:
The Python scripts do a lot of accessing of the Unit class so close look at them may be productive

danielrh:
I think there are 2 reasonable approaches
a) make the types indistinguishable: this means pasting every member var thing inside GameUnit inside Unit then adding a typedef Unit GameUnit; so that the rest of vega strike can't tell the difference
b) Modifying the game engine to take GameUnit everywhere. The only time "Unit" should appear is inside the unit_generic.hpp and game_unit.hpp headers. Once the game only talks about GameUnit and GamePlanet, etc.... at that point you can proceed to plan (a) by trying to merge the classes.

BenjamenMeyer:
I like approach `b` personally.

stephengtuggy:
I like approach `b` as well.

royfalk:
How is this progressing?

royfalk:
As I've not seen any response from @BottledByte , I've started working on a parallel course. I'm extracting code from Unit class to super classes. e.g. drawing code in Drawable. 
This has the advantage of making code easier to refactor and debug.

However, I'm at the point where I should also be making changes to other classes such as GameUnit. I'd appreciate either a status update from @BottledByte or a reassign from @BenjamenMeyer .

BenjamenMeyer:
@royfalk agree; it's yours.

royfalk:
A (not so) brief update of where I am right now. I finally figured how to change one thing in the unit area without breaking everything else. I switched from 
`GameX : GameUnit<X> and X : Unit `
to
[...]

PR #325, opened 2020-09-06
Move reactToCollision code to its own class

royfalk:
Purpose:
- What is this pull request trying to do? This is the refactoring of the reactToCollision functions throughout the unit class hierarchy. I've extracted the code and centralized it in a separate class. While it is still impossible to write unit tests for it, a second refactor will enable that too.  
- Is there a project or milestone we should apply this to? 0.7

Issues:
- Collision between units still isn't good, though better than before IMO. Units typically explode on impact.

Code Changes:
- [No ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

### DIFF
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -89,6 +89,17 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
         return;
     }
 
+    // Collision with a jump point does nothing
+    if(other_unit->isJumppoint())
+    {
+        other_unit->jumpReactToCollision(unit);
+        return;
+    }
+
+    if(unit->isJumppoint())
+    {
+        return;
+    }
 
     switch(unit_type)
     {
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -7,6 +7,7 @@
 #include "missile.h"
 #include "enhancement.h"
 #include <typeinfo>
+#include <boost/log/trivial.hpp>
 
 // TODO: convert all float to double and all Vector to QVector.
 
@@ -119,6 +120,7 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
         return;
 
     // Not sure what an enhancement is, but it looks like it's something that can increase the shields of the unit it collided with.
+    // TODO: refactor this.
     case ENHANCEMENTPTR:
         if (other_unit->isUnit() == ASTEROIDPTR)
         {
@@ -127,13 +129,13 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
         }
 
         double percent;
-        char tempdata[sizeof (unit->shield)];
-        memcpy( tempdata, &unit->shield, sizeof (unit->shield) );
+        char tempdata[sizeof(Shield)];
+        memcpy( tempdata, &unit->shield, sizeof(Shield));
         unit->shield.number = 0;     //don't want them getting our boosted shields!
         unit->shield.shield2fb.front = unit->shield.shield2fb.back =
                 unit->shield.shield2fb.frontmax = unit->shield.shield2fb.backmax = 0;
         other_unit->Upgrade( unit, 0, 0, true, true, percent );
-        memcpy( &unit->shield, tempdata, sizeof (unit->shield) );
+        memcpy( &unit->shield, tempdata, sizeof (Shield) );
         string fn( unit->filename );
         string fac( FactionUtil::GetFaction( unit->faction ) );
         unit->Kill();
@@ -204,9 +206,13 @@ void Collision::applyForce(double elasticity, float& m2, Vector& v2)
     This function adds an elasticity ratio to reduce velocity, due to heat, damage, etc.
     */
 
-    // Ratio must be between 0 and 1
-    assert(elasticity>=0);
-    assert(elasticity<=1);
+    // elasticity must be between 0 and 1
+    if(elasticity >1 || elasticity <0)
+    {
+        BOOST_LOG_TRIVIAL(warning) << "Collision::applyForce Expected an elasticity value between 0 and 1 but got " << elasticity;
+        elasticity = std::max(elasticity, 0.0);
+        elasticity = std::min(elasticity, 1.0);
+    }
 
     if(!apply_force)
     {
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -1,0 +1,405 @@
+#include "collision.h"
+
+#include "unit_generic.h"
+#include "universe.h"
+#include "universe_util.h"
+#include "game_config.h"
+#include "missile.h"
+#include "enhancement.h"
+#include <typeinfo>
+
+static const float kilojoules_per_damage = GameConfig::GetVariable( "physics", "kilojoules_per_unit_damage", 5400 );
+static const float collision_scale_factor =
+        GameConfig::GetVariable( "physics", "collision_damage_scale", 1.0f );
+static const float inelastic_scale = GameConfig::GetVariable( "physics", "inelastic_scale", 0.8f);
+static const float min_time =
+        GameConfig::GetVariable( "physics", "minimum_time_between_recorded_player_collisions", 0.1f);
+static const float minimum_mass = 1e-6f;
+
+//Collision force caps primarily for AI-AI collisions. Once the AIs get a real collision avoidance system, we can
+// turn damage for AI-AI collisions back on, and then we can remove these caps.
+//value, in seconds of desired maximum recovery time
+static const float max_torque_multiplier =
+        GameConfig::GetVariable( "physics", "maxCollisionTorqueMultiplier", 0.67f);
+//value, in seconds of desired maximum recovery time
+static const float max_force_multiplier  =
+        GameConfig::GetVariable( "physics", "maxCollisionForceMultiplier", 5);
+
+static const int upgrade_faction =
+        GameConfig::GetVariable( "physics", "cargo_deals_collide_damage",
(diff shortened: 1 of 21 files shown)


BenjamenMeyer:
I don't see anything obvious at least...please update and let's get some play testing done to verify.

evertvorster:
I have given this a little play test, bumping into Serenity at slow and high speeds. It performs as expected, bumping into it at 5 m/s does almost no damage, and at 100 m/s it's fatal, in the Llama at least. 
So, no worse than what's already there. In fact, I don't feel any difference at all. 

Collision detection is pretty terrible between missiles and Oswald, both with this new model and the previous one. I can be stationary 200m away from him, and fire four or five dumfire missiles straight through his craft without the game registering a hit. I don't know if this is the same between dumfires and other craft, I'll go fire some missiles at other craft shortly.

While we are messing with things that go bump in the night, it would be nice to have a repulsor effect from the ships shields, ie: your shields should act like a weak magnetic bumper, pushing you away from whatever you are flying close to. You should be able to easily overpower this force with just the craft's engines. 

I really don't know how easy this would be to implement, but if you do have a flash of inspiration or see an easy way to implement it I, for one, would be grateful for such a thing. Anyways, something to think about while you are playing with this bit of the game engine.

engine/src/cmd/movable.cpp
```diff
@@ -5,9 +5,40 @@
...
 #include "star_system.h"
 #include "universe.h"
+#include "game_config.h"
 
 #include <iostream>
 
+static const float VELOCITY_MAX = GameConfig::GetVariable( "physics", "velocity_max", 10000);
+
+//for the heck of it.
+static const float humanwarprampuptime = GameConfig::GetVariable( "physics", "warprampuptime", 5);
```
BenjamenMeyer, line 15: we should (possibly in another PR) make these variables more like `HUMAN_WARP_RAMP_UP_TIME` so they're more readable

royfalk: Let's do that in another PR. I'm actually still trying to work on #73 and this PR was just a way to solve an issue there.

royfalk:
Current status is that collision still works but is buggy. The previous version caused two units to get stuck together. This one has them flying through each other. This is not as bad as it sounds because for bases, the damage caused to the ship is significant and it dies. For ships, low speed causes damage and high speed kills. 
And of course, the code is much cleaner and easier to work on. 
My proposal:

- @evertvorster to test some more if available
- @BenjamenMeyer , @stephengtuggy , @nabaco  to review code
- @BenjamenMeyer  to approve for 0.9 maybe?

The advantage of this proposal is that it opens up the opportunity to further refactor the code and move in the right direction overall.

evertvorster:
Bumping around inside an asteroid seems to do what it's supposed to. When going slowly and driving my ship into a side, I get the audible bumping sound, and damage to the ship continually in small increments until I stop my foolish behavior. 
Perfect. 
When ramming full speed into the walls that's an instant kill. Also good. 

Ramming poor Oswald at the beginning, it "feels" like I am driving my Llama through his Robin. This does not quite feel right, but it's no worse than the previous code. It would be nice if the shields of both vessels did a bit of pushing aside like a magnetic bumper. Certainly not occupy the same space as they do now. But yeah, loads of damage and death when ramming another vessel at 100 m/s. (As it should be)

As a small niggle that can be fixed later, I am able to get my cockpit of the vendetta through the walls of the inside of an asteroid, allowing me to see the thing from the inside. This detracts from the realism of the simulation a bit. 
In short, I should not be able to push any part of the mesh of my vessel into an inflexible thing like an asteroid.

evertvorster:
Oh, and I am still able to shoot dumfires through Oswald without it hitting his ship. About a 50% hit ratio point blank.

royfalk:
_Bumping around inside an asteroid seems to do what it's supposed to. ... Perfect.
When ramming full speed into the walls that's an instant kill. Also good.
Ramming poor Oswald at the beginning, it "feels" like I am driving my Llama through his Robin. This does not quite feel right, but it's no worse than the previous code._ 
That's what I was aiming for in this PR. I haven't figured out how to bounce ships off each other yet. However, with the new code, if someone does understand how, it should be trivial to add.
 
_In short, I should not be able to push any part of the mesh of my vessel into an inflexible thing like an asteroid._
Yep. That's a bug. For the record, you could also do that with the old code. I managed to stick my ship half way through Oswald's one. 

_Oh, and I am still able to shoot dumfires through Oswald without it hitting his ship. About a 50% hit ratio point blank._
That's not a collision issue. That's a collider issue. I haven't touched that code at all.

BenjamenMeyer:
As long as we're improving things I'm good with the work going into 0.7.x with a long term goal of fixing things up in 0.9.x.

royfalk:
@BenjamenMeyer , @stephengtuggy , what is the status of this? Is there something else needed here or can we merge?

engine/src/cmd/collision.cpp
```diff
@@ -0,0 +1,263 @@
...
+    case ENHANCEMENTPTR:
+        if (other_unit->isUnit() == ASTEROIDPTR)
+        {
+            apply_force = true;
+            return;
+        }
+
+        double percent;
+        char tempdata[sizeof (unit->shield)];
+        memcpy( tempdata, &unit->shield, sizeof (unit->shield) );
```
BenjamenMeyer, line 131: This should use sizeof(type) instead of sizeof(var)

engine/src/cmd/collision.cpp
```diff
@@ -0,0 +1,263 @@
...
+    The current implementation is of an imperfectly elastic collision.
+    An elastic collision is an encounter between two bodies in which the total kinetic energy of the two bodies remains the same. In an ideal, perfectly elastic collision, there is no net conversion of kinetic energy into other forms such as heat, noise, or potential energy.
+    https://en.wikipedia.org/wiki/Elastic_collision
+
+    This function adds an elasticity ratio to reduce velocity, due to heat, damage, etc.
+    */
+
+    // Ratio must be between 0 and 1
+    assert(elasticity>=0);
+    assert(elasticity<=1);
```
BenjamenMeyer, line 209: I would not assert,  just log an error and do something reasonable (max, min, etc)

BenjamenMeyer (changes requested):
Couple minor changes; otherwise just very the play testing is sufficient.

engine/src/cmd/collision.cpp
```diff
@@ -0,0 +1,269 @@
...
+    case ENHANCEMENTPTR:
+        if (other_unit->isUnit() == ASTEROIDPTR)
+        {
+            apply_force = true;
+            return;
+        }
+
+        double percent;
+        char tempdata[sizeof(Shield)];
+        memcpy( tempdata, &unit->shield, sizeof(Shield));
```
BenjamenMeyer, line 133: not required right now, but the best pattern for this would be:
```
size_t tempdata_size = sizeof(struct Shield); // or would it be `class Shield`? - generally be as explicit as possible
char tempdata[tempdata_size];
memcpy(tempdata, &unit->shield, tempdata_size);
```
The main benefit of the extra variable is for debugging. The compiler will automatically do that in the background for the code, but this gives access to the value under gdb and other debuggers, as well as logging, etc if needed.

Not blocking.

engine/src/cmd/movable.cpp
BenjamenMeyer: That's fine. It's part of a larger code cleanup. Honestly, I'd like to see a global instance of the Game config and then have classes load the config from the global instance and minimize the static vars, or rather use a different pattern of initialization:
- single static var used as a gate (initialized, defaults to false)
- if false, the first class loads the config values

Allows the vars to be defaulted in code, and then have their config loaded at run-time more controllably. Long term.

BenjamenMeyer (approved):
couple comments, nothing blocking. :)
Thanks!

BenjamenMeyer:
If we have sufficient play test coverage then go ahead and merge.

engine/src/cmd/collision.cpp
royfalk: Done. Also, this is old code and I need to balance width with depth here. I could spend the next year polishing a few files.

engine/src/cmd/collision.cpp
royfalk: Done.

engine/src/cmd/collision.cpp
royfalk: Honestly, this is not my code and the best pattern wouldn't be using sizeof and memcpy.

royfalk:
Huh! One final, pretty major bug. It seems jump points are treated as buildings or units and blow ship up. Will need to fix that before merge.

BenjamenMeyer:
LGTM; approval still stands - waiting on sufficient play testing.

royfalk:
Did the complete test suite last night. Found a bug. Fixed it. Merging.

engine/src/cmd/movable.cpp
```diff
@@ -5,9 +5,40 @@
...
+
+//Pi^2
+
+static const float  warpMultiplierMin = GameConfig::GetVariable( "physics", "warpMultiplierMin", 9.86960440109f);
+
+//C
+static const float  warpMultiplierMax = GameConfig::GetVariable( "physics", "warpMultiplierMax", 300000000 );
+
+//Pi^2 * C
+static const float warpMaxEfVel = GameConfig::GetVariable( "physics", "warpMaxEfVel", 2960881320.0f);
```
stephengtuggy, line 38: Because this is both `static` and `const`, it gets initialized before vegastrike.config is actually loaded. So it never actually reads its value from the config file. This is what is causing #335 , I'm pretty sure.

Issue #259, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: mesh_tool.dir/mesher/to_BFXM.cpp


evertvorster:
[ 96%] Building CXX object objconv/CMakeFiles/mesh_tool.dir/mesher/to_BFXM.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp: In function 'void beginElement(const string&, const XMLSupport::AttributeList&, XML*)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp:484:59: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct line'; use assignment or value-initialization instead [-Wclass-memaccess]
  484 |         memset( &xml->linetemp, 0, sizeof (xml->linetemp) );
      |                                                           ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp:3:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.h:61:8: note: 'struct line' declared here
   61 | struct line
      |        ^~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp:490:67: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct triangle'; use assignment or value-initialization instead [-Wclass-memaccess]
  490 |         memset( &xml->triangletemp, 0, sizeof (xml->triangletemp) );
      |                                                                   ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp:3:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.h:83:8: note: 'struct triangle' declared here
   83 | struct triangle
      |        ^~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/objconv/mesher/to_BFXM.cpp:506:59: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct quad'; use assignment or value-initialization instead [-Wclass-memaccess]
[...]

Issue #254, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/gldrv/gl_light_state.cpp


evertvorster:
[ 44%] Building CXX object CMakeFiles/vegastrike.dir/src/gldrv/gl_light_state.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp: In member function 'GFXLight gfx_light::operator=(const GFXLight&)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp:158:43: warning: 'void* memcpy(void*, const void*, size_t)' copying an object of non-trivial type 'class gfx_light' from an array of 'const class GFXLight' [-Wclass-memaccess]
  158 |     memcpy( this, &tmp, sizeof (GFXLight) );
      |                                           ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp:6:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light.h:23:7: note: 'class gfx_light' declared here
   23 | class gfx_light : public GFXLight
      |       ^~~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp: In member function 'void gfx_light::ResetProperties(LIGHT_TARGET, const GFXColor&)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp:342:45: warning: 'void* memcpy(void*, const void*, size_t)' copying an object of non-trivial type 'class GFXLight' from an array of 'class gfx_light' [-Wclass-memaccess]
  342 |         memcpy( &t, this, sizeof (GFXLight) );
      |                                             ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfxlib.h:47,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light.h:3,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gldrv/gl_light_state.cpp:6:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfxlib_struct.h:585:7: note: 'class GFXLight' declared here
[...]

Issue #240, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/gfx/mesh_xml.cpp


evertvorster:
[  8%] Building CXX object CMakeFiles/engine_com.dir/src/gfx/mesh_xml.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/mesh_xml.cpp: In member function 'void Mesh::beginElement(MeshXML*, const string&, const XMLSupport::AttributeList&)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/mesh_xml.cpp:522:55: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct GFXVertex'; use assignment or value-initialization instead [-Wclass-memaccess]
  522 |         memset( &xml->vertex, 0, sizeof (xml->vertex) );
      |                                                       ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/star_system_generic.h:7,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/universe_generic.h:28,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/vs_globals.h:5,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/matrix.h:26,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/quaternion.h:5,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/mesh.h:27,
                 from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/mesh_xml.cpp:1:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfxlib_struct.h:53:8: note: 'struct GFXVertex' declared here
   53 | struct GFXVertex
      |        ^~~~~~~~~
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/gfx/mesh_xml.cpp:856:55: warning: 'void* memset(void*, int, size_t)' clearing an object of non-trivial type 'struct GFXVertex'; use assignment or value-initialization instead [-Wclass-memaccess]
  856 |         memset( &xml->vertex, 0, sizeof (xml->vertex) );
      |                                                       ^
[...]

Issue #231, opened 2020-08-05
Compiler (gcc 10.1.0) warning in: src/cmd/weapon_xml.cpp


evertvorster:
[  1%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/weapon_xml.cpp.o
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/weapon_xml.cpp: In function 'weapon_info getWeaponInfoFromBuffer(char*, int&)':
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/weapon_xml.cpp:33:45: warning: 'void* memcpy(void*, const void*, size_t)' writing to an object of type 'struct weapon_info' with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
   33 |     memcpy( &wi, netbuf+offset, sizeof (wi) );
      |                                             ^
In file included from /home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/weapon_xml.cpp:4:
/home/evert/vsmod/git/Vega-Strike-Engine-Source/engine/src/cmd/weapon_xml.h:11:8: note: 'struct weapon_info' declared here
   11 | struct weapon_info
      |        ^~~~~~~~~~~
[  1%] Building CXX object CMakeFiles/engine_com.dir/src/cmd/damageable.cpp.o

evertvorster:
This function is not called in our source. I suspect it's a hold-over from the networking code that was removed.

PR #326, opened 2020-09-07
Compiler fixes round 5

evertvorster:
General fixes to clean up compiler warnings, round 5

Please answer the following:

Code Changes:

Issues:
#231 
#240 
#254 
#259 

Purpose:
Clean up compiler warnings.
- What release is this for?
Any
Nope

### DIFF
--- a/engine/objconv/mesher/to_BFXM.cpp
+++ b/engine/objconv/mesher/to_BFXM.cpp
@@ -480,12 +480,12 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
     case XML::POLYGONS:
         break;
     case XML::LINE:
-        xml->linetemp = {};
+        xml->linetemp.clear();
         xml->curpolytype  = LINE;
         xml->curpolyindex = 0;
         break;
     case XML::TRI:
-        xml->triangletemp = {};
+        xml->triangletemp.clear();
         xml->curpolytype  = TRIANGLE;
         xml->curpolyindex = 0;
         for (iter = attributes.begin(); iter != attributes.end(); iter++) {
@@ -501,7 +501,7 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
         }
         break;
     case XML::QUAD:
-        xml->quadtemp = {};
+        xml->quadtemp.clear();
         xml->curpolytype  = QUAD;
         xml->curpolyindex = 0;
         for (iter = attributes.begin(); iter != attributes.end(); iter++) {
--- a/engine/objconv/mesher/to_BFXM.cpp
+++ b/engine/objconv/mesher/to_BFXM.cpp
@@ -1,7 +1,6 @@
 #include "PrecompiledHeaders/Converter.h"
 #include "mesh_io.h"
 #include "to_BFXM.h"
-
 #include <cstring>  //We are using C style string functions here
 
 //#define fprintf aprintf
@@ -481,19 +480,12 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
     case XML::POLYGONS:
         break;
     case XML::LINE:
-        xml->linetemp.flatshade = 0;
-        memset(xml->linetemp.indexref, 0, sizeof(xml->linetemp.indexref) );
-        memset(xml->linetemp.s, 0, sizeof(xml->linetemp.s));
-        memset(xml->linetemp.t, 0, sizeof(xml->linetemp.t));
+        xml->linetemp = {};
         xml->curpolytype  = LINE;
         xml->curpolyindex = 0;
-        xml->linetemp.flatshade = 0;
         break;
     case XML::TRI:
-        xml->triangletemp.flatshade = 0;
-        memset(xml->triangletemp.indexref, 0, sizeof(xml->triangletemp.indexref) );
-        memset(xml->triangletemp.s, 0, sizeof(xml->triangletemp.s));
-        memset(xml->triangletemp.t, 0, sizeof(xml->triangletemp.t));
+        xml->triangletemp = {};
         xml->curpolytype  = TRIANGLE;
         xml->curpolyindex = 0;
         for (iter = attributes.begin(); iter != attributes.end(); iter++) {
@@ -509,10 +501,7 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
         }
         break;
     case XML::QUAD:
-        xml->quadtemp.flatshade = 0;
-        memset(xml->quadtemp.indexref, 0, sizeof(xml->quadtemp.indexref) );
-        memset(xml->quadtemp.s, 0, sizeof(xml->quadtemp.s));
-        memset(xml->quadtemp.t, 0, sizeof(xml->quadtemp.t));
+        xml->quadtemp = {};
         xml->curpolytype  = QUAD;
         xml->curpolyindex = 0;
         for (iter = attributes.begin(); iter != attributes.end(); iter++) {
--- a/engine/objconv/mesher/to_BFXM.cpp
+++ b/engine/objconv/mesher/to_BFXM.cpp
@@ -481,13 +481,19 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
     case XML::POLYGONS:
         break;
     case XML::LINE:
-        memset( &xml->linetemp, 0, sizeof (xml->linetemp) );
+        xml->linetemp.flatshade = 0;
+        memset(xml->linetemp.indexref, 0, sizeof(xml->linetemp.indexref) );
+        memset(xml->linetemp.s, 0, sizeof(xml->linetemp.s));
+        memset(xml->linetemp.t, 0, sizeof(xml->linetemp.t));
         xml->curpolytype  = LINE;
         xml->curpolyindex = 0;
         xml->linetemp.flatshade = 0;
         break;
     case XML::TRI:
-        memset( &xml->triangletemp, 0, sizeof (xml->triangletemp) );
+        xml->triangletemp.flatshade = 0;
+        memset(xml->triangletemp.indexref, 0, sizeof(xml->triangletemp.indexref) );
+        memset(xml->triangletemp.s, 0, sizeof(xml->triangletemp.s));
+        memset(xml->triangletemp.t, 0, sizeof(xml->triangletemp.t));
         xml->curpolytype  = TRIANGLE;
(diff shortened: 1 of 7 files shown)


engine/src/cmd/weapon_xml.h
```diff
@@ -104,7 +104,7 @@ struct weapon_info
 private:
     float RefireRate;
 };
-weapon_info getWeaponInfoFromBuffer( char *netbuf, int &size );
+//weapon_info getWeaponInfoFromBuffer( char *netbuf, int &size );
```
BenjamenMeyer, line 107: is this simply not used? We should add a comment here why it's disabled other than compiler warning cleanups.

evertvorster: @Benjamin The text string "getWeaponInfoFromBuffer" occurs exactly twice in all of Vega Strike's source. Once in weapon_xml.h, where it prototypes the function, and once in weapon_xml.cpp where the function is defined. 

I think it's a leftover from the network code, as it is trying to read a netbuffer. 

I have actually removed that line completely in the next update that I am going to push when I have the next warning solved. So, come the next push that line will disappear completely and vex the next programmer no more.

I feel guilty of letting all those builds in Travis run for every small little update, so now I try to do the updates in batches.

BenjamenMeyer: let's try to make that clear at least in the commit comments then you should be golden.

evertvorster:
Not sure how to add more information to the commits. The second commit completely removes the line we chatted about yesterday. 
This commit also tries to implement a copy constructor. It's stolen from gl_light.cpp, and modified slightly. 
I have tried to implement it just as ass-backwards as the original one, returning the value that was sent to it and modifying the original. There is also the case of the "enable" If it's not there, the lights are off everywhere. I don't know how to check the incoming variable... everything I have tried to date has not worked. However, turning on the lights at this stage does not seem to have any ill effects in the playtesting that I have done on this modification.

engine/src/gldrv/gl_light_state.cpp
```diff
@@ -154,9 +154,25 @@ int _GLLightsEnabled    = 0;
...
+{   // Let's see if I can write a better copy operator
+//    memcpy( this, &tmp, sizeof (GFXLight) );
+            this->target  = -1;
+            this->options = 0;
+            memcpy( this->vect, tmp.vect, sizeof (float)*3 );
+            memcpy( this->diffuse, tmp.diffuse, sizeof (float)*4 );
+            memcpy( this->specular, tmp.specular, sizeof (float)*4 );
+            memcpy( this->ambient, tmp.ambient, sizeof (float)*4 );
+            memcpy( this->attenuate, tmp.attenuate, sizeof (float)*3 );
+            memcpy( this->direction, tmp.direction, sizeof (this->direction) );
```
BenjamenMeyer, line 166: this should still be `sizeof(float)*3`

engine/src/gldrv/gl_light_state.cpp
```diff
@@ -154,9 +154,25 @@ int _GLLightsEnabled    = 0;
...
+            this->exp    = tmp.exp;
+            this->cutoff = tmp.cutoff;
+            this->size   = tmp.size;
```
BenjamenMeyer, line 169: I think you missed `occlusion`

BenjamenMeyer (changes requested):
couple minor things to change; looking good overall

engine/src/gldrv/gl_light_state.cpp
evertvorster: Modified it to sizeof ( float ) * 3, as requested

engine/src/gldrv/gl_light_state.cpp
evertvorster: Indeed I did. Thanks for picking up on it.

engine/objconv/mesher/to_BFXM.cpp
```diff
@@ -481,13 +481,19 @@ void beginElement( const string &name, const AttributeList &attributes, XML *xml
     case XML::POLYGONS:
         break;
     case XML::LINE:
-        memset( &xml->linetemp, 0, sizeof (xml->linetemp) );
+        xml->linetemp.flatshade = 0;
+        memset(xml->linetemp.indexref, 0, sizeof(xml->linetemp.indexref) );
+        memset(xml->linetemp.s, 0, sizeof(xml->linetemp.s));
+        memset(xml->linetemp.t, 0, sizeof(xml->linetemp.t));
```
BenjamenMeyer, line 487: So, when doing `sizeof` it's better to use native types:
```
sizeof(float) * 5
sizeof(char)* 10
```
Than do use `sizeof(usertype)`. Reason being, that the variable being referenced (usertype) might be a pointer and will then give the wrong value. That is...

The intent is to clear the data (int :arrow_right: 0), but the parameter to `sizeof` is a pointer, so it calculates the size of a pointer (32-bit, 64-bit) and uses that instead.

For instance, on a 64-bit system:
```
sizeof(uint32) = 4 // (4 byte data type)
sizeof(*uint32) = 8 // (8 byte pointer)
```
Some times that happens to match; for instance the same code on a 32-bit system:
```
sizeof(uint32) = 4 // (4 byte data type)
sizeof(*uint32) = 4 // (4 byte pointer)
```

So I find it's always best to use built-in types to `sizeof` instead - explicit being better than implicit.

For this, code above, it'd be better to do:
```
memset(xml-linetemp.indexref, 0, sizeof(struct line))
```

Actually...you might want to just add a `clear()` on `struct line` insetad of using `memset` directly...probably similar in the other items (triangle, quad, etc) in `to_BFXM.h` too.

evertvorster: For memset(xml->linetemp.indexref, 0, sizeof(struct line)); would that not clear a block of memory the size of the entire struct? I just want to clear the array in this case. 
I've gone with your initial suggestion and specified the type and number at first, but then I searched stackoverload.

Because that clear()  was a little bit of a mystery, as there are a few libraries that provide different versions of it. In short, I did not know how to use it. 
Turns out, in C11, you can just tell the compiler with a plain old "struct = {};" that you want the structure to be empty.  I had a look at the structs, and they are properly defined to allow this sort of thing. Learn something every day!

evertvorster:
I am not going to add more compiler fixes to this PR. 

The only compiler warning that is still left from the bunch that I reported that is still un-addressed is issue #255. That will come in a different PR, as I have a suspicion that the fix won't be trivial. 

Any remaining issues with this PR I'll fix, of course. If the fix in the last commit is good enough, we should be ready to merge.

engine/objconv/mesher/to_BFXM.cpp
BenjamenMeyer: I'm not familiar with the `{}` method. I'll have to look that up before I approve.

Sorry if I wasn't clear enough, the `clear` suggestion was to add a function on the struct to do the work that would just get called.

evertvorster: @BenjamenMeyer , Ah, I see. Yes, that could work, too. If you are not happy with the = {} method, I'll implement that. As far as I know, the = {} method uses automatically generated destructors and constructors. So, it's implicit code, rather than explicit, and you have already expressed a preference for explicit code....

engine/src/gldrv/gl_light_state.cpp
```diff
@@ -338,11 +355,11 @@ void gfx_light::ResetProperties( const enum LIGHT_TARGET light_targ, const GFXCo
...
-        GFXLight t;
-        memcpy( &t, this, sizeof (GFXLight) );
-        t.SetProperties( light_targ, color );
-        changed = RemoveFromTable( false, t );
-        memcpy( this, &t, sizeof (GFXLight) );
+        //GFXLight t;
+        //memcpy( &t, this, sizeof (GFXLight) );
+        this->SetProperties( light_targ, color );
+        changed = RemoveFromTable( false, *this );
+        //memcpy( this, &t, sizeof (GFXLight) );
```
stephengtuggy, line 362: Not sure about this part. Everything else looks good to me.

stephengtuggy (commented):
Looking pretty good overall. I like the `= {}`, personally. It seems clear enough to me, but maybe that's just my opinion.

Perhaps we could add a comment in the code wherever that construct is used, briefly explaining what it does?

engine/src/gldrv/gl_light_state.cpp
evertvorster: @stephengtuggy I will look at it a little more closely now. Thanks for pointing it out. I thought it was just pointlessly copied back and fourth, but now this "changed" function is intriguing me.

engine/objconv/mesher/to_BFXM.cpp
evertvorster: @Benjamin, I have made the clear routine in the structure, and used it in this routine.

engine/src/gldrv/gl_light_state.cpp
evertvorster: @stephengtuggy I have reverted the code to previous behavior. However, I have implemented a copy construct to replace the memsets.

stephengtuggy: @evertvorster Cool, thanks. This looks correct now.

stephengtuggy (approved):
Looks good to me.

engine/src/gfxlib_struct.h
```diff
@@ -646,6 +646,36 @@ class GFXLight
               float exp = 0.0f,
               float cutoff = 180.0f,
               float size = 0.0f );
+    
+    GFXLight(const GFXLight& other)
+    {
+        vect[0] = other.vect[0];
+        vect[1] = other.vect[1];
+        vect[2] = other.vect[2];
```
BenjamenMeyer, line 654: A series of for loops for these would be good for simplicity; this works for now so not blocking.

BenjamenMeyer (approved):

engine/src/gfxlib_struct.h
evertvorster: I did consider for loops. Unfortunately, they take up more vertical space and requires the same amount of typing as the solution I went for here. Also, for the compiler for loops would require counters and more steps. From an optimization point of view what I have done is a manual loop unroll. 

In Lisp you would iterate over the names of the lists as well as the lists themselves, or have some sort of macro. 

Fortunately, this problem is too simple for such a complicated solution.

PR #345, opened 2020-11-23
Finish cpp boost logging

### LINKED ISSUE #204: TODO: Finish converting the C++ code to use Boost logging instead of console output directly
It would probably be a good idea to do this at some point when the codebase has settled down a little bit, so there aren't too many conflicting changes.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
**Closes #204** 

Purpose:
- What is this pull request trying to do? **Finish converting the C++ code to use Boost logging**
- What release is this for? **0.7.x, most likely**
- Is there a project or milestone we should apply this to? **0.7.x, most likely**

### DIFF
--- a/engine/src/aldrv/al_init.cpp
+++ b/engine/src/aldrv/al_init.cpp
@@ -1,3 +1,29 @@
+/**
+ * al_init.cpp
+ *
+ * Copyright (C) Daniel Horn
+ * Copyright (C) 2020 pyramid3d, Stephen G. Tuggy, and other Vega Strike
+ * contributors
+ *
+ * https://github.com/vegastrike/Vega-Strike-Engine-Source
+ *
+ * This file is part of Vega Strike.
+ *
+ * Vega Strike is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Vega Strike is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+
 #ifdef HAVE_AL
 #ifdef __APPLE__
 #include <al.h>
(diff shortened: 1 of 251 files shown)


engine/src/aldrv/al_init.cpp
```diff
@@ -135,7 +159,7 @@ static ALCcontext *context_id = NULL;
 #endif
 bool AUDInit()
 {
-    g_game.sound_enabled = 
+    g_game.sound_enabled =
```
BenjamenMeyer, line 162: Q: What is this suppose to equal?
I see it's on both sides just removing a space, but I don't see an rvalue, and if HAVE_AL is undefined then there won't be an rvalue either (otherwise the rvalue would be the value of `usedoppler`.

Seems like a bug.

BenjamenMeyer (changes requested):
After an initial first pass it looks okay. I did most of the review with whitespace changes disabled, but only after I had already reviewed the first few files. The one change I am making is from that first couple files where I reviewed with whitespace changes enabled, and appears to be a latent bug. I'll take another deeper review later on.

engine/src/aldrv/al_init.cpp
Loki1950: There have been unexplained sound failures for ages this maybe the source

evertvorster (approved):
@stephengtuggy You have been busy!

I just play-tested this code, and it plays really beautifully. I only use Py3, so that part of the codebase is getting a good test from my side. 

Overall impression is that gameplay is smooth, and nothing obviously untoward is happening. I picked a fight with Oswald. This time I was able to fire around ten dumbfires through him without hitting him, but that sort of thing has been happening before this patch. 

Flying around with a fully equipped Vendetta also feels really good, but the game engine starts showing some ... peculiarities at speeds even as low as 300kps, and is zany at 3000kps. Again, not introduced by this patch, it's been doing that before as well. 

Let's get this merged in so we can go bug-hunting!
-Evert-

engine/src/aldrv/al_init.cpp
BenjamenMeyer: let's at least file an issue so we can resolve this. I'll look later on to see if there are others that pop out from this too.

engine/src/aldrv/al_init.cpp
```diff
@@ -149,7 +173,7 @@ bool AUDInit()
     scalevel = game_options.DopplerScale;
     g_game.audio_frequency_mode = game_options.frequency;
     maxallowedsingle     = game_options.MaxSingleSounds;
-    g_game.max_sound_sources = 
+    g_game.max_sound_sources =
```
BenjamenMeyer, line 176: SImilar issue as https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/345/files#r528777404

engine/src/audio/codecs/FFStream.cpp
```diff
@@ -219,38 +244,38 @@ namespace Audio {
...
+
             void decodeFrame()
             {
                 if (!hasPacket())
                     throw PacketDecodeException();
-                
+
                 int dataSize = sampleBufferAlloc;
-                int used = 
+                int used =
```
BenjamenMeyer, line 278: SImilar issue as https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/345/files#r528777404

BenjamenMeyer (approved):
Looks good. Found a couple other spots similar to https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/345/files#r528777404.

Let's file some issues before this is merged and forgotten.

BenjamenMeyer:
Merging since these are latent bugs that were revealed because of this PR.

PR #348, opened 2020-11-28
Travis config: Enable deployment configuration; add Fedora 33

stephengtuggy:

Please answer the following:

Code Changes:
- [x] **This is a CD/build system change only**

Issues:
**N/A**

Purpose:
- What is this pull request trying to do? **Enable packaging/Travis CD for the 0.7.x branch**
- What release is this for? **0.7.0**
- Is there a project or milestone we should apply this to? **0.7.x**


engine/CMakeLists.txt
```diff
@@ -1370,13 +1370,13 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
                 ELSE (USE_PYTHON_3)
                     SET(CPACK_RPM_PACKAGE_REQUIRES "python27, boost-python2")
                 ENDIF (USE_PYTHON_3)
-            ELSE (LINUX_VERSION_ID VERSION_EQUAL 32)
```
BenjamenMeyer, line 1373: does this work? We're still using CMake 3.5 and the docs (https://cmake.org/cmake/help/v3.5/command/if.html) show that the `ELSE` and `ENDIF` need the expression from the first `IF` listed with them. It does seem later version simplified it and removed that requirement though.

BenjamenMeyer (changes requested):
Let's make sure this is compliant with our selected version of cmake.

engine/CMakeLists.txt
BenjamenMeyer: https://packages.ubuntu.com/xenial-updates/cmake shows Xenial (16.04) is still Cmake 3.5

stephengtuggy: Fixed

BenjamenMeyer (approved):
Thanks!

PR #356, opened 2020-12-04
Further refactoring of unit code

royfalk:
**Purpose**
This PR is trying to chip away at unit_generic and split it up into classes based on function: 

- Movable - everything related to movement
- Damageable - shield and armor
- Armed - weapons and maybe radar
- Intelligent - AI functions
- Energetic - mostly generator and energy usage

The idea here is that we could later create a pure virtual proto_movable class (interface in java) and a mock class and then start writing unit tests. 

**Issues**

- During combat, actual speed often starts going up incrementally. It's unclear if this is the actual speed or just a fault of the computer. Going to auto-pilot and spec resolves this.
- Auto pilot has difficulty getting to fighter barracks - this is not new.

Please answer the following:

Code Changes:

**Release**
The first issue is a serious one, so probably 9.0 or 10. @BenjamenMeyer to decide how and if to do this.

**Project or Milestone**
?


engine/src/cmd/ai/ikarus.cpp
```diff
@@ -47,8 +47,14 @@ void Ikarus::ExecuteStrategy( Unit *target )
 void Ikarus::WillFire( Unit *target )
 {
     bool  missilelockp = false;
-    if ( ShouldFire( target, missilelockp ) )     //this is a function from fire.cpp  you probably want to write a better one
+    if ( ShouldFire( target, missilelockp ) ) {
+        // this is a function from fire.cpp  you probably want to write a better one
+        // Roy Falk - This function was actually in unit_generic and moved to armed
+        // The false is coerced into unsigned int for the bitmask.
+        // I suspect that while this compiled, the boolean was not meant to do apply to the first parameter but to the second one -
+        // either beams_target_owner or listen_to_owner
```
BenjamenMeyer, line 55: Checking git history _might_ help reveal and answer

engine/src/cmd/damageable.cpp
```diff
@@ -27,15 +27,23 @@
...
 #include "vs_globals.h"
 #include "configxml.h"
 #include "unit_armorshield.h"
 #include "gfx/vec.h"
-#include "unit_damage.h"
+#include "lin_time.h"
 
 #include <algorithm>
 
+// TODO: make GameConfig support sub sections
```
BenjamenMeyer, line 39: Please make sure there is an issue filed, and then put a link to the issue in the code too.

engine/src/cmd/damageable.cpp
```diff
@@ -430,3 +438,31 @@ float Damageable::currentTotalShieldVal( const Shield &shield )
     return maxshield;
 }
 
+bool Damageable::flickerDamage()
+{
+    float damagelevel = GetHullPercent();
+    static double counter = getNewTime();
```
BenjamenMeyer, line 444: the types should be normalized so it's all working with floats or doubles so that things don't get lost in the mix-ups. Would prefer doubles.

engine/src/cmd/energetic.cpp
```diff
@@ -0,0 +1,197 @@
...
+    if (maxwarpenergy > 0)
+        return ( (float) warpenergy )/( (float) maxwarpenergy );
+    if (jump.energy > 0)
+        return ( (float) warpenergy )/( (float) jump.energy );
+    return 0.0f;
+}
+
+
+
+
```
BenjamenMeyer, line 183: can we please remove the dead space at the end of the file?

engine/src/cmd/unit_generic.cpp
```diff
@@ -1493,81 +1081,11 @@ const std::vector< std::string >& Unit::GetDestinations() const
...
-        if (steady_itts)
-            //** jay
-            curguess = posit+GetVelocity().Cast().Scale( time );
-        else
-            curguess = posit+velocity.Scale( time ).Cast();
-    }
-    return curguess+absposit;
-}
+
+
```
BenjamenMeyer, line 1085: there should only be 1-2 blank lines between functions

engine/src/cmd/unit_generic.h
```diff
@@ -109,114 +113,19 @@ class AsteroidGeneric;
  * Needed by star system to determine whether current unit
  * is orbitable
  */
-enum clsptr
+enum _UnitType
```
BenjamenMeyer, line 116: This would be better as `UnitType`; any particular reason for the `_`?

engine/src/cmd/script/mission.h
```diff
@@ -476,7 +476,7 @@ class Mission
     static double gametime;
     string mission_name;
     void terminateMission();
-    Unit  *call_unit_launch( class CreateFlightgroup*fg, int type /*clsptr type*/, const std::string&destinations );
+    Unit  *call_unit_launch( class CreateFlightgroup*fg, int type /*_UnitType type*/, const std::string&destinations );
```
BenjamenMeyer, line 479: this looks like the type of the parameter should be changed from `int` to `_UnitType`

BenjamenMeyer (changes requested):
some blank line cleanup; a type name correction, and a few other questions.

Generally looks good.

BenjamenMeyer:
@royfalk are the issue you listed new? Or were they existing behavior (even if not yet documented as such)?
If existing behavior, then I see no issue putting this into 0.8.x; but otherwise we need to wait until master is 0.9.x to merge.

royfalk:
> @royfalk are the issue you listed new? Or were they existing behavior (even if not yet documented as such)?
> If existing behavior, then I see no issue putting this into 0.8.x; but otherwise we need to wait until master is 0.9.x to merge.

It looks like the speed issue is new. The auto pilot is a known one.
I just played master and the speed issue doesn't exist. However, missiles are less likely to hit. Very confusing. 
I will continue to investigate.

stephengtuggy:
That increasing speed issue sounds like a showstopper to me. I would want to see it fixed before this gets merged in.

BenjamenMeyer:
@royfalk if you can fix the speed issue then we can put into 0.8.x.

royfalk:
OK. It turned out to be a simple typo. Bug is fixed. I'll try and work on the comments later today.

engine/src/cmd/ai/ikarus.cpp
royfalk: Most of this code is from svn still. The recent commit (11 years ago) is reformatting by uncrustify, which is a legit c code beautifier. The one before that is 18 years ago (!) and simply says: fixed should fire.
Whatever answers history holds are lost to us.

engine/src/cmd/damageable.cpp
royfalk: See #358. Done

engine/src/cmd/energetic.cpp
royfalk: Done.

engine/src/cmd/unit_generic.cpp
royfalk: Done.

engine/src/cmd/unit_generic.h
royfalk: This value was used in the template when I started refactoring. It was always the plan to drop it later. Opened #359 for it.

engine/src/cmd/damageable.cpp
royfalk: It should, but it's important to understand I'm refactoring the overall structure of the code, not the inside of specific functions (unless I have to). This is all "original" code. If I tried to refactor every function in unit class and its sub-classes, the next time you'd hear from me would be a decade from now.
I'd also be very careful about switching to double everywhere. I researched it before and there's some code in save game I think that reads a double and then converts it to two floats or some similar kludge. Exactly the sort of thing to generate the subtle bugs we all love.

engine/src/cmd/script/mission.h
royfalk: As noted above, I'm trying not to touch every kludge I see. Yes, it should be _UnitType and once we switch to enum class, this code right here would fail and we'd need to investigate and it would more resilient after we refactor it.
But this PR is big enough without more code in it and we have #359 for it.

engine/src/cmd/damageable.cpp
BenjamenMeyer: yes, we'll get those things fixed in time. Thanks!

BenjamenMeyer (approved):

BenjamenMeyer:
@royfalk fix up for CI and you should be good to go. Thanks for filing the additional issues.

royfalk:
I don't know why the build failed like this. It feels like the compiler randomly collects include directives from various other header files and adds them to files. Case in point, this failure was due to a missing include. Every other environment was fine with it. Go figure.
In any case, I tested it some more. Merging.

BenjamenMeyer:
@royfalk  that's why I learned long ago to make sure anything I needed in my own file (header/implementation) was directly included. It also lets you drop headers you don't need, or only include them on the implementation side.

Thanks!

Issue #387, opened 2021-01-06
Multiple joysticks cannot be used by a single player


viktorradnai:
Even though the `axis` and `binding` nodes support specifying the joystick number, this does not seem to work in practice. Only the first valid joystick is picked up and the rest ignored. This prevents using things like a HOTAS with separate stick and throttle devices, rudder pedals, etc.

PR #388, opened 2021-01-07
Assign all joysticks to player 0 by default

viktorradnai:

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? **Only basic flight controlled by multiple joysticks was tested.**

Issues:
 #387

Purpose:
This is a trivial workaround to make multiple joysticks work in a single player scenario. The old code must have been intended for split-screen multiplayer where each player has one joystick. Since trying to override it in vegastrike.config doesn't work (see  #387) this change just forces all joysticks to be owned by player 0 by default, which seems like a much more typical modern use case.


BenjamenMeyer (commented):
Looks okay, but sure if the effects yet.
We need a full play test to validate as well. Holding off on my approval until a full play test with several setups (mouse included) has been done by the group. I'll try to do so as well (with mouse) if I get the chance.

Also we might want to port this to 0.7.x too. If validated here we can do that and get it in the 0.7.0 release. Otherwise it'll be in the 0.8.0 release.

stephengtuggy:
We still need to play-test this, don't we? It's unfortunate that this PR has had to wait so long.

Loki1950:
IIRC allowing multiple joysticks was to make it possible for two players to share a single client with split screen and keyboard access adding an other joystick was an extension of that concept not sure if it ever worked correctly.

BenjamenMeyer:
This will need to get updated against current master before we can move forward on this.

LifWirser:
hmm first available or ]null\, ... I maybe missing something . two lstick on singfle yesbut how do two pilot use same joystick? two seperate players with difrerring sticks sure, but do we have multiplayer yet?

Loki1950:
There was an ability for 2 players to share a split keyboard and so play together not really multi-player

BenjamenMeyer:
Closing this due to age.

@viktorradnai Please feel free to update against current master and re-open.

PR #400, opened 2021-01-12
GH Actions - Release: Troubleshoot CD further

stephengtuggy:

Please answer the following:

Code Changes:
- [X] This is a CD change only

Issues:
N/A

Purpose:
- What is this pull request trying to do? Fix the CD for releases, which is still failing
- What release is this for? 0.7.0

### DIFF
--- a/.github/workflows/gh-actions-release.yml
+++ b/.github/workflows/gh-actions-release.yml
@@ -16,17 +16,11 @@ jobs:
       matrix:
         include:
           - FROM:     'ubuntu:focal'
-            COMPILER: 'gcc'
-            FLAGS:    '-DUSE_PYTHON_3=OFF'
-          - FROM:     'ubuntu:focal'
             COMPILER: 'clang'
             FLAGS:    '-DUSE_PYTHON_3=OFF'
           - FROM:     'ubuntu:focal'
             COMPILER: 'clang'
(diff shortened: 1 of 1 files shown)


stephengtuggy:
The last 0.7.x beta release failed with GitHub 502 errors. I suspect conflicts due to different builds trying to write to the same files at the same time. I'm trying to fix that by:

- Turning off fail-fast, so that we get **some** installers, even if others fail
- Removing a couple of the release sub-event-types, so that we don't have 3 release builds running at once
- Removing Ubuntu bionic and focal builds that are duplicates except for the choice of compiler, and thus, try to write to the same filename

.github/workflows/gh-actions-release.yml
```diff
@@ -3,32 +3,24 @@ name: 'GH Actions - Release'
 on:
   release:
     types:
-      - published
       - created
       - edited
-      - prereleased
```
BenjamenMeyer, line 9: We'll have to learn more about the event types; but I suspect we'll need those still. What we might want to do is have some kind of check to see if a target is already there before uploading unless a special flag is set to overwrite the target.

BenjamenMeyer (approved):
I think we may need published and prereleased.. not sure but not going to block on that.
Agreed we only need one build per platform target.

PR #403, opened 2021-01-15
CD: GH Actions - Release: Try a different way of including only .deb's & .rpm's

stephengtuggy:

Please answer the following:

Code Changes:
- [X] This is a build system/CI/CD change only

Issues:

Purpose:
- What is this pull request trying to do? Reduce the number of files per 0.7.x release, so it's not so confusing. Also test a concept that we may end up using with the assets as well.
- What release is this for? 0.7.x


engine/CMakeLists.txt
```diff
@@ -1181,17 +1190,16 @@ ELSEIF (XCODE)
...
-    IF (DEFINED ENV{TAG_NAME})
-        SET(VEGA_STRIKE_PACKAGE_VERSION_STR "$ENV{TAG_NAME}")
-    ELSE (DEFINED ENV{TAG_NAME})
-        SET(VEGA_STRIKE_PACKAGE_VERSION_STR "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
-    ENDIF (DEFINED ENV{TAG_NAME})
+    # IF (DEFINED ENV{TAG_NAME})
+    #     SET(VEGA_STRIKE_PACKAGE_VERSION_STR "$ENV{TAG_NAME}")
+    # ELSE (DEFINED ENV{TAG_NAME})
+    #     SET(VEGA_STRIKE_PACKAGE_VERSION_STR "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
+    # ENDIF (DEFINED ENV{TAG_NAME})
```
BenjamenMeyer, line 1202: since this is above is should go away here

engine/CMakeLists.txt
```diff
@@ -1181,17 +1190,16 @@ ELSEIF (XCODE)
     #  See https://cmake.org/cmake/help/v3.3/module/CPackDMG.html
     # Bundle -> Compressed Disk Image
     # PackageMaker - see https://cmake.org/cmake/help/v3.3/module/CPackPackageMaker.html
-    SET(CPACK_GENERATOR "DragNDrop" "BUNDLE" "TBZ2" "TGZ" "TXZ")
+    SET(CPACK_GENERATOR "DragNDrop" "BUNDLE" "TXZ")
 ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
     MESSAGE("-- Configuring Packaging for Linux")
-    SET(CPACK_GENERATOR "TBZ2" "TGZ" "TXZ")
-    # Linux
+    SET(CPACK_GENERATOR "TXZ")
```
BenjamenMeyer, line 1196: I'd prefer to keep TBZ2 on Linux as well. TGZ can get dropped since GH provides that.

.github/workflows/gh-actions-release.yml
```diff
@@ -83,5 +95,6 @@ jobs:
       uses: skx/github-action-publish-binaries@master
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        INSTALLER_EXT: ${{ matrix.INSTALLER_EXT }}
       with:
-        args: 'packages/Vega-Strike_v*.*'
+        args: 'packages/*.${INSTALLER_EXT}'
```
BenjamenMeyer, line 100: We should have another one that manages the tarballs; and just one of them - we don't need that for every platform.
GH provides the ZIP and TGZ files automatically; so we just need to add the TBZ2 and TXZ files; would probably be good to do that as a separate job line so that it could be as platform agnostic as possible too...so may be that should be a separate PR.

BenjamenMeyer (changes requested):
Couple minor changes and a suggestion for a follow-on PR.

.github/workflows/gh-actions-release.yml
stephengtuggy: Pardon me, but I don't understand why we need so many different tarball formats. What's wrong with ZIP and TGZ?

BenjamenMeyer: Zip is primarily used by Windows folks.
Traditionally open source used tgz and then tbz2 and more recently txz; however, it's typical that folks like us would provide all three so each distro can select the format they use.

Some Linux folks will only grab tgz; others only tbz2; and others yet only txz. so if we want to maintain a diverse set of users we'll want to maintain a diverse selection of archives formats too.

As we grow and get more integrated with distros we'll be able to refine that to what the distro level maintainers will use. We don't quite get the luxury yet that way while still achieving our goals in expanding the user and dev base.

engine/CMakeLists.txt
stephengtuggy: Done

.github/workflows/gh-actions-release.yml
stephengtuggy: Hmm. I see.

And what should be in these tarballs? Binaries? Source? Both?

stephengtuggy: Oh, I think I see. We need a single "source tarballs" build in the matrix. That would build TBZ2 and TXZ. Then we **also** need all 4 tarball formats with the binaries for unknown/not yet supported Linux distributions.

Moreover, for the Linux distros we support now, we should build **only** .RPM or **only** .DEB packages, as applicable.

Does that sound right? Take a look at what I've done in #404 , targeting the master branch, and let me know if it makes sense to you, @BenjamenMeyer .

BenjamenMeyer: Yep. Sounds about right.

engine/CMakeLists.txt
stephengtuggy: Fixed

.github/workflows/gh-actions-release.yml
stephengtuggy: OK, I think I'm done implementing these changes. Have a look when you get a chance.

engine/CMakeLists.txt
```diff
@@ -1407,7 +1409,7 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
 ELSE ()
     MESSAGE("-- Configuring Packaging for Unknown Platforms - \"${CMAKE_SYSTEM_NAME}\"")
     # Unknown Platform --> Just do the various compressed tarballs
-    SET(CPACK_GENERATOR "TBZ2" "TGZ" "TXZ")
+    SET(CPACK_GENERATOR "TXZ")
```
BenjamenMeyer, line 1412: Not blocking as it doesn't affect CI: This can really be any/all of them.
It doesn't affect CI as we don't do a build on an unknown platform; so this would really only apply for local builds at the moment.

BenjamenMeyer (approved):
One minor non-blocking comment. Let's get this merged and builds going again.

PR #421, opened 2021-02-09
Remove TemplateManager and VSFileXMLSerializer

royfalk:
Purpose:
- What is this pull request trying to do? Remove older code. In this case, my target was VSFileXMLSerializer. 
- What release is this for? The coming one.

Template Manager is part of the audio and does not appear to be used at the moment, but it's relatively clean and modern code. Audio works fine without it. I wonder if it was a replacement to current code.
I also had to gut the test file.

Code Changes:

Issues:
- One jump had a rendering freeze for 2-3 seconds. Following jump was fine.


Loki1950 (approved):
Not sure about the audio potion IIRC kluass's was trying to modernism the audio

BenjamenMeyer:
NOTE: Seems the builds that failed had an issue with Py3 and Boost-Python. Not sure why but that's the build error.

stephengtuggy:
I will try to review this PR this weekend. Can we get the builds passing before then?

BenjamenMeyer (changes requested):
Blocking until builds pass.

engine/CMakeLists.txt
```diff
@@ -808,9 +808,9 @@ IF (USE_SYSTEM_BOOST)
         MESSAGE("-- Boost_1_67_Or_Later_Result: ${Boost_1_67_Or_Later_Result}")
         MESSAGE("-- USE_PYTHON_3: ${USE_PYTHON_3}")
         IF (Boost_1_67_Or_Later_Result)
-            SET(BOOST_PYTHON_COMPONENT "python${PYTHON_VERSION_MAJOR}${PYTHON_VERSION_MINOR}")
+            SET(BOOST_PYTHON_COMPONENT "python38")
```
stephengtuggy, line 811: Why tie Boost-Python to a particular Python version, on a particular set of platforms/variants? Please revert this, or find another way to do it. These `python38` changes are probably why the build is breaking.

stephengtuggy (changes requested):
The Boost_Python changes here are probably what is breaking the build.

Apart from that, I'm worried that the positional audio functionality will no longer work. It works for me at present; does it not work for you guys?

Is there some other code elsewhere that takes care of positional audio, that I'm just not seeing?

royfalk:
Sorry. For some reason I got a whole bunch of emails about changes to the packages and none for this PR.
As I noted, this code is not actually used yet.
Now that I've noticed it, I'll try and get to it today or tomorrow.

royfalk:
OK. So build passes but CodeQL is complaining about code I didn't touch. 
I fixed the CMakeList file, so I think we're good to go.

stephengtuggy (approved):
OK. Looks fine to me at this point.

BenjamenMeyer (approved):

Issue #391, opened 2021-01-07
Vega Strike v0.7.0-beta1 Ubuntu bionic py3 installer missing at least one boost dependency


stephengtuggy:
It doesn't specify the Boost date-time library as a .deb dependency, even though the Vega Strike binary needs it to run.

stephengtuggy:
Well, unfortunately, this still isn't fixed. Vega Strike v0.7.0-beta10 encounters the same error as before.

Maybe I'm referencing the wrong Boost version? Maybe it should be 1.65.1 instead of 1.58.0? All the more reason to use a Boost version variable, I guess.

BenjamenMeyer:
@stephengtuggy does boost-datetime end up in the `dependency.list` file generated by `scripts/package`? It directly checks against what `ldd` reports are dependencies of the binary and then does a reverse lookup to determine which packages own the library, dropping the unique results into `dependency.list`.

stephengtuggy:
@BenjamenMeyer `boost-datetime` probably does end up in `dependency.list`. But for some reason, the `dependency.list` mechanism has never worked. It seems to fail at some point after the initial list gets generated, perhaps during sorting/deduplication. If you want to take a stab at troubleshooting that whole mechanism, that would probably be a worthwhile effort.

BenjamenMeyer:
I reviewed the `scripts/package` and it does seem to be working just fine; though it didn't have a couple changes with the binaries as it expected them in different locations. I've got a couple branches staged for the fix on that part.

BenjamenMeyer:
#423 fixes 0.7.x and #424 fixes master for the script/package; we should verify the package dependencies in the next release

BenjamenMeyer:
NOTE: This is the Debian dependencies using #424:
```
libpython3.8, libjpeg62, libpng16-16, freeglut3,
libgtk-3-0, libvorbis0a, libopenal1, libsdl-gfx1.2-5, xdg-utils,
libgl1, libopengl0,
libboost-python1.67.0,
libboost-log1.67.0,
libboost-date-time1.67.0,
libboost-chrono1.67.0,
libboost-atomic1.67.0
```

BenjamenMeyer:
@stephengtuggy it could be a order of operations issue.

[...]

PR #426, opened 2021-02-12
Bug Fix: run cmake before packaging

BenjamenMeyer:
There's a circular relationship between CMake and the dependency.list
file such that CMake has to be run first to build the binaries,
then script/package to build the dependencies, but then CMake has
to be run again for CPack to pick up the dependency list for the
packages to be correct as intended.

Please answer the following:

Code Changes:
- [x] CI only

Issues:
- #391

Purpose:
- Make the build system re-run cmake after building the dependencies so it picks up the right ones
- Synchronizes and additional change from #424 to 0.7.x

NOTE: This alleviates the issues external to the CMake configuration itself.

### DIFF
--- a/script/package
+++ b/script/package
@@ -154,6 +154,9 @@ function copyFiles()
 
 function buildPackages()
 {
+	# before running make package regen the Cmake data so that
+	# it picks up the dependencies just generated
+	cmake "${SRC_DIR}"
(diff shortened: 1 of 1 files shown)


Loki1950 (approved):
See that this just cleaning up on #424

PR #427, opened 2021-02-12
Windows 10 build support

ministerofinformation:

Please answer the following:

Code Changes:

Issues:

Purpose:
- What is this pull request trying to do? **Build VS tools/projects on Windows**
- What release is this for? 

Apologies for having built this in my fork's master -- will do all subsequent work in a branch off the fork (still retraining some of my old habits from pre-git development models)


engine/setup/src/c/setup.cpp
```diff
@@ -87,11 +87,31 @@ static void changeToProgramDirectory( char *argv0 )
...
-    GetModuleFileName( NULL, argv0, 65534 );
+    char **argv;
+    int  argc;
+    //strcpy( argv0, origpath );
+    //GetModuleFileName( NULL, argv0, 65534 );
+    
+    LPWSTR* szArglist;
+    int i;
+    argv = (char**)malloc(3*sizeof(char*));
+    szArglist = CommandLineToArgvW(GetCommandLineW(), &argc);
```
BenjamenMeyer, line 98: As these are wide-string (multi-byte) functions; then it should be sized based on `wchar_t` instead of `char`, same below. Otherwise the ASCII version of GetCommandLineW() should be used - GetCommandLineA().

It would certainly be preferable to stick to the wide-string version if we can; one less thing to do later if we want to support multiple languages all the way through.

ministerofinformation: Good catch - we should probably use the ascii version for the moment because I think the rest of setup is still assuming char*  in other places -- but would be best to shift everything to wide-char in the not-too-distant future.

ministerofinformation: Simplified that whole block to be ASCII (for the moment) and much shorter - there are actually intrinsics for __argc and __argv available, so the windows side just has to declare argv and argc and set them to the intrinsics.

engine/CMakeLists.txt
```diff
@@ -820,10 +843,9 @@ IF (USE_SYSTEM_BOOST)
                     SET(BOOST_PYTHON_COMPONENT "python-py3${PYTHON_VERSION_MINOR}")
                 ENDIF (LINUX_CODENAME STREQUAL xenial OR LINUX_CODENAME STREQUAL stretch)
             ENDIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
-
         ELSE (Boost_1_67_Or_Later_Result)
```
stephengtuggy, line 846: This needs to be changed to `ELSE (WIN32)`

engine/CMakeLists.txt
```diff
@@ -636,25 +644,37 @@ IF (ENABLE_PIE)
...
+    ADD_COMPILE_OPTIONS("-no-pie")
+    LINK_LIBRARIES("-no-pie")
+  ENDIF (NOT MSVC)
 ENDIF (ENABLE_PIE)
 
+IF (MSVC)
+    SET(COMPILER_SPECIFIC_FLAGS "/FI\"${Vega_Strike_BINARY_DIR}\"/config.h  -D__x86_64__ /Z7")
+    set(TOOLCHAIN_SPECIFIC_LINK_FLAGS "/DEBUG:FULL")
+ELSE (MSVC)
+    SET(COMPILER_SPECIFIC_FLAGS "-include config.h -pipe -Og -g3 -Wall -std=c++11 -fvisibility=hidden")
```
stephengtuggy, line 657: `-Og` and `-g3` are, technically, specific not only to the gcc compiler, but to the Debug configuration as well. This will do for now, I suppose, but we'll want to fix this up at some point.

engine/setup/src/c/setup.cpp
```diff
@@ -68,17 +68,17 @@ static void changeToProgramDirectory( char *argv0 )
     }
 #endif
     char *parentdir;
-    int   pathlen = strlen( program );
+    size_t   pathlen = strlen( program );
```
stephengtuggy, line 71: Good catch!

engine/setup/src/include/general.cpp
```diff
@@ -104,7 +104,7 @@ char *ptr_copy(const char *string)
 #ifdef __STDC_LIB_EXT1__
     strcpy_s(alloc, buf_size, string);
 #else
-    strncpy(alloc, string, buf_size);
+    strncpy_s(alloc, buf_size, string, buf_size);
```
stephengtuggy, line 107: Hmm. There may be a hidden compatibility issue here or two, to revisit at some point. But probably not a big deal. The code seems to be building on all presently supported platforms, at any rate.

engine/src/cmd/unit_const_cache.cpp
```diff
@@ -2,3 +2,15 @@
...
+{
+    return a.fac == b.fac && a.key == b.key;
+}
+
+bool operator<(const StringIntKey& a, const StringIntKey& b)
+{
+    if (a.fac != b.fac)
+        return a.fac < b.fac;
+    return a.key < b.key;
+}
```
stephengtuggy, line 16: git wants there to be an extra blank line at the end of the file. (Small quibble)

engine/src/gfx/mesh_gfx.cpp
```diff
@@ -57,7 +57,11 @@
 #include "universe_util.h"
 #include "universe.h"
 #include <utility>
+#ifdef _WIN32
+// What do we need from unistd? - let's find out :)
```
stephengtuggy, line 61: LOL

stephengtuggy (approved):
Found a couple of minor quibbles, but nothing blocking (IMO).

Overall, great job, @ministerofinformation . This represents a massive step forward for the project, to have Windows building again!

engine/setup/src/include/general.cpp
BenjamenMeyer: Just stick to `strncpy`; `strncpy_s` is an MSVCism and doesn't really buy you anything over `strncpy`; worse case we have to add a header to rename `_strncpy` to `strncpy` or something like that. But let's keep it platform agnostic as much as possible.

engine/setup/src/c/setup.cpp
```diff
@@ -160,11 +161,20 @@ int main( int argc, char *argv[] )
             //Test if the dir exist and contains config_file
             bogus_int = chdir( origpath );
             bogus_int = chdir( (*vsit).c_str() );
-            FILE *setupcfg = fopen( "setup.config", "r" );
+            FILE* setupcfg;
+#ifdef WIN32
+            fopen_s(&setupcfg, "setup.config", "r");
+#else
+            setupcfg = fopen("setup.config", "r");
+#endif
```
BenjamenMeyer, line 169: `fopen` should work on Windows too just fine. I know, MSVC complains about security when not using their special `_s` functions, but let's keep to cross-platform standards as much as we can. (The `_s` junk doesn't really buy you anything; just ties stuff to MSVC.)

engine/src/audio/CodecRegistry.cpp
```diff
@@ -57,22 +57,23 @@ namespace Audio {
             codecPriority[codec] = priority;
             nameCodec[codec->getName()] = codec;
 
-            BOOST_LOG_TRIVIAL(trace) << "Registering codec " << codec->getName().c_str();
+            // static initializer ordering issue between boost logger and other code?
+            std::cerr << "Registering codec " << codec->getName().c_str();
```
BenjamenMeyer, line 61: FYI - we're moving away from using this methodology to using the BOOST_LOG_TRIVIAL everywhere we can; the exception being the code before BOOST_LOG can get setup.

engine/src/cmd/unit_const_cache.h
```diff
@@ -3,35 +3,38 @@
 #include "hashtable.h"
 #include <string>
 #include <gnuhash.h>
+#include <functional>
 
 class Mesh;
-class ConstHasher;
 
-class StringIntKey
+struct StringIntKey
```
BenjamenMeyer, line 10: Per standard, struct is just a class with public by default. However, I've found that by convention (and compat with C) it's better to keep struct to data only stuff, and class for data+function stuff. $0.02

engine/CMakeLists.txt
BenjamenMeyer: `-0g` and `-g3` enable us to do release with debug info ;)

BenjamenMeyer (changes requested):
FYI - I have a particular aversion to Microsoft's `_s` functions; they don't add anything of value and just make it harder to have cross-platform code.  That said, return values can often differ between the POSIX functions between platforms (especially snprintf/vsnprintf where you need to detect buffer sizes) so often glue headers are needed any way. Any how...I'd rather see these kept to cross-platform functions than use MS specific functions.

Otherwise the biggest thing is moving reverting back to using the BOOST_LOG instead of std::cerr as we're trying to move to a unified logging system.

engine/src/audio/CodecRegistry.cpp
ministerofinformation: Yeah - I thought I was having an initializer ordering issue with the logger (because where it was failing, before main, didn't make much sense). However, reverting this back works just fine on the release target, so this was just another instance of the debug target being broken (with internet browsing for similar errors seeming to indicate the likely culprit being the boost libraries being compiled with a different iterator debugging level flag than the rest of the code. This causes failures in things like boost internal stringcopy functions and other inscrutable sadness). I'm making a pass over this and the other comments and should have an update committed later today.

engine/CMakeLists.txt
```diff
@@ -651,21 +651,27 @@ ELSE (ENABLE_PIE)
...
-    set(TOOLCHAIN_SPECIFIC_LINK_FLAGS "/DEBUG:FULL")
+    SET(COMPILER_SPECIFIC_FLAGS "/FI\"${Vega_Strike_BINARY_DIR}\"/config.h  -D__x86_64__ ")
+    SET(COMPILER_SPECIFIC_DEBUG_FLAGS "/Z7 ") # also used for release-with-debug -- can change later
+    SET(COMPILER_SPECIFIC_PROFILER_FLAGS "/Z7 /GL ") # not tested -- requires additional linker flag specialization for profiling run ( /FASTGENPROFILE) and profile use ( /USEPROFILE)
+    SET(COMPILER_SPECIFIC_RELEASE_FLAGS " ")
+    SET(TOOLCHAIN_SPECIFIC_LINK_FLAGS "/DEBUG:FULL ")
 ELSE (MSVC)
-    SET(COMPILER_SPECIFIC_FLAGS "-include config.h -pipe -Og -g3 -Wall -std=c++11 -fvisibility=hidden")
-    set(TOOLCHAIN_SPECIFIC_LINK_FLAGS "-Bsymbolic ")
+    SET(COMPILER_SPECIFIC_FLAGS "-include config.h -pipe -Og -g3 -Wall -std=c++11 -fvisibility=hidden ")
```
stephengtuggy, line 660: Still need to take out the `-Og -g3` from this line

engine/CMakeLists.txt
```diff
@@ -651,21 +651,27 @@ ELSE (ENABLE_PIE)
...
+    SET(COMPILER_SPECIFIC_DEBUG_FLAGS "-Og -g3 ")
```
stephengtuggy, line 661: That's a lot better

engine/CMakeLists.txt
```diff
@@ -843,7 +849,7 @@ IF (USE_SYSTEM_BOOST)
                     SET(BOOST_PYTHON_COMPONENT "python-py3${PYTHON_VERSION_MINOR}")
                 ENDIF (LINUX_CODENAME STREQUAL xenial OR LINUX_CODENAME STREQUAL stretch)
             ENDIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
-        ELSE (Boost_1_67_Or_Later_Result)
+        ELSE (WIN32)
```
stephengtuggy, line 852: Excellent.

engine/setup/src/include/general.cpp
```diff
@@ -101,11 +101,7 @@ char *ptr_copy(const char *string)
         fflush(stderr);
         exit(-1);
     }
-#ifdef __STDC_LIB_EXT1__
-    strcpy_s(alloc, buf_size, string);
-#else
-    strncpy_s(alloc, buf_size, string, buf_size);
-#endif
+    strncpy(alloc, string, buf_size);
```
stephengtuggy, line 104: @BenjamenMeyer , I disagree re. the `_s` functions. They do make for more secure code, by catching things like buffer overflow conditions.

_s functions are _slightly_ less cross-platform compatible. But only slightly, at this point in time, AFAIK. I believe any reasonably up-to-date Linux or macOS platform is going to `#define` `__STDC_LIB_EXT1__`, meaning that strcpy_s will work.

We can agree to disagree everywhere else. But in the case of this particular function, I spent a lot of time carefully crafting and researching it to be as secure and cross-platform-compatible as possible. I really would rather not see my code here simply swept away. 😜 

Instead, this particular `#if` block should probably read something like this:

```C++
#if defined(__STDC_LIB_EXT1__) || defined(WIN32)
    strcpy_s(alloc, buf_size, string);
#else
    strncpy(alloc, string, buf_size);
#endif
```

engine/src/audio/CodecRegistry.cpp
```diff
@@ -58,22 +58,22 @@ namespace Audio {
             nameCodec[codec->getName()] = codec;
 
             // static initializer ordering issue between boost logger and other code?
-            std::cerr << "Registering codec " << codec->getName().c_str();
+            BOOST_LOG_TRIVIAL(trace) << "Registering codec " << codec->getName().c_str();
```
stephengtuggy, line 61: @BenjamenMeyer @ministerofinformation This code may work in practice -- on all platforms, even -- but technically, it's incorrect. Even on Linux, I see this output before main() has had a chance to initialize Boost logging. That's not supposed to happen.

I don't think we need to spend any more time on it in this PR. This is fine for now. I think I'll file a follow-up issue though. The solution would be either to A) use cerr here; or B) put the first Boost logging init function in its own static initializer block, that (hopefully) runs before this one.

engine/src/cmd/unit_const_cache.h
```diff
@@ -7,8 +7,9 @@
 
 class Mesh;
 
-struct StringIntKey
+class StringIntKey
```
stephengtuggy, line 10: As long as this code builds and runs everywhere, I'm fine with it. I think the issue with a few of the classes vs. structs was that the Microsoft linker didn't like them. But if this is an exception, then that's cool.

stephengtuggy (changes requested):
I see a couple of things still that I'd like to see changed. Hopefully we can get them resolved quickly, and get this PR merged. 👍🏽

engine/setup/src/include/general.cpp
ministerofinformation: The _s 4 argument version seems to catch both overread as well as overwrite - you could also block overread by forcing the size argument to be the min of the two buffer sizes - either way, I'll wait for you and Benjamin to discuss this before making any changes.

engine/CMakeLists.txt
ministerofinformation: DOH.

engine/src/audio/CodecRegistry.cpp
ministerofinformation: This code definitely runs before main. Reverting to what we had pre-PR and filing a new issue does help to split the issues of Windows Compilation (which I originally _thought_ this was part of, but turns out to be distinct from) and initialization order dependencies for logging.

engine/CMakeLists.txt
ministerofinformation: (Waiting for convergence on the other outstanding strncpy_s discussion to fix this with a single commit)

engine/setup/src/include/general.cpp
ministerofinformation: @BenjamenMeyer - just making sure you saw this particular sub-thread, among the various reply sequences.

BenjamenMeyer: The 'n' version does buffer overflow detection just fine - that's the whole point of the n version, you tell it the maximum size of the buffer.

I know the history of the _s versions.

stephengtuggy: Okay. So I did do quite a bit of research and work on this code, but clearly there is still room for improvement.

Here is my latest thought: Since Microsoft C++ has [_strdup](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup-mbsdup?view=msvc-160), why don't we just use that when building on Windows? I.e., on line 94, add:

```C++
#elif defined (_WINDOWS)
    return _strdup(string);
```

before the `#else`.

Then leave lines 104 and following as they were before this PR. At least for now.

Does that solution seem like it will work? Is it acceptable to everyone?

Upon further consideration, I suppose `strncpy` isn't really any less secure than `strcpy_s`. At least in practice. In a future PR, we might be able to simply this code further to just use `strncpy` at line 104. As long as the Windows case is handled further up, then that shouldn't trigger any superfluous warnings. Hopefully.

stephengtuggy: Admittedly, the `__STDC_LIB_EXT1__` stuff is a pain to use. If we can get away without it, and without triggering any unnecessary compiler warnings or possible security vulnerabilities, then so much the better.

BenjamenMeyer: Reading through http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2573.pdf (latest draft), https://web.archive.org/web/20181230041359/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf (C17), and http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf (C11) - they basicalliy made it a part of the standard that is entirely optional, and of the various compilers out there most don't implement it; MS is one of the few that does but event they're noted as not complying with it (https://en.wikipedia.org/wiki/C11_(C_standard_revision)#cite_note-21 good discussion pointed to about the topic).

Ultimately we may find it more beneficial to use various parts of C++ Streams or Boost to negate the need for them any way, and as long as we're not doing network functionality there's isn't much of a security risk. As I said earlier, I'd rather us use something that is more widely available than just Microsoft compilers, and more standard. You'd have an easier time convincing me if GCC and Clang had implemented them too, but even then we'd still be running into the fact that Microsoft's implementation is different from the standard.

Per the MS warnings, I've had to mitigate them in the past, and did so using a Windows specific header that used macros to redefine the `_*` versions to their POSIX counter parts:

```
#ifndef ODD_BALL_WINDOWS_SUPPORT___
#define ODD_BALL_WINDOWS_SUPPORT___

#ifdef _WINDOWS

#define strdup _strdup

#endif

#endif ODD_BALL_WINDOWS_SUPPORT___ 
```

I also had to define custom versions of snprintf to make the return values more normative between GCC and MSVC as if you're checking them for buffer sizes (in order to allocate the right amount of memory) the MS version varies from the POSIX version (allowed per standard). Ultimately doing that kind of stuff becomes a really hairy situation so it's better to use a library that will just do it for you so we don't have to worry about it.

Doing those kinds of refactors should be one of our goals of 0.9.x as they'll probably be more extensive across the whole code-base.

stephengtuggy: All right. Sounds like this discussion deserves its own ticket (filed [here](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/430) ). It also sounds like we can leave the code as it is on this branch at the moment, and come back to it for 0.9.x. Seem reasonable?

BenjamenMeyer: @stephengtuggy sounds good to me.

ministerofinformation: @stephengtuggy and @BenjamenMeyer -- Updated. I believe I have now tagged every relevant location with a comment -- please give it all another pass over and let me know if I missed anything.

stephengtuggy (approved):
Whew! That's a lot of warnings and TODO's!

Okay, looks good for now.

stephengtuggy:
Two play-testing tasks remain:

1. Docking to a non-planet location
2. Buying a ship

stephengtuggy:
...And I have finished play-testing. The code works beautifully. This PR can be merged, as far as I'm concerned. Just needs approval from @BenjamenMeyer .

engine/CMakeLists.txt
```diff
@@ -1092,7 +1128,11 @@ SET(TST_LIBS vegastrike-engine_com vegastrike-OPcollide ${TST_LIBS})
 INCLUDE_DIRECTORIES(${TST_INCLUDES})
 
 TARGET_LINK_LIBRARIES(vegastrike ${TST_LIBS})
-SET_TARGET_PROPERTIES(vegastrike PROPERTIES LINK_FLAGS "-L/usr/lib ${TST_LFLAGS}")
+IF (MSVC)
+    SET_TARGET_PROPERTIES(vegastrike PROPERTIES LINK_FLAGS "${TST_LFLAGS}")
+ELSE (MSVC)
+    SET_TARGET_PROPERTIES(vegastrike PROPERTIES LINK_FLAGS "-L/usr/lib ${TST_LFLAGS}")
+endif (MSVC)
```
BenjamenMeyer, line 1135: All cmake directives should be uppercase

BenjamenMeyer (approved):
One minor quibble regarding cmake directives.

@ministerofinformation thanks!

engine/CMakeLists.txt
ministerofinformation: yeah, that's me fighting the editor when modifying the cmake file in Visual Studio 2019 (I should just use emacs instead :-P ) - it wants to "autocorrect" every cmake directive to lowercase -- :it is a mystery: -- and I must have missed that one :-|

Pushing the1-liner fix for this; assuming that the checks all re-run successfully, I'll do the merge once they complete.

BenjamenMeyer: @ministerofinformation I know there's a few others; but either way - feel free to merge once it passes the PR gates since it's just build tooling changes.

Issue #419, opened 2021-02-07
When the Vega Strike engine fails to find the assets, the resulting error message is very unclear


stephengtuggy:
We should have a clear, specific error message for this case.

stephengtuggy:
Relates to #321 
Relates to #417

BenjamenMeyer:
We really should have this in 0.7.x so I'm going to try to work through it.

BenjamenMeyer:
Okay so working through some code now to try to figure this out; prepped the CMake and docs to do two things:
1. Install both `vegastrike-engine` and `vegastrike` binaries
2. Updated the manpages for the delta between them
3. Install man pages as both `vegastrike.1` and `vegastrike-engine.1`

The difference between `vegastrike` and `vegastrike-engine` is that `vegastrike-engine` *requires* the data directory to be specified, while `vegastrike` continues with the legacy behavior.

As Cmake doesn't support symlinks it instead installs two copies of the same binary using both names. The code, however, checks hows it was called (in true Unix fashion) to decide which mode to operate in.

Still trying to figure out how to get it to behave as desired but the infrastructure is there to do it with minimal effort once I figure out the code.

Working on master but intending to backport these changes to 0.7.x after it's working.

Goal: Eventually turn over the `vegastrike` binary with it's legacy behavior to VS:UtCS Assets Package as a wrapper to `vegastrike-engine`. This is just the first step on the journey. If all goes well, PWCU could do a similar wrapper for their assets with their own desired binary name (privateer, pcwu, etc).

evertvorster:
Vegastrike-engine is used for different datasets, so it needs to be told upon which data set to operate.
This is why we have been adding a small script to the data, to find and run the game in the proper context. 

Running "vegastrike" on the command line should default to finding the vegastrike data, and operating on that. Currently, if the vegastrike data directory is not found, it has the unclear error message that is the cause of this report. 

In my humble opinion, vegastrike should also look in it's config file, to see if the data directory is specified there, and failing that, should give an error message like:

No game data found. 
Please specify where the data to load is located with "vegastrike -d[data_dir]" or run the script supplied with the game data.

stephengtuggy:
How about if we keep this ticket just about the error message? And discuss separating the binaries, etc. on the respective tickets for those

BenjamenMeyer:
[...]

PR #432, opened 2021-02-25
Bug Fix: Legacy Data Directory Mode Support

BenjamenMeyer:
If VS is called with the binary name `vegastrike` then do not
require that a data directory be specified and add the default
locations of the VS:UtCS data directories for use.

If VS is called with the binary name `vegastrike-engine` then
require that the data directory be specified and do not attempt
to search for it.

NOTE: the `vegastrike` command-line should eventually be
transferred to the VS:UtCS assets installation thereby negating
the need for this logic here.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation (to degree necessary to show it starts the application)

Issues:
- #419 

Purpose:
Introduces a legacy mode for helping the community transition from `vegastrike` being an executable representing the engine as a whole and it being that of the `VS:UtCS` game assets. In legacy mode `vegastrike` should behave like it did in previous releases; while `vegastrike-engine` exhibits the new behavior of requiring the data directory to be specified.

NOTE: I've tested this to show that it both the `vegastrike` and `vegastrike-engine` work as intended:
- `vegastrike`:
  - starts from any where
- `vegastrike-engine`
  - if current directory is not the assets then requires the directory to be specified
  - if current directory is the assets location then then start

NOTE: Some code is still needed for Win32 support. Added some documentation for what environment variables should be used to map it. However, I'm not familiar with the VS:UtCS install on Windows so it's just a best guess.

NOTE: This needs to get ported to 0.7.x once merged here on master.


BenjamenMeyer:
based on the failed GH Action.. I'm thinking something got missed in the CMakeLists.txt. my local build has the `vegastrike` binary available from the `make install` that `make package` does...so not sure what got missed.

NOTE: Make install is suppose to install a copy of `vegastrike-engine` as `vegastrike` since CMake doesn't support symlinking binaries.

Loki1950 (approved):

BenjamenMeyer:
Okay - found the issue; forgot to update the ancillary scripts. Hopefully should be fixed now.

doc/man/vegastrike-engine.1
```diff
@@ -27,11 +33,20 @@ vegasettings
 .PP
 Program to create and edit the settings for Vega Strike.
 .SS
-vegastrike
+vegastrike-engine
 .PP
 Launch Vega Strike and allows the user to select a campaign to run or load a saved game.
+This uses the legacy behavior of defaulting to the local directory for the game assets.
```
stephengtuggy, line 39: I'm confused. Aren't we repeating ourselves here?

doc/man/vegastrike-engine.1
```diff
@@ -27,11 +33,20 @@ vegasettings
...
 .PP
 Saved games must be saved at a base or planet.
+.SS
+vegastrike
+.PP
+Same as
+.ft I
+vegastrike-engine
+.ft P
+except that it uses the legacy behavior of defaulting to the local directory for the game assets.
```
stephengtuggy, line 49: I'm confused. Aren't we repeating ourselves here? ;-)

engine/src/vsfilesystem.cpp
```diff
@@ -707,14 +725,22 @@ void LoadConfig( string subdir )
         BOOST_LOG_TRIVIAL(info) << boost::format("DATADIR - Found a datadir in config, using : %1%") % game_options.datadir;
         datadir = game_options.datadir;
     } else {
-        BOOST_LOG_TRIVIAL(info) << boost::format("DATADIR - No datadir specified in config file, using : %1%") % datadir;
+        if (true == legacy_mode) {
+            BOOST_LOG_TRIVIAL(info) << boost::format("DATADIR - No datadir specified in config file, using : %1%") % datadir;
+        } else {
+            BOOST_LOG_TRIVIAL(fatal) << boost::format("DATADIR - No datadir specified in config file");
```
stephengtuggy, line 731: Just to make sure we're on the same page here, `BOOST_LOG_TRIVIAL(fatal)` does not, by itself, exit the program. Call `VSExit` to do that.

engine/src/vsfilesystem.cpp
```diff
@@ -707,14 +725,22 @@ void LoadConfig( string subdir )
...
 
     string universe_file = datadir + "/" \
         + vs_config->getVariable( "data", "universe_path", "universe" ) + "/" \
         + vs_config->getVariable( "general", "galaxy", "milky_way.xml" );
     BOOST_LOG_TRIVIAL(debug) << "Force galaxy to " << universe_file;
-    Galaxy galaxy = Galaxy(universe_file);
+    try {
+        Galaxy galaxy = Galaxy(universe_file);
+    } catch (std::exception &e) {
+        BOOST_LOG_TRIVIAL(fatal) << boost::format("Error while loading configuration. Did you specifcy the asset directory? Error: %1%") % e.what();
```
stephengtuggy, line 742: Three things:

1. Maybe catch a more specific exception type?
2. Call `VSExit`, as above.
3. `specifcy` should be `specify`.

engine/src/main.cpp
```diff
@@ -117,6 +117,11 @@ boost::shared_ptr<VSFileSystem::console_log_sink> VSFileSystem::pConsoleLogSink
 
 //false if command line option --net is given to start without network
 static bool ignore_network = true;
+// legacy_mode determines whether the application should require the data directory to be specified on the command-line or not
+// true - no parameter required, application should start regardless of where it's called from
+// false - parameter required, application should only start if specified
+// NOTE: This is tied to the executable name. `vegastrike` -> true; `vegastrike-engine` -> false
+bool legacy_mode;
```
stephengtuggy, line 124: Hmm. This variable name is OK, but I think we can do better. How about something like `legacy_data_dir_mode`?

stephengtuggy (changes requested):
I have a few suggestions. Mostly minor.

doc/man/vegastrike-engine.1
BenjamenMeyer: will fix to be less confusing and more accurate.

doc/man/vegastrike-engine.1
BenjamenMeyer: will clarify to be less confusing and more accurate.

engine/src/main.cpp
BenjamenMeyer: @stephengtuggy I was intending for it to be for anything that depended on the legacy behavior; the data directory is certainly one of them. Is it the only one?

engine/src/vsfilesystem.cpp
BenjamenMeyer: okay...so here's the error I'm getting:
```
GOT SUBDIR ARG = 
Found MODDIR = /mods
USING HOMEDIR : /home/bmeyer/.vegastrike As the home directory 
CONFIGFILE - Found a config file in home directory, using : /home/bmeyer/.vegastrike/vegastrike.config
Force galaxy to /universe/milky_way.xml
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::property_tree::xml_parser::xml_parser_error> >'
  what():  /universe/milky_way.xml: cannot open file
Aborted (core dumped)
```
however, I'm not sure that's an appropriate error to catch as it likely won't be the_only_ error that could possibly happen. That said, I had to remove a bunch of stuff to get the error to happen again; still catching the errors here would be a good idea.

engine/src/vsfilesystem.cpp
BenjamenMeyer: Thanks! done

engine/src/main.cpp
stephengtuggy: @BenjamenMeyer I can't think of any others at the moment.

engine/src/vsfilesystem.cpp
stephengtuggy: yeah, maybe just catching exceptions in general is OK here

engine/src/main.cpp
BenjamenMeyer: @Loki1950 are there others we may run into?

Loki1950: I think we have covered them all

BenjamenMeyer: okay - I'll put up a small change to rename it.

stephengtuggy (approved):
Assuming all the GH Actions pass, I think we're good to go.

engine/src/main.cpp
BenjamenMeyer: updated

PR #435, opened 2021-03-18
Enhancement: Re-enable dynamic dependency detection

BenjamenMeyer:
Dynamic dependency detection was fixed and the DEB files built
using `script/package` will properly use it. This re-enables the
CMake/CPack function to rely on it if it exists.

Dynamic dependency detection is preferred as it will make the
dependencies match up with what is actually needed based on an
examination of the binaries even if developers fail to specify it.
This is a more reliable method than static dependencies.

Please answer the following:

Code Changes:
- [x] Build system change only - verified the Deb files that were built

Issues:
- #391 

Purpose:
Re-enables the preference to use the dynamic dependencies. Should preserve the static dependencies as a fall-back mechanism.


Loki1950 (approved):

stephengtuggy (commented):
@BenjamenMeyer I have concerns that the auto-generated dependency list will be too specific. The fact that several Wayland libraries showed up in the list from your local system posted on #391  , seems to validate my concerns.

Thoughts?

stephengtuggy:
Thanks for the hard work on this, by the way. The auto-generated dependency list is very informative, even if we don't end up using it verbatim.

BenjamenMeyer:
If we know a better way to mix and match that'd be great. It just does an ldd analysis right now and grabs matching packages. Might need a tweak around grabbing the matching one that's installed versus and package; guessing that's why the wayland dependency might have shown up

BenjamenMeyer:
to answer @stephengtuggy 's question about the dependencies we're going to merge this, cut a pre-alpha release, check it, and if we don't like the result we may revert this change, or at least not port it yet to 0.7.x.

PR #445, opened 2021-05-08
cmake: Install man pages into matching sub directory

susnux:
Man pages usually are installed into sub directories matching the man page category.

Using `TYPE`  does not allow to specify a sub directory for the man pages, so using the still supported (and recommended for such tasks) `DESTINATION`

### DIFF
--- a/doc/man/CMakeLists.txt
+++ b/doc/man/CMakeLists.txt
@@ -1,23 +1,12 @@
 IF (UNIX)
 
-	# There is no good method for this. The preferred method of using the `TYPE` argument doesn't exist until CMake 3.14
-	# and there is no CMake Policy to detect the change or force only the older style. `TYPE` is preferred because the
-	# destination is held by CMake instead of being hard coded here. So once CMake 3.14 is the base version this can
-	# be dropped to just the one version.
-	IF (CMAKE_VERSION VERSION_LESS "3.14")
-		MESSAGE("Hard coded man page destinations")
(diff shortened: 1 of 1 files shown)


BenjamenMeyer:
> > @susnux Thanks for the PR; however, we are presently limited to CMake Version 3.5 where DESTINATION is not present due to platforms we support
> 
> On which platforms `DESTINATION` is not present?
> With CMake 3.5 only `DESTINATION` should be present, but not `TYPE` [cmake@v3.5#install](https://cmake.org/cmake/help/v3.5/command/install.html#install)

https://cmake.org/cmake/help/v3.5/command/install.html#installing-files
https://cmake.org/cmake/help/v3.14/command/install.html#installing-files

Sorry, it's the other way around - it's been a while since I wrote that functionality.

DESTINATION is actually a very generic tool where TYPE provides the ability to override more specifically; however, 3.14 is the minimum version to move over. Either way, once we can use 3.14 or later on all our official platforms we'll consolidate down to a single form.

susnux:
`DESTINATION` allows you to setup `DESTINATION` to `CMAKE_INSTALL_MANDIR/man1` for your man pages.
Because you know you only install man1 pages.

Packagers then could override `CMAKE_INSTALL_MANDIR` to something they want.

But with `TYPE` you can not setup `man1`, so the packager have to know that you only use `man1`, then can override `CMAKE_INSTALL_MANDIR`.
But what if you will add other man paged (man5...)? Then they get mixed up :/

BenjamenMeyer:
@susnux right now we only use man1; we'll cross that bridge if we ever add another man page group.

PR #447, opened 2021-05-08
cmake: Fix DATA_DIR quoting

susnux:
First level quoting is used for cmake,
second level is consumed by the shell executing the compiler,
so there needs to be a third level for the `DATA_DIR` macro to be recognized as a string inside the source files.

Without this patch currently it will get transformed like this:
```
cmake: "${DEFINES} -DDATA_DIR=\"${DATADIR}\""
calls gcc:
gcc ... -DDATA_DIR="/foo/bar"
defines:
#define DATA_DIR /foo/bar
```
So the DATA_DIR is no longer a string.


BenjamenMeyer (approved):
LGTM

PR #448, opened 2021-05-08
posh.h add Aarch64 support

susnux:
This is a patch from the openSUSE games repository to enable Aarch64 support.

### DIFF
--- a/engine/src/posh.h
+++ b/engine/src/posh.h
@@ -561,6 +561,17 @@
 #  endif
 #endif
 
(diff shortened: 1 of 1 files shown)


Loki1950 (approved):

BenjamenMeyer (approved):

BenjamenMeyer:
Do note that while this will help with AARCH64 systems they're not presently on our targetted list of supported system. I guess we'll need to add it for M1 Mac Support; but we'll also need a CI environment for it too.

Issue #149, opened 2020-06-19
Py3 engine segfaults when using a savefile from a py2 engine


evertvorster:
Hey there. I am playing with the latest vegastrike engine from master. 
It has been compiled with PYTHON3=ON, and using the system boost libs. 

So far, it's been going well. Due to the "jumping around issue" I personally don't think it's related to the SPEC, I have been trying a python2 version, as well as older branches of the engine. During this time I also advanced a little in the game, doing Janus' missions. 

Now I have reverted back to python3 and the latest master, and I get a segfault when the ship launches. When the engine is Python2, and the same savefile, there is no segfault. 

I have attached the savefile to this report.

[Evert_empty.tar.gz](https://github.com/vegastrike/Vega-Strike-Engine-Source/files/4805780/Evert_empty.tar.gz)

Here is the interesting bit of the error message:
-----snip---------

Adding news
finding nonloaded quest
quest_tutorial
finding quest
quest_tutorial
nonpfact
removing quest
Processing News
Compiling python module bases/launch_music.py
Traceback (most recent call last):
  File "/usr/share/vegastrike/modules/missions/privateer.py", line 34, in Execute
    i.Execute()
  File "/usr/share/vegastrike/modules/random_encounters.py", line 291, in Execute
    dynamic_battle.UpdateCombatTurn()
  File "/usr/share/vegastrike/modules/dynamic_battle.py", line 42, in UpdateCombatTurn
    lookorsiege=LookForTrouble (fac)
  File "/usr/share/vegastrike/modules/dynamic_battle.py", line 367, in LookForTrouble
    randomMovement (i,faction)
  File "/usr/share/vegastrike/modules/dynamic_battle.py", line 258, in randomMovement
    fg_util.TransferFG( fg,fac,suggestednewsys);
  File "/usr/share/vegastrike/modules/fg_util.py", line 425, in TransferFG
    _RemoveFGFromSystem(fgname,faction,starsystem)
  File "/usr/share/vegastrike/modules/fg_util.py", line 251, in _RemoveFGFromSystem
    if Director.getSaveString(ccp,key,index) == fgname:
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe3 in position 2: invalid continuation byte
terminate called after throwing an instance of 'boost::python::error_already_set'
/usr/bin/vs: line 7:  7015 Aborted                 (core dumped) vegastrike -d/usr/share/vegastrike
[...]

PR #453, opened 2021-05-21
Bug Fix: #149

BenjamenMeyer:
Loading an invalid saved game file should error out.
This goes part way, but hangs on the loading screens.

Please answer the following:

Code Changes:

Issues:
- #149 

Purpose:
- Prevent the game from crashing when loading an invalid saved game file.

This presently detects invalid characters properly however, it leaves the user hanging at the load screen instead of shunting them back to the screen to choose another saved game file.

### DIFF
--- a/engine/src/gfx/cockpit_generic.cpp
+++ b/engine/src/gfx/cockpit_generic.cpp
@@ -522,10 +522,12 @@ bool Cockpit::Update()
         QVector vec;
         DockToSavedBases( _Universe->CurrentCockpit(), vec );
     }
-    if (jumpok)
+    if (jumpok) {
         jumpok++;
-    if (jumpok > 5)
+    }
+    if (jumpok > 5) {
         jumpok = 0;
+    }
     UpdAutoPilot();
     Unit *par = GetParent();
     if (par != NULL) {
@@ -594,7 +596,9 @@ bool Cockpit::Update()
                     }
                 }
                 if (tmp == false) {
-                    if (tmpgot) index = 0;
+                    if (tmpgot) {
+                        index = 0;
+                    }
                     Unit *un = parentturret.GetUnit();
                     if ( un && ( !_Universe->isPlayerStarship( un ) ) ) {
                         SetParent( un, GetUnitFileName().c_str(), this->unitmodname.c_str(), savegame->GetPlayerLocation() );
@@ -620,8 +624,9 @@ bool Cockpit::Update()
                                                         par ) <= 0) )
                 && ( !( par->IsCleared( targ ) || targ->IsCleared( par ) || par->isDocked( targ )
                        || targ->isDocked( par ) ) ) && (par->getRelation( targ ) >= 0) && (targ->getRelation( par ) >= 0) ) {
-                if ( targ->isUnit() != _UnitType::planet || targ->GetDestinations().empty() )
+                if ( targ->isUnit() != _UnitType::planet || targ->GetDestinations().empty() ) {
                     RequestClearence( par, targ, 0 );                      //sex is always 0... don't know how to	 get it.
+                }
             } else if ( ( par->IsCleared( targ )
                          || targ->IsCleared( par ) ) && ( !( par->isDocked( targ ) ) || targ->isDocked( par ) )
                        && ( (targ->isUnit() == _UnitType::planet && UnitUtil::getSignificantDistance( par, targ ) > 0)
@@ -652,7 +657,7 @@ bool Cockpit::Update()
             Unit *un;
             bool  found = false;
             int   i     = 0;
-            for (un_iter ui = _Universe->activeStarSystem()->getUnitList().createIterator(); (un = *ui); ++ui)
+            for (un_iter ui = _Universe->activeStarSystem()->getUnitList().createIterator(); (un = *ui); ++ui) {
                 if (un->faction == this->unitfaction) {
 //this switches units UNLESS we're an ejected pilot. Instead, if we are an ejected
 //pilot, switch only if we're close enough.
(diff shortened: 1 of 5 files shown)


engine/src/main.cpp
```diff
@@ -672,16 +672,22 @@ void bootstrap_main_loop()
...
                                                                     mysystem,
                                                                     pos,
                                                                     setplayerXloc,
                                                                     credits,
                                                                     packedInfo,
                                                                     k );
-        } else {
+        }
+
+        if (!loadedSavedGame) {
```
P-D-E, line 690: I think the original else should be left alone, and the `if (!loadedSavedGame) {` brought into the upper `if`, where it should at the very least log the fail and bail out immediately; it might make easier to revert to the savegame loading screen from here rather than from inside the `main_loop` called further below.

engine/src/gfx/cockpit_generic.cpp
```diff
@@ -738,38 +751,49 @@ bool Cockpit::Update()
...
+                string  newsystem;
+                QVector pos;
+                bool    setplayerXloc;
+                savegame->SetStarSystem( "" );
+                QVector tmpoldpos = savegame->GetPlayerLocation();
+                savegame->SetPlayerLocation( QVector( FLT_MAX, FLT_MAX, FLT_MAX ) );
+                vector< string > packedInfo;
+                // `fatal` log level is used here so that it actually gets printed out without any buffering
+                // otherwise it'd be debug or info level
+                BOOST_LOG_TRIVIAL(fatal) << boost::format(" Loading %1% from Cockpit") % savegamefile;
```
BenjamenMeyer, line 783: This is the call I'm currently debugging where you select a game from the load screen.

engine/src/main.cpp
BenjamenMeyer: I think this is for when you tell it to load a game via the command-line. It's not in the current call path that I'm working on where you load a game from the load screen.

For simplicity I broke this out so that it could revert to the previous behavior easier; so in this case if it fails then it goes on to load as if no saved game was requested to load.

engine/src/gfx/cockpit_generic.cpp
P-D-E: Hmm, quite a lot has already happened before the save is actually parsed...
If `savegame.cpp` had a `bool CheckSaveGame( const std::string &filename )` that only checks files for utf-8 integrity, it could be called somewhere earlier in the stack, and the command line handling part could use it as well.

stephengtuggy: Please note that specifying log level `fatal` does not actually disable buffering, in and of itself. If you want to make sure that the output gets written right away, call `VSFileSystem::flushLogs();` as the next statement.

Looks like I need to document the Boost logging stuff in a bit more detail, with regard to this and a couple of other things.

engine/src/main.cpp
stephengtuggy: When the user tells VS to load a game via the command line, and it isn't able to load the game successfully, shouldn't VS bomb out and exit back to the command line with an error code? Whereas if the user tries to load a game via the GUI, failure to load should land the user right back at the same GUI screen, probably after an error dialog of some sort.

engine/src/gfx/cockpit_generic.cpp
BenjamenMeyer: @stephengtuggy I used fatal because it actually got it to the log dump so I could see it before killing the application because I knew it had hung. I tried other levels first but they didn't show up, so I figured it was less buffered.

My comment above may be wrong about why it worked; but it worked. We should have some levels that work that way without having to explicitly flush the logs after too, or perhaps we just need a wrapper to do that in one shot (even a C Macro would suffice).

stephengtuggy: @BenjamenMeyer So using `fatal` actually made a difference, huh? Interesting.

I think I agree that we need a wrapper macro or two. Maybe something like `VS_LOG_AND_FLUSH` and `VS_LOG` (without flushing). Each could take two arguments: the log level, such as `fatal` or `debug`; and the message to log. The message to log might be the result of running `boost::format` with some `%` operands. Simply enclosing the whole `boost::format` clause in an extra pair of parentheses should allow the macro to work, I think.

How does that sound?

P-D-E: I've put together a [command-line UTF-8 file checker](https://gist.github.com/P-D-E/fd3e23557461f0c87334ab68c39da0e4) based mostly on [this](https://helloacm.com/how-to-validate-utf-8-encoding-the-simple-utf-8-validation-algorithm/).

I've tested it on new and old saves, and on the infamous 0.5.x bzbr.txt asset and it reports correctly if those files can be read as UTF-8 or not.

If brought into `savegame` and called as early as possible, it could help avoiding all the new `ValidateString` logic.
Could this be a viable way?

P-D-E: I've found an alternative way with a similar void -> bool change, but in `UniverseUtil::loadGame` where I also added the UTF-8 checker; now `QuineComputer` in `computer_lib.py` can intercept it and show a message to the user after the failed load attempt. Should I put up an alternate PR with those changes?

P-D-E: @BenjamenMeyer Great, I invite you all to test the latest Py-only option, it won't even require compiling.

Meanwhile I'll investigate on the loadGame in `main_menu.py`: if the check is required there too, it might make sense to put it in a separate small module and import it anywhere needed.

P-D-E: Here's the Py-only [draft PR](https://github.com/vegastrike/Assets-Production/pull/60) for the assets; meanwhile I've found the in-game menu available when docked is handled by `basecomputer.cpp` in `cmd`, so the cpp check on the engine side is still needed.

P-D-E: Speaking of the C++ side, how does [this](https://github.com/P-D-E/VS-PoC/blob/main/utf8-check-in-game.png) look?

Loki1950: Short and all the info.

BenjamenMeyer:
Closing in favor of #463

PR #461, opened 2021-05-28
Don't build vegasettings on Windows (for now)?

stephengtuggy:

Please answer the following:

Code Changes:
- [X] This is a CI/build system change only

Issues:
- #53 

Purpose:
- What is this pull request trying to do? Get Windows CI passing, by leaving out vegasettings for the time being, since it doesn't build anyway
- Is there a project or milestone we should apply this to? 0.8.x, I guess?


Loki1950:
I assume this is until Microsoft clears that bug in the GTK lib's good idea

stephengtuggy:
@Loki1950 Yes. Although there were a couple of link errors before that too

stephengtuggy:
> I assume this is until Microsoft clears that bug in the GTK lib's good idea

For reference, here is the bug mentioned: https://github.com/microsoft/vcpkg/issues/18162

Loki1950 (approved):
You may want to wait for @BenjamenMeyer review

engine/setup/CMakeLists.txt
```diff
@@ -31,4 +31,4 @@ IF (NOT BEOS)
         ${GTK3_INCLUDE_DIRS}
     )
 	TARGET_LINK_LIBRARIES(vegasettings ${GTK_LIBS})
-ENDIF (NOT BEOS)
+ENDIF ((NOT BEOS) AND (NOT WIN32))
```
BenjamenMeyer, line 34: It would be good if we had an else clause that emitted a message about why we're skipping, something like:
```
ELSE ((NOT BEOS) AND (NOT WIN32))
    MESSAGE("SKIPPING vegasettings: Vega Settings Not support on the Platform at this time")
ENDIF ((NOT BEOS) AND (NOT WIN32))
```

BenjamenMeyer (changes requested):
One minor little change, but generally looks good - a builder  log message about why we're dropping it from the build.

Not sure how the GitHub warnings about tvector.h will impact...but right now this gets us a little farther along so more people can test.

engine/setup/CMakeLists.txt
stephengtuggy: Done.

BenjamenMeyer (approved):

PR #463, opened 2021-05-29
Utf-8 check, C++ initial version

P-D-E:

Please answer the following:

Code Changes:

Issues:
- #149

Purpose:
- What is this pull request trying to do?  **Preventing crashes from loading non-utf8 saved games via pre-emptive checks, alerting the user when it happens. Serving as a place for discussion of the issue, whatever direction the changes will go.**
- What release is this for?  **0.8.x, 0.9.x**
- Is there a project or milestone we should apply this to?  **0.8.x, 0.9.x**

### DIFF
--- a/engine/src/savegame.cpp
+++ b/engine/src/savegame.cpp
@@ -44,6 +44,12 @@
 #include "universe.h"
 #include "options.h"
 
+#include <iostream>
+#include <fstream>
+#include <iterator>
+#include <vector>
+
+typedef unsigned char BYTE;
 
 
 using namespace VSFileSystem;
@@ -171,6 +177,99 @@ void SaveFileCopy( const char *src, const char *dst )
         } else {
             BOOST_LOG_TRIVIAL(warning) << boost::format("WARNING : couldn't find the savegame to copy : %1% as SaveFile") % src;
         }
+    }
+}
+
+class Utf8Checker {
+public:
+    bool validUtf8(const std::vector<BYTE> &data) {
+        for (long unsigned int i = 0; i < data.size(); ++ i) {
+            // 0xxxxxxx
+            int bit1 = (data[i] >> 7) & 1;
+            if (bit1 == 0) continue;
+            // 110xxxxx 10xxxxxx
+            int bit2 = (data[i] >> 6) & 1;
+            if (bit2 == 0) return false;
+            // 11
+            int bit3 = (data[i] >> 5) & 1;
+            if (bit3 == 0) {
+                // 110xxxxx 10xxxxxx
+                if ((++ i) < data.size()) {
+                    if (is10x(data[i])) {
+                        continue;
+                    }
+                    return false;
+                } else {
+                    return false;
+                }
+            }
+            int bit4 = (data[i] >> 4) & 1;
+            if (bit4 == 0) {
+                // 1110xxxx 10xxxxxx 10xxxxxx
+                if (i + 2 < data.size()) {
+                    if (is10x(data[i + 1]) && is10x(data[i + 2])) {
+                        i += 2;
+                        continue;
+                    }
+                    return false;
+                } else {
+                    return false;
+                }
+            }
+            int bit5 = (data[i] >> 3) & 1;
+            if (bit5 == 1) return false;
+            if (i + 3 < data.size()) {
+                if (is10x(data[i + 1]) && is10x(data[i + 2]) && is10x(data[i + 3])) {
+                    i += 3;
+                    continue;
+                }
+                return false;
+            } else {
+                return false;
+            }
+        }
+        return true;
+    }
+private:
+    inline bool is10x(int a) {
+        int bit1 = (a >> 7) & 1;
+        int bit2 = (a >> 6) & 1;
+        return (bit1 == 1) && (bit2 == 0);
+    }
+};
+
+std::vector<BYTE> readFile(std::string filename)
+{
+    std::ifstream file(filename, std::ios::binary);
+    file.unsetf(std::ios::skipws);
+    std::streampos fileSize;
+
+    file.seekg(0, std::ios::end);
+    fileSize = file.tellg();
+    file.seekg(0, std::ios::beg);
+
+    std::vector<BYTE> vec;
+    vec.reserve(fileSize);
+
+    std::copy(std::istream_iterator<BYTE>(file),
+          std::istream_iterator<BYTE>(),
+          std::back_inserter(vec));
+
+    file.close();
+
+    return vec;
+}
+
+bool checkSaveGame(std::string filename)
+{
+    string path = GetSaveDir() + filename;
+    std::vector<BYTE> savegame = readFile(path);
+    Utf8Checker* check = new Utf8Checker;
+    if (check->validUtf8(savegame)) {
+        return true;
+    } else {
+        BOOST_LOG_TRIVIAL(fatal) << boost::format("ERROR: save file %1% is not UTF-8") % path;
+        return false;
     }
 }
 
--- a/engine/src/savegame.cpp
+++ b/engine/src/savegame.cpp
@@ -56,6 +56,7 @@ using namespace VSFileSystem;
 using std::vector;
 using std::string;
 using std::allocator;
+using std::unique_ptr;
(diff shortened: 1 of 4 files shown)


engine/src/savegame.cpp
```diff
@@ -174,6 +180,99 @@ void SaveFileCopy( const char *src, const char *dst )
     }
 }
 
+class Utf8Checker {
+public:
+    bool validUtf8(const std::vector<BYTE> &data) {
+        for (long unsigned int i = 0; i < data.size(); ++ i) {
+            // 0xxxxxxx
+            int bit1 = (data[i] >> 7) & 1;
```
stephengtuggy, line 188: Might want to make these `bitN` variables `unsigned int`s

engine/src/savegame.cpp
```diff
@@ -174,6 +180,99 @@ void SaveFileCopy( const char *src, const char *dst )
...
+    file.close();
+
+    return vec;
+}
+
+bool checkSaveGame(std::string filename)
+{
+    string path = GetSaveDir() + filename;
+    std::vector<BYTE> savegame = readFile(path);
+    Utf8Checker* check = new Utf8Checker;
```
stephengtuggy, line 267: Can we use a smart pointer here? Such as `boost::shared_ptr` or similar?

stephengtuggy (commented):
Just a couple of things. Looks good overall. The bit shifting is well documented, which I appreciate.

FWIW, I think savegame.cpp/.h is a perfectly fine place to put this functionality.

Thanks for the PR, @P-D-E !

engine/src/savegame.cpp
P-D-E: You're welcome to modify anything as you see fit, It's mostly code from [here](https://helloacm.com/how-to-validate-utf-8-encoding-the-simple-utf-8-validation-algorithm/), practically untouched.

engine/src/savegame.cpp
P-D-E: By the way, as you mentioned boost, could you investigate on the chance to use the Py check instead of this as an alternative like @ermo suggested? I've found a `PYTHONCALLBACK2` macro in `src/python/python_class.h` which should accept an argument (our filename complete with path) and return a value.

stephengtuggy: That does look promising. I also see a `PYTHONCALLBACK` macro, which takes one string argument instead of two, and `virtual void callFunction (std::string str)`, under `PythonClass`. I think the latter might be the way to go, perhaps.

P-D-E: Yes but the first argument is the module function, so we need another one to pass the game path, confrot these uses:
`PYTHONCALLBACK(void, this->self, "Execute");`
`PYTHONCALLBACK2(void, self, "UnPickle",s);`

stephengtuggy: Gotcha

P-D-E: ...meanwhile I managed to test the opposite, exposing the C++ check to Python; I put it in the `Director` module hosting the savegame data related functions, and the assets like `computer_lib.py` can do this:
```
import Director
...
if not Director.isUtf8SaveGame(game):
```

stephengtuggy: @P-D-E Cool. That sounds like the path of least resistance for moving forward.

stephengtuggy: If I understood correctly, the other proposed solution would have involved Python code calling C++ code calling more Python code. I think this way is a lot better, because it's simpler.

P-D-E: Correct indeed, unless the Python check code could be called both by the assets (directly instead of Py -> C++ -> Py) and by the engine (via boost's `call_method`), but this way is still to be experimented with.

engine/src/savegame.cpp
```diff
@@ -174,6 +180,99 @@ void SaveFileCopy( const char *src, const char *dst )
     }
 }
 
+class Utf8Checker {
```
royfalk, line 183: A bit late to the party, but I wonder if you could have used some boost facility instead.
Something like https://www.boost.org/doc/libs/1_48_0/libs/locale/doc/html/charset_handling.html. It's possible to ask it to throw an exception. Other than that, LGTM.

stephengtuggy (approved):
I like this code a lot. I'm ready to give it my approval. What do you guys say?

engine/src/savegame.cpp
ermo: @P-D-E just a quick check whether you actually added the smart pointer as @stephengtuggy suggested?

P-D-E: @ermo Not yet; if changing
```
    Utf8Checker* check = new Utf8Checker;
```
into
```
    std::unique_ptr<Utf8Checker> check(new Utf8Checker);
```
is okay for you all, I can do it now. It compiles and runs correctly.

BenjamenMeyer: @P-D-E that would be preferable yes; we're moving towards using more smart pointer storage objects

BenjamenMeyer (commented):
Generally this looks good; one minor change to move to the smart pointer object and we should be good to go

engine/src/savegame.cpp
```diff
@@ -174,6 +181,105 @@ void SaveFileCopy( const char *src, const char *dst )
...
+    file.close();
+
+    return vec;
+}
+
+bool isUtf8SaveGame(std::string filename)
+{
+    std::string path = GetSaveDir() + filename;
+    std::vector<BYTE> savegame = readFile(path);
+    std::unique_ptr<Utf8Checker> check(new Utf8Checker);
```
BenjamenMeyer, line 274: question: does this need to be a pointer to an instance? Or can it just be a local instance?
```
Utf8Checker check;
...
if (check.validUtf8(...
```
If we don't need to allocate, then let's not. If there's a reason we need to, then okay; but I think we may not need to - stack allocation should be sufficient.

NOTE: This wasn't quite clear before.

P-D-E: Yes it can, @vegastrike/code-developers do you all agree on this?

stephengtuggy: It could just be a local instance instead of a pointer, but personally, I don't think that it matters much one way or the other.

I say let's merge this PR as is.

ermo: If the checker is only ever used inside the scope of the isUtf8SaveGame function, then it makes sense to just use it directly since it doesn't need to be passed around to other consumers?

stephengtuggy: I thought we could save some time by leaving the code the way it is here, but I see that that is not the way this is going. So sure, let's change it to use the UTF8Checker directly, as a local instance.

BenjamenMeyer (approved):

BenjamenMeyer:
This needs to get ported over to the 0.8.x branch.

ermo:
> please port this over to the 0.8.x branch.

Who is supposed to do this?

BenjamenMeyer:
@ermo ideally the same person that did the PR; but anyone can do it:
```
$ git checkout 0.8.x
$ git checkout -b port_utf-8-heck
$ git cherry-pick <commit hash 1>
$ git cherry-pick <commit hash ...>
$ git cherry-pick <commit hash N>
$ git push origin port_utf-8-heck
```

Or something like that...some adjustments might need to be made, but that's the general process

ermo:
>>> Please port this over to the 0.8.x branch.
>> Who is supposed to do this?
> Ideally the same person that did the PR; but anyone can do it:
> 
> ```
> $ git checkout 0.8.x
> $ git checkout -b port_utf-8-heck
> $ git cherry-pick <commit hash 1>
> $ git cherry-pick <commit hash ...>
> $ git cherry-pick <commit hash N>
> $ git push origin port_utf-8-heck
> ```
> 
> Or something like that...some adjustments might need to be made, but that's the general process

@P-D-E: Would you mind taking a look at porting it to the 0.8.x branch since you're familiar with the code? :partying_face:

P-D-E:
@ermo Benjamen and I talked about it and agreed that he'll take care of it :+1:

PR #469, opened 2021-06-03
Unit_generic refactor

royfalk:
This release is part of the continuing unit classes refactoring. The main purpose of this PR is to move more and more code from unit_generic to dedicated classes.
This has several advantages:

1. It's easier to find code and when you look at a file, you know what it's about.
2. It's easier to compile. Smaller files mean fewer dependencies. 
3. It also makes it easier to write more elegant code.

Purpose:
- What release is this for? Whatever is upcoming

Code Changes:
- [Yes] Have the PR Validation Tests been run? 
- 
Issues:
- None identified


engine/src/cmd/jump_capable.cpp
```diff
@@ -0,0 +1,580 @@
...
+StarSystem* JumpCapable::getStarSystem()
+{
+    Unit *unit = static_cast<Unit*>(this);
+    if (activeStarSystem) {
+        return activeStarSystem;
+    } else {
+        Cockpit *cp = _Universe->isPlayerStarship( unit );
+        if (cp)
+            if (cp->activeStarSystem)
+                return cp->activeStarSystem;
```
BenjamenMeyer, line 464: Braces needed

engine/src/cmd/jump_capable.cpp
```diff
@@ -0,0 +1,580 @@
...
+const StarSystem* JumpCapable::getStarSystem() const
+{
+    const Unit *unit = static_cast<const Unit*>(this);
+    if (activeStarSystem) {
+        return activeStarSystem;
+    } else {
+        Cockpit *cp = _Universe->isPlayerStarship( unit );
+        if (cp)
+            if (cp->activeStarSystem)
+                return cp->activeStarSystem;
```
BenjamenMeyer, line 478: Same

BenjamenMeyer (changes requested):
Generally looking good from what I can tell on my phone.

jump_capable.cpp needs braces added in numerous blocks. Noted a couple, but those aren't all the locations.

engine/src/cmd/jump_capable.cpp
royfalk: I'll do a round of adding braces. However, I should note that this isn't a deep refactor of the code. I merely moved existing code to a new file.
If I were to try and also refactor every line in it, it would have taken me far more than a few hours.

engine/src/cmd/jump_capable.cpp
royfalk: Done

BenjamenMeyer (approved):

PR #470, opened 2021-06-06
Don't utf8-check New_Game

P-D-E:

Please answer the following:

Code Changes:

Issues:
- #149 

Purpose:
- What is this pull request trying to do?  **Skipping the utf-check for the New_Game special case as it's not a real file; it makes the Campaign (main menu) and New (while docked in game) buttons work properly**
- What release is this for?  **0.8.x**
- Is there a project or milestone we should apply this to?  **0.8.x**

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -5979,7 +5979,7 @@ bool BaseComputer::actionConfirmedLoadGame()
     if (desc) {
         std::string tmp = desc->text();
         if (tmp.length() > 0) {
-           if ((tmp != "New_Game") && (!isUtf8SaveGame(tmp))) {
+            if ((string( "New_Game" ) != tmp) && (!isUtf8SaveGame(tmp))) {
                 showAlert( tmp + " is not UTF-8, convert it before loading" );
                 return true;
             }
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -5979,7 +5979,7 @@ bool BaseComputer::actionConfirmedLoadGame()
     if (desc) {
         std::string tmp = desc->text();
         if (tmp.length() > 0) {
-            if (!isUtf8SaveGame(tmp)) {
+           if ((tmp != "New_Game") && (!isUtf8SaveGame(tmp))) {
                 showAlert( tmp + " is not UTF-8, convert it before loading" );
                 return true;
             }


Loki1950 (approved):

engine/src/cmd/basecomputer.cpp
```diff
@@ -5979,7 +5979,7 @@ bool BaseComputer::actionConfirmedLoadGame()
     if (desc) {
         std::string tmp = desc->text();
         if (tmp.length() > 0) {
-            if (!isUtf8SaveGame(tmp)) {
+           if ((tmp != "New_Game") && (!isUtf8SaveGame(tmp))) {
```
stephengtuggy, line 5982: I wonder if this check should be made case insensitive? Using, for example, `boost::iequals`.

Other than that, looks good.

stephengtuggy (commented):
Looks pretty good. Just have one question.

engine/src/cmd/basecomputer.cpp
P-D-E: New_Game is hard-coded, all other case variants are intentionally allowed as legit save names, and only this exact one is barred from manual save, so that the engine knows it's dealing with a new game.
You can see it for instance in the method above this one, `actionSaveGame`, which I just copied to `actionConfirmedLoadGame` for coherence: `(string( "New_Game" ) != tmp)`

stephengtuggy (approved):
Cool. Looks good to me then.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: @P-D-E is there a string constant that can be used for that? If not, we should add an issue to introduce one.

I'd just hate to see that string modified slightly and have things break.
Also, we probably should prohibit all case variants of that string (if we don't right now, that should be another issue).

BenjamenMeyer (commented):
Aside from my comments related to @stephengtuggy's review, I'm good for this too.

engine/src/cmd/basecomputer.cpp
P-D-E: Yes, there are quite a few independent instances of the string "New_Game":
```
engine/datascripts/setup.base.xml
engine/datascripts/021.nsi
engine/src/cmd/basecomputer.cpp
engine/src/options.cpp
engine/launcher/saveinterface.cpp
data/cgi-accountserver/db.py
data/bases/computer_lib.py
data/bases/main_menu.py
data/modules/GUI.py
```
and they are more than the usages of a convenient `UniverseUtil::getNewGameSaveName` function in `engine/src/universe_util_generic.cpp` also made available to Python via the `VS` module, e.g.
```
engine/src/cmd/basecomputer.cpp
6010:    desc->setText( UniverseUtil::getNewGameSaveName() );

data/bases/main_menu.py
16:    VS.loadGame(VS.getNewGameSaveName())
```
We might try replacing all the explicit New_Game mentions with calls to the function, especially where the needed includes are already there.

BenjamenMeyer: @P-D-E I'm good with a separate PR for that too; can you please open an issue with your findings? Thanks!

ermo: > We might try replacing all the explicit New_Game mentions with calls to the function, especially where the needed includes are already there.

That sounds like the correct approach to me.

P-D-E: @BenjamenMeyer @ermo Thank you both, [here](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/472)'s the issue.

BenjamenMeyer (approved):

PR #475, opened 2021-06-11
Refactor weapon_xml into weapon_factory and other classes (2)

royfalk:
**This PR replaces #459 as it got borked and fixing it is more complicated than doing it again.
This PR already includes fixes to the comments in the original PR.**

Refactor weapon_xml class into the following classes:

    weapon_factory
    weapon_info - moved to its own file
    mount_size - moved to its own file
    weapon_type - moved to its own file

Also create a faction_factory class to replace faction_xml.
Cleanup of related code to cleaner more modern standard. See for example firekeyboard.cpp:1670.

Code Changes:

    [Yes] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation Need to test again as I've written this code in April.

Issues:

    None yet

Purpose:

    What is this pull request trying to do? mostly get rid of the old xml parser with its unreadable code.
    What release is this for? 0.9
    Is there a project or milestone we should apply this to? no


royfalk:
There's a gnarly conflict that I need to deal with. 
I'll do that tomorrow and retest.

royfalk:
OK. I've merged from master and tested again - focusing now on a rather extensive 'clean sweep' mission. Everything seems to work and I think it's ready for merge (provided the builds pass).

engine/src/main.cpp
```diff
@@ -323,7 +323,7 @@ int main( int argc, char *argv[] )
 
     // when the program name is `vegastrike-engine` then enforce that the data directory must be specified
     // if the program name is `vegastrike` then enable legacy mode where the current path is assumed.
-    legacy_data_dir_mode = (program_name == "vegastrike") || (program_name == "vegastrike.exe");
+    legacy_data_dir_mode = true; //(program_name == "vegastrike") || (program_name == "vegastrike.exe");
```
stephengtuggy, line 326: Um, what? Why are we commenting out this logic? This was put here intentionally, by @BenjamenMeyer .

stephengtuggy (changes requested):

stephengtuggy:
Also, the CI builds didn't pass. Looks like they are throwing errors on two identical lines that have this:

```c++
     for (const auto& [key, value]: mount_map) {
```

engine/src/main.cpp
BenjamenMeyer: yes, that was a means of detecting legacy mode to fall back for old behavior with running it as the main binary is now `vegastrike-engine`. CMake takes care of creating the `vegastrike` binary as a copy during one of its steps (IIRC in the install process).

engine/doc/Makefile
```diff
@@ -0,0 +1,208 @@
+# CMAKE generated file: DO NOT EDIT!
+# Generated by "Unix Makefiles" Generator, CMake Version 3.16
+
+# Default target executed when no arguments are given to make.
+default_target: all
```
BenjamenMeyer, line 5: We shouldn't be commiting Makefiles as Cmake is the official source of truth for the build.

engine/doc/man/Makefile
```diff
@@ -0,0 +1,208 @@
+# CMAKE generated file: DO NOT EDIT!
+# Generated by "Unix Makefiles" Generator, CMake Version 3.16
```
BenjamenMeyer, line 2: Same

engine/config.h
```diff
@@ -0,0 +1,48 @@
+#ifndef _CONFIG_H_
+#define _CONFIG_H_
```
BenjamenMeyer, line 2: I believe this is a build artifact too; it doesn't have the `generated by` lines in it but it's communicating what CMake found during the `cmake` operation.

BenjamenMeyer (changes requested):
Generally looks good but...
- [ ] Need to remove build artifacations (config.h, Makefile)
- [ ] Need to restore the legacy check that @stephengtuggy mentioned.

Overall this should improve speed as text processing is expensive and moving to a proper enumeration (enum or int methodology for enum) will mean less data to pass around and faster comparison times; not to mention we can use switches instead of if blocks on the C/C++ side.

engine/src/main.cpp
royfalk: You are both absolutely right about this line. It's a hack I put because I was too lazy to figure out why the game wouldn't run and I was a good boy not to commit it until I missed it in the merge.

engine/doc/Makefile
royfalk: Sorry. The whole doc folder was downloaded by a merge and I assumed and didn't check it's documentation of the Makefile rather than an actual Makefile. I'll delete it.

engine/doc/man/Makefile
royfalk: Yep. Will do.

engine/config.h
royfalk: OK. Can we give this task to someone to do? There are a lot of these files and they really should be in gitignore.

royfalk:
> Also, the CI builds didn't pass. Looks like they are throwing errors on two identical lines that have this:
> 
> ```c++
>      for (const auto& [key, value]: mount_map) {
> ```

It's C++17. I'll modify the code to C++14 for now.

stephengtuggy (approved):
Much better. Thanks!

BenjamenMeyer (approved):

PR #478, opened 2021-06-19
More refactoring of (primaily) unit_generic

royfalk:
What is this pull request trying to do? Improve the code by refactoring it. This should not introduce new behavior.
What release is this for? Upcoming.
Is there a project or milestone we should apply this to? No

Code Changes:

Issues:
- None identified


engine/src/cmd/ai/aggressive.cpp
```diff
@@ -207,8 +207,8 @@ static AIEvents::ElemAttrMap * getProperScript( Unit *me, Unit *targ, bool inter
         }
         return getProperLogicOrInterruptScript( "default", fac, nam, interrupt, personalityseed );
     }
-    return getProperLogicOrInterruptScript( ROLES::getRoleEvents( me->attackPreference(),
-                                                                 targ->unitRole() ), me->faction, me->name, interrupt,
+    return getProperLogicOrInterruptScript( ROLES::getRoleEvents( me->attack_preference,
+                                                                 targ->unit_role ), me->faction, me->name, interrupt,
```
BenjamenMeyer, line 211: I'm not sure about these changes - if for performance, we could make the same gain by just using inline functions while maintaining the object portability and opaque variables.

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -316,7 +316,7 @@ bool AggressiveAI::ProcessLogicItem( const AIEvents::AIEvresult &item )
                 value = ( pdmag-parent->rSize()-targ->rSize() );
                 float  myvel = PosDifference.Dot( parent->GetVelocity()-targ->GetVelocity() )/value;        ///pdmag;
                 if (myvel > 0)
-                    value -= myvel*myvel/( 2*( parent->limits.retro/parent->GetMass() ) );
+                    value -= myvel*myvel/( 2*( parent->limits.retro/parent->Mass ) );
```
BenjamenMeyer, line 319: Same here

engine/src/cmd/armed.h
```diff
@@ -42,7 +42,7 @@ class Armed
     float gunspeed;
     float gunrange;
     float missilerange;
-    char  turretstatus;
+    char  turretstatus = 0;
```
BenjamenMeyer, line 45: This should be done in the implementation file, not the header.

engine/src/cmd/basecomputer.cpp
```diff
@@ -2717,10 +2717,10 @@ bool BaseComputer::sellSelectedCargo( int requestedQuantity )
...
         const int quantity = (requestedQuantity <= 0 ? item->quantity : requestedQuantity);
         if (item->mission) {
-            vector< Cargo >::iterator mycargo = std::find( playerUnit->pImage->cargo.begin(),
-                                                           playerUnit->pImage->cargo.end(), *item );
-            if ( mycargo != playerUnit->pImage->cargo.end() )
-                playerUnit->RemoveCargo( mycargo-playerUnit->pImage->cargo.begin(), quantity, true );
+            vector< Cargo >::iterator mycargo = std::find( playerUnit->cargo.begin(),
+                                                           playerUnit->cargo.end(), *item );
+            if ( mycargo != playerUnit->cargo.end() )
+                playerUnit->RemoveCargo( mycargo-playerUnit->cargo.begin(), quantity, true );
```
BenjamenMeyer, line 2723: Dropping the pImage layer should be fine/a good move assuming a 1:1 relationship.

engine/src/cmd/carrier.cpp
```diff
@@ -489,22 +489,22 @@ bool Carrier::CanAddCargo( const Cargo &carg ) const
...
 float Carrier::getCargoVolume( void ) const
 {
     const Unit *unit = static_cast<const Unit*>(this);
     float result = 0.0;
-    for (unsigned int i = 0; i < unit->pImage->cargo.size(); ++i)
-        if ( !cargoIsUpgrade( unit->pImage->cargo[i] ) )
-            result += unit->pImage->cargo[i].quantity*unit->pImage->cargo[i].volume;
+    for (unsigned int i = 0; i < unit->cargo.size(); ++i)
+        if ( !cargoIsUpgrade( unit->cargo[i] ) )
+            result += unit->cargo[i].quantity*unit->cargo[i].volume;
```
BenjamenMeyer, line 507: This needs braces for the various blocks

engine/src/cmd/carrier.cpp
```diff
@@ -570,9 +570,9 @@ float Carrier::getUpgradeVolume( void ) const
 {
     const Unit *unit = static_cast<const Unit*>(this);
     float result = 0.0;
-    for (unsigned int i = 0; i < unit->pImage->cargo.size(); ++i)
-        if ( cargoIsUpgrade( unit->pImage->cargo[i] ) )
-            result += unit->pImage->cargo[i].quantity*unit->pImage->cargo[i].volume;
+    for (unsigned int i = 0; i < unit->cargo.size(); ++i)
+        if ( cargoIsUpgrade( unit->cargo[i] ) )
+            result += unit->cargo[i].quantity*unit->cargo[i].volume;
```
BenjamenMeyer, line 575: same - braces required

engine/src/cmd/carrier.cpp
```diff
@@ -640,12 +640,15 @@ const Cargo* Carrier::GetCargo( const std::string &s, unsigned int &i ) const
...
         searchfor.content = s;
-        vector< Cargo >::iterator tmp = std::find( unit->pImage->cargo.begin(),
-                                                   unit->pImage->cargo.end(), searchfor );
-        if ( tmp == unit->pImage->cargo.end() )
-            return NULL;
+
+        // TODO: could not deduce the right var type. Resorted to auto
+        //vector< Cargo >::iterator tmp = std::find( unit->cargo.begin(),
+        //                                           unit->cargo.end(), searchfor );
+        auto tmp = std::find( unit->cargo.begin(), unit->cargo.end(), searchfor );
```
BenjamenMeyer, line 647: a good use of `auto` - really, one of the few uses for it IMHO

engine/src/cmd/computer.cpp
```diff
@@ -14,6 +15,19 @@ float Computer::max_ab_speed() const
     return (!combat_mode) ? combat_mode_mult*max_combat_speed : max_combat_ab_speed;
 }
 
+Computer::RADARLIM::RADARLIM() {
+    static float default_max_range = GameConfig::GetVariable( "graphics", "hud", "radarRange", 20000 );
+    static float default_tracking_cone = GameConfig::GetVariable( "physics", "autotracking", 0.93 );     //DO NOT CHANGE see unit_customize.cpp
+
+    // Note: below is probably a stale comment
+    // DO NOT CHANGE see unit_customize.cpp
+    static float default_lock_cone = GameConfig::GetVariable( "physics", "lock_cone", 0.8 );
```
BenjamenMeyer, line 24: Would prefer that we not use statics here; having class variables to hold the value would be better.

engine/src/cmd/container.h
```diff
@@ -8,7 +8,7 @@ class Unit;
 class UnitContainer
 {
 protected:
-    Unit *unit;
+    Unit *unit = nullptr;
```
BenjamenMeyer, line 11: this should be done in the constructor/implementation file, not the header file

engine/src/cmd/energetic.h
```diff
@@ -66,33 +69,33 @@ class Energetic
...
-        float insysenergy;                       //short fix
-        signed char   drive;
-        unsigned char delay;
-        unsigned char damage;
+        float warpDriveRating = 0;
+        float energy = 100;                            //short fix
+        float insysenergy;  //short fix
+        signed char   drive = -2; // disabled
+        unsigned char delay = 5;
+        unsigned char damage = 0;
```
BenjamenMeyer, line 77: this would be better in a constructor - and yes, `struct` can have constructors in C++.

engine/src/cmd/unit_generic.cpp
```diff
@@ -4881,10 +4588,10 @@ std::string Unit::massSerializer( const XMLType &input, void *mythis )
     Unit *un   = (Unit*) mythis;
     float mass = un->Mass;
     static bool usemass = XMLSupport::parse_bool( vs_config->getVariable( "physics", "use_cargo_mass", "true" ) );
-    for (unsigned int i = 0; i < un->pImage->cargo.size(); ++i)
-        if (un->pImage->cargo[i].quantity > 0)
+    for (unsigned int i = 0; i < un->cargo.size(); ++i)
+        if (un->cargo[i].quantity > 0)
             if (usemass)
-                mass -= un->pImage->cargo[i].mass*un->pImage->cargo[i].quantity;
+                mass -= un->cargo[i].mass*un->cargo[i].quantity;
```
BenjamenMeyer, line 4594: Braces needed

engine/src/star_system.cpp
```diff
@@ -892,121 +892,107 @@ void StarSystem::RequestPhysics( Unit *un, unsigned int queue )
...
+void StarSystem::UpdateUnitPhysics( bool firstframe, Unit *unit )
+{
+    int priority  = UnitUtil::getPhysicsPriority( unit );
+    //Doing spreading here and only on priority changes, so as to make AI easier
+    int predprior = unit->predicted_priority;
+    //If the priority has really changed (not an initial scattering, because prediction doesn't match)
+    if (priority != predprior) {
+        if (predprior == 0)
+            //Validate snapshot of current interpolated state (this is a reschedule)
+            unit->curr_physical_state = unit->cumulative_transformation;
```
BenjamenMeyer, line 964: Braces needed here

BenjamenMeyer (changes requested):
I'm mixed on this.

There some things I like - like remove the layer in the one case; but others I'm not so keen on (moving from property methods to directly accessing variables; initializing values in headers instead of implementation files/constructors.)

- [x] using inline methods would get us the same performance as direct access but maintain the indirect variable access so the variable can change underneath as needed for any class that may come through
- [ ] initialization really should happen in either inline constructors or in the implementation files (via constructors or otherwise)
- [ ] let's make sure braces are around all code blocks - I know it's just copy/paste but we're still updating the code to bring things into better readability/maintainability/a less error-prone state
- [ ] I'm not keen on using static vars to hold config values; we really need to move to using class variables to hold that and just initialize it on class creation. It gives us more flexibility going forward and makes it easier to reload the config if needed - I know, it's another story to move to that entirely, but let's not start adding new code that uses statics if we can help it.

royfalk:
> 
>     * [ ]  using inline methods would get us the same performance as direct access but maintain the indirect variable access so the variable can change underneath as needed for any class that may come through

As per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get , we should "Avoid trivial getters and setters. Reason: A trivial getter or setter adds no semantic value; the data item could just as well be public."
 
>     * [ ]  initialization really should happen in either inline constructors or in the implementation files (via constructors or otherwise)

As per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-default (and Kate Gregory), we shouldn't "Don’t define a default constructor that only initializes data members; use in-class member initializers instead. Reason: Using in-class member initializers lets the compiler generate the function for you. The compiler-generated function can be more efficient."
Also, it's far easier to understand the class when you see its variables and their defaults. Finally, doing it in a constructor violates DRY and in extreme case like ours, creates "Init" functions to be called from multiple consturctors.
 
>     * [ ]  let's make sure braces are around all code blocks - I know it's just copy/paste but we're still updating the code to bring things into better readability/maintainability/a less error-prone state

As I noted before, I'm moving a lot of code around and trying not to touch it. I won't add code without proper braces and will add some if I see it, but I won't go through every line. Of course, if you point it out in code I touched, I'll do it.

>     * [ ]  I'm not keen on using static vars to hold config values; we really need to move to using class variables to hold that and just initialize it on class creation. It gives us more flexibility going forward and makes it easier to reload the config if needed - I know, it's another story to move to that entirely, but let's not start adding new code that uses statics if we can help it.

Let's clarify this a bit. Global static var is one that can only be accessed in the file where it is created. This variable is said to have file scope. A local static var is persistent between function calls and is initialized once.

(Almost?) Every config I move from XMLSupport::parse_float( vs_config->getVariable(...) is a local static variable. I've started moving them to the top of the file when converting to GameConfig::GetVariable(...) precisely because I agree with you and want to make it easier to identify them all. I have a similar idea where we ditch the multiple settings files and go with one dynamic file/class, accessible from the GUI, maybe even mid-game. But that's a different story (literally) and one I'm not ready to tackle.

If you really hate it that much, I'll move them back inside the functions.

engine/src/cmd/basecomputer.cpp
royfalk: For a moment there, I had serious doubts.
But I don't see how it would compile if it weren't.

engine/src/cmd/armed.h
stephengtuggy: Hmm. I'm not sure I agree. See https://stackoverflow.com/a/11594963/5067822

engine/src/cmd/container.h
stephengtuggy: Again, I think I disagree. From what I've been reading, in C++11 and later, initializing fields in the header file is acceptable, even preferable.

stephengtuggy:
@royfalk I agree with you regarding using in-class member initializers. Everything I'm reading says the same thing.

Not sure I agree regarding the static vars to hold config values. IMHO, all the static vars should be moved from their respective classes to become members of GameConfig, and they should all be initialized when GameConfig is initialized. Then we can see all the config parameters in one place in the code, and also avoid bugs having to do with the order in which they are initialized.

However, perhaps the latter set of changes can happen in a future PR.

stephengtuggy:
Let's not remove `getElasticity()`, `attackPreference()`, or `unitRole()` just yet. Not until we understand their purpose better. I think these are some of those "fences" we should leave up for the time being.

royfalk:
> Not sure I agree regarding the static vars to hold config values. 
> However, perhaps the latter set of changes can happen in a future PR.

We are in complete agreement here. As I noted above, I am not happy with static variables either and think they should migrate to the config class in a later PR.

> Let's not remove getElasticity(), attackPreference(), or unitRole() just yet. Not until we understand their purpose better. I think these are some of those "fences" we should leave up for the time being.

Unless I missed something, these are trivial getters returning the variable. Where a function is referenced by python, this would not compile. I don't remember which is it, but I encountered one and left the function just for the python binding. Did you see something to make you think this could be a fence?

engine/src/cmd/armed.h
BenjamenMeyer: @stephengtuggy I read that, but I'm not sure I agree. I don't necessarily see all new things in C++11/14/17 as advantageous or worthwhile.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: if `pImage` was getting moved between different images for the same object, then it might create an inconsistency that won't be easily obvious; depends on what `pImage` is doing.

BenjamenMeyer:
@royfalk @stephengtuggy I take a very conservative approach to much of C++11/14/17, and in many respects prefer older methods. There's much in those standards that shouldn't be (IMHO; change for change sake, or politics or language perception).

regarding https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-default and https://stackoverflow.com/a/11594963/5067822 - I generally don't trust the compiler to do the right thing for generating constructors. Yes, I've read that's been Stroustrop's position for a long time - to just let the compiler do it, etc - but I don't agree. Again, I'm all for explicit over implicit and that creates an implicit situation. Don't leave the compiler to guess the intention; it may create a little more code but it makes for a lot easier ability to debug it.

regarding https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get - that depends; were they trivial already? Should they be trivial? Are they going to stay trivial going forward? I'm not sure we understand the later part of those questions yet. Locks can easily make them non-trivial. Though the key again in their statement there is that it should be a pure-data struct, not a class too. In that case, the function using the struct should know to use any locks that may be associated with it where a class can be expected to take care of that for you.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: I agree with @BenjamenMeyer here

engine/src/cmd/armed.h
stephengtuggy: Well, we need to make sure that every variable and every field is initialized before it is used. (With a few exceptions where we actually want random values.) Where better to initialize these fields than at the point where they are declared?

If not here, then we should make sure to initialize every field in each constructor, in an initializer list. And that is likely to be repetitive and error-prone, seems to me.

stephengtuggy:
@royfalk @BenjamenMeyer Regarding trivial getters and setters: Yeah, I would agree that we should avoid them in new code. But I hesitate to remove these getters and setters from existing code. At least until we can gain a little more confidence that indeed they are, have always been, and should always be trivial.

stephengtuggy:
@BenjamenMeyer Okay, I see your point regarding explicit vs. implicit constructor logic. It'll be more work to do it that way, but perhaps it's worth it.

stephengtuggy:
> @royfalk @BenjamenMeyer Regarding trivial getters and setters: Yeah, I would agree that we should avoid them in new code. But I hesitate to remove these getters and setters from existing code. At least until we can gain a little more confidence that indeed they are, have always been, and should always be trivial.

One thing I will say, though, is that the getter and setter for `attackPreference` should not use the same name! Same with `unitRole`. It's very confusing that way, at least to me.

engine/src/cmd/basecomputer.cpp
royfalk: It started as am image to be displayed in the targeting hud. Then stuff like this migrated there for some reason.

engine/src/cmd/energetic.h
royfalk: I am aware that structs and classes simply differ in the default visibility modifier, but in this case it's probably more relevant to do it in the outer class constructor anyhow. It's immediately used by the _jump_ field, so no other insntace can use it anyhow (not sure what this is called).

In any case, this is an excellent example of a struct that accumulated seemingly unrelated fields. It has ratings, energy and insysenergy (with a _short fix_ comment!) - all seem to be the same thing more or less. They had everything in unit_generic, including lots of variables, so their solution was to stick a couple in a struct.

royfalk:
The discussion went all over the place. I'd like to summarize where we are and arrive at a consensus.

- trivial setters with near identical names - how about we make the field private and make the getter get<field_name>? That would enforce use of getter outside the class and future proof it. 
  You should know there's no logic I could see why some variables get getters and others do not. Also, I'm sure there is a lot of direct access done even when there are getters. I'd note that this part of the code is not yet ready for anything that requires non-trivial getters - concurrency is probably the most relevant use case for non-trivial getters.
- Initialization - I'll move to a constructor/s. It's scut, but not terrible.
- (lack of) Braces - as per my policy, I'll do the ones I see and won't add any more.
- Static variables of configuration - I plan on working on a configuration class in a near PR. I'll concentrate exclusively on this in the PR and push it quickly.

What do you think?

royfalk:
Of course they had another function getUnitRole that returns a string. Why wouldn't you. That would become unitRoleDescription.

engine/src/cmd/armed.h
BenjamenMeyer: If we break things down properly it won't be so overwhelming for each constructor to initialize its variables.

BenjamenMeyer:
> The discussion went all over the place. I'd like to summarize where we are and arrive at a consensus.
> 
> * trivial setters with near identical names - how about we make the field private and make the getter get<field_name>? That would enforce use of getter outside the class and future proof it.
>   You should know there's no logic I could see why some variables get getters and others do not. Also, I'm sure there is a lot of direct access done even when there are getters. I'd note that this part of the code is not yet ready for anything that requires non-trivial getters - concurrency is probably the most relevant use case for non-trivial getters.

Yes, making the variables private/protected would go a long way to start, and enforcing the use of the getters/setters externally.

> * Initialization - I'll move to a constructor/s. It's scut, but not terrible.

Thanks. We'll see how the discussion goes in the other related issues. Again, thanks for driving that discussion.

> * (lack of) Braces - as per my policy, I'll do the ones I see and won't add any more.

Generally I'm fine with that. It's just hard to tell in a PR what is new and what isn't in this kind of refactor, so some things stick out more in PRs.

> * Static variables of configuration - I plan on working on a configuration class in a near PR. I'll concentrate exclusively on this in the PR and push it quickly.

I know there's a lot of the static variable stuff. If it was already in the function then we'll address in a round specific to the config stuff and can delay that; just asking that we don't introduce anything _new_ using that pattern if at all possible.

All-in-all good summary

stephengtuggy:
@royfalk I think your latest proposal is fine. Thanks for sticking with it.

engine/src/cmd/ai/aggressive.cpp
royfalk: As noted below, I've made changes to comply, except for config values, which would come in a near PR.

engine/src/cmd/container.h
royfalk: Missed that one.

engine/src/cmd/energetic.h
royfalk: Also missed this.

royfalk:
I think I'm done with all the comments. As a bonus, I think I figured out the whole damage issue. The fighter barracks has two types of green beams. If you're facing the weak side, you only get the weak beams and they can't hurt me. I moved sides and died almost instantly.

BenjamenMeyer (approved):
Thanks! and again thanks for starting the related conversations.

PR #480, opened 2021-06-27
Fix/opcode issue 342

### LINKED ISSUE #342: opcode collision system: Use `int32_t` or similar instead of `int`, where appropriate
According to a comment in engine/src/cmd/collide2/Stdafx.h, the OPCODE collision system assumes that `sizeof(float) == sizeof(int) == sizeof(unsigned int) == sizeof(dword)`. On 64-bit systems, I'm pretty sure that this assumption is incorrect. That may well be why the collision system hasn't been working correctly.

I think we should replace all instances of `int` with `int32_t`, and replace `unsigned int` with `uint32_t` (https://en.cppreference.com/w/cpp/types/integer). This appears to be the C++11 way to do it. Hopefully the collision system will then start working better again.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #342 

Purpose:
- What is this pull request trying to do? **Fix some issues with the collision detection system. Specifically, some incorrect assumptions about the size of data types such as `int`; and also getting rid of a bunch of warnings that have been suppressed via `#pragma` for some years now.**
- What release is this for? **0.9.x**
- Is there a project or milestone we should apply this to? **0.9.x**


engine/src/cmd/collide2/CSopcodecollider.cpp
```diff
@@ -128,7 +128,7 @@ csOPCODECollider::~csOPCODECollider ()
 }
 
 
-void csOPCODECollider::MeshCallback (udword triangle_index,
+void csOPCODECollider::MeshCallback (ice_udword triangle_index,
```
stephengtuggy, line 131: Most of the changes in this PR are similar to this one, where I just renamed some integer type aliases used by `ICE` and `opcode` or `OPC`. Didn't want them to conflict with each other's (or anyone else's) typedefs.

engine/src/cmd/collide2/Ice/IceFPU.h
```diff
@@ -14,17 +14,24 @@
 
 	#define	SIGN_BITMASK			0x80000000
 
+    union ice_union
+    {
+        ice_sdword sdword_member;
+        ice_udword udword_member;
+        float      float_member;
+    };
+
```
stephengtuggy, line 23: This is one of the key changes in this PR.

engine/src/cmd/collide2/Ice/IceFPU.h
```diff
@@ -14,17 +14,24 @@
...
 	//! Integer representation of a floating-point value.
-	#define IR(x)					((udword&)(x))
+	#define IR(x)					((static_cast<ice_union*>((void *)&(x)))->udword_member)
```
stephengtuggy, line 25: Another key set of changes. Before, everywhere these four macros were used (and they are used a LOT), they were triggering warnings about type-punning with pointers violating strict aliasing, or something like that. A previous developer suppressed these warnings with a `#pragma` GCC directive in Stdafx.h. Now the warnings are no longer suppressed. They don't need to be, because the code is more correct. :-)

engine/src/cmd/collide2/Ice/IceTypes.h
```diff
@@ -45,38 +46,38 @@
...
+
+    // ICE_COMPILE_TIME_ASSERT(sizeof(bool)==1);	// ...otherwise things might fail with VC++ 4.2 !
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_ubyte)==1);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_sbyte)==1);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_sword)==2);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_uword)==2);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_udword)==4);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_sdword)==4);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_uqword)==8);
+    ICE_COMPILE_TIME_ASSERT(sizeof(ice_sqword)==8);
```
stephengtuggy, line 68: Here's where I renamed these typedefs. Also, please, let's use the fixed-width integer types that C++11 provides, instead of making the faulty assumptions that the code made before.

engine/src/cmd/collide2/OPC_HybridModel.cpp
```diff
@@ -301,11 +301,11 @@ bool HybridModel::Build(const OPCODECREATE& create)
  *	\return		amount of bytes used
  */
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-udword HybridModel::GetUsedBytes() const
+ice_udword HybridModel::GetUsedBytes() const
```
stephengtuggy, line 304: Hmm. This method isn't correct. I'll have to revisit it.

engine/src/cmd/collide2/OPC_SweepAndPrune.cpp
```diff
@@ -201,9 +201,9 @@ SAP_Element* SAP_PairData::GetFreeElem(udword id, SAP_Element* next, udword* rem
 
 			// Remap everything
 			{
-				udword Delta = uintptr_t(NewElems) - uintptr_t(mElementPool);
+				ice_udword Delta = uintptr_t(NewElems) - uintptr_t(mElementPool);
```
stephengtuggy, line 204: Hmm. Didn't catch this. `Delta` should probably be of type `ice_ptrdiff_t` or whatever it's called.

engine/src/cmd/collide2/OPC_TriBoxOverlap.h
```diff
@@ -104,7 +104,7 @@ inline_ bool planeBoxOverlap(const Point& normal, const float d, const Point& ma
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 /**
  *	Triangle-Box overlap test using the separating axis theorem.
- *	This is the code from Tomas M�ller, a bit optimized:
+ *	This is the code from Tomas M�ller, a bit optimized:
```
stephengtuggy, line 107: Whoops, what happened here?

engine/src/cmd/collide2/Stdafx.h
```diff
@@ -29,10 +30,11 @@
...
 
-//chuck starchaser shutting up the warnings about violations of strict aliasing...
-#ifndef _WIN32 // pheonixstorm shutting up pragma warnings on win32
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-#endif
+// Stephen G. Tuggy getting rid of this hack entirely
+// chuck starchaser shutting up the warnings about violations of strict aliasing...
+// #ifndef _WIN32 // pheonixstorm shutting up pragma warnings on win32
+// #pragma GCC diagnostic ignored "-Wstrict-aliasing"
+// #endif
```
stephengtuggy, line 37: Here's where I turned the warnings back on.

engine/src/cmd/collide2/opcodetypes.h
```diff
@@ -67,46 +67,14 @@
...
-typedef int64_t int64;
-#endif
+typedef uint8_t opc_uint8;
+typedef int8_t opc_int8;
+typedef uint16_t opc_uint16;
+typedef int16_t opc_int16;
+typedef uint32_t opc_uint32;
+typedef int32_t opc_int32;
+typedef uint64_t opc_uint64;
+typedef int64_t opc_int64;
```
stephengtuggy, line 77: Another key change. Let's just use the C++11 standard fixed-width integer types, regardless of platform. It should work.

stephengtuggy (commented):
Looks like there are a couple of things I'll have to come back to.

engine/src/cmd/collide2/OPC_TriBoxOverlap.h
stephengtuggy: Fixed.

engine/src/cmd/collide2/Ice/IceFPU.h
BenjamenMeyer: Q: do we need versions for pointers and versions for non-pointers? (Just curious)
Also, might it make sense to change these from C Macros to C++ inline methods?

engine/src/cmd/collide2/Ice/IceTypes.h
```diff
@@ -45,38 +46,38 @@
...
+    typedef int8_t              ice_sbyte;      //!< sizeof(ice_sbyte)      must be 1
+    typedef uint8_t             ice_ubyte;      //!< sizeof(ice_ubyte)      must be 1
+    typedef int16_t             ice_sword;      //!< sizeof(ice_sword)      must be 2
+    typedef uint16_t            ice_uword;      //!< sizeof(ice_uword)      must be 2
+    typedef int32_t             ice_sdword;     //!< sizeof(ice_sdword)     must be 4
+    typedef uint32_t            ice_udword;     //!< sizeof(ice_udword)     must be 4
+    typedef int64_t             ice_sqword;     //!< sizeof(ice_sqword)	    must be 8
+    typedef uint64_t            ice_uqword;     //!< sizeof(ice_uqword)     must be 8
+    typedef float               ice_float32;    //!< sizeof(ice_float32)    must be 4
+    typedef double              ice_float64;    //!< sizeof(ice_float64)    must be 4
```
BenjamenMeyer, line 58: comment copy/pasta? double is 8 bytes, not 4.

engine/src/cmd/collide2/Ice/IceTypes.h
```diff
@@ -45,38 +46,38 @@
...
-	ICE_COMPILE_TIME_ASSERT(sizeof(uqword)==8);
-	ICE_COMPILE_TIME_ASSERT(sizeof(sqword)==8);
+    typedef int8_t              ice_sbyte;      //!< sizeof(ice_sbyte)      must be 1
+    typedef uint8_t             ice_ubyte;      //!< sizeof(ice_ubyte)      must be 1
+    typedef int16_t             ice_sword;      //!< sizeof(ice_sword)      must be 2
+    typedef uint16_t            ice_uword;      //!< sizeof(ice_uword)      must be 2
+    typedef int32_t             ice_sdword;     //!< sizeof(ice_sdword)     must be 4
+    typedef uint32_t            ice_udword;     //!< sizeof(ice_udword)     must be 4
+    typedef int64_t             ice_sqword;     //!< sizeof(ice_sqword)	    must be 8
+    typedef uint64_t            ice_uqword;     //!< sizeof(ice_uqword)     must be 8
```
BenjamenMeyer, line 56: Question: what is relevant to convey here?

I don't want to create churn for the sake of churn so before any change is made (if any) it'd like to have a proper discussion on that aspect.

Personally I think the bit sizes are more important than the word/dword/qword information as it's more meaningful. If that's the case, then `ice_ubyte` should be `ice_uint8`. If there is a necessary meaning to using word/dword/qword then it's fine AS-IS; but I want to make sure there's good reason. Most have moved away from the word/dword/qword nomenclature AFAICT; so do we need to stick with it too? Or should we move on?

Again, raising the question at the moment, not necessary asking for an immediate change.

engine/src/cmd/collide2/Ice/IceTypes.h
```diff
@@ -90,31 +91,31 @@
...
+	#define	MAX_UBYTE				0xff						//!< max possible ice_ubyte value
+	#define	MIN_UBYTE				0x00						//!< min possible ice_ubyte value
+	#define	MAX_SWORD				0x7fff						//!< max possible ice_sword value
+	#define	MIN_SWORD				0x8000						//!< min possible ice_sword value
+	#define	MAX_UWORD				0xffff						//!< max possible ice_uword value
+	#define	MIN_UWORD				0x0000						//!< min possible ice_uword value
+	#define	MAX_SDWORD				0x7fffffff					//!< max possible ice_sdword value
+	#define	MIN_SDWORD				0x80000000					//!< min possible ice_sdword value
+	#define	MAX_UDWORD				0xffffffff					//!< max possible ice_udword value
+	#define	MIN_UDWORD				0x00000000					//!< min possible ice_udword value
```
BenjamenMeyer, line 118: These probably should be cast to ensure they don't get accidentally converted to another value - e.g  an 8-bit signed max getting stored into a 16-bit value. I'd stick with the standard c11 types:
```
#define MAX_SBYTE int8_t(0x7F)
```
though if this is only for the ICE subsystem then using the ICE types would be okay too - `ice_sbyte(0x7F)`.

engine/src/cmd/collide2/OPC_HybridModel.cpp
BenjamenMeyer: If you're not going to do so before this gets merged, then please file an issue so it doesn't get forgotten

engine/src/cmd/collide2/OPC_SweepAndPrune.cpp
BenjamenMeyer: probably should have a comment about the importance of the pointer arithmetic too

BenjamenMeyer (changes requested):
Left a few comments; namely blocking to have a discussion about the word/dword/qword resolved before this goes forward.

engine/src/cmd/collide2/Ice/IceFPU.h
stephengtuggy: @BenjamenMeyer no, we don't need separate versions for pointers. They wouldn't be used anywhere.

It might make sense to change these to C++ inline methods at some point, but that seems like a potentially risky move. One that I'm not ready to make in this PR.

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: Hmm. Yeah, that's a mistake. I'll fix it

engine/src/cmd/collide2/Ice/IceTypes.h
royfalk: Unfortunately, I agree. Word is 32 bit(?). Dword and qword are terms I'd have to check.
Honestly, ~~with the introduction of bit size variables~~, why are we using typedefs?
The codebase has a fascination with adding complexity, which just makes it more and more difficult to add stuff.

stephengtuggy: @BenjamenMeyer That's a good point. `ice_uint8` is probably a lot better, actually. And that change should be easy enough to make. I guess I'll go ahead and make that correction.

engine/src/cmd/collide2/Ice/IceTypes.h
royfalk: If we used bit sized variables, we could probably use the system definitions and get rid of this. I'm assuming this matches some kind of system def.

engine/src/cmd/collide2/Ice/IceFPU.h
BenjamenMeyer: Changing to C++ inline methods shouldn't make a difference for risk; but would expose places where they're changed as linting would be a bit more effective.

engine/src/cmd/collide2/Ice/IceTypes.h
BenjamenMeyer: @royfalk yes, we could potentially declare these as C++ constants instead of C Macros; I'm not sure we need to right now though we might want to in the future; right now, I'm mostly wanting to make sure that the values don't get sign extended (https://en.wikipedia.org/wiki/Sign_extension), adding the types on each should at least make that visible to us via warnings.

engine/src/cmd/collide2/Ice/IceTypes.h
BenjamenMeyer: sounds like there's some agreement.

@royfalk yes, `WORD` is 16 bits, `DWORD` = Double `WORD`, 32-bits; `QWORD` = Quad `WORD` so 64-bits; but that assumes you know that a `WORD` is two `bytes`. Typically WORD/DWORD/QWORD are all unsigned values too. https://en.wikipedia.org/wiki/Word_(computer_architecture)#Size_families has some good info on this whole thing, which while making some of it clearer will also make things a lot more confusing.

stephengtuggy: OK. I've gone ahead and changed the type names to `ice_uint8`, `ice_sint8` (signed 8-bit integer for the ICE subsystem), etc. But should we get rid of the typedefs, and just go ahead and use `uint8_t`, etc. directly?

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: OK, I mostly switched to using the c++11 standard `INT32_MIN`, `INT32_MAX`, `UINT32_MAX`, etc. For the different width unsigned zeros, I do use `uint32_t(0x00000000)`, etc.

Again, though, do we really need these constants that are specific to ICE and OPCODE? Or can we just use `INT32_MAX` etc., directly?

engine/src/cmd/collide2/Ice/IceTypes.h
BenjamenMeyer: @stephengtuggy does it add value? if not, change it.

engine/src/cmd/collide2/Ice/IceTypes.h
BenjamenMeyer: is there value added? If not, migrate.

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: 👍🏽 I'll change it.

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: Fixed

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: Done

engine/src/cmd/collide2/OPC_HybridModel.cpp
stephengtuggy: Filed as #488

engine/src/cmd/collide2/OPC_SweepAndPrune.cpp
stephengtuggy: Both done

engine/src/cmd/collide2/Ice/IceTypes.h
stephengtuggy: Migrated

engine/src/cmd/collide2/CSopcodecollider.cpp
stephengtuggy: Update: No longer applicable

engine/src/cmd/collide2/Ice/IceMemoryMacros.h
```diff
@@ -85,14 +91,14 @@
...
 	#define SIZEOFOBJECT		sizeof(*this)									//!< Gives the size of current object. Avoid some mistakes (e.g. "sizeof(this)").
 	//#define CLEAROBJECT		{ memset(this, 0, SIZEOFOBJECT);	}			//!< Clears current object. Laziness is my business. HANDLE WITH CARE.
-	#define DELETESINGLE(x)		{ delete x; x = null; }		//!< Deletes an instance of a class.
-	#define DELETEARRAY(x)		{ delete []x; x = null; }		//!< Deletes an array.
-	#define SAFE_RELEASE(x)		if (x) { (x)->Release();		(x) = null; }	//!< Safe D3D-style release
-	#define SAFE_DESTRUCT(x)	if (x) { (x)->SelfDestruct();	(x) = null; }	//!< Safe ICE-style release
+	#define DELETESINGLE(x)		{ delete x; x = nullptr; }		//!< Deletes an instance of a class.
+	#define DELETEARRAY(x)		{ delete []x; x = nullptr; }		//!< Deletes an array.
+	#define SAFE_RELEASE(x)		if (x) { (x)->Release();		(x) = nullptr; }	//!< Safe D3D-style release
+	#define SAFE_DESTRUCT(x)	if (x) { (x)->SelfDestruct();	(x) = nullptr; }	//!< Safe ICE-style release
```
BenjamenMeyer, line 101: can we turn these into inline functions?

engine/src/cmd/collide2/opcodetypes.h
```diff
@@ -150,34 +121,10 @@ typedef int64_t int64;
 // <stddef.h> provided the types by checking if _INTPTR_T_DEFINED has been
 // #defined; newer versions of MSVC will provide them; older ones will not.  If
 // all else fails, then we fake up these types on our own.
+//
+// 2021-06-26 SGT: I believe these are standard types now in C++11 and later.
```
BenjamenMeyer, line 125: VS2010 or VS2012 started including hte `stddef.h` which has long defined the standard types on Unix/Linux systems. MS was the only oddball compiler left not providing it. I don't recall if the bit-types were pulled into ISO C or ISO C++ or not; it may still be by convention and everyone finally got onboard with providing them.

engine/src/cmd/collide2/Ice/IceMemoryMacros.h
stephengtuggy: @BenjamenMeyer Can you give an example of what one of these would look like as an inline function?

BenjamenMeyer: @stephengtuggy 
```C++
inline void DELETESINGLE(void** x) { delete *x; *x = nullptr; }
inline void SAFE_RELEASE(void** x) { if (*x) { (x)->Release(); (x) = nullptr; }
```
Or better yet (one of the few uses of a template):
```C++
template<typename T>
void DELETESINGLE(T x) {
    delete x;
    x = nullptr;
}
template<typename T>
void SAFE_RELEASE(T x) {
    if (x) {
        (x)->Release();
        (x) = nullptr;
    }
}
```

Or something like that...haven't tested the code but that should be about right.

BenjamenMeyer (changes requested):
If possible, please change the Macros to inline C++ functions. If not possible; just note it and we'll move on.

engine/src/cmd/collide2/Ice/IceMemoryMacros.h
royfalk: I'd also consider renaming to lower case as this would no longer be a macro.

engine/src/cmd/collide2/Ice/IceTypes.h
```diff
@@ -90,31 +96,31 @@
...
+	#define	MIN_UBYTE				uint8_t(0x00)				//!< min possible uint8_t value
+	// #define	MAX_SWORD				INT16_MAX       			//!< max possible int16_t value
+	// #define	MIN_SWORD				INT16_MIN       			//!< min possible int16_t value
+	// #define	MAX_UWORD				UINT16_MAX					//!< max possible uint16_t value
+	#define	MIN_UWORD				uint16_t(0x0000)			//!< min possible uint16_t value
+	// #define	MAX_SDWORD				INT32_MAX					//!< max possible int32_t value
+	// #define	MIN_SDWORD				INT32_MIN					//!< min possible int32_t value
+	// #define	MAX_UDWORD				UINT32_MAX					//!< max possible uint32_t value
+	#define	MIN_UDWORD				uint32_t(0x00000000)	    //!< min possible uint32_t value
 	#define	MAX_FLOAT				FLT_MAX						//!< max possible float value
```
royfalk, line 124: Nice!

engine/src/cmd/collide2/opcodesysdef.h
```diff
@@ -1,7 +1,9 @@
 /**
  * opcodesysdef.h
  *
+ * Copyright (C) Daniel Horn
  * Copyright (C) 2020 pyramid3d, Stephen G. Tuggy, and other Vega Strike contributors.
+ * Copyright (C) 2021 Stephen G. Tuggy
```
royfalk, line 6: It does not look as if you've made any changes to this file.
I don't care that you added the line but I wonder if you meant to make more changes and something got lost somewhere.

royfalk:
Overall, it looks good to me. I'll do a play test or two later today before I approve.

royfalk:
OK. Cloned the repo and built it. 
I did not notice any improvement in the collider. I could still fire missiles at point blank and miss.

The game had an issue with excessive flash effects following the destruction of ships. I would consider holding off on merging if others are seeing this too. Then again, the whole flash effect from engines isn't a show stopper.

Also I noticed some slow downs compared to my build. This can be explained by the influx of ships after I killed Oswald. There were more than usual and after killing a few there was a lot of cargo debris. Again, only an issue if others are seeing it.

ministerofinformation:
> The game had an issue with excessive flash effects following the destruction of ships. I would consider holding off on merging if others are seeing this too. Then again, the whole flash effect from engines isn't a show stopper.

I saw this as well, but in an older branch off of master that I'm doing the collision work on, so I don't think this is an issue introduced with these changes - I see it on any substantial damage, not just destructions. We should file a a separate issue for the flash effects - my take is that something seems off with the scaling size for the animation.

ministerofinformation (approved):
Did a playtest - collision detection still isn't perfect, and I think I noticed some coordinate system dependence on accuracy, but those seem to be deeper library/interfacing issues for a later day and not germane to the PR, which seems to be in good shape to me. Collision detection is at least as good as it was before and the code is safer and saner for the efforts.

engine/src/cmd/collide2/OPC_PlanesAABBOverlap.h
```diff
@@ -20,14 +28,14 @@ inline_ bool PlanesCollider::PlanesAABBOverlap(const Point& center, const Point&
 
 	// Evaluate through all active frustum planes. We determine the relation 
 	// between the AABB and a plane by using the concept of "near" and "far"
-	// vertices originally described by Zhang (and later by M�ller). Our
+	// vertices originally described by Zhang (and later by M�ller). Our
 	// variant here uses 3 fabs ops, 6 muls, 7 adds and two floating point
```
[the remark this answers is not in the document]
stephengtuggy: Is it still?? I thought I fixed that already!

engine/src/cmd/collide2/OPC_MeshInterface.cpp
```diff
@@ -114,6 +120,12 @@
 */
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+/**
+ * Updated by Stephen G. Tuggy 2021-07-03
+ */
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
```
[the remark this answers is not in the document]
stephengtuggy: Removed duplicate.

engine/src/cmd/collide2/OPC_HybridModel.cpp
```diff
@@ -80,6 +86,12 @@
 */
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+/**
+ * Updated by Stephen G. Tuggy 2021-07-03
+ */
+///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
stephengtuggy: Removed duplicate.

engine/src/cmd/collide2/opcodesysdef.h
stephengtuggy: No, I don't think there are any changes missing. Thanks for checking

engine/src/cmd/collide2/Ice/IceMemoryMacros.h
stephengtuggy: Converted to inline functions (template versions).

BenjamenMeyer (approved):
LGTM! Thanks!

engine/src/cmd/collide2/Ice/IceMemoryMacros.h
BenjamenMeyer: Thanks! Looks great!

engine/src/cmd/collide2/OPC_PlanesAABBOverlap.h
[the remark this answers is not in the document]
BenjamenMeyer: Looks fixed now

PR #484, opened 2021-06-30
Draft - Initial work on a configuration library

royfalk:
These is a preliminary draft of the work I've promised to do in the previous PR.
I've intentionally separated the files to their own folder and library, as they are standalone and do not require any vegastrike headers.

The idea here is that GameConfig::LoadConfig is called very early in the game, and loads all settings to a private variable.
Then, when a struct is accessed for one of the fields in it, it is lazily populated by calling its constructor.

I may separate the structs to a different file (e.g. configuration.h) or several files (weapons_config.h), depending on its size.
I hope to add unit test to this and so have more robust tests. I am aware it currently has a bug that disabled movement.

**The ask here is for you to review this early in the process, to make sure you agree with the plan.** I've added sample config to armed.cpp so it'll be clearer what I intend to do and how it will look.

Please answer the following:

Code Changes:
- [ not yet ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Pending

Purpose:
- What release is this for?

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -84,6 +84,10 @@ INCLUDE_DIRECTORIES(
 )
 ENDIF (MSVC)
 
+SET(LIBCONFIG
+    src/configuration/game_config.cpp
+)
+
 SET(LIBGUI_SOURCES
     src/gui/button.cpp
     src/gui/control.cpp
@@ -368,7 +372,6 @@ SET(LIBGFXSERVER_SOURCES
 )
 
 SET(LIBROOTGENERIC_SOURCES
-    src/game_config.cpp
     src/configxml.cpp
     src/easydom.cpp
     src/endianness.cpp
@@ -467,6 +470,7 @@ SET(LIBAUDIO_SOURCES
 
 
 ADD_LIBRARY(vegastrike-engine_com
+    ${LIBCONFIG}
     ${LIBAI_SOURCES}
     ${LIBCMD_SOURCES}
     ${LIBNET_SOURCES}
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -550,6 +550,7 @@ SET(VEGASTRIKE_SOURCES
     src/python/briefing_wrapper.cpp
 )
 
+
 IF (COMMAND cmake_policy)
     CMAKE_POLICY(SET CMP0003 NEW)
 ENDIF (COMMAND cmake_policy)
@@ -1535,3 +1536,34 @@ ELSE ()
 ENDIF (WIN32 AND NOT UNIX)
 
 INCLUDE(CPack)
+
+INCLUDE(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+)
+# For Windows: Prevent overriding the parent project's compiler/linker settings
+SET(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
+
+enable_testing()
+
+file(COPY "src/configuration/tests/vegastrike.config" DESTINATION ${CMAKE_BINARY_DIR})
+
+add_executable(
+    unit_tests
+    src/configuration/tests/configuration_tests.cpp
+)
+
+ADD_LIBRARY(vegastrike-testing
+    ${LIBCONFIG}
+)
+
+target_link_libraries(
+  unit_tests
+  gtest_main
+  vegastrike-testing
+)
+
+include(GoogleTest)
+gtest_discover_tests(unit_tests)
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1537,35 +1537,3 @@ ENDIF (WIN32 AND NOT UNIX)
 
 INCLUDE(CPack)
 
-IF (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 2.10)
-    INCLUDE(FetchContent)
-    FetchContent_Declare(
-      googletest
-      URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
-    )
-    # For Windows: Prevent overriding the parent project's compiler/linker settings
-    SET(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-    FetchContent_MakeAvailable(googletest)
-
-    enable_testing()
-
-    file(COPY "src/configuration/tests/vegastrike.config" DESTINATION "${CMAKE_BINARY_DIR}/test_assets")
-
-    add_executable(
-        unit_tests
-        src/configuration/tests/configuration_tests.cpp
-    )
-
-    ADD_LIBRARY(vegastrike-testing
-        ${LIBCONFIG}
-    )
-
-    target_link_libraries(
-      unit_tests
-      gtest_main
-      vegastrike-testing
-    )
-
-    include(GoogleTest)
-    gtest_discover_tests(unit_tests)
-ENDIF (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 2.10)
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1537,33 +1537,35 @@ ENDIF (WIN32 AND NOT UNIX)
 
 INCLUDE(CPack)
 
-INCLUDE(FetchContent)
-FetchContent_Declare(
-  googletest
-  URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
-)
-# For Windows: Prevent overriding the parent project's compiler/linker settings
-SET(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
-FetchContent_MakeAvailable(googletest)
+IF (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 2.10)
+    INCLUDE(FetchContent)
+    FetchContent_Declare(
+      googletest
+      URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+    )
+    # For Windows: Prevent overriding the parent project's compiler/linker settings
+    SET(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+    FetchContent_MakeAvailable(googletest)
 
-enable_testing()
+    enable_testing()
 
-file(COPY "src/configuration/tests/vegastrike.config" DESTINATION "${CMAKE_BINARY_DIR}/test_assets")
+    file(COPY "src/configuration/tests/vegastrike.config" DESTINATION "${CMAKE_BINARY_DIR}/test_assets")
 
-add_executable(
-    unit_tests
-    src/configuration/tests/configuration_tests.cpp
-)
+    add_executable(
+        unit_tests
+        src/configuration/tests/configuration_tests.cpp
+    )
 
-ADD_LIBRARY(vegastrike-testing
-    ${LIBCONFIG}
-)
+    ADD_LIBRARY(vegastrike-testing
+        ${LIBCONFIG}
+    )
 
-target_link_libraries(
-  unit_tests
-  gtest_main
-  vegastrike-testing
-)
+    target_link_libraries(
+      unit_tests
+      gtest_main
+      vegastrike-testing
+    )
 
-include(GoogleTest)
-gtest_discover_tests(unit_tests)
+    include(GoogleTest)
+    gtest_discover_tests(unit_tests)
+ENDIF (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 2.10)
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1548,7 +1548,7 @@ FetchContent_MakeAvailable(googletest)
 
 enable_testing()
 
-file(COPY "src/configuration/tests/vegastrike.config" DESTINATION ${CMAKE_BINARY_DIR})
+file(COPY "src/configuration/tests/vegastrike.config" DESTINATION "${CMAKE_BINARY_DIR}/test_assets")
 
 add_executable(
     unit_tests
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -85,6 +85,7 @@ INCLUDE_DIRECTORIES(
 ENDIF (MSVC)
 
 SET(LIBCONFIG
+    src/configuration/configuration.cpp
     src/configuration/game_config.cpp
 )
 
--- a/engine/src/cmd/armed.cpp
+++ b/engine/src/cmd/armed.cpp
@@ -25,7 +25,7 @@
 
 #include "armed.h"
 
-#include "configuration/game_config.h"
+#include "configuration/configuration.h"
 #include "mount.h"
 #include "mount_size.h"
 #include "weapon_info.h"
@@ -190,11 +190,9 @@ void Armed::ActivateGuns( const weapon_info *sz, bool ms )
 void Armed::Fire( unsigned int weapon_type_bitmask, bool listen_to_owner)
 {
     Unit *unit = static_cast<Unit*>(this);
-    /*static bool can_fire_in_spec  = GameConfig::GetVariable( "physics", "can_fire_in_spec", false);
-    static bool can_fire_in_cloak = GameConfig::GetVariable( "physics", "can_fire_in_cloak", false);*/
-    static bool verbose_debug     = GameConfig::GetVariable("data", "verbose_debug", false);
 
-    if ( (unit->cloaking >= 0 && weapons_config.can_fire_in_cloak == false) || (unit->graphicOptions.InWarp && weapons_config.can_fire_in_spec == false) ) {
+    if ( (unit->cloaking >= 0 && configuration.weapons.can_fire_in_cloak == false) ||
+         (unit->graphicOptions.InWarp && configuration.weapons.can_fire_in_spec == false) ) {
         UnFire();
         return;
     }
@@ -236,7 +234,7 @@ void Armed::Fire( unsigned int weapon_type_bitmask, bool listen_to_owner)
         const bool locked_missile   = (mis && locked_on && lockable_weapon);
         const bool missile_and_want_to_fire_missiles = ( mis && (weapon_type_bitmask&ROLES::FIRE_MISSILES) );
         const bool gun_and_want_to_fire_guns = ( (!mis) && (weapon_type_bitmask&ROLES::FIRE_GUNS) );
-        if (verbose_debug && missile_and_want_to_fire_missiles && locked_missile)
+        if (configuration.logging.verbose_debug && missile_and_want_to_fire_missiles && locked_missile)
             VSFileSystem::vs_fprintf( stderr, "\n about to fire locked missile \n" );
         bool want_to_fire = (fire_non_autotrackers || autotracking_gun || locked_missile) &&
             //&& ( (ROLES::EVERYTHING_ELSE&weapon_type_bitmask&i->type->role_bits) || i->type->role_bits == 0 )
--- a/engine/src/cmd/armed.cpp
+++ b/engine/src/cmd/armed.cpp
@@ -25,7 +25,7 @@
 
 #include "armed.h"
 
-#include "game_config.h"
+#include "configuration/game_config.h"
 #include "mount.h"
 #include "mount_size.h"
 #include "weapon_info.h"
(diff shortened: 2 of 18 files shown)


engine/src/cmd/armed.cpp
```diff
@@ -190,11 +190,11 @@ void Armed::ActivateGuns( const weapon_info *sz, bool ms )
...
 {
     Unit *unit = static_cast<Unit*>(this);
-    static bool can_fire_in_spec  = GameConfig::GetVariable( "physics", "can_fire_in_spec", false);
-    static bool can_fire_in_cloak = GameConfig::GetVariable( "physics", "can_fire_in_cloak", false);
+    /*static bool can_fire_in_spec  = GameConfig::GetVariable( "physics", "can_fire_in_spec", false);
+    static bool can_fire_in_cloak = GameConfig::GetVariable( "physics", "can_fire_in_cloak", false);*/
     static bool verbose_debug     = GameConfig::GetVariable("data", "verbose_debug", false);
 
-    if ( (unit->cloaking >= 0 && can_fire_in_cloak == false) || (unit->graphicOptions.InWarp && can_fire_in_spec == false) ) {
+    if ( (unit->cloaking >= 0 && weapons_config.can_fire_in_cloak == false) || (unit->graphicOptions.InWarp && weapons_config.can_fire_in_spec == false) ) {
```
BenjamenMeyer, line 197: This can be cached in the object data as it shouldn't change during runtime, however object state might change in relevant ways so only the object's stored version should then get modified
Object can load the information in it's constructor (or a common function if there are multiple constructors or if the object needs to be reset).

Iow it's rare that objects would touch the global object.

engine/src/configuration/game_config.cpp
```diff
@@ -1,4 +1,13 @@
-#include "game_config.h"
+#include "configuration/game_config.h"
+
+// Global Variable Definitions
+WeaponsConfig weapons_config;
```
BenjamenMeyer, line 4: We should have a single config object for everything, and weapons config can be a sub-object of that object.

BenjamenMeyer, line 4: We should have a single config object for everything, and weapons config can be a sub-object of that object.

BenjamenMeyer (commented):
@royfalk good starts a couple minor tweaks from what I was thinking, but a good first round. thanks!

engine/src/cmd/armed.cpp
royfalk: Just to clarify what you mean - it can be cached within Armed as a field or if it is in only one function, within that function as a static variable?

There was also the question of const-ness. Currently, all settings are loaded from a file and are therefore const. However, in the future, I imagine a GUI to modify these in-game.

Regarding the caching, I'm not sure why we would do such a thing. Wouldn't accessing the config object be  extremely fast?! There's no function at all. You're accessing a variable.

Having said all of this, I actually thought about injecting these during initialization. One of the issues I have encountered when refactoring unit_generic is how interconnected everything in the game is to everything else. The Armed class, for example, doesn't need to know about GameConfig and vice-versa. However, something like this would not be trivial to implement, especially if we factor in changes in configuration mid-game.

At the end of this design review, I'd like to push an update just for Armed, showing the final form of phase 1. It won't be everything I've talked about such as DI and in-game configuration. Just the basic stuff.

BenjamenMeyer: no, don't use a static variable - use a member variable. This would be faster than accessing the global config all the time as it would be stack local with the object versus having to shunt out to another page to get the global information, invoke another function call, etc.

BenjamenMeyer: I wouldn't make them constant either; no need for that.

stephengtuggy: I have no idea why we would need caching. This mechanism would be very fast regardless, seems to me.

engine/src/configuration/game_config.cpp
stephengtuggy: I do agree on this point.

stephengtuggy:
Yeah, looks good to me so far.

engine/src/cmd/armed.cpp
BenjamenMeyer: @stephengtuggy we should only access the global objects as few times as possible; caching to a member variable will speed up access b/c it'll be object local memory as opposed to requiring another function call to access.

engine/src/configuration/configuration.h
```diff
@@ -0,0 +1,55 @@
...
+        //static bool does_missile_bounce = GameConfig::GetVariable( "physics", "missile_bounce", false);
+
+        Physics();
+    } physics;
+
+    struct Weapons {
+        const bool can_fire_in_cloak;
+        const bool can_fire_in_spec;
+        Weapons();
+    } weapons;
```
BenjamenMeyer, line 48: It would be better to declare these in a namespace (Vegastrike::Config), outside the class object, and then reference it.

engine/src/cmd/armed.cpp
ministerofinformation: (The (overall) performance impact of caching vs. not caching is going to largely be dependent on the tradeoff between two things: 1) total size of global config structure 2) degree of duplication among object instances + relative increase in object size. 
Considering extreme cases for both is simple: 

1. if the global config is very large and accessed with poor locality and infrequently, then the cost per access will be high for specifically those loads into the global config. If the locality/size/reuse distance is sufficiently bad, then, as Benjamen alluded to, the translation for the page(s) the global config is stored in may have been evicted from the L1 TLB or even the L2 TLB by the time the next field is accessed, which will incur additional overheads
2. if a substantial portion of the total fields of a given object are copies of the global config values and there are many instances of that object that are concurrently active, then the working set may have poor (HW) cache locality compared to a case where only one copy of these globals has to be in the (HW) cache and all of the objects themselves are smaller, which may make the average memory access latency for all memory operations worse.
 
As for the overhead of a call vs. caching and using, that depends on how the global struct is exposed. If it's full-on raw access to a fully public struct, there won't even be a function call - but you've also given away write access. Also, even without there being a function call, there may still be a table indirection, depending on how the global space is being managed. If there's a getter that returns a const ref to the struct, it's not clear that this will actually be compiled to a function call either, since it's really just static checking on the uses of a pointer you passed. If it is, and it's say, in a computationally intense inner loop, then the cached version will likely be much better. Again, however, the total performance impact will be dependent on what fraction of code benefits from caching vs. is penalized for the per-object overheads.

ministerofinformation: Re: local modifications --

While the mechanism would certainly support using copies of global configuration values for locally modifiable fields, I think we should consider whether that's a good paradigm in general to allow it to always be used that way. To that end, perhaps it's worth splitting the sets of global configuration values at the top level into a few bins, and this will make the semantics of each set cleaner:

1. Global constants for the loaded dataset, e.g. physics properties. There are some number of global, dataset specific constants (say, the conversion between VSU damage units and Joules). These are actually constant within the scope of a dataset. 

2. User-specific settings/preferences, e.g. game resolution, default volume. These are game options and, even if they don't (and can't currently) change at run-time, these are logically fungible -- and changes to these values should be global changes, because the properties are global.

3. Globally specified default / useful state values for member variables of objects. These would be used to initialize, reset, adjust member variables... but said member variables are not (at a higher logical semantics level) copy-on-write versions of the global field , but just member values that happen, at some point in time to be set to values from a global field. To give a more concrete example, we might initialize "fuelamount" to a global value for "initial_fuel_amount" - but we wouldn't really want a copy of initial_fuel_amount hanging around, because we already have a member variable that encodes the current state of that local property

BenjamenMeyer: I suspect we'll see more of case 1 than case 2 per https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/484#discussion_r663310115. We may even end up with mixtures of the two - e.g some code will exhibit case 1 while other code will exhibit case 2. Ultimately we won't know until we do real performance analysis.

That said, if we want to preserve our ability to later do writes on the config, then I suggest we do inline C++ getters instead of publicly exposing the values. For the moment,  it'll compile down to fast getters that will be equivalent of the current code but give us the flexibility to change to change it later easier.

ministerofinformation: I agree - let's do something that uses a clean abstraction and has the features we want first (like: modifiable, but not public), and then worry about the constant factors of the specific performance impact once we can measure it. If there's specific performance critical code that turns out to need to cache values locally, that can still be done on top of the fast getters as the usually-good-enough case (for actually-constants, it's even completely safe) ... and if everything/too much needs caching, then we can always revisit the organization in general.

engine/src/configuration/configuration.h
royfalk: Do you mean split the struct definition and move it outside and then only have X x; inside the class?

engine/src/configuration/game_config.cpp
royfalk: Done.

royfalk: Done.

engine/src/configuration/configuration.h
royfalk: If this is the case, I'm not sure I see the value. 
It's in a header file and we don't want to expose it, so we use a namespace. 
Why not just place it within the struct under private? 
Or keep it the way it is (anonymous structs). After all, we don't foresee any other instance of these inner structs.

This really is a question - what is the value of changing this?

BenjamenMeyer: It helps with unit testing for one.

Secondly its easier to read.

Honestly I don't see an issue exposing it too.

If we want to keep it private then we can stick it into an internal only header and do PIMPL...but for the moment that's too much work.

Personally I don't care for declaring structures and classes within other structures and classes. It's messy.

I also don't like a anonymous types, makes unit testing harder.

That's my opinion; go with what our code standard says as far actual usage as it may differ from my opinion.

royfalk:
> I wouldn't make them constant either; no need for that.

I use constant as a default. I know the settings aren't supposed to change in this version, so why not. This way I get an error if something does try to change. It's good habits.

> caching to a member variable will speed up access b/c it'll be object local memory as opposed to requiring another function call to access.

I'd like to do that only if absolutely necessary. Before we die optimizing on this hill, I'd like to profile it in specific cases, rather than everywhere. Mind you, this is already an improvement over parsing a string and accessing a data structure in:

`XMLSupport::parse_bool( vs_config->getVariable( "section", "key", "default" ) );

> That said, if we want to preserve our ability to later do writes on the config, then I suggest we do inline C++ getters instead of publicly exposing the values.

Let's wait with this. Write access was never the plan for phase 1 of settings. Also, consider creating getters/setters for potentially **1300** settings. When we do, we'll probably provide write access, setters and GUI for a much smaller subset.

> Declaring structs in the namespace

I tried that, but the struct names are too generic. Computer, Fuel, etc. They'd collide with the actual Computer class, for example. As a compromise, I made them non-anonymous.

I just play tested this. It seems to be working fine. Also the unit tests are passing.

I realize the temptation to discuss getters vs direct access a few more days, but I'd really like to move on to something else. Unless you feel really passionate about it, how about we ~merge~ approve? Still need to figure out why build is failing on some machines.

royfalk:
Build is failing on bionic ubuntu because it is using cmake 3.10 and FetchContent was introduced in 3.11.
I will add an if condition to check for it, and disable tests on old systems.

ministerofinformation:
> Mind you, this is already an improvement over parsing a string and accessing a data structure in:
> 
> `XMLSupport::parse_bool( vs_config->getVariable( "section", "key", "default" ) );

As a quick (pedantic) aside, while I agree that we can worry about performance optimization later, the above comparison is inapt - the pre-refactoring code cached everything in statics, so the performance of code touching config vars was never I/O limited. So, while any of the things we've discussed are better than reading an XML file every time we want some constant, that's not what any of the implementations have ever done.

engine/CMakeLists.txt
```diff
@@ -1530,3 +1536,36 @@ ELSE ()
 ENDIF (WIN32 AND NOT UNIX)
 
 INCLUDE(CPack)
+
+IF (${CMAKE_MAJOR_VERSION}.${CMAKE_MINOR_VERSION} GREATER 2.10)
```
stephengtuggy, line 1540: This should be `3.10`, not `2.10`. And `VERSION_GREATER` for comparing versions, rather than simply `GREATER`.

Let's see if those two changes fix it.

stephengtuggy:
I like the idea of optimizing only as necessary. With today's hardware, I doubt we'll save more than a few microseconds here or there anyway. And we're spending a lot more time than that discussing this. 🙃 

If there really are on the order of 1000 settings -- and I suspect that number is as close as any we'll come up with -- then it doesn't seem to make sense to create getters for all of them.

I'm ready to approve this PR as soon as the unit test thing gets fixed and the builds are all passing.

engine/src/configuration/configuration.h
```diff
@@ -0,0 +1,95 @@
...
+
+
+struct Configuration
+{
+    // Config Structs Declaration
+
+    // These are settings that are relevant to collision.cpp and are marked as hacks
+    // or their comments are unclear or they are unclear
+    struct CollisionHacks {
+        const float collision_hack_distance;
```
BenjamenMeyer, line 12: I would suggest we put this in the `VegaStrike::Configuration` or `VegaStrikeEngine::Configuration` namespace, and then have all the structures in parallel there. It's cleaner overall and easier to verify the structures that way.

The actual global configuration object instance can be in a different namespace if desired.

BenjamenMeyer (changes requested):
One small modification request - though I won't be hard line on it if others disagree; I tend away from embedding definitions within other definitions.

Otherwise, LGTM

ministerofinformation:
> I like the idea of optimizing only as necessary. With today's hardware, I doubt we'll save more than a few microseconds here or there anyway. And we're spending a lot more time than that discussing this. 🙃

Well, depending on where in the loop nests the references to config vars are, performance impacts could still be tangible, and enforcing good usage by PR review vs. inherent code structure is a tradeoff. I don't think we should let performance concerns block us on a direction we are all in general agreement on, but I wouldn't be completely nonchalant about the _potential_ variance in implementation performance outcomes that may need eventual addressing.

> If there really are on the order of 1000 settings -- and I suspect that number is as close as any we'll come up with -- then it doesn't seem to make sense to create getters for all of them.

I don't think we need getters for every variable - a const ref getter to a higher level of the struct should be fine. Something like one getter per section/subsection should be sufficient to achieve the goal of the compiler helping us to prevent writes outside of the class.

engine/src/configuration/configuration.h
BenjamenMeyer: Resolving this in favor of continuing the discussion in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/484#pullrequestreview-700156858 as this is now referencing out-of-date code

engine/src/configuration/configuration.h
stephengtuggy: Personally, I think this is fine the way it is.

ministerofinformation: So, to separate the two comments:

1) From the documentation on gtest, I suspect we'd be well served to make a habit of putting things in namespaces, because it will make subsequent test code easier to write

2) I'd prefer to not have the other structs declared within the top level just to make the top level config class a bit easier to read, especially since it will be the only one that will eventually become a class with getter/setter methods, rather than a naked public struct. It's not a blocking issue for me though.

engine/src/configuration/configuration.h
```diff
@@ -0,0 +1,95 @@
...
+        const float default_lock_cone;
+        const float default_max_range;
+        const float default_tracking_cone;
+
+        Computer();
+    };
+
+    struct Fuel {
+        const float afterburner_fuel_usage;
+        const float fmec_exit_velocity_inverse; // Well, that's very clear
```
ministerofinformation, line 37: Re: Comment block - there are a pair of "FMEC" variables - they both involve "Fuel Mass to Energy Conversion" - this one happens to specify the inverse (it's only ever used as 1/Value, so just encode 1/Value, not Value) of the assumed exit velocity of the mass ejected as thrust, calculated based on energy-possible (if not necessarily plausible) outcomes from a Li-6 + Deuterium fusion reaction.

The other variable (not present here) FMEC_factor, is used in reactor --> energy production. As the comment in the code next to the variable init says, it specifies how many metric tons (1 VS mass unit) of fuel are used to produce 100MJ (one units.csv energy recharge unit) of recharge.

At some point, it wouldn't kill us to renormalize the engine and dataset to both just use SI units, but that's not a priority.

engine/src/configuration/configuration.h
```diff
@@ -0,0 +1,95 @@
...
+        const float fuel_efficiency;
```
ministerofinformation, line 38: There's some relevant context that's been removed from the original name of this variable "Lithium6constant" -- a better name would ultimately be "FuelToEnergyConversionRelativeToLithium6DeuterideFusion" - that fully encodes what the efficiency is relative to. This isn't a request to change anything, just writing this down while it's fresh(ish) in my head. I plan on doing a pass at some point to go through all the config vars adding comments, invariants, SI units, etc. and changing names if appropriate.

ministerofinformation: To clarify - Lithium6constant isn't a good name for what it is - but it's one that's easy for me to identify and subsequently fix. I'd prefer if we limited the renaming of thing prior to consolidation

engine/src/configuration/configuration.h
```diff
@@ -0,0 +1,95 @@
...
+        const float kilojoules_per_damage;
+        const float max_force_multiplier;
+        const bool max_shield_lowers_capacitance;
+        const float max_torque_multiplier;
+        const float minimum_mass;
+        const float minimum_time;             // between recorded player collisions
+
+
+        // Disabled bouncing missile option. Missiles always explode when colliding with something.
+        //static bool does_missile_bounce = GameConfig::GetVariable( "physics", "missile_bounce", false);
```
ministerofinformation, line 64: I'd like this eventually to be be enabled again - kinetic kill missiles impacting with small targets will shouldn't auto-explode, and I'd like to be able to model that for datasets that need it.

engine/src/configuration/configuration.h
stephengtuggy: 1. Oh. Yes, let's put this in a namespace. I didn't notice until now that this code was not in any namespace at all.
2. We can separate out the other structs if you guys insist. Perhaps that is better.

My primary goal here is just to get this PR approved, passing, and merged. It's been sitting here long enough, IMHO.

To that end, perhaps the CTest additions can be omitted for now? It might be better to figure them out in another PR.

BenjamenMeyer: separating out the testing into its own PR might be a good idea

engine/CMakeLists.txt
royfalk: Unfortunately, I tested a simple if(false) { jhfdskhdsklh } and cmake failed to parse. So this is not a preprocessor. You can't have stuff that's unknown to the cmake version anywhere in the file.

engine/src/configuration/configuration.h
royfalk: I'm going to resolve this discussion with the following:
1. I'll add a namespace and move the structs outside.
2. I'll keep the tests, but drop google test for now. 
3. Where the call is made from a nested loop and there are performance implications, it can be placed in a static local variable, which will only be called once. The person refactoring that code will make that decision.

engine/src/configuration/configuration.h
royfalk: I will add this info. Thanks.

engine/src/configuration/configuration.cpp
```diff
@@ -11,7 +11,7 @@ Configuration::Configuration()
     //logging.verbose_debug = GameConfig::GetVariable("data", "verbose_debug", false);
 }
 
-Configuration::CollisionHacks::CollisionHacks():
+vegastrike::configuration::CollisionHacks::CollisionHacks():
```
BenjamenMeyer, line 14: You should use a `using namespace vegastrike::configuration` at the top then this doesn't have to be here. There's a few places where that doesn't work but it's extremely rare.

Not a blocker.

engine/src/configuration/configuration.h
```diff
@@ -1,95 +1,114 @@
...
-    Physics physics;
-    Warp warp;
-    Weapons weapons;
+    vegastrike::configuration::CollisionHacks collision_hacks;
+    vegastrike::configuration::Computer computer;
+    vegastrike::configuration::Fuel fuel;
+    vegastrike::configuration::Logging logging;
+    vegastrike::configuration::Physics physics;
+    vegastrike::configuration::Warp warp;
+    vegastrike::configuration::Weapons weapons;
```
BenjamenMeyer, line 108: if you're within the vegastrike::configuration namespace you shouldn't need this.
Not a blocker.

BenjamenMeyer (approved):
The namespace functionality can be cleaned up a little, but not a blocker AS-IS.

BenjamenMeyer:
While I gave my approval, it looks like it still needs to pass the PR gates

engine/src/configuration/configuration.cpp
royfalk: Unfortunately, this is one of these rare cases. As I mentioned before, one of the structs in the namespace is Computer and we also have a Computer class outside.

engine/src/configuration/configuration.h
royfalk: I specifically placed the parent struct outside the namespace, so other parts of the code accessing it won't need vegastrike_configuration::Configuration or resort to a  _using_ directive.

stephengtuggy:
The Windows build still failed, with the following error:

```log
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\configuration\configuration.cpp(19,42): error C2065: 'M_PI': undeclared identifier [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine_com.vcxproj]
```

That looks fairly simple to fix.

PR #485, opened 2021-07-02
Draft - Introduce unit testing using google tests

royfalk:
This draft PR introduces unit tests based on gtests by google. 
If approved, I will begin introducing unit tests to the other ongoing PR's, primarily #484.

Code Changes:
- [ No ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- None

Purpose:
- What release is this for? 0.9

### DIFF
--- a/engine/src/tests/test.cpp
+++ b/engine/src/tests/test.cpp
@@ -1,0 +1,9 @@
+#include <gtest/gtest.h>
+
+// Demonstrate some basic assertions.
+TEST(HelloTest, BasicAssertions) {
+  // Expect two strings not to be equal.
+  EXPECT_STRNE("hello", "world");
+  // Expect equality.
+  EXPECT_EQ(7 * 6, 42);
+}
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -545,6 +545,7 @@ SET(VEGASTRIKE_SOURCES
     src/python/briefing_wrapper.cpp
 )
 
+
 IF (COMMAND cmake_policy)
     CMAKE_POLICY(SET CMP0003 NEW)
 ENDIF (COMMAND cmake_policy)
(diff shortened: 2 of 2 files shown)


engine/src/tests/test.cpp
```diff
@@ -0,0 +1,9 @@
+#include <gtest/gtest.h>
+
+// Demonstrate some basic assertions.
+TEST(HelloTest, BasicAssertions) {
+  // Expect two strings not to be equal.
+  EXPECT_STRNE("hello", "world");
+  // Expect equality.
+  EXPECT_EQ(7 * 6, 42);
+}
```
BenjamenMeyer, line 9: For comparison Boost::Test equivalent:

```C++
#define BOOST_TEST_DYN_LINK
#define BOOST_TEST_NO_MAIN

#include <boost/test/unit_test.hpp>

BOOST_AUTO_TEST_SUITE( GroupName );

BOOST_AUTO_TEST( IndividualTestName )
{
    // For whatever reason Boost doesn't do a not-equal directly
    BOOST_CHECK("hello" != "world");
    // but they do do equal directly
    BOOST_CHECK_EQUAL(7*6, 42);
}
```

BenjamenMeyer (commented):
Added an example of how it would be done using Boost::Test for comparison.
Haven't used Google's gtest before, so it's interesting to compare.

gtest looks a little simpler from some perspectives; though it'd be nice to see how it works with multiple C++ files and if there's anything special required for that.

ministerofinformation:
Chatting with some folks who've worked on Google product teams, they're pretty happy with the Google testing infrastructure. I got the impression that the feature set is deep, especially in  the "mock" portions, but that the learning curve is appropriately proportional to the complexity of the codebase and behaviors being tested (specifically, if your code is more template metaprogramming than actual C++, then you'll have more fun understanding the associated errors coming out of the framework, but you also brought that complexity on yourself...)

ministerofinformation:
whoops - sorry about that, mis-clicked while typing on my phone.

Loki1950:
So it really depends on how much work we put into our testing paradigm and it's integration with our code base.

stephengtuggy (commented):
I like this. gtest looks like a good option for unit testing.

BTW, the existing code does seem to have some unit tests of sorts, here and there. Nothing particularly organized though. It would be good to translate those into the unit testing framework that we end up settling on.

engine/src/tests/test.cpp
royfalk: Yeah. Like I said, at the basic level, they're all the same and I have looked at Boost.

royfalk:
@BenjamenMeyer I suggest we close this PR. 
#484 has cherry picked this commit and implemented some meaningful first unit tests.

stephengtuggy:
Closing, per @royfalk 's request.

Issue #498, opened 2021-07-10
Segfault when trying to load a game from main menu


stephengtuggy:
Very annoying, to be sure!

Issue #494, opened 2021-07-10
configuration.cpp: Undeclared identifier `M_PI` on Windows


stephengtuggy:
The Windows build still failed, with the following error:

```log
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\configuration\configuration.cpp(19,42): error C2065: 'M_PI': undeclared identifier [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine_com.vcxproj]
```

That looks fairly simple to fix.

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/484#issuecomment-877527376_

ministerofinformation:
https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-160

Issue #487, opened 2021-07-03
Static initialization ordering dependence for static std::map<string, string> GameConfig::variables


ministerofinformation:
std::map<string, string> variables in GameConfig.h is not guaranteed to be initialized before use in any call to 
static inline string _GetVariable(string const &section, string const &name) - such calls exist for lines such as 
Moveable.cpp:52 --
static Vector default_angular_velocity( GameConfig::GetVariable( "general", "pitch", 0.0f ),
                                        GameConfig::GetVariable( "general", "yaw", 0.0f ),
                                        GameConfig::GetVariable( "general", "roll", 0.0f));

Where GameConfig::GetVariable will call _GetVariable

This causes a crash on running due to null pointers in the std::map on (my) Windows build during the evaluation of the initializer for default_angular_velocity.

stephengtuggy:
Would something like this be a potential solution? https://stackoverflow.com/a/17642967

ministerofinformation:
Well, I suspect this particular issue will be fixed when the config library is finished; The example shown is actually broken even without static initialization ordering issues - it just isn't crash inducing (it will always get the default value because the config hasn't been loaded yet). While having an explicitly lazy initialization interface may be necessary in some cases,  we should avoid statics that depend on logically dynamic things wherever possible. I suspect that there will usually be another way to write most codes that isn't ordering sensitive, and we'll want to document explicitly any cases where it really needs to be done lazily.

BenjamenMeyer:
@ministerofinformation that's one reason why we shouldn't be using statics and objects should get their configuration data at creation, which should all happen after `main` loads the configuration.

ministerofinformation:
@BenjamenMeyer Yup. The old "standard" was always intra-method static caching in non-static methods, i.e. done on first call to the method, which would be after config load, but that left no way to guarantee that the defaults specified in two different files were set the same if it wasn't specified in the config file and no way to change the value consistently across scopes. So, centralization is just plain better, but intermediate points between the two approaches have undesirable properties :)

Issue #486, opened 2021-07-03
Uninitialized Boolean variable "resolveforces" in Moveable; Use without initialization.


ministerofinformation:
Boolean resolveforces is used without initialization. It has no default value and the setter is only called for planets, but all Moveables use it. Value should default to true.

I would change this with a 1-liner PR, but HEAD is currently non-runnable for me and I don't want to do PRs for code I haven't tested, no matter how trivial.

PR #501, opened 2021-07-10
Collision revamp and necessary bug fixes to enable playtesting of collision revamp

ministerofinformation:

Please answer the following:

Code Changes:

Issues:
#415
#502  
#498
#494 
#487 
#486 

Purpose:
- What is this pull request trying to do? 1) Fix errors in collision response code (velocity != force) 2) re-add torque to collision response after removal during refactoring 3) fix compile/crash/critical errors that prevented playtesting after merging in recent PRs.
- What release is this for?


ministerofinformation:
Known caveats:

I expect to make some changes on review - at the very least, the fix for the global config is ... a duct tape level workaround dropped in so I could test, commit, and go to sleep before it gets light out :-P. Let's figure out what we want to do to have it be clean - but I wanted to at least have a working Windows build option again.

Likewise, the #ifdef for M_PI -- if we want to be pushing the defines to CMAKE, that's fine too.

The collision code has some validation logic that should eventually be factored out to testing code, and it can be made leaner and cleaner.

The collision code changes were designed to be localized to collision.cpp/h - as such, there are some both fundamental and dataset issues that have not been resolved with respect to moment of inertia, as well as invariants that would be better enforced in other portions of the code (e.g. getMass() should always return a non-zero value, either by ensuring that minimum mass is specified at initialization or by hiding the std::max operation behind the getter, etc.) These represent some more systemic changes to enhance cleanliness, but they'll have to wait for another PR. I'll try to be sure to file issues on all of the things that came up during this feature redevelopment.

engine/src/main.cpp
```diff
@@ -384,7 +384,8 @@ int main( int argc, char *argv[] )
         //Specify the config file and the possible mod subdir to play
         VSFileSystem::InitPaths( CONFIGFILE, subdir );
     }
-
+    Configuration postConfigLoad;
+    memcpy(&configuration, &postConfigLoad, sizeof(configuration)); // should define an = operator as this, but, eh - just for testing.
```
stephengtuggy, line 388: No `memcpy`. Bad `memcpy`. Not for complex types like structs and classes, at any rate.

The = operator sounds like a better idea.

engine/src/configuration/configuration.cpp
```diff
@@ -1,6 +1,8 @@
 #include "configuration.h"
 #include "game_config.h"
-
+#ifndef _USE_MATH_DEFINES
+#define _USE_MATH_DEFINES
```
stephengtuggy, line 4: Nice

engine/src/cmd/collision.cpp
```diff
@@ -268,12 +262,78 @@ void Collision::collide( Unit* unit1,
...
+    QVector force_on_1 = -1 * impulse * (1.0f / (unit1->sim_atom_multiplier * SIMULATION_ATOM)); // divide impulse by time over which it will be applied to derive force -- allow different units to have different physics fidelity
+    QVector force_on_2 = impulse * (1.0f / (unit2->sim_atom_multiplier * SIMULATION_ATOM)); // divide impulse by time over which it will be applied to derive force 
+
+    collision1.adjustInterpenetration(v1_new, w1_new, normal1); // Call this BEFORE applying force
+    collision1.applyForce(force_on_1, location1_local); // handles both force-on-center-of-mass and torque
+    collision2.adjustInterpenetration(v2_new, w2_new, normal1); // uses normal1, not normal2
+    collision2.applyForce(force_on_2, location2_local);
+
+    collision1.dealDamage(collision2, delta_kinetic_energy_linear, delta_kinetic_energy_angular);
+    collision2.dealDamage(collision1, delta_kinetic_energy_linear, delta_kinetic_energy_angular);
```
stephengtuggy, line 338: Wow. My head is spinning right about now. Nice job, @ministerofinformation

stephengtuggy (changes requested):
I would like to request that the `memcpy` be replaced with something else before we merge this. Otherwise, looks good!

engine/src/main.cpp
ministerofinformation: Yup. Quick and dirty 3am code is ... exactly-as-awful-as-expected 3am code. I don't actually _want_ this accepted before I fix this, but I also don't want to delay the rest of this PR being reviewed/tested any further. I expect to be on baby duty for much of the day, so I'll try to get to this to tonight/tomorrow. This is known-flawed-and-dangerous for non-POD structs, but it enables testing of the config functionality, etc. 

Re = operator -- that will work for now and be generally potentially useful indefinitely. Longer term, I think the discussion was moving towards upgrading Configuration to have methods, while keeping its members as simple structs, and (re)init/setters would then be methods in Configuration. If we want it to initialize at construction, Configuration can only init to defaults, and we should be explicit in documenting this in the code and using obvious patterns (i.e. If construction is default, just use the default, not the call that looks like it's getting a value, and to keep default specification singular, the XML reader function used to init should then not have a default field at all)

engine/src/cmd/collision.cpp
ministerofinformation: If your head is spinning, are there more comments I could add to make it more self-contained without dumping the whole Wikipedia article contents in (the figure is actually quite useful for building an intuition of what's going on)? I'd like this to be as clearly documented as possible, so as to separate the math of the collision from future renovations on the moment of inertia implementation and various getters/non-local invariants we want to adjust/enforce (applying Newtonian physics to zero-mass objects tends to end poorly in more than the collision code, etc.)

stephengtuggy: No, it's fine, thanks. Just a lot of code to read through. :-)

Loki1950: Just standard vector arithmetic @stephengtuggy

stephengtuggy:
I ran a bunch more of the play tests today. Looking good so far. Ships actually bounce and spin away now 💯

ministerofinformation:
> I ran a bunch more of the play tests today. Looking good so far. Ships actually bounce and spin away now 💯

If you have the time, I'd also recommend testing with some different values for the elasticity constant "inelastic_scale" (for historical reasons, 1-elasticity) in the physics section of the config xml to make sure you're seeing the expected differences, to test the #502 aspect of the PR.

ministerofinformation:
After sleeping on it, I went with adding a new member function versus putting in an = operator. I think this more clearly embeds the pattern of what we want to happen at each point in time: A specific default value at construction time (everything is initialized to some valid value), followed by an adjustment to said values after the user configuration file has been loaded (existing value used as default in the GameConfig::GetVariable function, so that there aren't two copies of the default value in play). We may want to consider cleaning this up further by writing a companion to GameConfig::GetVariable, say, GameConfig::OverrideVariableIfSpecified, or somesuch, that takes a mutable reference to the existing value and only modifies it if there's an overriding value. The current pattern is clear enough, and flexible enough, but a little more verbose.

stephengtuggy:
@ministerofinformation I like the latest change a lot. Makes things much clearer. I will plan on doing some more playtesting here shortly.

engine/src/cmd/collision.cpp
```diff
@@ -156,96 +141,104 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
...
-    float& m1 = mass;
-    float& m2 = other_collision.mass;
-    Vector& v1 = velocity;
-    Vector& v2 = other_collision.velocity;
-
-    float relative_velocity = (v1-v2).Magnitude();
-    float damage = relative_velocity * relative_velocity *
-            (m2)/(m1+m2) * factor;
-
+    float damage = 0.5 * ( deltaKE_linear + deltaKE_angular) /* deltaKE is in KiloJoules, due to mass being in units of 1000 Kg, not Kg - so convert accordingly */ / configuration.physics.kilojoules_per_damage; // assign half the change in energy to this unit, convert from KJ to VSD
```
BenjamenMeyer, line 152: The `0.5` should be a constant declared somewhere; magic values shouldn't appear as anything else.

engine/src/cmd/collision.cpp
```diff
@@ -156,96 +141,104 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
...
     {
-        BOOST_LOG_TRIVIAL(warning) << "Collision::applyForce Expected an elasticity value between 0 and 1 but got " << elasticity;
-        elasticity = std::max(elasticity, 0.0);
-        elasticity = std::min(elasticity, 1.0);
+        return;
     }
+    
+    double magnitude = new_velocity.Magnitude();
+    QVector velocity_norm = new_velocity.Normalize(); // want to perform positional displacement in direction of new velocity vector
+    double movementKludgeMagnitude = 2 * (new_angular_velocity.Dot(velocity_norm) * unit->radial_size / PI + magnitude);
```
BenjamenMeyer, line 175: same with the `2`

engine/src/cmd/collision.cpp
```diff
@@ -156,96 +141,104 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
...
+        force *= (max_force / force_magnitude);
+    }
+    if (torque_magnitude > max_torque) {
+        torque *= (max_torque / torque_magnitude);
+    }
+    
+    */
+
+    // Apply force at location, should just be one call to ApplyLocalTorque.... but there's a catch - our current moment of inertia units(as in SI units, not unit class) are a bit b0rked, so we need to do a workaround :-(
+    Vector torqueForce = force * (1 / (0.667 * unit->radial_size * unit->radial_size)); // use shell approximation
```
BenjamenMeyer, line 212: same for the 0.667

BenjamenMeyer: possibly the `1` as well

engine/src/cmd/collision.cpp
```diff
@@ -156,96 +141,104 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
...
+    const QVector& v1_new,
+    const QVector& v2_new,
+    const QVector& w1_new,
+    const QVector& w2_new) {
+    //following two values should be identical. Due to FP math, should be nearly identical. Checks for both absolute and relative error, the former to shield slightly larger errors on very small values - can set tighter/looser bounds later
+    double RestorativeVelAlongNormal = -1 * (1 - configuration.physics.inelastic_scale) * relative_velocity.Dot(normal1);
+    double ResultantVelAlongNormal = ((v2_new + w2_new.Cross(location2_local)) - (v1_new + w1_new.Cross(location1_local))).Dot(normal1);
+    double normalizedError = abs(1.0 - ResultantVelAlongNormal / RestorativeVelAlongNormal);
+    double absoluteError = abs(ResultantVelAlongNormal - RestorativeVelAlongNormal);
+    if (absoluteError > 0.001 && normalizedError > 0.01) { // absolute error > 1 milimeter/second along collision normal AND normalized error > 1%
```
BenjamenMeyer, line 235: would be nice if the fuzzy compare was abstracted to its own inline function for re-use

BenjamenMeyer (approved):
would be nice to move the `magic numbers` to constants so we can understand their meaning better; but as there's a lot of them in the code right now I'm not going to block.
Also would be nice to abstract the fuzzy float/double comparison, but again, not blocking.

engine/src/cmd/collision.cpp
ministerofinformation: The 1 is just an inversion - I'd leave the 1 as is.

engine/src/cmd/collision.cpp
ministerofinformation: Agreed - albeit in this specific function everything is currently magic, so it'll just be moved to a local var just to not be nameless, rather than something more configurable.

engine/src/cmd/collision.cpp
ministerofinformation: > same for the 0.667

Agreed -  the 0.667, as well as the need to access radial_size is a bandaid over a longer-term fix; it will eventually go away when we renovate the moment of inertia to not be scalar in the first place and also fix the dataset accordingly to specify the proper 3x3 matrices.

engine/src/cmd/collision.cpp
ministerofinformation: Probably a good idea to do so later, make it an inline function with both a percent and absolute bound - where in the codebase do you think is the best place to have testing/logging/etc. adjacent functions specified?

engine/src/cmd/collision.cpp
ministerofinformation: Agreed - should factor this one out into a something representing the fraction of deltaKE assigned to this unit vs. the other unit.

engine/src/cmd/collision.cpp
BenjamenMeyer: Not sure; right now, it could just site in a function above this one.

engine/src/cmd/collision.cpp
BenjamenMeyer: doesn't have to be configurable - the point of not having magic numbers is that there's no meaning in the number; using a variable (configurable) or a constant (non-configurable) value instead conveys the meaning of the value.
For now, even a simple constant right before usage is good. If we find a lot of the same, then it should be moved to a file header scope as appropriate.

stephengtuggy (approved):
Looks good to me! Sorry, I didn't realize I was holding up this PR.

Can we go ahead and merge this? I think it's urgent that we get master fixed.

stephengtuggy:
I have finished play testing this code to my satisfaction.

PR #505, opened 2021-08-01
Draft: lib_damage

royfalk:
This is an implementation of a stand-alone, damage library and initial integration into the game. 
This is my first real attempt at such an ambitious refactoring. It's also the first one where the work is really half ready - there are lots of bugs in many parts of the code I'm unfamiliar with. 
I've started this as I've noticed a lot of repeated code in the area of game damage and decided to "do it right" once.

Obviously, this is not for any near release, as it requires more work, and not just by me.

**Issues**
Obviously, even minimal integration would break many things and this isn't minimal.
Some things I've noticed:

1. Ship images (sprite?) is missing. This includes ship in dock and ships in VDU.
2. Weapons are too powerful - one shot kills compared to before.
3. I can't die or even receive damage. Not sure what's going on here and if this is true or the VDU is not registering damage.
4. Ship deaths are different from before. Shorter maybe?


royfalk:
A brief into into the lib_damage parts:

**CoreVector** - a struct that holds an x,y,z float values. We don't use Vector here because that would expose too much of the main engine for no reason. All we need are these three floats.

**Damage** - a struct that holds multiple types of damage and how much of it has already been expanded. This replaces a whole bunch of hacks, primarily negative for non-lethal damage. As this structs goes through shield, armor and hull, damage is reduced from it and transferred to them. Currently, it is also used for vulnerability. What layer is vulnerable to what damage. For example, phase damage does not damage the shield and so keeps on going deeper.

**DamageableObject** - something that can be damaged and be destroyed. Holds multiple DamageableLayers in a vector.

**DamageableLayer** - a layer within an object. Typically shield/armor/hull. Holds multiple DamageableFacets in a vector. The layer also evaluates which facet receives the damage.

**DamageableFacet** - if layer is shield, a facet is front, rear, etc. Holds a health object.

**Health** - holds the maximum health, current health etc.

BenjamenMeyer:
I'll take a look later, but awesome. If like to see us split more stuff up this way and create interfaces between things while doing so.

Underlying implementation should be able to change and only require a re-link after compiling just that one library once all is said and done. Obviously it may take a while to get to that point, and might not be feasible with the first few breakouts. Just setting a long term goal.

Glad to see this. ;)🎉🎉🎉

ministerofinformation:
re: The VDU bug,  the target was previously color-shifted red as a function of how damaged the hull was; there may be some blending issues to resolve.

engine/src/cmd/unit_generic.cpp
```diff
@@ -5288,10 +5252,10 @@ float Unit::DealDamageToHull( const Vector &pnt, float damage)
       if ( (!isplayer) || _Universe->AccessCockpit()->godliness <= 0 ) { // || system_damage_on_armor ) {
           static float system_failure =
               XMLSupport::parse_float( vs_config->getVariable( "physics", "indiscriminate_system_destruction", ".25" ) );
-          if ( (!isplayer) && DestroySystem( hull, maxhull, 1 ) )
-            DamageRandSys( system_failure*rand01()+(1-system_failure)*( 1-(hull > 0 ? damage/hull : 1.0f) ), pnt );
-          else if ( isplayer && DestroyPlayerSystem( hull, maxhull, 1 ) )
-            DamageRandSys( system_failure*rand01()+(1-system_failure)*( 1-(hull > 0 ? damage/hull : 1.0f) ), pnt );
+          if ( (!isplayer) && DestroySystem( health.health, health.max_health, 1 ) )
+            DamageRandSys( system_failure*rand01()+(1-system_failure)*( 1-(health.health > 0 ? damage/health.health : 1.0f) ), pnt );
+          else if ( isplayer && DestroyPlayerSystem( health.health, health.max_health, 1 ) )
```
stephengtuggy, line 5257: What is `health.health`? This nomenclature is a bit difficult to understand

engine/src/damage/core_vector.h
```diff
@@ -0,0 +1,11 @@
+#ifndef COREVECTOR_H
+#define COREVECTOR_H
+
+struct CoreVector {
```
stephengtuggy, line 4: I think the Google Code Style would say that all these new structs, classes, etc. should go in one or more namespaces. E.g., `namespace VSDamage` or `namespace VegaStrikeDamage`

engine/src/main.cpp
```diff
@@ -323,7 +323,7 @@ int main( int argc, char *argv[] )
 
     // when the program name is `vegastrike-engine` then enforce that the data directory must be specified
     // if the program name is `vegastrike` then enable legacy mode where the current path is assumed.
-    legacy_data_dir_mode = (program_name == "vegastrike") || (program_name == "vegastrike.exe");
+    legacy_data_dir_mode = true; //(program_name == "vegastrike") || (program_name == "vegastrike.exe");
```
stephengtuggy, line 326: Whoops, looks like this snuck in again.

stephengtuggy (commented):
A few comments. Overall, I am very excited to see this kind of refactor happening. Nice work, @royfalk !

engine/src/cmd/collision.cpp
```diff
@@ -150,18 +150,9 @@ void Collision::dealDamage(Collision other_collision, double deltaKE_linear, dou
...
-    Vector& v1 = velocity;
-    Vector& v2 = other_collision.velocity;
-
-    float relative_velocity = (v1-v2).Magnitude();
-
     Damage damage;
-    damage.normal_damage = relative_velocity * relative_velocity *
-            (m2)/(m1+m2) * factor;
-
+    damage.normal_damage = 0.5 * ( deltaKE_linear + deltaKE_angular) /* deltaKE is in KiloJoules, due to mass being in units of 1000 Kg, not Kg - so convert accordingly */ / configuration.physics.kilojoules_per_damage; // assign half the change in energy to this unit, convert from KJ to VSD
```
BenjamenMeyer, line 154: Would prefer the comment to not be in-line with the equation:
```
// deltaKE = in KiloJoules due to mass being in units of 1,000 Kg, not simply Kg; convert accordingly
// and then assign half the change in energy to convert from KJ to VSD
damage.normal_damage = 0.5 * ( deltaKE_linear + deltaKE_angular) / configuration.physics.kilojoules_per_damage;
```

It's cleaner, easier to read and more  useful.

engine/src/damage/core_vector.h
BenjamenMeyer: I'd vote for VegaStrikeDamage.

engine/src/cmd/unit_generic.cpp
```diff
@@ -5147,13 +5145,11 @@ void Unit::UpdatePhysics3(const Transformation &trans,
       }
   }
 
-  // TODO: this should never happen
-  if (health.max_health <= 0) {
+  // TODO: lib_damage
+  // I merged two conditions for explode and kill here
```
stephengtuggy, line 5149: Hmm. I hope that is correct

royfalk:
Pushed another update - damage is mostly handled correctly.

**Current issues:**
* Ship in dock is not seen - not sure if this is related to current.
* Ships don't get angry quickly enough. Enemy ships don't engage me.
* Damage doesn't seem to affect player ship.
* Planets show flickering shields in VDU (right bottom screen).

engine/src/cmd/basecomputer.cpp
```diff
@@ -5250,53 +5250,53 @@ void showUnitStats( Unit *playerUnit, string &text, int subunitlevel, int mode,
             break;
         }
     }
-    if (!mode || MODIFIES(replacement_mode, playerUnit, blankUnit, armor.facets[2].health.health)) {
+    if (!mode || MODIFIES(replacement_mode, playerUnit, blankUnit, layers[1].facets[2].health.health)) {
```
BenjamenMeyer, line 5255: `armor` was more descriptive; though I like the array. The descriptiveness can be continued if you use a constant or enum for the index instead. Same for the shields below.

engine/src/cmd/damageable.h
```diff
@@ -57,10 +49,41 @@ class Damageable
   Damageable& operator=( const Damageable& ) = delete;
 
 public:
-  float GetHull() const {
-      return health.health;
+  // We follow the existing convention of GetX for the actual health value
+  // because we are constrained by existing python interfaces, which cannot
+  // be easily changed.
+  const float GetHull() const {
+      return layers[0].facets[0].health.health;
```
BenjamenMeyer, line 59: same here - the constants/enum should bescoped to the same as `layers`.

engine/src/cmd/unit_generic.cpp
BenjamenMeyer: should be more or less, it's just a matter of perspective.

BenjamenMeyer (changes requested):

engine/src/damage/damageable_object.cpp
```diff
@@ -28,18 +30,27 @@ DamageableObject::DamageableObject(std::vector<DamageableLayer> layers,
...
-void DamageableObject::DealDamage( const CoreVector &attack_vector, Damage &damage ) {
-    for(DamageableLayer layer : layers) {
-        layer.DealDamage(attack_vector, damage);
+InflictedDamage DamageableObject::DealDamage( const CoreVector &attack_vector, Damage &damage ) {
+    InflictedDamage inflicted_damage(3); // Currently hard-coded default is 3!
 
-        if(layer.core_layer) {
+    // Higher index layers are outer layers. We therefore need to reverse the order.
+    for (DamageableLayer& layer : boost::adaptors::reverse(layers)) {
+        std::cout << "Damaging layer " << layer.layer_index << "\n";
```
stephengtuggy, line 38: Can this be changed to use the Boost logging please?

engine/src/cmd/damageable.cpp
```diff
@@ -267,11 +333,155 @@ void Damageable::ApplyDamage( const Vector &pnt,
...
+
+    static std::string restricted_items = vs_config->getVariable( "physics", "indestructable_cargo_items", "" );
+    int cargo_to_damage_index = rand() % unit->numCargo();
+    Cargo& cargo = unit->GetCargo( cargo_to_damage_index );
+    const std::string& cargo_category = cargo.GetCategory();
+
+    bool is_upgrade = cargo_category.find( "upgrades/" ) == 0;
+    bool already_damaged = cargo_category.find( "upgrades/Damaged/" ) == 0;
+    bool is_multiple = cargo_category.find( "mult_" ) == 0;
+    bool is_restricted = restricted_items.find(cargo.GetContent()) == string::npos;
```
stephengtuggy, line 452: Hmm. It almost seems like the logic for each of these lines is reversed. But I could be wrong.

engine/src/cmd/damageable.h
```diff
@@ -45,106 +44,107 @@ class Damageable : public DamageableObject
...
-  bool ShieldUp( const Vector& ) const;
+    // Is the shield up from this position
+    bool ShieldUp( const Vector& ) const;
 
-  //applies damage from the given pnt to the shield, and returns % damage applied and applies lighitn
-  virtual float DealDamageToShield( const Vector &pnt, float &Damage );
-  virtual float DealDamageToHull( const Vector &pnt, float Damage );
+    // Currently available in case someone wants to call from python
+    // but there isn't an interface
+    float DealDamageToShield( const Vector &pnt, float &Damage );
```
stephengtuggy, line 104: Is it safe to make this no longer virtual, or will that break backwards compatibility or anything?

engine/src/cmd/unit.cpp
```diff
@@ -220,7 +220,7 @@ void GameUnit::DrawNow( const Matrix &mato, float lod )
     if (rootunit == NULL) rootunit = (const void*) this;
     float damagelevel = 1.0;
     unsigned char chardamage    = 0;
-    if (this->layers[0].facets[0].health.health < this->layers[0].facets[0].health.max_health) {
+    if (this->GetHull() < this->layers[0].facets[0].health.max_health) {
```
stephengtuggy, line 223: Thank you. Much easier to understand.

stephengtuggy:
Overall, looks like this is coming right along. Thanks, @royfalk

engine/src/cmd/damageable.cpp
```diff
@@ -45,242 +53,484 @@ static float  flickeronprob = 0.66f; // ( "graphics", "glowflicker", "num-times-
...
+    Damageable *damageable = const_cast<Damageable*>(this);
+    DamageableLayer armor_layer = damageable->GetArmorLayer();
+    armor[0] = armor_layer.facets[0].health;
+    armor[1] = armor_layer.facets[1].health;
+    armor[2] = armor_layer.facets[2].health;
+    armor[3] = armor_layer.facets[3].health;
+    armor[4] = armor_layer.facets[4].health;
+    armor[5] = armor_layer.facets[5].health;
+    armor[6] = armor_layer.facets[6].health;
+    armor[7] = armor_layer.facets[7].health;
```
BenjamenMeyer, line 525: might a for-loop do better here:
```
for int i = 0; i < 8; i++ {
    armor[i] = armor_layer.facets[i].health;
}
```
the compiler might unroll it to what you have, but it's just as effective and clear.

engine/src/damage/damageable_object.cpp
BenjamenMeyer: yes, we should have any calls to `std::cout`  or `std::cerr`.

engine/src/cmd/collision.cpp
royfalk: Not actually my commit, but fixed anyway.

engine/src/cmd/basecomputer.cpp
royfalk: Fixed by adding shorthand pointers hull, armor and shield to DamageableLayer.

engine/src/cmd/damageable.cpp
royfalk: I was sure you were right and then I found the original code and we're both wrong :). The find function returns the index of the substring. It's only is_upgrade for example, if the string starts with "upgrades/"

engine/src/cmd/damageable.h
royfalk: Excellent question. DealDamageToHull is no longer implemented in a subclass. Therefore, I don't think we need a virtual to specify which function is called. 

Please resolve issue if you think this is correct or continue on gitter if not.

engine/src/cmd/damageable.cpp
royfalk: You are absolutely right. I suspect I originally needed to implement 

`armor[i] = armor_layer.facets[j].health;

where i != j and fixed it along the way. Honestly, I no longer remember.

engine/src/damage/facet_configuration.cpp
```diff
@@ -0,0 +1,29 @@
...
+    if(configuration ==  FacetConfiguration::eight) { return 8; }
+}
+
+const FacetName* FacetConfigurationByName(FacetConfiguration configuration) {
+    if(configuration ==  FacetConfiguration::one) { return one_configuration; }
+    if(configuration ==  FacetConfiguration::two) { return two_configuration; }
+    if(configuration ==  FacetConfiguration::four) { return four_configuration; }
+    if(configuration ==  FacetConfiguration::six) { return six_configuration; }
+    if(configuration ==  FacetConfiguration::eight) { return eight_configuration; }
+}
```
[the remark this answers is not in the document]
royfalk: The compiler is wrong here. With an enum class, this satisfies all paths. I guess it simply cannot understand it. Fixed it with an extra line to silence it.

engine/src/cmd/damageable.cpp
```diff
@@ -479,10 +477,25 @@ void Damageable::DamageCargo(InflictedDamage inflicted_damage) {
     }
 }
 
+// TODO: get rid of extern
+extern void DestroyMount( Mount* );
```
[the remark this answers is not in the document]
royfalk: Ordinarily I'd agree with you. However, there are numerous sections of the code where this isn't declared in a header file. I prefer not to focus on stuff like this but just comment. 
After I'm done with most or all of the unit code refactoring, I'll do a quick TODO hunting for whatever is left.
This TODO hunting might also be a nice starter task for a junior programmer.

royfalk:
> The Windows build is failing with a series of linker errors, of the form ...
> 
> ```
> vegastrike-engine_com.lib(unit_functions_generic.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(...)" 

The next fix will make the parameters names match. This should be a warning but maybe MSVC is configured more strictly. If not, I'm drawing a blank.

stephengtuggy (commented):
Looks like it's coming along well. I just had one question on the latest set of changes.

stephengtuggy:
> > The Windows build is failing with a series of linker errors, of the form ...
> > ```
> > vegastrike-engine_com.lib(unit_functions_generic.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(...)" 
> > ```
> 
> The next fix will make the parameters names match. This should be a warning but maybe MSVC is configured more strictly. If not, I'm drawing a blank.

Scrolling up a ways in the latest Windows build log, I see the following:

```log
  vegastrike-engine_com.lib(unit_functions_generic.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  vegastrike-engine_com.lib(beam_generic.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  vegastrike-engine_com.lib(unit_wrapper.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  vegastrike-engine_com.lib(collision.obj) : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  bolt.obj : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  cont_terrain.obj : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  missile.obj : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  terrain.obj : error LNK2001: unresolved external symbol "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,struct GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBUGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  vegastrike-engine_com.lib(damageable.obj) : error LNK2019: unresolved external symbol "public: void __cdecl Drawable::LightShields(class TVector<float,double> const &,class TVector<float,double> const &,float,struct GFXColor const &)" (?LightShields@Drawable@@QEAAXAEBV?$TVector@MN@@0MAEBUGFXColor@@@Z) referenced in function "public: void __cdecl Damageable::ApplyDamage(class TVector<float,double> const &,class TVector<float,double> const &,struct Damage,class Unit *,class GFXColor const &,void *)" (?ApplyDamage@Damageable@@QEAAXAEBV?$TVector@MN@@0UDamage@@PEAVUnit@@AEBVGFXColor@@PEAX@Z) [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\Release\vegastrike-engine.exe : fatal error LNK1120: 2 unresolved externals [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
```

So now we have two unresolved symbols: `ApplyDamage` and `LightShields`.

engine/src/cmd/unit_csv.cpp
```diff
@@ -1140,6 +1140,14 @@ void Unit::LoadRow( CSVRow &row, string modification, string *netxml )
 
     // End shield section
 
+    // Assign shorthand pointers;
+    hull = &layers[2];
+    armor = &layers[2];
+    shield = &layers[2];
```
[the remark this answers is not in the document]
royfalk: Of course I am, because I'm a tired idiot.

engine/src/cmd/basecomputer.cpp
```diff
@@ -5234,7 +5233,7 @@ void showUnitStats( Unit *playerUnit, string &text, int subunitlevel, int mode,
         text += "#n##n##c0:1:.5#"+prefix+"[DURABILITY STATISTICS]#n##-c";
         text += "#n#"+prefix+statcolor+"Armor damage resistance:#-c";
     }
-    if (mode && MODIFIES(replacement_mode, playerUnit, blankUnit, armor.frontlefttop)) {
+    if (mode && MODIFIES(replacement_mode, playerUnit, blankUnit, armor->facets[0].health)) {
```
BenjamenMeyer, line 5236: what does `facets[0]` represent? Using an enum instead of a raw integer would be better at conveying that information

engine/src/cmd/damageable.h
```diff
@@ -26,78 +26,142 @@
...
+    DamageableLayer *armor;
+    DamageableLayer *shield;
+
+    float *current_hull;
+    float *max_hull;
+
 
+    bool dying;
+    //Is dead already?
+    bool  killed;
```
BenjamenMeyer, line 51: would an enum be better than 2 booleans?
```
enum LifeStatus { ALIVE, DYING, DEAD };
state LifeStatus;
```

engine/src/cmd/damageable.h
```diff
@@ -26,78 +26,142 @@
...
-  bool withinShield( const ShieldFacing &facing, float theta, float rho );
-
-  static float totalShieldVal( const Shield &shield );
-  static float currentTotalShieldVal( const Shield &shield );
-  static float totalShieldEnergyCapacitance( const Shield &shield );
+    // We follow the existing convention of GetX for the actual health value
+    // because we are constrained by existing python interfaces, which cannot
+    // be easily changed.
+    const float GetHull() const {
+        return hull->facets[0].health;
```
BenjamenMeyer, line 75: what is `faces[0]`? Using an enum instead of the raw integer would better convey that.

engine/src/cmd/damageable.h
```diff
@@ -26,78 +26,142 @@
...
+    }
+
+    const float GetArmor(int facet = 0) const {
+        return armor->facets[facet].health;
```
BenjamenMeyer, line 79: what are the valid values of `facets`? Using an enum would convey that and help limit to valid values.

engine/src/cmd/damageable.h
```diff
@@ -26,78 +26,142 @@
...
+    }
+
+    const float GetShield(int facet = 0) const {
+        return shield->facets[facet].health;
+    }
+
+    DamageableLayer& GetHullLayer() {
+        return layers[0];
```
BenjamenMeyer, line 87: Awesome that it's wrapped in the function; though using an enum instead of a raw integer would still help convey the meaning of that layer.

The function should probably be declared as `inline` too.

engine/src/cmd/damageable.h
```diff
@@ -26,78 +26,142 @@
...
+
+    const float GetHullPercent() const
+    {
+        return hull->facets[0].Percent();
+    }
+
+    const float GetShieldPercent() const
+    {
+        return shield->facets[0].Percent();
+    }
```
BenjamenMeyer, line 110: The above functions should probably be declared inline.

engine/src/cmd/damageable.h
BenjamenMeyer: Saw the gitter discussion (@royfalk  thanks!) and pretty sure it does't need to be virtual.

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
 }
+
+
+void Drawable::Split( int level )
+{
+    Unit *unit = static_cast<Unit*>(this);
+
+    if (game_options.split_dead_subunits)
+        for (un_iter su = unit->getSubUnits(); *su; ++su)
+            (*su)->Split( level );
```
BenjamenMeyer, line 278: braces at all the code-block levels please

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
+    Vector PlaneNorm;
+    for (unsigned int i = 0; i < nummesh();) {
+        if (this->meshdata[i]) {
+            if (this->meshdata[i]->getBlendDst() == ONE) {
+                delete this->meshdata[i];
+                this->meshdata.erase( this->meshdata.begin()+i );
+            } else {i++; }} else {this->meshdata.erase( this->meshdata.begin()+i ); }}
```
BenjamenMeyer, line 285: we don't need to shorten the code block; please expand them out for easier comprehension

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
+    int    nm = this->nummesh();
+    string fac = FactionUtil::GetFaction( unit->faction );
+
+    CSVRow unit_stats( LookupUnitRow( unit->name, fac ) );
+    unsigned int num_chunks = unit_stats.success() ? atoi( unit_stats["Num_Chunks"].c_str() ) : 0;
+    if (nm <= 0 && num_chunks == 0)
+        return;
```
BenjamenMeyer, line 292: braces please

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
+                old[i]->Fork( nw[nw.size()-2], nw.back(), PlaneNorm.i, PlaneNorm.j, PlaneNorm.k,
+                             -PlaneNorm.Dot( old[i]->Position() ) );                                                                              //splits somehow right down the middle.
+                delete old[i];
+                old[i] = NULL;
+                if (nw[nw.size()-2] == NULL) {
+                    nw[nw.size()-2] = nw.back();
+                    nw.pop_back();
+                }
+                if (nw.back() == NULL)
+                    nw.pop_back();
```
BenjamenMeyer, line 339: braces please

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
+        string scalestr     = UniverseUtil::LookupUnitStat( unit->name, fac, "Unit_Scale" );
+        int   scale = atoi( scalestr.c_str() );
+        if (scale == 0) scale = 1;
+        AddMeshes( nw, randomstartframe, randomstartseconds, scale, chunkname, unit->faction,
+                   unit->getFlightgroup(), &meshsizes );
+        VSFileSystem::current_type.pop_back();
+        VSFileSystem::current_subdirectory.pop_back();
+        VSFileSystem::current_path.pop_back();
+        for (i = 0; i < old.size(); ++i)
+            delete old[i];
```
BenjamenMeyer, line 318: braces please

engine/src/cmd/drawable.cpp
```diff
@@ -262,3 +267,122 @@ bool Drawable::animationRuns() const
...
+        string  chunkname   = UniverseUtil::LookupUnitStat( unit->name, fac, "Chunk_"+XMLSupport::tostring( which_chunk ) );
+        string  dir = UniverseUtil::LookupUnitStat( unit->name, fac, "Directory" );
+        VSFileSystem::current_path.push_back( unit_stats.getRoot() );
+        VSFileSystem::current_subdirectory.push_back( "/"+dir );
+        VSFileSystem::current_type.push_back( UnitFile );
+        float randomstartframe   = 0;
+        float randomstartseconds = 0;
+        string scalestr     = UniverseUtil::LookupUnitStat( unit->name, fac, "Unit_Scale" );
+        int   scale = atoi( scalestr.c_str() );
+        if (scale == 0) scale = 1;
```
BenjamenMeyer, line 311: braces please

engine/src/damage/damageable_layer.cpp
```diff
@@ -0,0 +1,250 @@
...
+        return 0;
+    } else if(configuration == FacetConfiguration::two) {
+        return Find(facet_name, two_configuration, 2);
+    } else if(configuration == FacetConfiguration::four) {
+        return Find(facet_name, four_configuration, 4);
+    } else if(configuration == FacetConfiguration::six) {
+        return Find(facet_name, six_configuration, 6);
+    } else if(configuration == FacetConfiguration::eight) {
+        return Find(facet_name, eight_configuration, 8);
+    }
```
BenjamenMeyer, line 106: A `switch` block would probably be better:
```C++
switch (configuration) {
    case FaceConfiguration::one:
        return 0;
        break;
    case FaceConfiguration::two:
        return Find(face_name, two_configuration, 2);
        break;
    ...
}
```
Alternatively using a map for the second two parameters would be just as good, possibly even better.
```C++
// map version would be something like:
if (configuration == FacetConfiguration::one) {
    return 0;
} else {
   // needs a check to make sure that the configuration is in the map
   // parameters is a pair with the configuration and integer value
   parameters = parameterMap[configuration];
   return Find(face_name, parameters.first, parameters.second);
}
```
Though there may be a better way to map the values out too.
Both would yield higher performance than the multiple IF-blocks.

engine/src/damage/damageable_layer.h
```diff
@@ -0,0 +1,66 @@
...
+#include "health.h"
+
+/**
+ * @brief The DamageableLayer class represents an object to be damaged.
+ * This can be shields, armor, hull or subsystem.
+ * A layer can be regenerative (e.g. shields) or not (pretty much everything else).
+ */
+struct DamageableLayer
+{
+    int layer_index;
```
BenjamenMeyer, line 17: it would be better if this was an enum

engine/src/damage/facet_configuration.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+    return -1;
+}
+
+int FacetConfigurationSize(FacetConfiguration configuration) {
+    if(configuration ==  FacetConfiguration::one) { return 1; }
+    if(configuration ==  FacetConfiguration::two) { return 2; }
+    if(configuration ==  FacetConfiguration::four) { return 4; }
+    if(configuration ==  FacetConfiguration::six) { return 6; }
+    if(configuration ==  FacetConfiguration::eight) { return 8; }
```
BenjamenMeyer, line 20: this would be better as a switch

engine/src/damage/facet_configuration.cpp
```diff
@@ -0,0 +1,31 @@
...
+    return 1; // Literally cannot happen
+}
+
+const FacetName* FacetConfigurationByName(FacetConfiguration configuration) {
+    if(configuration ==  FacetConfiguration::one) { return one_configuration; }
+    if(configuration ==  FacetConfiguration::two) { return two_configuration; }
+    if(configuration ==  FacetConfiguration::four) { return four_configuration; }
+    if(configuration ==  FacetConfiguration::six) { return six_configuration; }
+    if(configuration ==  FacetConfiguration::eight) { return eight_configuration; }
```
BenjamenMeyer, line 29: this would be better as a switch, which would also help avoid the warnings that @stephengtuggy mentioned

BenjamenMeyer (changes requested):
there's a few code improvements still needed:
- braces for blocks in various points
- use of a switch instead of multiple If-blocks in several places
- use of enum instead of integer values for a few of the arrays to help convey meaning.

engine/src/cmd/basecomputer.cpp
royfalk: It would, but in this case, it really is just any facet. It's doing a check on a single facet and extrapolating the result for all others.

engine/src/cmd/damageable.h
royfalk: The problem is the underlying "death" code is unclear and complicated. You would think a unit would just die and be gone. Instead, there's a dying process when it's exploding. This leads to all sorts of complicated code that I'll admit I don't fully understand. 
It would probably be better to kill the Unit object and have it eject a DyingUnit object along with cargo and pilot.
It also seems I never used dying, so I removed it.

engine/src/cmd/damageable.h
royfalk: It would, except here. GetHull is already the convenience method hiding the messy implementation. Yes, we could do something like: `const int SINGLE_FACET=0;` and use that but would that really tell you that much more?

engine/src/cmd/damageable.h
royfalk: GetArmor and GetShield are not used. I added them for completeness sake because we have GetHull and it is exposed to python. We can probably remove this but if we expose hull to python, we should also expose these. Because it's python, you can't limit to enum class anyway (can you?!?). I added a TODO to check for ranges.

engine/src/cmd/damageable.h
royfalk: Same as above, this is hiding the messy implementation. It should really go in cpp file. I wouldn't assume it should be inlined. Certainly not yet.

engine/src/cmd/drawable.cpp
royfalk: Done.

engine/src/damage/damageable_layer.cpp
royfalk: Sorry. For some reason I assumed enum class does not support switches. Will fix.

engine/src/cmd/damageable.h
stephengtuggy: OK. Resolving.

engine/src/cmd/damageable.h
stephengtuggy: @royfalk I would go for the `SINGLE_FACET` constant. I think it would make the code clearer, actually

engine/src/cmd/damageable.h
stephengtuggy: I agree with @royfalk here. A unit dying is quite the process. I don't fully understand it either

engine/src/damage/damageable_layer.h
royfalk: Not in this case. Remember that lib_damage doesn't understand hull, armor and shield. In theory you could use it in some other way - say with multiple layers of armor.

engine/src/damage/facet_configuration.cpp
royfalk: Removed function, as using as_integer is better.

engine/src/damage/facet_configuration.cpp
royfalk: Done.

stephengtuggy:
Woot! The Windows build succeeded

engine/src/cmd/basecomputer.cpp
royfalk: Fixed by adding value to enum.

engine/src/cmd/damageable.h
royfalk: Actually went with return *current_hull. We still need GetHull as it's part of the python API.

engine/src/cmd/damageable.h
royfalk: It's short and it's in the header but I'm not convinced it should be inlined. I'd prefer someone optimizes things after profiling them. In any case, the low hanging fruit is introducing threads.

engine/src/damage/damageable_layer.cpp
royfalk: Done.

engine/src/cmd/damageable.h
BenjamenMeyer: okay.

I'm guessing the "death" code has to do with explosions and such. But we can leave it for now and come back to it in a later refactor.

engine/src/cmd/damageable.h
BenjamenMeyer: You should still be able to limit to the enum class too, it's just a matter of mapping; it might be easier to do in the Python C API as opposed to Boost::Python though...not sure. But if it's too complex we can come back to it in a later refactor.  I'm mostly looking to make sure we can understand the code; for now, comments might be best so we an convey intent.

BenjamenMeyer (approved):

BenjamenMeyer:
Approved per gitter conversation

engine/CMakeLists.txt
```diff
@@ -1539,3 +1550,33 @@ ENDIF (WIN32 AND NOT UNIX)
...
 INCLUDE(CPack)
 
+INCLUDE(FetchContent)
+FetchContent_Declare(
+    googletest
+    URL https://github.com/google/googletest/archive/609281088cfefc76f9d0ce82e1ff6c30cc3591e5.zip
+    )
+# For Windows: Prevent overriding the parent project's compiler/linker settings
+SET(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+FetchContent_MakeAvailable(googletest)
```
stephengtuggy, line 1560: Since we'll probably still need to support CMake version 3.11, this line will probably need to be changed to fit the pattern:

```cmake
FetchContent_GetProperties(googletest)
if(NOT googletest_POPULATED)
  FetchContent_Populate(googletest)
  ...
endif()
```

See https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_getproperties

stephengtuggy: I think this is the only change remaining to be made, other than merging in the latest from master after #536 gets merged. Then we should be good to go.

stephengtuggy: ... and #536 has now been merged into master. @royfalk those changes should be ready for you to merge into this branch, whenever you get the chance.

stephengtuggy:
Hmm. I'm getting the following when running on Windows 10:

```log
Exception thrown at 0x00007FF705B76550 in vegastrike-engine.exe: 0xC0000005: Access violation reading location 0x0000000000000010.
```

That's three times in a row that I've gotten the same error in the same place. Something isn't right.

engine/src/damage/tests/health_tests.cpp
```diff
@@ -0,0 +1,81 @@
+#include <gtest/gtest.h>
+
+#include "health.h"
+
+
+// Demonstrate some basic assertions.
```
stephengtuggy, line 6: I assume that disabling these tests is only temporary?

engine/src/cmd/unit_csv.cpp
```diff
@@ -1048,87 +1047,80 @@ void Unit::LoadRow( CSVRow &row, string modification, string *netxml )
...
-    shield_values[7] = ::stof(OPTIM_GET( row, table, Shield_Back_Bottom_Right ));
+    // TODO: this mapping should really go away
+    shield_values[0] = ::stof(OPTIM_GET( row, table, Shield_Front_Top_Right ));
+    shield_values[1] = ::stof(OPTIM_GET( row, table, Shield_Back_Top_Left ));
+    shield_values[2] = ::stof(OPTIM_GET( row, table, Shield_Front_Bottom_Right ));
+    shield_values[3] = ::stof(OPTIM_GET( row, table, Shield_Front_Bottom_Left ));
 
 
-    for(int i=0;i<8;i++) {
+    for(int i=0;i<4;i++) {
```
BenjamenMeyer, line 1059: perhaps `sizeof(shield_values)` (or something like that)  instead of `4`?
Ultimately we may want to change this to an iterable type so we could just use iterators:
```
for (auto sv = shield_values.begin(); sv != shield_values.end(); sv++) {
   if (*sv) {
       shield_count++;
   }
}
```

but it's probably not possible to make that change yet.

engine/src/cmd/damageable.cpp
```diff
@@ -495,28 +494,56 @@ void Damageable::Destroy() {
...
+            (unit->shield->facets[facet_index].max_health);
+}
+
+
 float Damageable::FShieldData() const
 {
-    Damageable *damageable = const_cast<Damageable*>(this);
-    return static_cast<DamageableLayer>(damageable->GetShieldLayer()).GetPercent(FacetName::front);
+    switch(shield->number_of_facets) {
+    case 2:
```
BenjamenMeyer, line 508: would it be possible to use enums instead of 2/4 in these?

engine/src/cmd/damageable.h
BenjamenMeyer: functions definitions in headers should be inline; if they shouldn't be, then they don't belong in headers.

engine/src/python/unit_wrapper_class.h
```diff
@@ -272,9 +272,8 @@ class UnitWrapper : public UnitContainer
     }
     void ApplyDamage( Vector pnt,
                       Vector normal,
-                      float amt,
+                      Damage damage,
```
stephengtuggy, line 275: Hmm. I hope this doesn't break anything Python-wise.

engine/src/cmd/damageable.cpp
royfalk: I'd leave that to the next PR. It's better than it has been and let's leave it at that.

royfalk:
This last round of fixes sorted nearly everything out. 
At this point I think we can either merge this to master or to some kind of beta for people to test.

I've done extensive testing both during development and after. Here are the remaining issues:

- Can't kill asteroids when firing at them with guns. It used to be possible.
- I once fired a missile at a ship in an asteroid field and the whole field detonated. I had a similar issue some time ago in another PR.
- There was an issue where the shield turned off at some point and wouldn't turn on. Not sure why and not reproduced in later tests.
- When ejecting, the pilot has better radar than skyscope01. It has colors and stuff. It should have less.
- Escort ships of bounty missions don't turn red when I blow the bounty ship up. They are marked as hostile and yellow. This may be due to fear. I did notice them running away.

engine/src/damage/damageable_layer.cpp
```diff
@@ -283,14 +348,57 @@ float DamageableLayer::GetPercent(FacetName facet_name) {
...
+        facets[0].health = facets[0].max_health = new_facets[0];
+        break;
+
+    case 4:
+        facets[0].health = facets[0].max_health = new_facets[3];
+        facets[1].health = facets[1].max_health = new_facets[2];
+        facets[2].health = facets[2].max_health = new_facets[0];
+        facets[3].health = facets[3].max_health = new_facets[1];
+        break;
+    case 2:
```
stephengtuggy, line 387: So case 2 falls through to case 8? Might want to add a comment to that effect. (Small nitpick.)

engine/src/cmd/energetic.cpp
```diff
@@ -230,3 +256,187 @@ float Energetic::warpEnergyData() const
...
+
+void Energetic::ExpendEnergy(float usage) {
+    energy = std::max(0.0f, energy - usage);
+}
+
+// The original code was a continuation of the comment above and simply unclear.
+// I replaced it with a very simple model.
+void Energetic::ExpendFuel() {
+    //Fuel Mass in metric tons expended per generation of 100MJ
+    static float FMEC_factor = GameConfig::GetVariable( "physics", "FMEC_factor", 0.000000008 );
```
BenjamenMeyer, line 307: Just to not: this is pushing the boundaries of what a float can compute, depending on hardware. Most hardware should be 9 decimal now, but it's possible it could be just 6 or 7 digit. (Not a blocker...just noting it)

PR #521, opened 2021-08-20
Fix/issue 513

### LINKED ISSUE #513: gfx/matrix.h: VC++ warning about loss of precision from `double` to `float`, on 9 separate lines in this file
Compiling Vega Strike on Windows generates a whole slew of compiler warnings. Many of them are repeated over and over again, from the same header files included multiple times in different .cpp files.

One such repeated warning mentions a possible loss of precision converting from `double` to `float`, in `engine/src/gfx/matrix.h`. This warning occurs on **nine separate lines** in this file, for a total number of occurrences equal to 9 times however many files matrix.h is included in.

We could cut down the number of compiler warnings a LOT by just fixing this file. And it looks like a simple fix to me: Just change

```C++
    double c = cos( angle );
    double s = sin( angle );
```

in inline function Rotate to

```C++
    float c = cos( angle );
    float s = sin( angle );
```

The overloaded `cos` and `sin` functions return `float`s in this case anyway. Then those `float` values are converted to `double`s, and then back to `float`s. It doesn't make much sense, at least as far as I can see.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #513 

Purpose:
- What is this pull request trying to do? Fix a set of compiler warnings
- What release is this for? 0.9.x / 1.0.x
- Is there a project or milestone we should apply this to? 0.9.x / 1.0.x

### DIFF
--- a/engine/src/gfx/matrix.h
+++ b/engine/src/gfx/matrix.h
@@ -355,8 +355,8 @@ inline void InvertMatrix( Matrix &o, const Matrix &trans )
 
 inline void Rotate( Matrix &tmp, const Vector &axis, float angle )
 {
-    double c = cos( angle );
-    double s = sin( angle );
+    float c = cos( angle );
+    float s = sin( angle );
 //Row, COl
 #define M( a, b ) (tmp.r[b*3+a])
     M( 0, 0 ) = axis.i*axis.i*(1-c)+c;
--- a/engine/src/gfx/matrix.h
+++ b/engine/src/gfx/matrix.h
@@ -1,23 +1,28 @@
-/*
- * Vega Strike
- * Copyright (C) 2001-2002 Alan Shieh Rewritten by Daniel Horn for Double precision
+/**
+ * matrix.h
  *
- * http://vegastrike.sourceforge.net/
+ * Copyright (C) 2001-2002 Alan Shieh and Daniel Horn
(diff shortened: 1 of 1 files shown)


stephengtuggy:
I may be imagining it, but it seems like my ship actually rolls more smoothly with this patch.

engine/src/gfx/matrix.h
```diff
@@ -355,8 +360,8 @@ inline void InvertMatrix( Matrix &o, const Matrix &trans )
 
 inline void Rotate( Matrix &tmp, const Vector &axis, float angle )
 {
-    double c = cos( angle );
-    double s = sin( angle );
+    float c = cos( angle );
+    float s = sin( angle );
```
BenjamenMeyer, line 364: Okay, so this means we don't lose precision on the cos/sin of angle.
Question: do this mean we end up losing precision in M() when processing Matrix and Vector data? Is any of that data stored as `double`? or is it it all `float`?

BenjamenMeyer (commented):
Just a question before I approve

engine/src/gfx/matrix.h
stephengtuggy: @BenjamenMeyer I just checked, and all of the Matrix and Vector data is stored as `float`.

BenjamenMeyer (approved):

PR #536, opened 2021-08-30
On master, drop Ubuntu 16.04 "Xenial" LTS and Debian 9 "stretch"

### LINKED ISSUE #528: On master, drop support for Ubuntu 16.04 "xenial" and Debian 9 "stretch"
As discussed on Gitter.

This move will allow us to update our minimum supported CMake version to 3.10 or 3.11. That, in turn, should allow us to include unit tests in our CI/CD with GoogleTest.

### LINKED ISSUE #529: script/bootstrap: Update openSUSE version detection logic to match what's on build-system-docker-images

stephengtuggy:

Please answer the following:

Code Changes:
- [X] This is a CI/CD/build system change only

Issues:
- Closes #528 
- Closes #529 

Purpose:
- What is this pull request trying to do? Drop support for Xenial and stretch on master, to pave the way for CMake v3.11 and gtest unit tests
- What release is this for? 0.9.x/1.0.x
- Is there a project or milestone we should apply this to? 0.9.x/1.0.x


stephengtuggy:
Should I drop Xenial and stretch from the build scripts and from CMakeLists.txt in the same PR? I suppose I should.

nabaco (approved):

stephengtuggy:
Thanks, @nabaco . Unfortunately, I had added 4 more commits on my local machine last night, and another one this morning. Hopefully it's not too much trouble to review those too?

script/build-authors.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
BenjamenMeyer, line 1: `/bin/bash` is actually guaranteed to be there per Linux Standard Base.
`/usr/bin/env` is not.

There's various arguments either way for most things, and most scripts engines it can make sense. However, `/bin/bash` and `/bin/sh` are two that have to be around for a basic system to function and can be directly relied upon so they don't make sense to export out to `/usr/bin/env`.

$0.02

BenjamenMeyer (approved):
I'd prefer we change the scripts back to using `/bin/bash` instead of using `/usr/bin/env bash`; but I'm not going to block. See my comment on the one case for more info/discussion.

script/build-authors.sh
stephengtuggy: Hmm. I can look this up later, I suppose. Anyway, thanks for approving the PR

PR #539, opened 2021-09-03
Add unit tests

### LINKED ISSUE #532: Add unit tests via GoogleTest and CTest
On master only

stephengtuggy:

Please answer the following:

Code Changes:
- [-] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Closes #532 

Purpose:
- What is this pull request trying to do? Add unit tests to the project, using GoogleTest and CTest

This is a CI/CD/build system change. As such, I have run some quick sanity tests on Linux and Windows, and I think we are good to go.


stephengtuggy:
Note that, when I tried to use gtest v1.10, CodeQL flagged two new security problems in the gtest-related Python code. So I switched to using gtest v1.11, and the problems went away.

Using the latest version is probably best anyway, right?

.github/workflows/gh-actions-pr.yml
```diff
@@ -7,8 +7,12 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
 
+    defaults:
+      run:
+        shell: bash
+
     strategy:
-      fail-fast: true
+      fail-fast: false
```
BenjamenMeyer, line 15: we probably want fast-fail as it prevents other builders from running if earlier ones fail.

BenjamenMeyer (changes requested):

.github/workflows/gh-actions-pr.yml
stephengtuggy: Fixed. That was supposed to be temporary. Good catch

BenjamenMeyer (approved):

PR #540, opened 2021-09-04
cmd/collision.cpp: Make it so that Nav 8 cannot be crashed into

### LINKED ISSUE #516: PWCU: Unexpectedly, Nav 8 in Troy is an object that can be collided with

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Fixes #516 for 0.8.x branch

Purpose:
- What is this pull request trying to do? Fix a bug where, with PWCU, Nav 8 is unexpectedly an object that you can crash into
- What release is this for? 0.8.x; 0.9.x pending
- Is there a project or milestone we should apply this to? 0.8.x

### DIFF
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -292,10 +292,7 @@ void Collision::collide( Unit* unit1,
     // See commit a66bdcfa1e00bf039183603913567d48e52c7a8e method Unit->jumpReactToCollision for an example
     // I assume this is for "always open" jump gates that you pass through
 
-    collision1.apply_force = false;
-    collision1.deal_damage = false;
-    collision2.apply_force = false;
-    collision2.deal_damage = false;
+    // apply_force and deal_damage are both initialized to false when the Collision object is constructed. (See collision.h.)
 
     collision1.shouldApplyForceAndDealDamage(unit2);
     collision2.shouldApplyForceAndDealDamage(unit1);
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -112,7 +112,7 @@ void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
             return;
         case _UnitType::planet:
             // Handle the "Nav 8" case
-            if (other_unit->getFilename().find("invisible")) {
+            if (other_unit->getFilename().find("invisible") != std::string::npos) {
                 BOOST_LOG_TRIVIAL(debug) << "Found a Nav_8-type object";
                 return;
             } else {
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -101,17 +101,25 @@ _UnitType::enhancement,
 _UnitType::missile*/
 void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
 {
-    // Collision with a nebula does nothing
-    if(other_unit->isUnit() == _UnitType::nebula)
+    switch (other_unit->isUnit())
     {
-        return;
-    }
-
-    // Collision with a enhancement improves your shield apparently
-    if(other_unit->isUnit() == _UnitType::enhancement)
-    {
-        apply_force = true;
-        return;
+        case _UnitType::nebula:
+            // Collision with a nebula does nothing
+            return;
+        case _UnitType::enhancement:
+            // Collision with a enhancement improves your shield apparently
+            apply_force = true;
+            return;
+        case _UnitType::planet:
+            // Handle the "Nav 8" case
+            if (other_unit->getFilename().find("invisible")) {
+                BOOST_LOG_TRIVIAL(debug) << "Found a Nav_8-type object";
+                return;
+            } else {
+                break;
+            }
+        default:
+            break;
     }
 
     // Collision with a jump point does nothing
@@ -298,15 +306,15 @@ void Collision::collide( Unit* unit1,
     collision2.shouldApplyForceAndDealDamage(unit1);
 
     float elasticity = 0.8f;
-    if (collision1.apply_force && collision2.apply_force) {
+    // if (collision1.apply_force && collision2.apply_force) {
         collision1.applyForce(elasticity, collision2.mass, collision2.velocity);
         collision2.applyForce(elasticity, collision1.mass, collision1.velocity);
-    }
+    // }
 
-    if (collision1.deal_damage && collision2.deal_damage) {
+    // if (collision1.deal_damage && collision2.deal_damage) {
         collision1.dealDamage(collision2);
         collision2.dealDamage(collision1);
-    }
+    // }
 
     // Is this still a TODO? -- add torque
 }
--- a/engine/src/cmd/collision.cpp
+++ b/engine/src/cmd/collision.cpp
@@ -1,7 +1,7 @@
 /**
  * collision.cpp
  *
- * Copyright (C) 2020 Roy Falk, Stephen G. Tuggy and other Vega Strike
+ * Copyright (C) 2020-2021 Roy Falk, Stephen G. Tuggy and other Vega Strike
  * contributors
  *
  * https://github.com/vegastrike/Vega-Strike-Engine-Source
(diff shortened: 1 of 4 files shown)


engine/src/cmd/collision.cpp
```diff
@@ -292,15 +292,24 @@ void Collision::collide( Unit* unit1,
     // See commit a66bdcfa1e00bf039183603913567d48e52c7a8e method Unit->jumpReactToCollision for an example
     // I assume this is for "always open" jump gates that you pass through
 
+    collision1.apply_force = false;
```
royfalk, line 295: This should not be necessary. The variables are initialized in the header.

engine/src/cmd/collision.cpp
```diff
@@ -292,15 +292,24 @@ void Collision::collide( Unit* unit1,
...
+    collision2.apply_force = false;
+    collision2.deal_damage = false;
+
     collision1.shouldApplyForceAndDealDamage(unit2);
     collision2.shouldApplyForceAndDealDamage(unit1);
 
     float elasticity = 0.8f;
-    collision1.applyForce(elasticity, collision2.mass, collision2.velocity);
-    collision2.applyForce(elasticity, collision1.mass, collision1.velocity);
+    if (collision1.apply_force && collision2.apply_force) {
```
royfalk, line 304: Also not needed. First line of applyForce function returns if false.

engine/src/cmd/collision.cpp
```diff
@@ -292,15 +292,24 @@ void Collision::collide( Unit* unit1,
...
+        collision1.applyForce(elasticity, collision2.mass, collision2.velocity);
+        collision2.applyForce(elasticity, collision1.mass, collision1.velocity);
+    }
 
-    collision1.dealDamage(collision2);
-    collision2.dealDamage(collision1);
+    if (collision1.deal_damage && collision2.deal_damage) {
```
royfalk, line 309: Same here. First line of dealDamage checks the variable.
In short, I don't see what your changes actually do to fix the issue. Also, I don't see how the issue is caused based on this. Unless you come up with a better hypothesis, I think the root cause is the function isJumppoint. It should return without setting either apply_force or apply_damage to true.
As this appears to work, we should probably check the data or WC to see what's there.

engine/src/cmd/collision.cpp
stephengtuggy: You might be right. But it wasn't obvious from looking at the code whether these fields were initialized to `false`, or not.

engine/src/cmd/collision.cpp
stephengtuggy: The behavior definitely changed when I made this change to the code. The difference is that now, no force is applied to **either** object/unit unless **both** pass the "shouldApplyForce" test. Whereas before, the conditions applied to the units individually.

engine/src/cmd/collision.cpp
stephengtuggy: See previous comment, above.

stephengtuggy: You might be right about isJumppoint though. I'll take a look at that as well.

stephengtuggy (commented):
@royfalk See responses.

engine/src/cmd/collision.cpp
stephengtuggy: You're right. I see that now. I just took those redundant initializations out.

stephengtuggy:
All right, let me work on this issue some more. I think I can find a better solution.

engine/src/cmd/collision.cpp
stephengtuggy: OK, I went ahead and undid this. It didn't turn out to be necessary after all.

engine/src/cmd/collision.cpp
stephengtuggy: Undid this too

stephengtuggy:
OK, I think I like this solution better. To determine whether a planet can be collided with, It literally checks for the string `"invisible"` in the planet's `getFullname()` property. So actual planets -- like Atlantis or Troy -- can be collided with, while invisible `Nav` objects -- such as `Nav_8` cannot.

It's a bit of an ugly hack, but it works.

stephengtuggy:
@royfalk you were correct -- none of my 3 earlier code changes were necessary. I commented out or removed each one of them, and took a different approach.

stephengtuggy:
Probably the ultimate solution here would be to have each game / dataset define its own types of objects, such as planets, warp points, etc., and define which ones can be collided with and which ones can't.

engine/src/cmd/collision.cpp
```diff
@@ -101,17 +101,26 @@ _UnitType::enhancement,
...
+            // Handle the "Nav 8" case
+            // BOOST_LOG_TRIVIAL(debug) << "shouldApplyForceAndDealDamage(): other_unit is a planet, with full name " << other_unit->getFullname();
+            if (other_unit->getFullname().find("invisible") != std::string::npos) {
+                // BOOST_LOG_TRIVIAL(debug) << "Found a Nav_8-type object";
+                return;
+            } else {
+                break;
+            }
+        default:
+            break;
```
BenjamenMeyer, line 123: should probably do a debug warning on default

engine/src/cmd/collision.cpp
```diff
@@ -101,17 +101,26 @@ _UnitType::enhancement,
 _UnitType::missile*/
 void Collision::shouldApplyForceAndDealDamage(Unit* other_unit)
 {
-    // Collision with a nebula does nothing
-    if(other_unit->isUnit() == _UnitType::nebula)
+    switch (other_unit->isUnit())
```
BenjamenMeyer, line 104: switch over if-block is 100% better!

BenjamenMeyer (changes requested):

BenjamenMeyer:
@stephengtuggy  we probably just need to add another object type for the NAV Point object to the engine.

The data sets should be able to change the displayable name various types, but should just be using what is provided by the engine.

stephengtuggy:
> @stephengtuggy we probably just need to add another object type for the NAV Point object to the engine.
> 
> The data sets should be able to change the displayable name various types, but should just be using what is provided by the engine.

@BenjamenMeyer do you want to add the Nav_Point object type in 0.8.x? I suppose we could do so in a fairly backwards-compatible manner, since we would be adding a unit type, not subtracting anything.

BenjamenMeyer:
> > @stephengtuggy we probably just need to add another object type for the NAV Point object to the engine.
> > The data sets should be able to change the displayable name various types, but should just be using what is provided > @stephengtuggy we probably just need to add another object type for the NAV Point object to the engine.
> The data sets should be able to change the displayable name various types, but should just be using what is provided by the engine.

@BenjamenMeyer do you want to add the Nav_Point object type in 0.8.x? I suppose we could do so in a fairly backwards-compatible manner, since we would be adding a unit type, not subtracting anything.by the engine.
> 
> @BenjamenMeyer do you want to add the Nav_Point object type in 0.8.x? I suppose we could do so in a fairly backwards-compatible manner, since we would be adding a unit type, not subtracting anything.

if it's not hard to do then we may have the time to; but otherwise it could be 0.9.x.

stephengtuggy:
@BenjamenMeyer so then what changes are required for 0.8.x? Just the warning message if a unit type does not match one of the expected values?

BenjamenMeyer:
> @BenjamenMeyer so then what changes are required for 0.8.x? Just the warning message if a unit type does not match one of the expected values?

At a minimum yes; but let's see how much effort it would be to add a NavPoint object that could be utilized. If it's too much work, we'll push that to 0.9.x; if not, it can go in 0.8.x but let's see what it would take first.

engine/src/cmd/collision.cpp
```diff
@@ -101,17 +101,26 @@ _UnitType::enhancement,
...
+            // Collision with a nebula does nothing
+            return;
+        case _UnitType::enhancement:
+            // Collision with a enhancement improves your shield apparently
+            apply_force = true;
+            return;
+        case _UnitType::planet:
+            // Handle the "Nav 8" case
+            // BOOST_LOG_TRIVIAL(debug) << "shouldApplyForceAndDealDamage(): other_unit is a planet, with full name " << other_unit->getFullname();
+            if (other_unit->getFullname().find("invisible") != std::string::npos) {
```
royfalk, line 116: It's an ugly hack, but at least it's minimal.

royfalk:
I was looking for an approve button when I realized there isn't one. LGTM.

stephengtuggy:
As I'm thinking some more about this, it occurs to me that if we add a NavPoint object to 0.8.x, then the game engine will be backwards compatible, but the PWCU data set won't be. That may not be desirable. I think that for the 0.8.x timeframe, we should keep both (all?) game datasets compatible with previous versions of the Vega Strike Game Engine. Which would mean leaving the NavPoint and similar objects for later.

What do you all think?

engine/src/cmd/collision.cpp
stephengtuggy: @royfalk Yep.

engine/src/cmd/collision.cpp
stephengtuggy: Done. For both the current unit's type and the other unit's type, I now handle all possible values of the `_UnitType` enum, and warn if an unexpected value is encountered.

stephengtuggy:
Play tests rerun, with both PWCU and VS:UtCS, and on Windows as well as Linux. Seems to be working well.

stephengtuggy:
The more I think about it, the more complicated I think it will be to add a NavPoint unit type to the code. I vote we wait until 0.9.x to do that.

stephengtuggy:
> The more I think about it, the more complicated I think it will be to add a NavPoint unit type to the code. I vote we wait until 0.9.x to do that.

We will need to add some more possible values to the `_UnitType` enum. Add code to handle the additional values in each place where it's relevant. Add classes for the new unit types. Implement physics and drawing behavior. Then use the new functionality in the PWCU dataset. And test, test, test.

BenjamenMeyer (approved):

BenjamenMeyer:
@stephengtuggy thanks for the analysis. We can let it slide for 0.8.x

Loki1950 (approved):

PR #551, opened 2021-09-11
Fix/CodeQL Warnings 2021-09 (Part 1)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:

Purpose:
- What is this pull request trying to do? Fix some CodeQL warnings. And in the process, refactor some code so that, for example, the startup sequence on Win32 matches that on other platforms more closely.

### DIFF
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -248,10 +248,10 @@ int main( int argc, char *argv[] )
     // Change to program directory if not already
     // std::string program_as_called();
     const boost::filesystem::path  program_path(argv[0]);
-    // const boost::filesystem::path& canonical_program_path = boost::filesystem::canonical(program_path);
+    // const boost::filesystem::path canonical_program_path = boost::filesystem::canonical(program_path);
 
-    const boost::filesystem::path& program_name = program_path.filename();  //canonical_program_path.filename();
-    const boost::filesystem::path& program_directory_path = program_path.parent_path(); //.parent_path();
+    const boost::filesystem::path program_name{program_path.filename()};  //canonical_program_path.filename();
+    const boost::filesystem::path program_directory_path{program_path.parent_path()};
 
     // This will be set later
     boost::filesystem::path home_subdir_path{};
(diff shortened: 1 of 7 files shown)


engine/src/main.cpp
```diff
@@ -273,9 +273,11 @@ int main( int argc, char *argv[] )
...
     VegaStrikeLogging::VegaStrikeLogger::InitLoggingPart1();
 
-#ifdef WIN32
-    VSFileSystem::InitHomeDirectory();
-#endif
+    // stephengtuggy 2021-09-10 Let's try initializing the directories in the same order
+    // on Windows as on other platforms, and see what happens
+// #ifdef WIN32
+//     VSFileSystem::InitHomeDirectory();
+// #endif
```
BenjamenMeyer, line 280: :+1:

engine/src/universe_util_generic.cpp
```diff
@@ -898,43 +907,73 @@ string getSaveInfo( const std::string &filename, bool formatForTextbox )
...
+                ss << "  " << simplePrettySystem( Ships[i] ) << lf;
+            }
         }
     }
     if (!game_options.quick_savegame_summaries) {
         bool hit = false;
         for (set< string >::const_iterator it = campaign_score_vars.begin(); it != campaign_score_vars.end(); ++it) {
             string var = *it;
-            unsigned int curscore = savegame.getMissionData( var ).size()+savegame.getMissionStringData( var ).size();
+            int_fast64_t curscore = savegame.getMissionData( var ).size() + savegame.getMissionStringData( var ).size();
```
BenjamenMeyer, line 963: why `int_fast64_t` instead of `int64_t`?

BenjamenMeyer (commented):
One question, generally looks good

engine/src/main.cpp
stephengtuggy: As a nice bonus, this change seems to have enabled PWCU to start on Windows without crashing.

engine/src/universe_util_generic.cpp
stephengtuggy: OK, I went with `int64_t`.

BenjamenMeyer (approved):

PR #553, opened 2021-09-14
Bug Fix: properly detect DEB package tooling

### LINKED ISSUE #510: cmake fails on Gentoo-based systems that have a program installed called `apt`
The current revision of the code fails to configure with cmake on non-Debian systems. Cmake looks for /etc/debian_version and if that file doesn't exist, it fails. That file does not exist on systems that are not based on Debian.

Trying to build on Funtoo (Gentoo-based distribution) results in:

-- Configuring Packaging for Linux
-- Configuring Debian Packaging
CMake Error at CMakeLists.txt:1316 (FILE):
  FILE failed to open for reading (No such file or directory):

    /etc/debian_version

Looking for /home/silviu/vegastrike/Vega-Strike-Engine-Source/engine/build1/dependency.list
Still need to generate dependency list
SEND_WARNING!! Unsupported Debian version.
See also "/home/silviu/vegastrike/Vega-Strike-Engine-Source/engine/build1/CMakeFiles/CMakeOutput.log".
See also "/home/silviu/vegastrike/Vega-Strike-Engine-Source/engine/build1/CMakeFiles/CMakeError.log".

This wasn't happening the last time I built it, about a month ago.

BenjamenMeyer:
- Enhancement: Move some of the packaging logic out of the primary CMakeLists.txt and into its own CMake module file.
- Bug Fix: Apparently `apt` is a program other than the Debian packager so use `dpkg-buildpackage` instead which is more accurate as well.

Please answer the following:

Code Changes:
- [x] Build System Change

Issues:
- #510 for master (0.9.x); will need to be ported to 0.8.x

Purpose:
- What is this pull request trying to do? Fix #510, improve the process across all distros

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -105,14 +105,14 @@ IF (MSVC)
         ${Vega_Strike_SOURCE_DIR}/src
         ${Vega_Strike_SOURCE_DIR}/src/cmd
         ${Vega_Strike_BINARY_DIR}
-)
+    )
 ELSE (MSVC)
     INCLUDE_DIRECTORIES(
         ${Vega_Strike_SOURCE_DIR}/src
         ${Vega_Strike_SOURCE_DIR}/src/cmd
         ${Vega_Strike_BINARY_DIR}
         /usr/include/harfbuzz/
-)
+    )
 ENDIF (MSVC)
 
 # The source files used to be listed here...
@@ -177,19 +177,6 @@ IF (DATADIR)
     SET(DEFINES "${DEFINES} -DDATA_DIR='\"${DATADIR}\"'")
 ENDIF (DATADIR)
 
-# Gather info about Linux distro and release (if applicable) for later use down below.
-IF (CMAKE_SYSTEM_NAME STREQUAL Linux)
-    EXECUTE_PROCESS(
-        COMMAND sh -c "cat /etc/os-release | grep ^ID= | sed 's/^ID=//' | tr -d '\"\n'"
-        OUTPUT_VARIABLE LINUX_ID)
-    EXECUTE_PROCESS(
-        COMMAND sh -c "cat /etc/os-release | grep ^VERSION_CODENAME= | sed 's/^VERSION_CODENAME=//' | tr -d '\"\n'"
-        OUTPUT_VARIABLE LINUX_CODENAME)
-    EXECUTE_PROCESS(
-        COMMAND sh -c "cat /etc/os-release | grep ^VERSION_ID= | sed 's/^VERSION_ID=//' | tr -d '\"\n'"
-        OUTPUT_VARIABLE LINUX_VERSION_ID)
-ENDIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
-
 # On some Ubuntu versions and derivatives, a bug exists whereby enabling
 # PIE compilation (Position Independent Executables) results in the
 # `file` utility incorrectly recognising the compiled vegastrike-engine binary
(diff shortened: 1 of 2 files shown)


engine/CMakeLists.txt
```diff
@@ -1339,31 +1339,11 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
-            OUTPUT_VARIABLE LSB_LINUX_DISTRIBUTION_CODENAME
-            OUTPUT_STRIP_TRAILING_WHITESPACE)
-        EXECUTE_PROCESS(COMMAND ${LSB_RELEASE_EXEC} --id --short
-            OUTPUT_VARIABLE LSB_LINUX_DISTRIBUTION
-            OUTPUT_STRIP_TRAILING_WHITESPACE)
-        SET(LSB_EXISTS TRUE)
-        MESSAGE("Found Linux Distribution Release Name: ${LSB_LINUX_DISTRIBUTION_CODENAME}")
-    ENDIF (LSB_RELEASE_EXEC)
+	# Detect available Linux Distros that can be built for
+	FIND_PACKAGE(LinuxDistro REQUIRED)
```
stephengtuggy, line 1343: Hmm. Not sure if I like the `FIND_PACKAGE` semantics here. Almost makes it sound like the Linux Distro is just another dependency, like SDL or expat. It's not. It's the entire operating system.

Could we change the usage to something like `INCLUDE(LinuxDistroDetect)` ?

engine/FindLinuxDistro.cmake
```diff
@@ -0,0 +1,51 @@
...
+# we can probably more reliably detect via checking for different packager commands
+# `lsb_release` doesn't always work correctly, so it's easier to read the underlying file
+# if it exists. It doesn't always exist. Debian doesn't use it, but Ubuntu does.
+SET(LSB_EXISTS FALSE)
+SET(LSB_LINUX_DISTRIBUTION "Unknown")
+SET(LSB_LINUX_DISTRIBUTION_CODENAME "Unknown")
+
+# Lookup the LSB Data - this is known for Ubuntu, but also used by other distros too.
+# and may help with distro-sub-selection
+FIND_PROGRAM(LSB_RELEASE_EXEC lsb_release)
```
stephengtuggy, line 35: What about also including the functionality from lines 180-191 of the main CMakeLists.txt file, that checks for and reads the `os-release` file, if present?

stephengtuggy (changes requested):
I noted a couple of questions, and one additional variable that needs to get documented.

engine/CMakeLists.txt
BenjamenMeyer: I modeled it after the other Find* dependencies; so it's following the same pattern.

Also, per the documentation FindPackage has many more options and would even give us the ability to pass parameters (via COMPONENTS) if we needed to.
- https://cmake.org/cmake/help/latest/command/find_package.html
- https://cmake.org/cmake/help/latest/command/include.html 

So I get what you're saying, but I think FIND_PACKAGE is still a better tool for the long run. I only did a few things we currently do now; if we start adding more distros then we may want to start adding options to it.

engine/FindLinuxDistro.cmake
BenjamenMeyer: I didn't see that. Thanks for pointing it out.

engine/CMakeLists.txt
stephengtuggy: OK, fair enough.

stephengtuggy (approved):
LGTM

PR #559, opened 2021-09-23
Bugfix gentoo apt get 0.8.x

### LINKED ISSUE #510: cmake fails on Gentoo-based systems that have a program installed called `apt`

BenjamenMeyer:

Please answer the following:

Code Changes:
- [x] Build System Change

Issues:
- #510 

Purpose:
- Port of #553 to 0.8.x
- One of the two commits on #553 was able to be cherry-picked; had to manually port the other commit.

Closes #510


engine/CMakeLists.txt
```diff
@@ -602,19 +602,6 @@ IF (DATADIR)
     SET(DEFINES "${DEFINES} -DDATA_DIR='\"${DATADIR}\"'")
 ENDIF (DATADIR)
 
-# Gather info about Linux distro and release (if applicable) for later use down below.
```
stephengtuggy, line 605: The `FIND_PACKAGE(LinuxDistro)` line needs to go here, not down below. Otherwise, locating python3 fails on Xenial and stretch. (Looks like that's what caused the CI builds to fail.)

stephengtuggy (changes requested):
The logic to determine the Linux distro needs to be moved further up in CMakeLists.txt. Other than that, it looks good.

engine/CMakeLists.txt
BenjamenMeyer: Interesting since it built locally (though not on that Ubuntu version, since I'm on 21.04 IIRC)

stephengtuggy: Yeah, it's only those particular distro versions that it fails on. At least to my knowledge.

BenjamenMeyer: Updated against master, which removes Xenial. Let's see if this is still and issue after that.

stephengtuggy: Looks like it's no longer an issue. Cool.

PR #561, opened 2021-09-26
Minor refactoring of weapons (beam/bolt)

royfalk:
This PR is part of an ongoing refactoring of the code base. It is for the next release.

Please answer the following:

Code Changes:

Issues: None due to this PR. I will be opening multiple bugs later.


stephengtuggy (changes requested):
This code seems functional enough, and it definitely fixes the "invisible beam" problem. However, I think we need to address the 4 "Multiplication result converted to larger type" issues before we merge this.

royfalk:
> This code seems functional enough, and it definitely fixes the "invisible beam" problem. However, I think we need to address the 4 "Multiplication result converted to larger type" issues before we merge this.

I usually don't mind doing some minor edits of nearby code, but this ain't it. I moved these functions from beam_generic to beam without touching them at all. You can tell there are 4 "fixed" warnings in beam_generic.

It's tempting to cast each part of the expression to double just to get rid of the warning but I vote against. This is a part of the code I don't understand at all and the original developers had a fondness for pushing the envelope in terms of what can be done. My concern here is someone originally said "let's do this by having it overflow in float" and fixing it would introduce some subtle or not so subtle bug.

I am fine with a new issue to deal with this, assigned to me. I'll get to it this week and test it separately.  I'll also try and understand the code a little and maybe polish some of it.

stephengtuggy (approved):
OK, fair enough. Approving.

BenjamenMeyer (approved):

PR #602, opened 2021-10-30
Under CMake v3.14 and up, use built-in PIE setting support

### LINKED ISSUE #599: Use CMake 3.14 PIE logic when available?
Understood that we don't want to move to 3.14 yet as our minimum supported CMake version. However, in those cases where we do have CMake 3.14 available, I think we should use the PIE support that it natively provides. (An `IF...ELSE` kind of thing.)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #599 

Purpose:
- What is this pull request trying to do? Improve support for Position Independent Executable (PIE) compiling and linking under CMake v3.14 and up


BenjamenMeyer (approved):

stephengtuggy:
I really can't die. I got Ataraxia Fighter Base to fire at me, such that I died 6 or 7 times, but each time, I just came back immediately / automatically with a new pilot escape pod.

This issue was noted previously. We'll have to address it separately, I think.

PR #605, opened 2021-11-11
Refactor CSV code for unit creation

royfalk:
This is part of the general unit refactoring effort and a prerequisite to the following features:
- Ship components refactoring (shield, armor, etc.)
- JSON unit files

Code Changes:
- [yes] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues (none probably related to PR):
- 20 second freeze during jump.
- Shrmn(?) stratofortress has a mesh glitch due to overlapping areas.
- Save sometimes starts in space and not docked.


engine/src/cmd/unit_csv.cpp
```diff
@@ -1088,162 +850,142 @@ void Unit::LoadRow( CSVRow &row, string modification, string *netxml )
...
+    limits.yaw = UnitCSVFactory::GetVariable(unit_key, "Maneuver_Yaw", 0.0f) *VS_PI/180.0;
+    limits.pitch     = UnitCSVFactory::GetVariable(unit_key, "Maneuver_Pitch", 0.0f) *VS_PI/180.0;
+    limits.roll      = UnitCSVFactory::GetVariable(unit_key, "Maneuver_Roll", 0.0f) *VS_PI/180.0;
+
+    YawPitchRollParser(unit_key, "Yaw_Governor", "Yaw_Governor", "Yaw_Governor", computer.max_yaw_right, computer.max_yaw_left);
+    YawPitchRollParser(unit_key, "Pitch_Governor", "Pitch_Governor_Up", "Pitch_Governor_Down", computer.max_pitch_up, computer.max_pitch_down);
+    YawPitchRollParser(unit_key, "Roll_Governor", "Roll_Governor_Right", "Roll_Governor_Left", computer.max_roll_right, computer.max_roll_left);
+
+    static float game_accel = GameConfig::GetVariable( "physics", "game_accel", 1.0 );
+    static float game_speed = GameConfig::GetVariable( "physics", "game_speed", 1.0 );
```
BenjamenMeyer, line 878: did you mean `1.0f` on these since you're casting to a float?

engine/src/cmd/unit_csv.cpp
```diff
@@ -1088,162 +850,142 @@ void Unit::LoadRow( CSVRow &row, string modification, string *netxml )
...
-            }
-            else
-            {
-                computer.radar.capability = value;
-            }
+    } else {
+        unsigned int value = stoi(iffval, 0);
+        if (value == 0) {
+            // Unknown value
+            assert(false);
```
BenjamenMeyer, line 929: assert should have a good message with it for the logs so that we can know why it broke...or was this just for your testing and it slipped through?

royfalk: I was actually confused by it, as I thought it was mine, but it's existing code. I think I'll just delete it, as it's clearly not intended to be there. I think someone else added it for exactly the same reason - to catch some edge case and accidentally left it there.

engine/src/cmd/unit_csv.cpp
royfalk: Fixed.

engine/src/cmd/tests/csv_tests.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "unit_csv_factory.h"
+#include <iostream>
+#include <fstream>
+#include <iterator>
+
+TEST(CSV, Sanity) {
+    // This may not work for all deployments.
+    // Consider standardizing this.
+    std::ifstream ifs ("../../data/units/units.csv", std::ifstream::in);
```
stephengtuggy, line 11: Yeah, I think we might need to figure out something so that this works more reliably across different machines and for different users.

stephengtuggy (approved):
Just one, nonblocking comment. Looks good!

engine/src/cmd/tests/csv_tests.cpp
royfalk: I agree but have no idea at the moment. Hopefully, this is an excellent task for a beginner. Well encapsulated and fairly easy.

PR #607, opened 2021-11-11
script/build.sh: Improve handling of paths and parameters with spaces

### LINKED ISSUE #606: build.sh errors out if the starting path has spaces in it
We need to wrap $BUILD_DIR in both braces {} and quotes "", ideally, to solve this problem.

stephengtuggy:

Please answer the following:

Code Changes:
- [Minimally, yes] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Addresses #606 

Purpose:
- What is this pull request trying to do? Improve the build script's handling of paths and parameters that contain spaces
- What release is this for? 0.9.x; should perhaps address this in 0.8.x as well
- Is there a project or milestone we should apply this to? 0.9.x; maybe 0.8.x in a later PR

### DIFF
--- a/script/build.sh
+++ b/script/build.sh
@@ -36,7 +36,7 @@
 set -e
 
 echo "-----------------------------"
-echo "--- build.sh | 2021-09-03 ---"
+echo "--- build.sh | 2021-11-11 ---"
 echo "-----------------------------"
 
 #----------------------------------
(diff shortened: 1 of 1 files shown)


script/build.sh
```diff
@@ -36,36 +36,38 @@
...
 
 ROOT_DIR=$(pwd)
-BUILD_DIR=$ROOT_DIR/build
-BIN_DIR=$ROOT_DIR/bin
-SRC_DIR=$ROOT_DIR/engine
+echo "ROOT_DIR: ${ROOT_DIR}"
+BUILD_DIR="${ROOT_DIR}/build"
+echo "BUILD_DIR: ${BUILD_DIR}"
+BIN_DIR="${ROOT_DIR}/bin"
+SRC_DIR="${ROOT_DIR}/engine"
```
BenjamenMeyer, line 51: not blocking: the `echo`'s can be done after the definitions as a block. Just makes for easier reasing.

BenjamenMeyer (approved):
One minor improvement possible, but not a blocker.

script/build.sh
stephengtuggy: Well, I was going to make this change, but then this PR sat here for two weeks. So I decided just to merge it.

Issue #550, opened 2021-09-11
Possible alternate locations for .vegastrike subfolder on Windows


stephengtuggy:
On Windows, Vega Strike puts .vegastrike subfolder under `C:\Users\<user name>\Documents`. IMHO, this is less than ideal. I would like to see us put `.vegastrike` directly under the user's profile folder, i.e. `C:\Users\<user name>\.vegastrike`.

If we're going to make this change, we had better make it soon, because it won't be backwards compatible. After the change, any existing Windows users will "lose" their Vega Strike data. It will be in a different folder.

Note that I think legacy Vega Strike (v0.5.1r1 and prior) put the data under C:\Program Files or something strange like that. We won't be able to keep **that** same folder path, whatever we do. But with 0.8.x and onward, I would like to see us get it right.

stephengtuggy:
Another possibility is that we should put the `.vegastrike` folder under `C:\Users\<user name>\AppData`, then either `Local` or `Roaming` underneath that. Between Local and Roaming, I'm not sure which is the best choice.

See discussion of this general issue on StackOverflow [here](https://stackoverflow.com/questions/57019/where-should-cross-platform-apps-keep-their-data?rq=1) .

stephengtuggy:
To end up with:

- `C:\Users\<user name>\.vegastrike`: Change the following in vsfilesystem.cpp to use `CSIDL_PROFILE` instead of `CSIDL_PERSONAL`:

```c++11
    // Get user's home directory the MS way
    if (SUCCEEDED(SHGetFolderPathA(NULL,
                                CSIDL_PERSONAL,
                                NULL,
                                0,
                                szPath)))
```

- To end up with `C:\Users\<user name>\AppData\Roaming`: Use `CSIDL_APPDATA`
- To end up with `C:\Users\<user name>\AppData\Local`: Use `CSIDL_LOCAL_APPDATA`

royfalk:
I generally think it should be under some kind of appdata folder. I'll comment on specific implementation when I review it, but am fine with using a windows specific API as per s/o post.

stephengtuggy:
@royfalk Sounds good.

I'm thinking we go with `CSIDL_LOCAL_APPDATA`. That will most closely match the existing behavior of having the generated star systems, etc. be unique to a specific machine.

ministerofinformation:
Seems good to me.

ermo:
[...]

PR #615, opened 2021-12-05
Change location of .vegastrike folder on Windows - 0.8.x

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Implements #550 for 0.8.x

Purpose:
- What is this pull request trying to do? Port #551 to 0.8.x, and then change location of .vegastrike and .pu folders on Windows, from "My Documents" to `C:\Users\<user>\AppData\Local`
- What release is this for? 0.8.x; 0.9.x pending
- Is there a project or milestone we should apply this to? 0.8.x; 0.9.x pending


Loki1950:
BTW that is the location that 0.05 placed the config file so there should be #Define with the relevant code still about.

BenjamenMeyer (commented):
LGTM, but one question: does it preserve the legacy location?

stephengtuggy:
> BTW that is the location that 0.05 placed the config file so there should be #Define with the relevant code still about.

Really? I thought I remembered 0.5.x trying to store all the data under Program Files.

Sure enough, looks like you're right. The My Documents location was introduced by @pheonixstorm in https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/0d8aaddf719ab107120122d1b411043cafc233e9 , in 2012. However, did that make it into 0.5.1r1, the last release from SVN?

stephengtuggy:
I see that 0.5.1.r1 was released April 1, 2012: https://forums.vega-strike.org/viewtopic.php?f=0&t=18088 https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/0d8aaddf719ab107120122d1b411043cafc233e9 was committed July 22, 2012. So I don't think that change made it into any of the releases until we switched over to GitHub, at which point VS was no longer playable on Windows anyway.

stephengtuggy:
The commit before that, https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/53de7524648fe43a8093490a86ab2f874986bfd2 , dated April 16, 2012, used `getenv("USERPROFILE")` . And the commit before that is the one with the behavior I remember: https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/97fb92cea69c1eb74694122515339a54dc043888 , dated October 1, 2003.

Loki1950 (approved):

engine/src/vs_globals.h
```diff
@@ -72,41 +98,41 @@ extern Mission    *mission;
 template < class MyType >
 class LeakVector
 {
-    std::vector< MyType > *active_missions;
+    std::vector< MyType > *internal_vector_p_;
```
BenjamenMeyer, line 101: for clarity, this should be declared private (default for class) or protected.

stephengtuggy: Done.

BenjamenMeyer (approved):

PR #618, opened 2021-12-07
Change location of .vegastrike folder on Windows - master

### LINKED ISSUE #550: Possible alternate locations for .vegastrike subfolder on Windows

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #550 

Purpose:
- What is this pull request trying to do? Port #615 to master, changing location of .vegastrike and .pu folders on Windows, from "My Documents" to `C:\Users\<user>\AppData\Local`


Loki1950 (approved):
Looks good @stephengtuggy

PR #619, opened 2021-12-09
Fix issue preventing player from dying

### LINKED ISSUE #600: When the ship is destroyed, the pilot should auto-eject some of the time.
Currently, it looks like the pilot auto-ejects all/most of the time. This does not appear to be the correct behavior.

**Required**

1. Find the old code and figure out the correct ratio.
2. Find the new code and fix the ratio to match.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #600 

Purpose:
- What is this pull request trying to do? Fix #600 , the issue where the player simply can't die

### DIFF
--- a/engine/src/cmd/damageable.cpp
+++ b/engine/src/cmd/damageable.cpp
@@ -228,7 +228,7 @@ void Damageable::ApplyDamage( const Vector &pnt,
 
         // Eject cargo
         static float cargo_eject_percent =
-            GameConfig::GetVariable( "physics", "eject_cargo_percent", 1.0 );
+            GameConfig::GetVariable( "physics", "eject_cargo_percent", 1.0f );
         static unsigned int max_dump_cargo =
             GameConfig::GetVariable( "physics", "max_dumped_cargo", 15 );
         unsigned int dumped_cargo = 0;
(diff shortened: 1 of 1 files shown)


engine/src/cmd/damageable.cpp
```diff
@@ -251,10 +251,8 @@ void Damageable::ApplyDamage( const Vector &pnt,
         static bool player_autoeject =
                 GameConfig::GetVariable( "physics", "player_autoeject", true );
 
-        if(shot_at_is_player && player_autoeject) {
-            unit->EjectCargo( (unsigned int) -1 );
-        } else if(rand() < (RAND_MAX*auto_eject_percent) && unit->isUnit() == _UnitType::unit &&
-                  unit->faction != neutralfac && unit->faction != upgradesfac) {
+        if (shot_at_is_player && player_autoeject && rand() < (RAND_MAX*auto_eject_percent)
```
royfalk, line 254: This will only apply to the player. The current behavior is that other units auto eject some of the time.

stephengtuggy: Not sure I follow?

royfalk: The deleted code was split in two:
if(player && player_auto_eject) eject
else something...

Your new code has merged both so the if statement will only apply to the player ship. Other pilots will not eject.

stephengtuggy: Hmm. You're probably right. Let me try this again

engine/src/cmd/damageable.cpp
```diff
@@ -247,15 +247,26 @@ void Damageable::ApplyDamage( const Vector &pnt,
...
+                VS_LOG(debug, "Not auto ejecting player");
+            }
+        } else {
+            if (unit->isUnit() == _UnitType::unit
+                    && rand() < (RAND_MAX * auto_eject_percent)) {
+                VS_LOG(debug, "Auto ejecting NPC");
+                unit->EjectCargo( (unsigned int) -1 );
+            } else {
+                VS_LOG(debug, "Not auto ejecting NPC");
+            }
```
stephengtuggy, line 269: OK, try it now?

I tested this code just now, and I was able to confirm that all four cases were exercised:

1. Auto ejecting NPC
2. Not auto ejecting NPC
3. Auto ejecting player
4. Not auto ejecting player

royfalk: OK. This looks good and I'll approve it.

I have some minor nitpicks:

- This is very readable but I'd shorten it and lose readability for something like:
`if(unit->isUnit && ((shot_at_is_player && player_auto_eject) || !shot_at_is_player)`
   This would collapse a lot of the code into something shorter. But, like I said, it's a balance and taste issue.
- I just realized that the if (unit->isUnit() == _UnitType::unit) is potentially going to be a big issue moving forward. If we move ships to their own class as we should and is planned, we'll need to go through the code and find all these instances.

royfalk: Oops. I can't approve as I'm not a reviewer.

stephengtuggy: I just re-requested a review from you, @royfalk . Hopefully that allows you to approve it.

engine/src/cmd/damageable.cpp
```diff
@@ -247,15 +247,26 @@ void Damageable::ApplyDamage( const Vector &pnt,
...
                 GameConfig::GetVariable( "physics", "player_autoeject", true );
 
-        if(shot_at_is_player && player_autoeject) {
-            unit->EjectCargo( (unsigned int) -1 );
-        } else if(rand() < (RAND_MAX*auto_eject_percent) && unit->isUnit() == _UnitType::unit &&
-                  unit->faction != neutralfac && unit->faction != upgradesfac) {
-            unit->EjectCargo( (unsigned int) -1 );
+        if (shot_at_is_player) {
+            if (player_autoeject
+                    && rand() < (RAND_MAX * auto_eject_percent)) {
```
BenjamenMeyer, line 256: This would be better as a function of some sort, especially since you use it twice right here - or calculate it ahead of time and store in a variable

BenjamenMeyer (approved):
Would prefer the rand lines be less duplicative, but not going to block on it.

PR #627, opened 2021-12-23
Further refactoring of bolt.cpp

royfalk:
Move away from the concept of decals to a simpler model where texture manager holds a master of every texture and copies it per shot.
Can't do the same for balls and animations because of weird life cycle issue of animations. They're destroyed before they should.

Please answer the following:

Code Changes:

Issues:
- Shooting within the shield bubble of a base cause it to attack you. 
- Balls animation only use the first animation displayed. This is most likely a regression in the previous refactoring of bolt.cpp.
- Can't die in a collision. 
I'll open issues to the above as they are probably unrelated to this PR.

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -968,6 +968,7 @@ SET(LIBCMDSERVER_SOURCES
 )
 
 SET(LIBGFXGENERIC_SOURCES
+    src/gfx/texture_manager.cpp
     src/gfx/boltdrawmanager.cpp
     src/gfx/cockpit_generic.cpp
     src/gfx/lerp.cpp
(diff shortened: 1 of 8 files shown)


stephengtuggy:
- Shooting within the shield bubble of a base causes it to attack you -- I think this has been this way for a long time.
- Balls animation only use the first animation displayed -- Hmm. That's concerning. Can we double-check when this bug was introduced?

BenjamenMeyer (approved):
LGTM

royfalk:
>Shooting within the shield bubble of a base cause it to attack you.

Not sure what to do about this. Probably make all units with shield tight?

> Balls animation only use the first animation displayed. This is most likely a regression in the previous refactoring of bolt.cpp.

Opened #630 for this.

> Can't die in a collision.

Unable to reproduce.

PR #643, opened 2022-01-23
find python3 directly

maddanio:

Please answer the following:

Code Changes:

Issues:

Purpose:

make python3 detection from cmake more robust by using new mechanisms

- What release is this for?

head

your call


stephengtuggy:
Thanks for the pull request, @maddanio ! I did not know about the `FIND_PACKAGE(Python3...)` syntax. That's good to know.

It looks like this syntax is supported only by CMake 3.12 and up. Our current minimum supported CMake version is 3.11. We will need to discuss it further before updating our minimum version, I think.

stephengtuggy (dismissed):
This doesn't seem to work on macOS for some reason. It's also failing on Fedora 34. Looks like more work is needed.

Again, thanks for the contribution. It would be nice to get this change functional.

maddanio:
> This doesn't seem to work on macOS for some reason. It's also failing on Fedora 34. Looks like more work is needed.

I thinl thats the link step. I will have a look. Curious: how do you manage to build on macos at all? I had problems with the gl functions for which it seems vs expects prototype defimnitions so they can be loaded individually, but osx seems to just have them available as linked symbols.

maddanio:
Hmm, thats weird. Just looked at the main cmake file, and I dont see any linking to python from vegastrike-engine. Why did this work before? I would expect some kind of lin king against boost_python and/or python itself. and aren't there also lots of other dependencies like png etc? I only see this link line
`    ADD_DEPENDENCIES(vegastrike-engine OpenGL::GL OpenGL::GLU)`
and conditionally
`TARGET_LINK_LIBRARIES(vegastrike-engine m)`

maddanio:
stupid me, it's all in TST_LIBS, was confused by the name...

stephengtuggy:
You got all the Continuous Integration checks to pass this time! I'm impressed.

engine/src/gldrv/gl_globals.h
```diff
@@ -39,6 +39,16 @@ const static bool GFX_BUFFER_MAP_UNMAP = false;
...
 
+#ifndef APIENTRY
+#define APIENTRY
+#endif
+#ifndef APIENTRYP
+#define APIENTRYP APIENTRY *
+#endif
+#ifndef GLAPI
+#define GLAPI extern
+#endif
```
stephengtuggy, line 50: I'm curious where you got this block of code from.

engine/src/gldrv/gl_globals.h
```diff
@@ -147,6 +157,198 @@ struct GFXStats
 #define GL_MAX_CUBE_MAP_TEXTURE_SIZE_EXT 0x851C
 #endif
 
+typedef void (APIENTRYP PFNGLBINDBUFFERARBPROC) (GLenum target, GLuint buffer);
```
stephengtuggy, line 160: And this one. Is this standard practice with OpenGL on macOS?

BenjamenMeyer (commented):
I'm all for this kind of work as long as it's in our supported tool chain versions.
@maddanio as @stephengtuggy  said - thank you for the contribution.

engine/src/gldrv/gl_globals.h
maddanio: oh, that wasnt supposed to be in there. that was trying to get gl to work on my system. still curious why osx works on ci at all

stephengtuggy:
> I'm all for this kind of work as long as it's in our supported tool chain versions. @maddanio as @stephengtuggy said - thank you for the contribution.

@BenjamenMeyer I thought we were still using CMake 3.10/3.11 in some places, but it turns out we're not. On Ubuntu 18.04 "bionic", we would be using CMake 3.10, except that we're installing it using pip now, as of 5 months ago. See https://github.com/vegastrike/build-system-docker-images/blob/0d40650bea3d2653403804a33307f4e95e1d04dc/script/bootstrap#L290

The next oldest version of CMake we're using is 3.13, on Debian 10 "buster". (See https://pkgs.org/download/cmake for reference.) So, in practice, using CMake 3.12 features works fine.

Shall we go ahead and bump up our official min supported version to cmake 3.12 or 3.13, in the spirit of moving things forward and encouraging participation from new contributors?

BenjamenMeyer:
> > I'm all for this kind of work as long as it's in our supported tool chain versions. @maddanio as @stephengtuggy said - thank you for the contribution.
> 
> @BenjamenMeyer I thought we were still using CMake 3.10/3.11 in some places, but it turns out we're not. On Ubuntu 18.04 "bionic", we would be using CMake 3.10, except that we're installing it using pip now, as of 5 months ago. See https://github.com/vegastrike/build-system-docker-images/blob/0d40650bea3d2653403804a33307f4e95e1d04dc/script/bootstrap#L290
> 
> The next oldest version of CMake we're using is 3.13, on Debian 10 "buster". (See https://pkgs.org/download/cmake for reference.) So, in practice, using CMake 3.12 features works fine.
> 
> Shall we go ahead and bump up our official min supported version to cmake 3.12 or 3.13, in the spirit of moving things forward and encouraging participation from new contributors?

@stephengtuggy  can you add which platforms use which CMake to https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Platform-Support? I'd say just adding `CMake <version>` next to the name is fine...though we need to differentiate between native and `pip` installed support too - maybe `CMake <version>` and `pip/CMake <version>`?

stephengtuggy:
@BenjamenMeyer Sure. I've gone ahead and added the CMake version next to each Linux distro on the Platform Support wiki page. I'm pretty sure that the version of CMake available on macOS will be the latest, so probably 3.22; and likewise on Windows 10.

stephengtuggy (approved):
LGTM!

BenjamenMeyer (approved):

BenjamenMeyer:
@stephengtuggy thanks - we can go ahead and update to a newer minimum version - looks like 3.13 is the oldest required right now. @maddanio sorry for the extra discussion here.

maddanio:
Easy, this was meant to instill discussion. I wasn't sure about "the right way"

maddanio:
Cool i could contribute:)

PR #655, opened 2022-02-27
Task cockpit

royfalk:
This is a refactoring of the cockpit code. It merges cockpit_generic with cockpit.
It's more of a first and second pass than a full refactoring, as we're talking about a lot of code.

Code Changes:
[in progress] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Capital ships can't dock in space stations. When leaving, they can overlap, resulting in death. Probably not related to PR. Open bug.
- Dock rectangles showing as lock shapes instead.
- ~~Nav cross seemed warped. Need to reproduce and fix.~~
- Second ship in the fleet will not display cockpit.


engine/src/gfx/cockpit.cpp
```diff
@@ -102,749 +106,80 @@ using VSFileSystem::SoundFile;
...
+    string key( string( "visited_" )+systemname );
+    vector< float > *v = &cp->savegame->getMissionData( key );
+    if ( v->empty() )
+        v->push_back( num );
+    else if ( (*v)[0] != 1.0 && num == 1 )
+        (*v)[0] = num;
 }
 
-soundContainer::~soundContainer()
+#define sqr( x ) ( ( (x) )*( (x) ) )
```
BenjamenMeyer, line 129: Probably should use `pow(x, 2)` (https://www.cplusplus.com/reference/cmath/pow/) instead

BenjamenMeyer (changes requested):
one minor change; you'll also want to update against @stephengtuggy 's #648 once that's merged

royfalk:
> 

@stephengtuggy it looks like your changes were made by an automated tool. My PR merged cockpit_generic into cockpit and yours reformatted all of cockpit. Trying to merge this is an exercise in futility. I hate to ask you to repeat this and no idea how much of an effort it is to run the tool again (hopefully this is automatic).

Can you talk here or in gitter about how difficult is it to redo clion so we can figure out what to do? As you can see from the merge, I've been working for the last month or two on cockpit code and unaware of the progress you've made.

stephengtuggy:
@royfalk Yes, sorry about that. My changes were indeed made by an automated tool, and I am happy to say that it will be relatively easy to rerun the tool to make the same formatting changes. Why don't we just use your version of each of the conflicting files. Then we can redo the CLion stuff on top of that.

stephengtuggy:
@royfalk I think I have successfully merged the conflicting files above, just using your version of each. Do you mind if I push that merge commit to your branch, and you can look it over?

stephengtuggy:
OK, merged files are here: https://github.com/stephengtuggy/Vega-Strike-Engine-Source/tree/task_cockpit

engine/src/gfx/cockpit_gfx_utils.cpp
```diff
@@ -0,0 +1,255 @@
...
+    static const float lock_line = GameConfig::GetVariable( "graphics", "hud", "LockConfirmLineLength", 1.5 );
+    static const float diamond_size = GameConfig::GetVariable( "graphics", "hud", "DiamondSize", 2.05 );
+    static const float theta_speed = GameConfig::GetVariable( "graphics", "hud", "DiamondRotationSpeed", 1.0 );
+
+    const float max = diamond_size * r_size * 0.75 * end_reticle;
+    const float coord = end_reticle+(start_reticle-end_reticle)*lock_percent;
+    const double rtot = 1./sqrtf( 2 );
+
+    //this causes the rotation!
+    const float theta = 4*M_PI*lock_percent*theta_speed;
```
stephengtuggy, line 181: I'm having issues with `M_PI` not being defined, at least on Windows. The fix appears to be to add the following 3 lines at the top of the file:

```c++11
#undef __STRICT_ANSI__
#define _USE_MATH_DEFINES
#include <cmath>
```

stephengtuggy (changes requested):
I found one small issue when building on Windows. It looks like it will be an easy fix.

stephengtuggy:
Also please note: On Windows, this code crashes on startup, at this line: https://github.com/royfalk/Vega-Strike-Engine-Source/blob/684539b80e11a7ae452f7518cbcce4d5cf83a05c/engine/src/configuration/game_config.cpp#L29 

That will need to be fixed before merging this, I think.

engine/src/gfx/cockpit_gfx_utils.cpp
BenjamenMeyer: @stephengtuggy perhaps that should be wrapped by:
```
#ifdef WIN32
...
#endif // WIN32
```
http://www.cplusplus.com/forum/beginner/105176/

Or better yet - we should move the Boost version: https://www.boost.org/doc/libs/1_78_0/libs/math/doc/html/math_toolkit/constants.html

royfalk:
Closing this and replacing it with #659. Merging cockpit_generic and cockpit turned out to be much harder than expected.

PR #659, opened 2022-04-01
Move cockpit draw code to cockpit_gfx

royfalk:
Also try to fix IDE complaints about unknown size_t and FLT_MAX by adding relevant headers. Doesn't seem to solve it at my end but leaving anyway, as it's good practice.

This replaces the much more ambitious but ultimately unsuccessful #655. I've copied some of the code from it and packaged it in this PR.

Please answer the following:

Code Changes:

Issues:
- None

Purpose:
- What is this pull request trying to do? Just general code refactoring
- What release is this for? Any

### DIFF
--- a/engine/src/gfx/cockpit.cpp
+++ b/engine/src/gfx/cockpit.cpp
@@ -86,6 +86,10 @@
 #include "planet.h"
 #include "mount_size.h"
 #include "weapon_info.h"
+#include "cockpit_gfx.h"
+
+#include <cstddef>
+#include <cfloat>
 
 using std::min;
 using std::max;
@@ -101,7 +105,6 @@ using VSFileSystem::SoundFile;
 static soundContainer disableautosound;
 static soundContainer enableautosound;
 
-#define sqr(x) ( ( (x) )*( (x) ) )
 
 std::string GameCockpit::GetNavSelectedSystem() {
     return AccessNavSystem()->getSelectedSystem();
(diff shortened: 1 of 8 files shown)


stephengtuggy:
The Windows build failed with the following two errors:

```
D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\gfx\cockpit_gfx_utils.cpp(209,27): error C2065: 'M_PI': undeclared identifier [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\gfx\cockpit_gfx_utils.cpp(209,17): error C2737: 'theta': const object must be initialized [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\vegastrike-engine.vcxproj]
```

We'll need to fix those before merging this.

stephengtuggy:
That latest change fixed the Windows build! Sweet.

engine/src/gfx/cockpit.cpp
```diff
@@ -2618,9 +1664,9 @@ void GameCockpit::Draw() {
                 headtrans.back().p.j = shake_mag * shakin * cos(1.3731 * theta) * cockpitradial
                         / 100;  //AccessCamera()->GetPosition().j+shakin*cos(theta);
                 headtrans.back().p.k = 0;                 //AccessCamera()->GetPosition().k;
-                headtrans.back().p.i += warp_shake_mag * cos(wtheta) * sqr(warp_strength) * cockpitradial
+                headtrans.back().p.i += warp_shake_mag * cos(wtheta) * sqrtf(warp_strength) * cockpitradial
```
stephengtuggy, line 1667: Why `sqrtf`? Why not just `sqrt`?

engine/src/gfx/cockpit.cpp
```diff
@@ -2618,9 +1664,9 @@ void GameCockpit::Draw() {
...
                         / 100;           //AccessCamera()->GetPosition().i+shakin*cos(theta);
-                headtrans.back().p.j += warp_shake_mag * cos(1.165864 * wtheta) * sqr(warp_strength) * cockpitradial
+                headtrans.back().p.j += warp_shake_mag * cos(1.165864 * wtheta) * sqrtf(warp_strength) * cockpitradial
```
stephengtuggy, line 1669: Same on this line

stephengtuggy (commented):
I just had a question about a couple of lines. I'm not sure if I'll block for this, but it would be good to at least discuss it briefly

engine/src/gfx/cockpit.cpp
royfalk: It was originally a macro here in the previous MR and I replaced it with a standard function. I usually use std::sqrt in such places. For some reason, the IDE didn't like either that or sqrt and I didn't spend much time on it. 
It's possible the reason for it was that warp_strength is a float. Most variables in VS are floats for reasons of performance. I'm not sure if this is still a valid reason in today's architectures, but converting them en masse is asking for trouble.

engine/src/gfx/cockpit.cpp
royfalk: Same as above.

engine/src/gfx/cockpit.cpp
stephengtuggy: OK, we can leave these as is then.

stephengtuggy (approved):

PR #683, opened 2022-05-17
Refactor: Static Game Config Part B, Remix

stephengtuggy:

Please answer the following:

Code Changes:

Issues:

Purpose:
- What is this pull request trying to do? Convert over the majority of the game configuration settings to use the new configuration()->… approach. Also address a few other things in the process, such as refactoring the CMake stuff to use [CMake generator expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html?highlight=generator%20expressions#id4), and refactoring the vs_logging stuff to use a Meyers singleton as well.
- What release is this for? 0.9.x, presumably
- Is there a project or milestone we should apply this to? 0.9.x, presumably


stephengtuggy:
Known Issues: The game crashed just now, after I left my ship on autopilot for about 20 minutes. It hit a vector-out-of-bounds exception on [this line](https://github.com/stephengtuggy/Vega-Strike-Engine-Source/blob/fa144c9184c71f2741a9e2cc286d1d033f520ab5/engine/src/gfx/mesh_gfx.cpp#L442). Hopefully I will have this fixed within another couple of PRs or so. At least we're catching it now, rather than blithely going on into unallocated memory or something of the sort.

Another known issue: The "speed limit" is back to 9.87 C. I need to double-check to verify this, but I think this is the game engine doing what it's supposed to do. If we want to change this setting, I think we'll want to look at the assets or vegastrike.config. But I'm not 100% sure. I'll get back to you guys on that.

engine/CMakeLists.txt
```diff
@@ -208,14 +208,14 @@ IF(NOT POW_FUNCTION_EXISTS AND NOT NEED_LINKING_AGAINST_LIBM)
...
-# Set the default build type
-IF (NOT CMAKE_BUILD_TYPE)
-    SET(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Release, RelWithDebInfo, Debug, Profiler" FORCE )
-ENDIF (NOT CMAKE_BUILD_TYPE)
-
-IF (NOT BUILD_OPT)
-    SET(BUILD_OPT -O2 CACHE STRING "-O0, -O1, -O2, -O3, -Os, -Ofast" FORCE )
-ENDIF (NOT BUILD_OPT)
+## Set the default build type
+#IF (NOT CMAKE_BUILD_TYPE)
```
BenjamenMeyer, line 212: Assume these changes are temporary while you work on the branch

BenjamenMeyer (commented):
looking good; I like where this is going

engine/CMakeLists.txt
stephengtuggy: @BenjamenMeyer I was actually planning to make these changes permanent. As it stands currently, the default value for `BUILD_OPT` overrides the proper compiler optimization level for the current Debug, Release, etc. configuration -- and it does so without any intervention from the user of CMake. Not ideal.

I'm also not sure that we should be setting a default build type. Especially with CMake generators that cover multiple build types at once, like the Visual Studio one. However, we can leave that discussion for later if you like.

BenjamenMeyer: Our CI needs a default built type, and it's useful for most environments. I'd consider VS to be the abnormal case.

Our workflow should simply be:
```
$ mkdir build
$ cd build
$ cmake ../engine
$ make
```
And that should get you a binary. The dev shouldn't have to care about the default build type.

BTW, IIRC cmake defaults to the release type, which doesn't give us any symbols which is why we were defaulting to the RelWithDebugInfo type and overriding it.

stephengtuggy: @BenjamenMeyer Okay. I'll leave `BUILD_OPT` out, but put the default `CMAKE_BUILD_TYPE` back in.

stephengtuggy: Done.

stephengtuggy:
Closing in favor of #687 , which is the continuation of this PR, but with the "speed limit" issue fixed (among other things).

PR #686, opened 2022-05-22
Make game load units from JSON instead of CSV.

royfalk:
This affects main game and not save game (player ships).
This goes together with a change to asset-production that will include a unit.json file and conversion script.
This must be paired with https://github.com/vegastrike/Assets-Production/pull/59.

Code Changes:

Issues:
- None identified.


engine/src/cmd/json.h
```diff
@@ -0,0 +1,850 @@
...
+#include <cstdlib>
+#include <string>
+#include <vector>
+#include <cstdio>
+#include <utility>
+#include <stdexcept>
+#include <cctype>
+
+/*! \brief Base namespace for simpleson */
+namespace json
```
BenjamenMeyer, line 17: Is there a reason we're implementing our own JSON parser?
Can't we just use Boost's JSON Parser (https://www.boost.org/doc/libs/1_79_0/boost/property_tree/json_parser.hpp)?

BenjamenMeyer (changes requested):
Looks good; though I'd prefer to use an existing C/C++ JSON parser than implement our own.

engine/src/cmd/json.h
stephengtuggy: @royfalk are you using this library? https://github.com/gregjesl/simpleson

BenjamenMeyer: That may make sense...but then we need to document it; however, that Boost::Json_Parser is more actively maintained and used by a larger audience, so less likely to have issues and we wouldn't have to maintain/sync it.

stephengtuggy: I'm fine with using simpleson, personally. The code is already written to use that library. Why bother changing it? Also, I seem to recall reading that Boost's JSON parser leaves something to be desired.

We do need to figure out why the CI is failing, though.

engine/src/cmd/tests/json_tests.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <gtest/gtest.h>
+
+#include "unit_json_factory.h"
+#include "unit_csv_factory.h"
+
+#include <iostream>
+#include <fstream>
+#include <iterator>
+
+#include <unistd.h>
```
stephengtuggy, line 35: This header is not available on Windows systems. This is what is causing the Windows CI build to break, with the following error message:

```
D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\tests\json_tests.cpp(35,10): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory [D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\build\Vega_Strike_tests.vcxproj]
```

stephengtuggy:
On macOS, looks like the failure is that the JSON Sanity test fails with the following message:

```
C++ exception with description "Input was not an array" thrown in the test body.
```

royfalk:
Interesting. The build fails because the unit test is unable to find the units.json file. However, I already merged the complementing assets PR. Does anyone have an idea why it's not picking up the file?

Loki1950 (approved):

stephengtuggy:
Aw, man! The tests failed again. Perhaps the only thing missing is to copy the JSON file in CMakeLists.txt, as is done with another test file or two already.

engine/src/cmd/json.h
stephengtuggy: @BenjamenMeyer Issue filed as #686

stephengtuggy (approved):
LGTM!

Loki1950 (approved):

engine/src/cmd/json.h
royfalk: @BenjamenMeyer you will need to approve the PR, or I will not be able to merge.

BenjamenMeyer: @stephengtuggy thanks! @royfalk will approve

BenjamenMeyer:
@royfalk thanks - my last hold was waiting for the issue that @stephengtuggy filed; so all good now! Thanks!

PR #687, opened 2022-05-26
Config refactor - Speed limit troubleshooting

### LINKED ISSUE #684: "Speed Limit" (warp max effective velocity) is not loaded correctly from vegastrike.config
Another known issue: The "speed limit" is back to 9.87 C. I need to double-check to verify this, but I think this is the game engine doing what it's supposed to do. If we want to change this setting, I think we'll want to look at the assets or vegastrike.config. But I'm not 100% sure. I'll get back to you guys on that.

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/683#issuecomment-1128397673_

Update: As it turns out, the new code is *not* correct. It does not replicate the functionality of `domNodeType *LoadXML(const char *filename)` in easydom.h, which searches an entire series of file paths to find the config file it's looking for. We will need to have it replicate that functionality in order to get all the settings right.

stephengtuggy:

Please answer the following:

Code Changes:
- [Mostly] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Closes #684 

Purpose:
- What is this pull request trying to do? Continue refactoring game settings, while fixing "speed limit" issue


engine/src/cmd/ai/aggressive.cpp
```diff
@@ -769,14 +761,8 @@ bool AggressiveAI::ProcessCurrentFgDirective(Flightgroup *fg) {
...
-                            static float esc_percent = XMLSupport::parse_float(vs_config->getVariable("AI",
-                                    "Targetting",
-                                    "EscortDistance",
-                                    "10.0"));
-                            static float turn_leader = XMLSupport::parse_float(vs_config->getVariable("AI",
-                                    "Targetting",
-                                    "TurnLeaderDist",
-                                    "5.0"));
+                            const float esc_percent = configuration()->ai.escort_distance;
+                            const float turn_leader = configuration()->ai.turn_leader_distance;
```
BenjamenMeyer, line 765: so much cleaner!!!! :)

engine/src/configuration/configuration.h
```diff
@@ -372,10 +435,12 @@ class Configuration {
     vega_config::AudioConfig audio_config_;
     vega_config::CollisionHacks collision_hacks;
     vega_config::ComputerConfig computer_config_;
+    vega_config::EjectConfig eject_config_;
```
BenjamenMeyer, line 438: I'd be in favor of dropping the `_` at the end....it'd make it a little easier to read

BenjamenMeyer (commented):
Looking good - love how this is cleaning up the config options and reducing the code.
One minor comment...but otherwise looking good IMHO.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: I know, right? :-)

engine/src/configuration/configuration.h
stephengtuggy: Done.

stephengtuggy:
Known issue: #695 . On Linux, the aera and confed ships aren't fighting each other anymore. At least, not very much.

I don't know what specifically is causing this issue. For now, I think I'm just going to leave it. This PR has gotten big enough as it is.

engine/src/cmd/planet.cpp
```diff
@@ -317,30 +314,28 @@ Planet::Planet(QVector x,
...
             }
-            static vector<string>
+
+            // TODO: Refactor the below section of code
+            vector<string>
                     shines = ParseDestinations(vs_config->getVariable("graphics", "star_shine", "shine.ani"));
             if (shines.empty()) {
-                shines.push_back("shine.ani");
+                shines.emplace_back("shine.ani");
             }
```
BenjamenMeyer, line 337: would this ever be true? shouldn't the new config stuff make sure that `shine.ani` is provided as a default when read from the configuration?

engine/src/vs_logging.cpp
```diff
@@ -117,4 +102,77 @@ void VegaStrikeLogger::FlushLogs() {
...
+
+void VegaStrikeLogger::Log(const vega_log_level level, const char *message) {
+    boost::log::record rec = slg_->open_record(boost::log::keywords::severity = level);
+    if (rec)
+    {
+        boost::log::record_ostream strm(rec);
+        strm << message;
+        strm.flush();
+        slg_->push_record(boost::move(rec));
+    }
```
BenjamenMeyer, line 149: nice - so much better than us handling it directly.

engine/src/vs_logging.h
```diff
@@ -25,51 +26,66 @@
...
+    void FlushLogs();
+    void Log(const vega_log_level level, const std::string& message);
+    void Log(const vega_log_level level, const char * message);
+    void Log(const vega_log_level level, const boost::basic_format<char>& message);
+    void LogAndFlush(const vega_log_level level, const std::string& message);
+    void LogAndFlush(const vega_log_level level, const char * message);
+    void LogAndFlush(const vega_log_level level, const boost::basic_format<char>& message);
 };
 
+extern boost::shared_ptr<VegaStrikeLogger> vega_logger();
```
BenjamenMeyer, line 87: awesome - just so long as it's thread safe we should be good

engine/src/options.h
```diff
@@ -41,52 +41,39 @@ class vs_options {
...
-    int garbagecollectfrequency;
-    uint numoldsystems;
-    bool deleteoldsystems;
-    char vsdebug;
-    bool while_loading_starsystem;
     std::string default_mission;
     std::string galaxy;
-    bool command_interpretor;
-    bool load_last_savegame;
+    bool command_interpretor{};
```
BenjamenMeyer, line 46: okay - so I know I gave @royfalk some grief previously about using this format vs a more normal legacy method, and I'm going to take the moment to say that at the time that was the right call; but now I think it's a good time to start using this format more. We've had a couple releases done and established and it's slowly becoming more of our - your - own code.

I'd also like to call out that this will probably be the first of many rounds like this. Overall we'll keep changes this way a bit more conservative and as you all get us older folk use to reading it we'll start allowing more and more of it in as it makes sense.

@royfalk I haven't forgotten ;) just want to call out the above and not let it seem like I'm letting is slide under the radar. I've noticed.

BenjamenMeyer (approved):
looking good; love the direction this is going. With the changes I would imagine we'd see a slight boost in performance.

engine/src/cmd/planet.cpp
stephengtuggy: Yeah, this would probably only be true if there was a setting present in the config file that specifically set this to an empty/null value. I figure it's better to be safe than sorry.

engine/src/vs_logging.cpp
stephengtuggy: This is taken from the Boost.Log documentation, incidentally. I basically copied the code for one of their macros, albeit with one or two adjustments.

engine/src/options.h
stephengtuggy: I'll be happy to be able to use this format more going forward. ;-)

stephengtuggy:
Darn it. Once again, I forgot to squash these commits when I merged. Oh well, too late now.

PR #691, opened 2022-06-03
Convert weapon_list.xml to weapons.json

royfalk:
Engine code to support the switch.
Note that this change and the previous one to units will break Wing Commander.

Code Changes:

Issues:
- There are previous issues in the weapon_list that should be addressed at some point.
- The JSON library does not conform to the standard, making the code sloppier than I would have liked.
- Old files (units.csv and weapon_list.xml) have not been deleted yet.

### DIFF
--- a/engine/src/cmd/weapon_factory.cpp
+++ b/engine/src/cmd/weapon_factory.cpp
@@ -32,6 +32,7 @@
 #include "role_bitmask.h"
 #include "audiolib.h"
 #include "hashtable.h"
+#include "json.h"
 
 #include <iostream>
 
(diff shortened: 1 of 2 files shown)


BenjamenMeyer (approved):

stephengtuggy:
Can we be sure to increment the Assets API version to 2 in this PR, as well?

stephengtuggy:
Can we be sure to increment the Assets API version to 2 in this PR, as well?

stephengtuggy:
Other than that, it looks all right.

royfalk:
> Can we be sure to increment the Assets API version to 2 in this PR, as well?

How do I do that?

BenjamenMeyer:
@royfalk increment the value at https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/CMakeLists.txt#L79

royfalk:
As these are all part of the same "Epic", I'll increment once I'm done with it.
For anyone interested, I still need to modify the save game and the planets. Hopefully, that will get rid of legacy file formats.

BenjamenMeyer:
We can increment at any time, but it should be done sooner than later so 9.8.x assets detect that master isn't necessarily compatible any more and we can see how things play work too

PR #711, opened 2022-07-13
Convert master_part_list read code from csv to json

royfalk:
This makes carrier read the master parts list from json and not csv.

See also https://github.com/vegastrike/Assets-Production/pull/95

Code Changes:

Issues:
- None identified

### DIFF
--- a/engine/src/cmd/carrier.cpp
+++ b/engine/src/cmd/carrier.cpp
@@ -30,10 +30,11 @@
 #include "universe_util.h"
 
 #include "ai/aggressive.h"
-#include "csv.h"
 #include "missile.h"
 #include "vs_random.h"
 #include "vs_logging.h"
+
+#include "json.h"
 
 // TODO: find out where this is and maybe refactor
 extern int SelectDockPort(Unit *, Unit *parent);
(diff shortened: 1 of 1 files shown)


BenjamenMeyer (approved):

engine/src/cmd/carrier.cpp
```diff
@@ -542,29 +543,36 @@ float Carrier::getCargoVolume(void) const {
...
-            carg.volume = stof(row["volume"], 1);
-            carg.mass = stof(row["mass"], 1);
-            carg.quantity = 1;
-            carg.price = stoi(row["price"], 1);
-            carg.description = row["description"];
-            ret->cargo.push_back(carg);
-        }
-        delete table;
+
+    static std::string json_filename = "master_part_list.json"; //vs_config->getVariable("data", "master_part_list", "master_part_list");
```
stephengtuggy, line 554: I'd rather we didn't hard-code the filename if we can help it. But perhaps we can't help it, for this iteration. What do you think, @royfalk ?

BenjamenMeyer: @stephengtuggy agreed it'd be better not to have it hard-coded especially at this level - it should come in from the command-line parameters and/or configuration file where it can be overridden. As a first pass I'm okay with it - especially since the CSV was hard-coded, but we should at least mark it as something to be improved upon via the settings.

stephengtuggy: OK. Filed an issue, #712 . Approving this PR.

stephengtuggy (approved):
Looks good to me!

engine/src/cmd/carrier.cpp
royfalk: I'm not a fan of either of the two options I've used:
1. In the first one, I renamed the csv file provided by some other code to json.
2. In this one, I just hard coded it.

I'll add a TODO here and move on. At some point we'll want an asset manager to give us the files and filenames.

royfalk: Opened an issue instead, as this affects more than just this part of the code.
https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/714

Issue #716, opened 2022-07-22
Unit destructor crashes the game


royfalk:
Sometimes, when a unit is destroyed, the game crashes.

Stack Trace:
`#13 0x000055a645097aaa in std::vector<Mesh*, std::allocator<Mesh*> >::_M_range_check (
    __n=18446744073709551615, this=0x55a653939740)
    at /usr/include/c++/11/bits/hashtable_policy.h:387
#14 std::vector<Mesh*, std::allocator<Mesh*> >::at (__n=18446744073709551615, 
    this=0x55a653939740) at /usr/include/c++/11/bits/stl_vector.h:1094
#15 Mesh::~Mesh (this=0x55a65393a498, __in_chrg=<optimized out>)
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/engine/src/gfx/mesh_gfx.cpp:442
#16 0x000055a645097ce2 in Mesh::~Mesh (this=0x55a6539a6c50, __in_chrg=<optimized out>)
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/engine/src/gfx/mesh_gfx.cpp:461
#17 0x000055a645097d5f in Mesh::~Mesh (this=0x55a6539a6c50, __in_chrg=<optimized out>)
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/engine/src/gfx/mesh_gfx.cpp:407
#18 0x000055a64544481a in Unit::~Unit (this=0x55a653938670, __in_chrg=<optimized out>)
    at /usr/include/c++/11/bits/stl_vector.h:1043
#19 0x000055a645444c5f in Unit::~Unit (this=0x55a653938670, __in_chrg=<optimized out>)
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/engine/src/cmd/unit_generic.cpp:326`

stephengtuggy:
Yes, I'm pretty sure I've noticed this same crash. It's on my to-do list to fix. Hopefully sooner rather than later.

stephengtuggy:
Well, #754 had to be rolled back in its entirety. So now this crash is back. Argh.

BenjamenMeyer:
are we still seeing this? If so we probably need to fix for the 0.9.x release.

stephengtuggy:
> are we still seeing this? If so we probably need to fix for the 0.9.x release.

[...]

PR #717, opened 2022-07-22
Draft: Remove unused code related to csv

royfalk:
The file VegaPersonalities.csv isn't used.

@ministerofinformation / @Loki1950 - please let me know if you think this file should have existed and used recently. 

Code Changes:

Issues:
- Game crashed twice. Opened https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/716 for second crash.

Purpose:
- What is this pull request trying to do? Remove CSV code by either converting to JSON or deleting if not used.

### DIFF
--- a/engine/src/cmd/ai/aggressive.cpp
+++ b/engine/src/cmd/ai/aggressive.cpp
@@ -109,31 +109,6 @@ static void TurretFAW(Unit *parent) {
     }
 }
 
-static vsUMap<string, string> getAITypes() {
-    vsUMap<string, string> ret;
-    VSFileSystem::VSFile f;
-    VSFileSystem::VSError err = f.OpenReadOnly("VegaPersonalities.csv", VSFileSystem::AiFile);
-    if (err <= VSFileSystem::Ok) {
-        CSVTable table(f, f.GetRoot());
-        vsUMap<std::string, int>::iterator browser = table.rows.begin();
-        for (; browser != table.rows.end(); ++browser) {
-            string rowname = (*browser).first;
-            CSVRow row(&table, rowname);
-            for (unsigned int i = 1; i < table.key.size(); ++i) {
-                string hasher = rowname;
-                if (i != 1) {
-                    hasher = rowname + "%" + table.key[i];
-                }
-                string rawrow = row[i];
-                if (rawrow.length() > 0) {
-                    ret[hasher] = rawrow;
-                }
-            }
-        }
-        f.Close();
-    }
-    return ret;
-}
 
 static string select_from_space_list(string inp, unsigned int seed) {
     if (inp.length() == 0) {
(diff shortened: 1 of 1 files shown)


Loki1950 (commented):
not sure about the personality file but it may be related to the story line

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -173,7 +148,7 @@ static AIEvents::ElemAttrMap *getLogicOrInterrupt(string name,
         vsUMap<string, AIEvents::ElemAttrMap *> &mymap,
         int personalityseed) {
     string append = "agg";
-    static vsUMap<string, string> myappend = getAITypes();
+    static vsUMap<string, string> myappend;
```
BenjamenMeyer, line 151: how will this get populated now?

stephengtuggy (changes requested):
I would rather not make this change at this time. This seems like one of those "fences" that we don't fully understand the purpose of yet. Once we do, we can come back and revisit.

engine/src/cmd/ai/aggressive.cpp
royfalk: If you look at line 113-115, this is what's happening anyway. The file 'VegaPersonalities' doesn't exist in either VS or WCPU. I've trimmed the code but nothing's changed. At least, that was the intention.
I specifically asked @Loki1950, as he's our resident historian for such things. 
As I can't figure out from the code what they wanted to do and this appears to be a stub for some future feature, I vote to delete it. If/When we get to developing, we'll hopefully document it as well :)

royfalk:
> 

I absolutely don't understand the purpose of this. However, it also isn't active yet and never will be. The only two ways to figure this out are to ask one of the original devs and hope he remembers after a decade or meditate on the entire AI code and hope it becomes clearer then.
As noted above, this code doesn't do anything if the file is missing.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: I just checked. The file 'VegaPersonalities.csv' does, in fact, exist in PWCU. Three copies of it. One under the 'ai' folder, another under the 'ai.easy' folder, and the third under the 'ai.hard' folder.

royfalk: Huh. I actually checked in what I assumed was the repository. Is it something you're running on your computer or is it in the repo as well? If it is, can you point me to it?

stephengtuggy: https://github.com/pwcu/privateer_wcu/blob/master/ai.easy/VegaPersonalities.csv

Loki1950:
I have a possible reason for it's existence as @stephengtuggy has mentioned it holds the AI profiles of NPC's in the story line since Upon a the Coldest Sea is just a stub ATM that file is not needed yet as it is in Privateer which does have a full story.

royfalk:
Based on the example in the link above, I'll start working on converting it instead of simply deleting the whole function.

royfalk:
OK. So I looked at the PU data and tried to make it work for VS and it crashed nicely. Looks like I'm going to have to run PU and debug that.

royfalk:
I'm going to close this. The code is needed and I may as well start from scratch. There's nothing here to salvage.

PR #722, opened 2022-07-28
Switch to SDL2

royfalk:
**For testing purposes only.** 

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation No. Minimal game play only.

Issues:
- Can't run the game in non-legacy mode.  Not sure what changed though.


engine/src/main.cpp
```diff
@@ -247,7 +247,7 @@ int main(int argc, char *argv[]) {
 
     // when the program name is `vegastrike-engine` then enforce that the data directory must be specified
     // if the program name is `vegastrike` then enable legacy mode where the current path is assumed.
-    legacy_data_dir_mode = (program_name == "vegastrike") || (program_name == "vegastrike.exe");
+    legacy_data_dir_mode = true; //(program_name == "vegastrike") || (program_name == "vegastrike.exe");
```
BenjamenMeyer, line 250: this shouldn't be in here

BenjamenMeyer (changes requested):

BenjamenMeyer:
Is SDL1 and SDL2 that compatible that all we need to do is change the library to start the conversion? If so, awesome!

engine/src/main.cpp
royfalk: You're right. However, I noted this and I couldn't make the game run without it.

What I want from you is three things:
1. Confirm it is using SDL2. I'm not a cmake guy and conversion went too smooth for me to prove it is SDL2.
2. Figure out why non-legacy doesn't work.
3. Test it too if you have the time.

BenjamenMeyer: I'll try to take a look sometime soon; I'm quite glad to move us to SDL2

royfalk:
I think the reason they are so compatible is that we're doing very little with SDL. I think the original devs even considered it an optional component. Given the capabilities, we should probably look what else we can use from it can replace custom solutions.

Loki1950:
Your right about how little we use SDL 1 it was a very new library at that time ;) so still not mature as it is now SDL 2 handles monitors much better than it's earlier incarnation with dpi and multi-monitor control more fine grained.

engine/src/main.cpp
royfalk: _Confirm it is using SDL2._ 

It wasn't. :) But I'm back to working on it.

BenjamenMeyer: So legacy mode was simply for whether it would automatically try to find the assets using the old algorithm (legacy mode) or require the value to be specified on the command line (new behavior).

For simplicity legacy mode gets enabled if you call the original executable name (vegastrike), and the new mode is enabled if you are calling our new executable name (vegastrike-engine) in line with our build target naming conventions.

I'll pull latest master and check it again, and perhaps check against your branch too.

BenjamenMeyer: Okay - checked master on Linux (can't speak to Windows or Mac).

`legacy_data_dir_mode` is tied to the paths from the `InitDataDirectory()` (see https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/vsfilesystem.cpp#L516).

To test:
```
$ mkdir build
$ cd build
$ cmake ../engine
...
$ make -j 10
$ ln -s vegastrike-engine vegastrike
```
Out the gate without any other setup this works as expected - both `vegastrike` and `vegastrike-engine` fail to find the assets unless the `-d` parameter were specified.

However, once I set the assets to be in an expected location (like doing a symlink of my Assets Production to `/usr/share/vegastrike` per https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/vsfilesystem.cpp#L535) then `vegastrike` works without any parameters specified.

I then tried your branch, and there seems to be something that messes up the command-line parsing as it won't read the value off the command-line. It does find it using legacy mode just fine though.

I verified this by using the PWCU assets on the command-line via the `-d` parameter while `/usr/share/vegastrike` pointed to my VS:UtCS assets, and had the following result:
- `vegastrike-engine` failed as it could not find assets even though the `-d` parameter was specified
- `vegastrike` failed if I specified the `-d` parameter
- `vegastrike` worked if I did not use the `-d` parameter which would cause it to then load the assets from `/usr/share/vegastrike`.

So I think we need to look at what happened in this branch regarding the processing of the `datadir` configuration  and its relationship to the command-line processing.

BenjamenMeyer: Oh - I also discovered that master is also broken with respect to reading the `datadir` variable out of the configuration file, or at least I don't know how to get it in there properly.

BenjamenMeyer: @royfalk how are you trying to start the application? for my testing I was using:
```
$ ./vegastrike
$ ./vegastrike -d/path/to/assets/without/leading/space
$ ./vegastrike-engine -d/path/to/assets/without/leading/space
```
All worked for me with both master and your branch.

royfalk:
Closing this. It wasn't actually using SDL2. It was simply not using SDL1.
Another PR will add SDL2.

PR #732, opened 2022-08-23
Add code for store and customer

royfalk:
- Fix bug in resource
- Add tests for resource and store

I have added unit tests instead of the standard tests as the code has not yet been integrated with engine.
I am waiting for @stephengtuggy 's changes, as he's touching the same areas of the code.

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation No. 

Issues:
- None

Purpose:
- What is this pull request trying to do? Make the code cleaner.
- What release is this for?

### DIFF
--- a/engine/src/resource/store.cpp
+++ b/engine/src/resource/store.cpp
@@ -1,0 +1,131 @@
+/*
+ * Copyright (C) 2001-2022 Daniel Horn, pyramid3d, Stephen G. Tuggy,
+ * and other Vega Strike contributors.
+ *
+ * https://github.com/vegastrike/Vega-Strike-Engine-Source
+ *
+ * This file is part of Vega Strike.
+ *
+ * Vega Strike is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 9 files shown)


engine/src/resource/store.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return -1;
+}
+
+
+
+bool Store::Sell(Customer& buyer, std::string product_name, double quantity)
+{
+    buyer.Buy(*this, product_name, quantity);
```
BenjamenMeyer, line 119: missing return value (thus the GH check failures)

BenjamenMeyer (changes requested):
I like the direction this is going

engine/src/resource/store.cpp
royfalk: Yeah, I noticed that. Weird though that **_my_** IDE and compiler were fine with it.

royfalk: Are you good with my changes? I need an approval please.

BenjamenMeyer (approved):

stephengtuggy (approved):

PR #736, opened 2022-09-30
Remove old unit csv code and fully replace with modern json/csv support

royfalk:
This PR introduces full support for both JSON and CSV unit files. 
At present, save games will use CSV.
Unless a release is planned soon, the coming one. 

Code Changes:

Issues:
- When died and respawned, ship was missing all modules and texture. Radar range was either random or max int. Non-reproducible.
- When landing on a planet, ship was incorrectly drawn. Non-reproducible.
- Multiple issues related to legacy code and correctness of assets (to be fixed there).

### DIFF
--- a/engine/src/cmd/ai/aggressive.cpp
+++ b/engine/src/cmd/ai/aggressive.cpp
@@ -109,31 +109,6 @@ static void TurretFAW(Unit *parent) {
     }
 }
 
-static vsUMap<string, string> getAITypes() {
-    vsUMap<string, string> ret;
-    VSFileSystem::VSFile f;
-    VSFileSystem::VSError err = f.OpenReadOnly("VegaPersonalities.csv", VSFileSystem::AiFile);
-    if (err <= VSFileSystem::Ok) {
-        CSVTable table(f, f.GetRoot());
-        vsUMap<std::string, int>::iterator browser = table.rows.begin();
-        for (; browser != table.rows.end(); ++browser) {
-            string rowname = (*browser).first;
-            CSVRow row(&table, rowname);
-            for (unsigned int i = 1; i < table.key.size(); ++i) {
-                string hasher = rowname;
-                if (i != 1) {
-                    hasher = rowname + "%" + table.key[i];
-                }
-                string rawrow = row[i];
-                if (rawrow.length() > 0) {
-                    ret[hasher] = rawrow;
-                }
-            }
-        }
-        f.Close();
-    }
-    return ret;
-}
 
 static string select_from_space_list(string inp, unsigned int seed) {
     if (inp.length() == 0) {
@@ -173,7 +148,7 @@ static AIEvents::ElemAttrMap *getLogicOrInterrupt(string name,
         vsUMap<string, AIEvents::ElemAttrMap *> &mymap,
         int personalityseed) {
     string append = "agg";
-    static vsUMap<string, string> myappend = getAITypes();
+    static vsUMap<string, string> myappend;
     vsUMap<string, string>::iterator iter;
     string factionname = FactionUtil::GetFaction(faction);
     if ((iter = myappend.find(factionname + "%" + unittype)) != myappend.end()) {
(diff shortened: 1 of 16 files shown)


engine/src/cmd/unit_csv_factory.h
```diff
@@ -30,9 +30,42 @@
...
+                            "Lifesupport_Functionality",	"Max_Lifesupport_Functionality",	"Comm_Functionality",
+                            "Max_Comm_Functionality",	"FireControl_Functionality",	"Max_FireControl_Functionality",
+                            "SPECDrive_Functionality",	"Max_SPECDrive_Functionality",	"Slide_Start",	"Slide_End",
+                            "Activation_Accel",	"Activation_Speed",	"Upgrades",	"Prohibited_Upgrades",
+                            "Sub_Units",	"Sound",	"Light",	"Mounts",	"Net_Comm",	"Dock",	"Cargo_Import",	"Cargo",
+                            "Explosion",	"Num_Animation_Stages",	"Upgrade_Storage_Volume",	"Heat_Sink_Rating",
+                            "Shield_Efficiency",	"Num_Chunks",	"Chunk_0",	"Collide_Subunits",	"Spec_Interdiction",
+                            "Tractorability",
+                            // These values are not in units.csv! There are probably more but I stopped mapping.
+                            // TODO: map all missing values using the commented out code below!
```
BenjamenMeyer, line 65: we should probably file an issue for that too so we don't lose track of it

BenjamenMeyer (commented):
generally LGTM; but please file an issue for the one mentioned TODO

engine/src/cmd/unit_csv_factory.h
royfalk: https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/737

engine/src/cmd/unit_csv.cpp
```diff
@@ -120,19 +108,21 @@ void AddMeshes(std::vector<Mesh *> &xmeshes,
...
         counts->clear();
     }
+
+    // Reserve space in counts and xmeshes vectors based on number of opening braces
     {
-        int nelem = 0;
-        while ((ofs = meshes.find('{', ofs)) != string::npos) {
-            nelem++, ofs++;
-        }
+        // This code is proven to work for one mesh.
```
stephengtuggy, line 119: @royfalk can you expand on this comment a little? Has the code not been tested with more than one mesh?

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -173,7 +148,7 @@ static AIEvents::ElemAttrMap *getLogicOrInterrupt(string name,
         vsUMap<string, AIEvents::ElemAttrMap *> &mymap,
         int personalityseed) {
     string append = "agg";
-    static vsUMap<string, string> myappend = getAITypes();
```
stephengtuggy, line 176: I thought we discussed previously that the AI personality types are still needed. Is there code elsewhere that supports them? If so, I don't see it.

stephengtuggy (commented):
I have a couple of questions

engine/src/cmd/unit_csv.cpp
royfalk: If you look at the json, you'll see a mesh field like this:
"Mesh": "{turretlauncher.bfxm;;}"
Originally, they planned for multi-mesh support like this:
"Mesh": "{turretlauncher.bfxm;;}{another_mesh.bfxm;;}"

AFAIK, there isn't an example of multi-mesh for me to test, hence the comment. As for the old code, the original devs were a lot looser with the quality of what they put in. Specifically, they had a tendency to put half baked projects in. This makes it difficult to guess what some code is used for.

engine/src/cmd/ai/aggressive.cpp
royfalk: This line change should not be in this PR. I'll revert.

royfalk: Reverted.

stephengtuggy (approved):
Looks good to me. Thanks, @royfalk .

PR #739, opened 2022-10-15
Add ability to parse from ships.json, a hierarchical unit structure

royfalk:
This is a first attempt to reduce units.json file size by moving ships to another file and making it hierarchical.
The aim is to write a identical value once. For example, all variants of the same ship (e.g. llama) share the folder, name, description, etc. If you write these in the parent llama, all variants will apply them. Have a different description for llama.begin, overwrite it for that variant specifically.

Currently, implemented only for Areus (and only for Areus variant) and Llama.begin.

See also related asset [98](https://github.com/vegastrike/Assets-Production/pull/98)
Code Changes:
- [not yet] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- ~~In loaded game, shield reduce in power~~ (fixed)
- In loaded game, Oswald started as neutral or hostile (non reproducible)
- When re-spawned after dying, I was in an unidentified ship with no mesh and icepicks for weapons. Couldn't dock or die either. (non reproducible)

### DIFF
--- a/engine/src/cmd/unit_optimize_factory.h
+++ b/engine/src/cmd/unit_optimize_factory.h
@@ -1,0 +1,38 @@
+#ifndef UNITOPTIMIZEFACTORY_H
+#define UNITOPTIMIZEFACTORY_H
+
+/*
+ * unit_optimize_factory.cpp
+ *
+ * Copyright (C) 2020-2022 Daniel Horn, Roy Falk, Stephen G. Tuggy, and
+ * other Vega Strike contributors
+ *
+ * https://github.com/vegastrike/Vega-Strike-Engine-Source
+ *
+ * This file is part of Vega Strike.
+ *
+ * Vega Strike is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Vega Strike is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "vsfilesystem.h"
+
+class UnitOptimizeFactory
+{
+public:
+    static void ParseJSON(VSFileSystem::VSFile &file);
+    static void RecuriveParse(std::map<std::string, std::string> unit_attributes,
+                       const std::string& json_text, bool is_root);
+};
+
+#endif // UNITOPTIMIZEFACTORY_H
--- a/engine/src/cmd/unit_optimize_factory.h
+++ b/engine/src/cmd/unit_optimize_factory.h
@@ -26,6 +26,7 @@
  */
 
 #include <map>
+#include <string>
 
 #include "vsfilesystem.h"
 
--- a/engine/src/cmd/unit_optimize_factory.h
+++ b/engine/src/cmd/unit_optimize_factory.h
@@ -25,6 +25,8 @@
  * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+#include <map>
+
 #include "vsfilesystem.h"
 
 class UnitOptimizeFactory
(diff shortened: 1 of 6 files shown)


stephengtuggy:
With regard to the issues you noted: I've seen problems with shields not regenerating to full capacity with current `master`, as well. I think this was introduced with one of the recent PRs.

stephengtuggy:
Looks like the Windows builds are failing; we'll need to take a look at that.

stephengtuggy:
The build errors on Windows are as follows:

```
  Error: D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.h(34): error C2039: 'map': is not a member of 'std'
  C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\include\fstream(26): note: see declaration of 'std'
  Error: D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.h(34): error C2061: syntax error: identifier 'map'
  Error: D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.cpp(30): error C2511: 'void UnitOptimizeFactory::RecuriveParse(std::map<std::string,std::string,std::less<std::string>,std::allocator<std::pair<const std::string,std::string>>>,const std::string &,bool)': overloaded member function not found in 'UnitOptimizeFactory'
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.h(30): note: see declaration of 'UnitOptimizeFactory'
  Error: D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.cpp(57): error C2660: 'UnitOptimizeFactory::RecuriveParse': function does not take 3 arguments
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.h(34): note: see declaration of 'UnitOptimizeFactory::RecuriveParse'
  Error: D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.cpp(82): error C2660: 'UnitOptimizeFactory::RecuriveParse': function does not take 3 arguments
  D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\unit_optimize_factory.h(34): note: see declaration of 'UnitOptimizeFactory::RecuriveParse'
```

engine/src/cmd/unit_optimize_factory.h
```diff
@@ -0,0 +1,38 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "vsfilesystem.h"
+
```
BenjamenMeyer, line 29: missing the `#include <map>` - which would probably resolve the various errors

BenjamenMeyer (changes requested):
Missing a header include;  always good practice to have any dependent headers directly included.

royfalk:
> With regard to the issues you noted: I've seen problems with shields not regenerating to full capacity with current master, as well. I think this was introduced with one of the recent PRs.

Opened #743 to track this.

engine/src/cmd/unit_optimize_factory.h
royfalk: Done and fixed.

engine/src/cmd/unit_optimize_factory.h
```diff
@@ -0,0 +1,40 @@
...
+#include <map>
+
+#include "vsfilesystem.h"
+
+class UnitOptimizeFactory
+{
+public:
+    static void ParseJSON(VSFileSystem::VSFile &file);
+    static void RecuriveParse(std::map<std::string, std::string> unit_attributes,
+                       const std::string& json_text, bool is_root);
```
BenjamenMeyer, line 37: Probably need `#include <string>` too; would appreciate the add but not going to block per se on just this

BenjamenMeyer (approved):
Another missing head; would appreciate the add. Otherwise LGTM.
Approving so you don't need to wait further.

stephengtuggy (approved):
LGTM

Issue #710, opened 2022-07-09
String-management wrapper around `PyRun_SimpleString`?


stephengtuggy:
Perhaps we need a wrapper around `PyRun_SimpleString` to manage the `strdup`/`free` functionality?
Also, it's possible `strdup` could return a `nullptr` - unlikely but possible.

_Originally posted by @BenjamenMeyer in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/709#discussion_r913864011_

BenjamenMeyer:
Example from the commented about code above:
```
    std::string python_snippet_to_run_1("import sys\nprint(sys.path)\n");
    VS_LOG(info, (boost::format("running %1%") % python_snippet_to_run_1));
    char *temppython = strdup(python_snippet_to_run_1.c_str());
    PyRun_SimpleString(temppython);
    Python::reseterrors();
    free(temppython);
```

stephengtuggy:
I was thinking of maybe calling it something like `VegaPyRunString`, since `VS` can always potentially be mistaken to mean "Visual Studio." Other than that, I agree with the proposal.

PR #740, opened 2022-10-16
VegaPyRunString

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Implements #710 

Purpose:
- What is this pull request trying to do? DRY up the code around running Python snippets; also (hack) allow the game to load further in Debug mode on Windows, by not running the Python snippets.


Loki1950 (approved):
Looks good @stephengtuggy

engine/CMakeLists.txt
```diff
@@ -861,7 +861,7 @@ SET(LIBCMD_SOURCES
     src/cmd/weapon_info.cpp
     src/cmd/mount_size.cpp
     src/cmd/weapon_type.cpp
-)
+        src/vega_py_run.cpp src/vega_py_run.hpp)
```
royfalk, line 864: Headers aren't typically included in these lists. Instead, we have INCLUDE_DIRECTORIES at around line 157. Also, we usually use h files.

royfalk (commented):
Generally looks good, though I'm less familiar with the python part of things. One comment about the hpp to look at. I leave it in your capable hands.

BenjamenMeyer (commented):
generally looks good; aside from @royfalk 's comment about the header being compiled, let's verify `strdup` being the same on the various platforms, otherwise subtle bugs could show up.

We may want to look at changing `strdup` and other C style functions into an interface layer or finding a Boost equivalent to avoid these kinds of issues. `fsprintf` family has a ton of issues this way due to subtle and allowed changes by the POSIX standard and Microsoft's deviation from everyone else's implementation.

engine/src/vega_py_run.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include <Python.h>
+#include <pyerrors.h>
+#include <pythonrun.h>
+#include "vega_py_run.hpp"
+#include "vs_logging.h"
+
+void VegaPyRunString(const std::string &py_snippet) {
+    VS_LOG(info, (boost::format("running %1%") % py_snippet));
+    char * py_snippet_temp = strdup(py_snippet.c_str());
```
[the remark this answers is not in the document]
stephengtuggy: OK, as part of this PR, I will also implement and use the `vega_str_dup` function that we've discussed before.

engine/CMakeLists.txt
BenjamenMeyer: We're pretty much a C++ project, and others don't include our files so whether we use `.hpp`, `.hxx`, `.h`, etc doesn't really matter; we should, however, remain consistent so for the time being we should be using `.h` files.

As I noted in https://github.com/isocpp/CppCoreGuidelines/issues/1965#issuecomment-1218460543, if we need to block C or C++ we can do so through pragma directives; we shouldn't here since we're entirely C/C++ and don't compile them as pure C or pure C++.

For our purposes here, consistency is more key.

Loki1950 (approved):

stephengtuggy:
Uh oh. The macOS builds are failing. Probably because macOS is the only OS in our CI/CD where vega_str_dup can call VSExit.

Let me see if I can fix that.

Loki1950 (approved):
Seems that for every touch you have to find the hidden references all over again ;)

BenjamenMeyer (approved):

PR #748, opened 2022-11-21
Make cargo a subclass of product

royfalk:
There are some minor code improvements, but the main thing here is the separation of concerns.
A product deals with the commercial issues (buying/selling).
Cargo is carried and has mass and volume.
Upgrades can be attached to the ship and do something (future).

Please answer the following:

Code Changes:

Issues:
- Game crashed when flying a newly bought Progeny. This may to be an issue with progeny, as it doesn't have enough upgrade slots.

### DIFF
--- a/engine/src/cmd/ai/firekeyboard.cpp
+++ b/engine/src/cmd/ai/firekeyboard.cpp
@@ -1600,7 +1600,7 @@ void Arrested(Unit *parent) {
                 Cargo *ci = &parent->GetCargo(i);
                 for (unsigned int j = 0; j < contra->numCargo(); ++j) {
                     Cargo *cj = &contra->GetCargo(j);
-                    if (ci->content == cj->content) {
+                    if (ci->name == cj->name) {
                         attack = true;
                         break;
                     }
@@ -2149,7 +2149,7 @@ void FireKeyboard::Execute() {
         }
         for (; offset < static_cast<int>(parent->numCargo()); ++offset) {
             Cargo *tmp = &parent->GetCargo(offset);
-            if (tmp->GetCategory().find("upgrades") == string::npos && (missiontoo || tmp->mission == false)) {
+            if (tmp->category.find("upgrades") == string::npos && (missiontoo || tmp->mission == false)) {
                 parent->EjectCargo(offset);
                 break;
             }
--- a/engine/src/cmd/ai/firekeyboard.cpp
+++ b/engine/src/cmd/ai/firekeyboard.cpp
@@ -1600,7 +1600,7 @@ void Arrested(Unit *parent) {
                 Cargo *ci = &parent->GetCargo(i);
                 for (unsigned int j = 0; j < contra->numCargo(); ++j) {
                     Cargo *cj = &contra->GetCargo(j);
-                    if (ci->name == cj->name) {
+                    if (ci->GetName() == cj->GetName()) {
                         attack = true;
                         break;
                     }
@@ -1659,7 +1659,7 @@ void Arrested(Unit *parent) {
                 parent->PrimeOrders(new Orders::DockingOps(owner, tmp, true, true));
                 arrested_list_do_not_dereference.insert(parent);
                 for (int i = parent->numCargo() - 1; i >= 0; --i) {
-                    parent->RemoveCargo(i, parent->GetCargo((unsigned int) i).quantity, true);
+                    parent->RemoveCargo(i, parent->GetCargo((unsigned int) i).GetQuantity(), true);
                 }
                 UniverseUtil::IOmessage(
                         0,
--- a/engine/src/cmd/ai/firekeyboard.cpp
+++ b/engine/src/cmd/ai/firekeyboard.cpp
@@ -2149,7 +2149,7 @@ void FireKeyboard::Execute() {
         }
         for (; offset < static_cast<int>(parent->numCargo()); ++offset) {
             Cargo *tmp = &parent->GetCargo(offset);
-            if (tmp->category.find("upgrades") == string::npos && (missiontoo || tmp->mission == false)) {
+            if (tmp->GetCategory().find("upgrades") == string::npos && (missiontoo || tmp->GetMissionFlag() == false)) {
                 parent->EjectCargo(offset);
                 break;
             }
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -1690,7 +1690,7 @@ bool BaseComputer::scrollToItem(Picker *picker, const Cargo &item, bool select, 
     }
     //We have the parent category, now we need the child itself.
     assert(cells != NULL);
-    PickerCell *cell = cells->cellWithId(item.name);
+    PickerCell *cell = cells->cellWithId(item.GetName());
     picker->setMustRecalc();
     if (!cell) {
         //Item is not here.
@@ -1705,7 +1705,7 @@ bool BaseComputer::scrollToItem(Picker *picker, const Cargo &item, bool select, 
         //original, un-beautified name.
         int i = 0;
         for (; i < count; i++) {
-            if (item.name < cells->cellAt(i)->id()) {
+            if (item.GetName() < cells->cellAt(i)->id()) {
                 break;
             }
         }
@@ -1784,10 +1784,10 @@ void BaseComputer::configureCargoCommitControls(const Cargo &item, TransactionTy
         commitAllButton->setLabel("Buy");
         commitAllButton->setCommand("BuyAllCargo");
 
-        const int maxQuantity = maxQuantityForPlayer(item, item.quantity.Value());
+        const int maxQuantity = maxQuantityForPlayer(item, item.GetQuantity());
 
         //Total price display.
-        const double totalPrice = item.price * maxQuantity;
+        const double totalPrice = item.GetPrice() * maxQuantity;
         std::string tempString = (boost::format("Total: #b#%1$.2f#-b") % totalPrice).str();
         StaticDisplay *totalDisplay = static_cast< StaticDisplay * > ( window()->findControlById("TotalPrice"));
         assert(totalDisplay != NULL);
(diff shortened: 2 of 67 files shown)


engine/src/cmd/ai/firekeyboard.cpp
```diff
@@ -2149,7 +2149,7 @@ void FireKeyboard::Execute() {
         }
         for (; offset < static_cast<int>(parent->numCargo()); ++offset) {
             Cargo *tmp = &parent->GetCargo(offset);
-            if (tmp->GetCategory().find("upgrades") == string::npos && (missiontoo || tmp->mission == false)) {
+            if (tmp->category.find("upgrades") == string::npos && (missiontoo || tmp->mission == false)) {
```
stephengtuggy, line 2152: I think this line was better the way it was, actually. Using property getters and setters consistently will allow us to change the underlying implementation without breaking stuff.

engine/src/cmd/basecomputer.cpp
```diff
@@ -1869,8 +1869,8 @@ bool BaseComputer::configureUpgradeCommitControls(const Cargo &item, Transaction
                 //weapons can always be sold
                 for (unsigned int i = 0; i < numc; ++i) {
                     Cargo *c = &player->GetCargo(i);
-                    if (c->GetCategory().find("upgrades/") == 0 && !isWeapon(c->category)) {
-                        float po = UnitUtil::PercentOperational(player, c->content, c->category, false);
+                    if (c->category.find("upgrades/") == 0 && !isWeapon(c->category)) {
```
stephengtuggy, line 1872: I think `GetCategory()` was better here too.

engine/src/cmd/beam.cpp
```diff
@@ -576,22 +576,22 @@ bool Beam::Collide(Unit *target, Unit *firer, Unit *superunit) {
                     Cargo tmp;
                     bool isnotcargo = (c == NULL);
                     if (!isnotcargo) {
-                        if (c->GetCategory().find("upgrades") == 0) {
+                        if (c->category.find("upgrades") == 0) {
```
stephengtuggy, line 579: As above

engine/src/cmd/cargo.cpp
```diff
@@ -32,50 +32,36 @@
 Cargo::Cargo() {
     mass = 0;
     volume = 0;
-    price = 0;
-    quantity = 1;
     mission = false;
     installed = false;
-    functionality = maxfunctionality = 1.0f;
+    functionality = max_functionality = 1.0f;
```
stephengtuggy, line 37: Nice

engine/src/cmd/unit_optimize_factory.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include <map>
+
+#include "vsfilesystem.h"
+
+class UnitOptimizeFactory
+{
+public:
+    static void ParseJSON(VSFileSystem::VSFile &file);
+    static void RecuriveParse(std::map<std::string, std::string> unit_attributes,
```
stephengtuggy, line 36: Should be `RecursiveParse`, I believe

engine/src/resource/product.h
```diff
@@ -27,24 +27,15 @@
 
 #include "resource.h"
 
-class Product
+struct Product
```
stephengtuggy, line 30: I think this is better as a class, with its fields `protected` or even `private`. Prefer not to expose internals to other parts of the code except when necessary.

stephengtuggy (changes requested):
I left a few comments. In general, I like the direction this is heading. Thanks for the PR.

engine/src/cmd/ai/firekeyboard.cpp
royfalk: I followed the C++ Guidelines - https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-get 

If we need to change it in the future, we will. With the cleaner code, these changes will be limited to a few specific files. However, I honestly don't see this changing.

engine/src/resource/product.h
royfalk: I don't actually remember where I found the recommendation to use struct. I used struct as the fields are public and there is no reason right now to hide them. If it becomes necessary, we can switch.
We're paying a cost in readability here. A lot of the old code is harder to read and for no good reason. It isn't doing something fancy or complicated. It's just written that way.

engine/src/cmd/carrier.cpp
```diff
@@ -61,10 +61,10 @@ static int stoi(const string &inp, int def = 0) {
 class CatCompare {
 public:
     bool operator()(const Cargo &a, const Cargo &b) {
-        std::string::const_iterator aiter = a.GetCategory().begin();
-        std::string::const_iterator aend = a.GetCategory().end();
-        std::string::const_iterator biter = b.GetCategory().begin();
-        std::string::const_iterator bend = b.GetCategory().end();
+        std::string::const_iterator aiter = a.category.begin();
```
BenjamenMeyer, line 64: This is where the `GetCategory()` can show a difference; this would make more sense to me as `GetCategories()` where `GetCategory()` returns a single or raw value, and `GetCategories()` returns a `std::list` (or `std::deque`) of all the categories that could be encoded within the category data since the use of the iterator seems to imply we're splitting the string data up for some reason.

Similarly a `HasCategory()` could be implemented so that `c.category.find("upgrades")` would be `c.HasCategory("upgrades")` instead where `HasCategory(name)` would be implemented like:
```
categories = c.GetCategories()
if name in categories return true
else return false
```

engine/src/resource/product.h
BenjamenMeyer: I'm mixed on this one. If a structure only has data, then the `struct` is 100% the right call. If there's associated methods, then it should be a class and the data should be more hidden by default with proper getters/setters.

I do generally agree with @stephengtuggy that the data should be protected.

That said, the old code has some issues:
1. we shouldn't be friending if we can help it - just implement the functions directly or wrappers to do so, especially with operators.
2. we shouldn't friend other class objects; those should adhere to the public interface and if necessary improve the public interface.

It may be that we need to open up to general `struct` for a while in order to fix things up before closing them back down after we fix a bunch of relationships, etc - if that's the case, I'm okay with making this kind of change.

engine/src/resource/resource.cpp
```diff
@@ -143,19 +143,19 @@ void Resource<T>::Zero() {
...
-    std::cout << value_ << "\n";
-    if(max_value_ != -1) {
-        value_ = std::min(max_value_, value_);
-    }
-    std::cout << value_ << "\n";
-    value_ = std::max(min_value_, value_);
-    std::cout << value_ << "\n";
-    return *this;
-}*/
+//template<typename T>
```
BenjamenMeyer, line 146: Looks like we either should take this out or it should be `#ifdef/#endif` enabled, but most likely it needs to just go away

BenjamenMeyer (changes requested):

engine/src/cmd/ai/firekeyboard.cpp
stephengtuggy: When I attempted that big refactor back in August, I refactored StringPool as part of it. Then I found that I had to change a bunch of places in the code that referenced object properties that were StringPool references. This is where getters and setters would have come in handy.

engine/src/resource/product.h
BenjamenMeyer: NOTE: Another option if you want to keep it a `struct` is having non-associated methods - straight functional/free-functions instead of objects.

engine/src/cmd/ai/firekeyboard.cpp
royfalk: Consolidating all discussions here.

How about a "protected" class? Subclasses like upgrades will have access and I'll switch back to class. Everything else will use getters and we'll drop the friend definitions.

stephengtuggy: @royfalk Sure, sounds good. 👍🏽

engine/src/cmd/carrier.cpp
royfalk: Maybe later. Changing this and back has been expensive time-wise. I'll refactor this when I revisit the class.

engine/src/resource/product.h
royfalk: I think it makes sense to use friend functionality where it makes sense. Carrier/Cargo and Store/Product are two examples that come to mind. What do you think?

stephengtuggy: @royfalk Yeah, I think in certain cases using `friend` functionality is OK. That would probably include the cases you mentioned. At least if it turns out to be necessary.

stephengtuggy:
Looking much better. I'm trying to figure out why the CI checks failed this most recent run. It looks like a lot of them were canceled? Were they canceled manually, or was it a GitHub system issue or something?

stephengtuggy:
Also, we'll need to resolve those two conflicting files

royfalk:
I don't think the window build crashes are related to anything I did. Somehow cmake is crashing or failing and I don't see anything relevant to me.

stephengtuggy:
> I don't think the window build crashes are related to anything I did. Somehow cmake is crashing or failing and I don't see anything relevant to me.

@royfalk I think you're right. I think maybe we need to hurry up and merge https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/749 ; it might fix those Windows build errors.

Other than those, this PR is lookin' pretty good.

stephengtuggy (approved):

BenjamenMeyer (changes requested):
let's fix those overflow warnings

stephengtuggy:
#749 has been merged into `master`. I think it will solve your Windows build problems here. It seems like Microsoft must have changed something on their end recently, and we had to change our code a bit to compensate.

stephengtuggy:
All checks are now passing. Good enough for me. @BenjamenMeyer can you re-review please? Thanks

PR #749, opened 2022-11-29
Refactor gnuhash.h

stephengtuggy:

Please answer the following:

Code Changes:
- [Mostly] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:

Purpose:
- What is this pull request trying to do? Move part of the contents of gnuhash.h out of the `std` namespace, in order to follow best practices, and possibly also in order to fix some compile errors that have cropped up recently.

### DIFF
--- a/engine/src/gnuhash.h
+++ b/engine/src/gnuhash.h
@@ -20,9 +20,10 @@
  * along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
  */
 
-#ifndef _GNUHASH_H_
-#define _GNUHASH_H_
+#ifndef VEGA_STRIKE_GNUHASH_H_
+#define VEGA_STRIKE_GNUHASH_H_
 #include <unordered_map>
+#include <cstddef>
 
 #define vsUMap     std::unordered_map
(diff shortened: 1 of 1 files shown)


engine/src/gnuhash.h
```diff
@@ -66,16 +67,19 @@ struct hash<std::pair<Unit *, Unit *> > {
                 ^ ((size_t) a((int) (((size_t) key.second) >> 4))));
     }
 };
-#ifdef __GNUC__
+
+}
```
BenjamenMeyer, line 71: NOTE: We really should derive a new type from `std::hash` and add the above to it instead of extending `std::hash` to do what we want. Not sure if that's viable in this PR or not, but something to think about.

stephengtuggy: @BenjamenMeyer I agree with the concept. I'm trying to make one small change at a time though, to minimize breakage. :-)

BenjamenMeyer (approved):

PR #752, opened 2023-01-13
Refactoring of upgrade code

royfalk:

Please answer the following:

Code Changes:

Issues:
- None identified

Purpose:
- What is this pull request trying to do? Refactor upgrade code to be better and easier to read
  
- What release is this for? Any


royfalk:
Looks like the check is failing on _Unable to connect to azure.archive.ubuntu.com:http: [IP: 40.119.46.219 80]_.

stephengtuggy:
> Looks like the check is failing on _Unable to connect to azure.archive.ubuntu.com:http: [IP: 40.119.46.219 80]_.

Looks like [this issue](https://github.com/actions/runner-images/issues/675#issuecomment-1382027536)

stephengtuggy:
> > Looks like the check is failing on _Unable to connect to azure.archive.ubuntu.com:http: [IP: 40.119.46.219 80]_.
> 
> Looks like [this issue](https://github.com/actions/runner-images/issues/675#issuecomment-1382027536)

I reran the CodeQL build. We shall see if rerunning fixes the problem.

Meanwhile, the Windows builds seem to be failing with a different error:

```
D:\a\Vega-Strike-Engine-Source\Vega-Strike-Engine-Source\engine\src\cmd\upgradeable_unit.cpp(201,21): error C3861: 'strcasecmp': identifier not found
```

engine/src/cmd/upgradeable_unit.h
```diff
@@ -0,0 +1,52 @@
...
+#include <string>
+#include <vector>
+
+
+class Unit;
+class Mount;
+
+// TODO: make this into a subclass of unit later
+
+class UpgradeableUnit
```
stephengtuggy, line 39: Since we are inheriting from this class elsewhere, it needs a public virtual destructor.

engine/src/cmd/upgradeable_unit.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+
+bool UpgradeableUnit::UpgradeMounts(const Unit *up,
+                                    int mountoffset,
+                                    bool touchme,
+                                    bool downgrade,
+                                    int &numave,        // number of used parts?
+                                    double &percentage) // percentage of used parts
+{
+    Unit* unit = static_cast<Unit *>(this);
```
stephengtuggy, line 129: I would prefer to see a dynamic pointer cast here, rather than a `static_cast`. You should be able to use `vega_dynamic_cast_ptr` from "vega_cast_utils.h".

stephengtuggy (changes requested):
Looks good overall. Just a couple of changes I'd like to see.

engine/src/cmd/upgradeable_unit.cpp
royfalk: @BenjamenMeyer what do you think
Why use our own cast when we can use a well known, part of the language, one? 
Also, I've used a ton of static casts already in other parts of unit classes. If we make a decision here, I'll open an issue to change them all.

engine/src/cmd/upgradeable_unit.h
stephengtuggy: This is a general C++ guideline. "Default to having a virtual destructor if your class is meant to be used as a base class." (See <https://softwareengineering.stackexchange.com/a/284564>)

engine/src/cmd/upgradeable_unit.cpp
stephengtuggy: You can use the built-in dynamic pointer cast operator if you want. The `vega_...` one is just designed to be a convenience wrapper for the built-in one. It logs a message and aborts the program if the cast attempt fails.

engine/src/cmd/upgradeable_unit.h
stephengtuggy: Elsewhere on that question page, various commenters recommend using composition instead of inheritance. Perhaps it would be a good idea for us to investigate what that might look like in practice.

stephengtuggy:
Thanks for fixing the Windows build, by the way. It seems to be working fine now.

engine/src/cmd/upgradeable_unit.h
stephengtuggy: This looks relevant: https://stackoverflow.com/questions/49890491/a-request-for-simple-c-composition-vs-inheritance-examples

engine/src/cmd/upgradeable_unit.cpp
stephengtuggy: Much better, thanks

engine/src/cmd/upgradeable_unit.h
stephengtuggy: Perfect, thanks!

stephengtuggy (approved):
Looking good! The Windows builds are failing again, but that doesn't appear to have anything to do with this MR. Rather, one of our dependencies, `boost-tokenizer`, is failing to download.

PR #753, opened 2023-01-23
Fix bug - Store::Subtract adds quantity instead of subtracting it

### LINKED ISSUE #751: `Store::Subtract` adds `quantity` instead of subtracting it
The fix for this should include at least one unit test, to make sure the code is fixed and working properly in this regard.

royfalk:
closes #751
Also add tests and a few additional methods to store.

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation **No. Only unit tests on store. Should be fine, as store code has not been linked to main game yet.**

### DIFF
--- a/engine/src/resource/store.cpp
+++ b/engine/src/resource/store.cpp
@@ -41,7 +41,7 @@ void Store::Add(int index, int quantity) {
 }
 
 void Store::Subtract(int index, int quantity) {
-    stock[index].quantity += quantity;
+    stock[index].quantity -= quantity;
 }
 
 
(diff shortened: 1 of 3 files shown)


engine/src/resource/store.cpp
```diff
@@ -90,14 +90,32 @@ bool Store::Buy(Customer& seller, std::string product_name, double quantity)
...
         }
     }
 
     return false;
 }
 
+double Store::GetStock(std::string product_name) {
+    for(Product &in_store_product : stock) {
+        if(in_store_product == product_name ) {
+            return true;
```
BenjamenMeyer, line 104: shouldn't this be returning a number?

engine/src/resource/store.cpp
```diff
@@ -90,14 +90,32 @@ bool Store::Buy(Customer& seller, std::string product_name, double quantity)
...
+        }
+    }
+
+    return -1;
```
BenjamenMeyer, line 108: Shouldn't this be `0` or is there some logic about it being `-1` for when there's no stock? If so, we should document that here too.

BenjamenMeyer (changes requested):
Small bug - return double instead of double(boolean) for one code path.

stephengtuggy (approved):
Looks good to me. @BenjamenMeyer , can you re-review? TIA

engine/src/resource/store.cpp
royfalk: It's an easy way to return an error, as product quantity is a resource and therefore, cannot be less than 0.

BenjamenMeyer: @royfalk thanks. We probably should make a standard for this across the code-base and ensure we get it in the documentation

BenjamenMeyer (approved):

Issue #755, opened 2023-01-28
Save games do not clear out the ship folder


royfalk:
Every save game also saves the related ships in a ship folder.
It looks something like:
/home/.vegastrike/serialized_xml/<save_game_name>/<ship name>

If you overwrite the save game with another and with a different ship, the old ship file is still there.

PR #756, opened 2023-01-28
Fix save game bug

### LINKED ISSUE #747: Savegames, when loaded, are missing the player's ship
When I load a savegame that was generated using a recent build of Vega Strike, my ship is gone.

Savegames generated using previous builds do not have this problem.

royfalk:
closes #747 

Code Changes:
    _I've also ran some basic tests and ran extensive tests on the specific problem in question._

Issues:
- Overwriting save game, leaves the old ship file in the ship folder if it is not the same. #755

### DIFF
--- a/engine/src/cmd/csv.cpp
+++ b/engine/src/cmd/csv.cpp
@@ -87,6 +87,8 @@ std::vector<string> readCSV(const string &s, string delim) {
     return l;
 }
 
+// TODO: do we need this? I assume this escapes the delimiter within the
+// current csv value.
 static string addQuote(string s, string delim = ",;") {
     if (s.find_first_of(delim + "\"") != string::npos) {
         if (s.find('\"') != string::npos) {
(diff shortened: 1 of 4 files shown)


royfalk:
What's up with the windows CI fails? I doubt this is related to my PR.

stephengtuggy:
> What's up with the windows CI fails? I doubt this is related to my PR.

The Windows CI builds are failing on [line 38 of csv.h](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/task_fix_736/engine/src/cmd/csv.h#L38) , while building `vegastrike-testing`. The error is `error C2039: 'map': is not a member of 'std'`.

TBH, I'm surprised that this file is compiling at all, on any platform. It ought to `#include <map>`, but it doesn't. I'm not finding any transitive includes of the `map` header file, either -- just `<unordered_map>`, in gnuhash.h.

I think adding the `#include <map>` line to csv.h will probably fix the errors.

stephengtuggy:
Other than that, it's looking good, as far as I can see. I would like to see one of us run through the standard set of PR Validation Tests though, before we merge this.

royfalk:
> Other than that, it's looking good, as far as I can see. I would like to see one of us run through the standard set of PR Validation Tests though, before we merge this.

Ran the full tests successfully. If the windows CI passes, I think we're good to merge.

stephengtuggy (approved):

PR #754, opened 2023-01-25
Big refactor to fix crash in mesh destructor

stephengtuggy:
Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [Partially] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Fixes #716 

Purpose:
- What is this pull request trying to do? Refactor `Mesh`; `AnimatedTexture`; and related classes, to use smart pointers (among other things), for improved memory safety and fewer crashes
- What release is this for? 0.9.x, preferably
- Is there a project or milestone we should apply this to? 0.9.x

PR #765, opened 2023-02-20
Further refactoring of unit code

royfalk:
- Split master_part_list to multiple files
- Make ship into a separate super class of unit
- Move additional methods to movable

Please answer the following:

Code Changes:

Issues:
- Ship was damaged after respawning

**Note: this required the complementing changes to data files.**

### DIFF
--- a/engine/src/cmd/energetic.h
+++ b/engine/src/cmd/energetic.h
@@ -100,7 +100,6 @@ public:
     float warpenergy;    //short fix
     float constrained_charge_to_shields;
     bool sufficient_energy_to_recharge_shields;
-public:
 
     //fuel of this unit
     float fuel;
--- a/engine/src/cmd/energetic.h
+++ b/engine/src/cmd/energetic.h
@@ -100,7 +100,7 @@ public:
     float warpenergy;    //short fix
     float constrained_charge_to_shields;
     bool sufficient_energy_to_recharge_shields;
-protected:
+public:
 
     //fuel of this unit
     float fuel;
--- a/engine/src/cmd/movable.cpp
+++ b/engine/src/cmd/movable.cpp
@@ -476,7 +476,7 @@ void Movable::FireEngines(const Vector &Direction /*unit vector... might default
         float FMass) {
     Energetic *energetic = dynamic_cast<Energetic*>(this);
     FMass = energetic->ExpendFuel(FMass);
-    NetForce += Direction * (FuelSpeed * FMass / GetElapsedTime());
+    NetForce += Direction * ((double)FuelSpeed * (double)FMass / GetElapsedTime());
 }
 
 //applies a force for the whole gameturn upon the center of mass
--- a/engine/src/cmd/movable.cpp
+++ b/engine/src/cmd/movable.cpp
@@ -30,10 +30,14 @@
 #include "universe.h"
 #include "configuration/game_config.h"
 #include "vs_logging.h"
+#include "audiolib.h"
+#include "audible.h"
+
 
 #include <iostream>
 #include <string>
 #include <vega_cast_utils.h>
+#include <limits.h>
 
 float accelStarHandler(float &input) {
     return input / (configuration()->physics_config.game_speed * configuration()->physics_config.game_accel);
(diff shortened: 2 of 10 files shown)


engine/src/cmd/energetic.h
```diff
@@ -100,7 +100,7 @@ class Energetic {
     float warpenergy;    //short fix
     float constrained_charge_to_shields;
     bool sufficient_energy_to_recharge_shields;
-protected:
+public:
```
BenjamenMeyer, line 103: everything prior to this is public, so if we're going to change it then we should just drop it.

engine/src/cmd/movable.cpp
```diff
@@ -466,3 +470,362 @@
     return minimum_multiplier;
 }
 
+
+void Movable::FireEngines(const Vector &Direction /*unit vector... might default to "r"*/,
+        float FuelSpeed,
+        float FMass) {
+    Energetic *energetic = dynamic_cast<Energetic*>(this);
+    FMass = energetic->ExpendFuel(FMass);
+    NetForce += Direction * (FuelSpeed * FMass / GetElapsedTime());
```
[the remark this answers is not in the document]
BenjamenMeyer: Please fix this warning so we don't lose data or improperly cut-off data and lose precision.

engine/src/cmd/movable.cpp
```diff
@@ -466,3 +470,362 @@ double Movable::GetMaxWarpFieldStrength(float rampmult) const {
...
+    if (unit->afterburntype == 0) {
+        //fuel-burning afterburner - uses default efficiency - appears to check for available energy? FIXME
+        //HACK this forces the reaction to be Li-6+Li-6 fusion with efficiency governed by the getFuelUsage function
+        unit->fuel -= unit->getFuelUsage(false) * simulation_atom_var * Res.Magnitude() * FMEC_exit_vel_inverse / Lithium6constant;
+#ifndef __APPLE__
+        if (ISNAN(unit->fuel)) {
+            VS_LOG(error, "Fuel is NAN B");
+            unit->fuel = 0;
+        }
+#endif
```
BenjamenMeyer, line 697: Perhaps we need to change compiler flags instead of opting out platforms from this logic check?
https://developer.apple.com/forums/thread/656056

BenjamenMeyer (changes requested):

engine/src/cmd/movable.cpp
royfalk: I dug a bit into this and I don't know why it's here. The original [commit](https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/58be34265c8382c848eadfce58aea7d35cd391e0) is from 2006. OSX was introduced in 2001. Apple switched to Intel computers in 2006 but that should have made it more consistent with Linux, not break it. There's a good chance that if we get the game running on Mac, we won't need this.

BenjamenMeyer: @royfalk if we can get rid of it that'd be awesome!

stephengtuggy: @BenjamenMeyer I'd like to recommend that we leave this code as is for now. Then once we get #773 reviewed and merged (hint, hint) then we can come back and revisit this topic.

BenjamenMeyer: @stephengtuggy that's fine. It'd be nice to simplify this so it's the same on all platforms or easily swappable via compiler flags instead.  However we get there - or whichever PR gets us there - doesn't matter to me. Thanks!

engine/src/cmd/movable.cpp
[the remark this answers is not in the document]
stephengtuggy: I agree that it's probably best to fix this warning.

engine/src/cmd/movable.cpp
royfalk: Moved to #787. Resolving conversation.

royfalk:
I think I did everything needed here. Can I get an approval?

stephengtuggy:
@royfalk You do realize that this includes the changes from the big #754 refactor, right? I don't think those are ready to go into the `master` branch yet, unless I am mistaken.

stephengtuggy:
I would be okay with merging these changes into a 754-specific working branch.

stephengtuggy (approved):
OK, my mistake. Upon checking the commit list more carefully, I realized that this PR includes the commit(s) rolling back 754, as well.

Looks good. Approving and merging.

royfalk:
> OK, my mistake. Upon checking the commit list more carefully, I realized that this PR includes the commit(s) rolling back 754, as well.
> 
> Looks good. Approving and merging.

Note the warning at the top. It is coupled to a change in assets that I still need to submit. Once approved, I'll push it.

stephengtuggy:
Actually, I can't merge this until @BenjamenMeyer re-reviews and approves it. Crap. Ben, would you be so kind as to review this again?

BenjamenMeyer:
Been away this weekend will take a look as soon as I can

BenjamenMeyer (approved):

royfalk:
I'm not sure why git keeps asking me to rebase the same changes over and over. 
I finally forced push these in the hopes this will do the trick.

Issue #713, opened 2022-07-15
Better Command-line Argument Processing


BenjamenMeyer:
We need to revisit how the command-line arguments are processed and update to using a new tooling system that can provide consistency across platforms and even between arguments. We have quite a few different things that are listing issues with the command-line arguments in one way or another:
- #212 
- #316
- #378 
- #533 
- #581, #584
- #583 
- #706 
- #712

As a result, this one option may help clean up several outstanding bugs (though not that whole list).

There are a variety of options out there, to list a few:
- GNU getopt
- Boost:Program_Options

As we are already integrated with Boost it may make sense to use/prefer the Boost:Program_Options.

PR #774, opened 2023-03-27
Use boost::program_options for parsing cmd args

ondrejkajinek:

Please answer the following:

Code Changes:

Issues:
- closes: #713 

Purpose:
- Instead of manual parsing of cmd args, boost::program_options is used. Modified binaries: vegastrike-engine, vegesetup
- Originally labeled as 0.9.x milestone but 0.8.x would also benefit from it


engine/src/main.cpp
```diff
@@ -760,139 +761,134 @@ void bootstrap_main_loop()
...
+    boost::program_options::options_description vs_options;
+    vs_options.add(vs_switches).add(vs_modules);
+
+    boost::program_options::positional_options_description vs_positionals;
+    vs_positionals.add("mission_name", 1);
+
+    boost::program_options::command_line_style::style_t style = boost::program_options::command_line_style::style_t(
+        boost::program_options::command_line_style::unix_style |
+        boost::program_options::command_line_style::case_insensitive
+    );
```
BenjamenMeyer, line 806: :+1: Awesome.
We might relax from `unix_style` later so it matches each platform properly, but this is good for transition to the new support.

BenjamenMeyer (approved):
LGTM - @ondrejkajinek thanks; this will really help going forward.

engine/src/main.cpp
ondrejkajinek: Actually, I hate all the code around checking which option was passed :) Too much code for very little action.

Yes, windows users might be happier with `/` instead of `-`. This might be easily changed with some `#ifdef` or something like that.

BenjamenMeyer: yeah - but one step at a time. I could see us moving this over to the main function at some point so that we don't get this far into the code if we can't pass the options. Regardless this is a good first step. thanks!

BenjamenMeyer: also for 0.8.x we need to maintain full compatibility - no changes like swapping parameter styles. for 0.9.x (current master) we could break that compatibility and allow the platforms to use their native parameter styles.

ondrejkajinek: > for 0.9.x (current master) we could break that compatibility and allow the platforms to use their native parameter styles.

Should I create new Issue for this?

engine/src/main.cpp
```diff
@@ -760,139 +761,134 @@ void bootstrap_main_loop()
...
+        // (",p", boost::program_options::value<std::string>(), "Specify player location")  // wasn't handled, don't know the exact format
+        (",j", boost::program_options::value<std::string>(), "Start in a specific system")
+        (",l", boost::program_options::value<std::string>(), "Specify player location")
+        (",a", "Normal resolution (800x600)")
+        (",h", "High resolution (1024x768)")
+        (",v", "Super high resolution (1280x1024)")
+        ("help", "Show this help")
+        ("benchmark", boost::program_options::value<float>(), "???")
+        ("net", "Networking Enabled (Experimental)")
+        ("debug", boost::program_options::value<int>()->implicit_value(1), "Enable debugging output, 1 major warnings, 2 medium, 3 developer notes")
```
stephengtuggy, line 787: Now, that is cool!

engine/src/python/python_compile.cpp
```diff
@@ -99,6 +101,11 @@ PyObject * CompilePython( const std::string &name )
 extern PyObject* PyInit_VS;
 void CompileRunPython( const std::string &filename )
 {
+#if (PY_VERSION_HEX >= 0x03110000)
```
stephengtuggy, line 104: Remember, it's `0x030B0000`, not `0x03110000`. :-) That tripped me up at first, too.

engine/vcpkg.json
```diff
@@ -14,6 +14,7 @@
     "boost-chrono",
     "boost-atomic",
     "boost-assign",
+    "boost-program-options",
```
stephengtuggy, line 17: 👍🏽

stephengtuggy (changes requested):
This looks really good! The only item I can see that really needs to be changed is one occurrence of `0x03110000` for the Python version. (Needs to be `0x030B0000` instead.)

Other than that, a nice-to-have would be a more complete command-line reference help message for vegasettings. But that can come later.

All in all, very nice work!

engine/src/python/python_compile.cpp
ondrejkajinek: Yeah, I happened to fall in the same trap by copy-pasting. Very, very bad.

Fixed.

stephengtuggy: No worries. :-) It's looking better now.

engine/src/main.cpp
stephengtuggy: @ondrejkajinek Sure, feel free to do so. You can create an issue directly off any of the above comments.

ondrejkajinek: Here: https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/775

stephengtuggy (approved):

Issue #761, opened 2023-02-07
This save game makes Vega Strike Core Dump with Floating point error exception


evertvorster:
Hi there!

I thought I would play test VegaStrike a little to see what everyone was up to lately. 
It loads really fast, and some of my save games I am able to load, but this one makes the engine core dump:
```
[evert@Evert ~]$ vegastrike
Registering codec ogg
.
Legacy Mode: TRUE
Saving current directory (/home/evert) as DATA_DIR
 In path /home/evert
Vega Strike  
See http://www.gnu.org/copyleft/gpl.html for license details.

GOT SUBDIR ARG = 
Found data in /usr/share/vegastrike
Using /usr/share/vegastrike as data directory
Using .vegastrike as the home directory
Found MODDIR = /usr/share/vegastrike/mods
USING HOMEDIR : /home/evert/.vegastrike As the home directory 
CONFIGFILE - Found a config file in home directory, using : /home/evert/.vegastrike/vegastrike.config
LoadGameConfig: Found game config at '/home/evert/.vegastrike/vegastrike.config'
LoadXML called with filename '/home/evert/.vegastrike/vegastrike.config'
DATADIR - No datadir specified in config file, using : /usr/share/vegastrike
Force galaxy to /usr/share/vegastrike/universe/milky_way.xml
SIMULATION_ATOM: 0.01
log directory : '/home/evert/.vegastrike/logs'
['/usr/lib/python310.zip', '/usr/lib/python3.10', '/usr/lib/python3.10/lib-dynload']
['', '/usr/share/vegastrike/modules/builtin', '/usr/share/vegastrike/modules/quests', '/usr/share/vegastrike/modules/missions', '/usr/share/vegastrike/modules/ai', '/usr/share/vegastrike/modules', '/usr/share/vegastrike/bases', '/usr/lib/python310.zip', '/usr/lib/python3.10', '/usr/lib/python3.10/lib-dynload']
Engine Version: 0.9.0.d0e6b767f 
Asset API Version: 2
[...]

PR #780, opened 2023-05-24
Get the game working on macOS again (without PR 754 changes)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Relates to #533 

Purpose:
- What is this pull request trying to do? Get the game working on macOS again, this time without all the changes from the PR 754 refactor. Also, update several of the GitHub Actions to the latest version available.
- What release is this for? 0.9.x, presumably
- Is there a project or milestone we should apply this to? 0.9.x, presumably

Additional Notes:

During play testing, I kept encountering crashes after loading certain savegames. (I think that at least some of these may be the same issue previously reported as #761 .) I also hit the crash in the `Mesh` destructor at least once. (#716 ) And finally, there's the issue with no text on macOS (#786)

### DIFF
--- a/engine/src/gldrv/gl_init.cpp
+++ b/engine/src/gldrv/gl_init.cpp
@@ -285,12 +285,27 @@ void init_opengl_extensions() {
 #ifndef __APPLE_PANTHER_GCC33_CLI__
         glLockArraysEXT_p   = nullptr;
         glUnlockArraysEXT_p = nullptr;
+#else
+        glLockArraysEXT_p = nullptr;
+        glUnlockArraysEXT_p = nullptr;
 #endif /*__APPLE_PANTHER_GCC33_CLI__*/
 #endif
         VS_LOG(debug, "OpenGL::GL_EXT_compiled_vertex_array unsupported");
     }
 #endif
-#ifndef __APPLE__
+#if defined (__MACOSX__)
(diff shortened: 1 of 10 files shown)


engine/src/gldrv/gl_init.cpp
```diff
@@ -271,39 +283,45 @@ void init_opengl_extensions() {
     } else {
 #ifdef __APPLE__
 #ifndef __APPLE_PANTHER_GCC33_CLI__
-        glLockArraysEXT_p   = 0;
-        glUnlockArraysEXT_p = 0;
+        glLockArraysEXT_p   = nullptr;
+        glUnlockArraysEXT_p = nullptr;
+#else
+        glLockArraysEXT_p = nullptr;
+        glUnlockArraysEXT_p = nullptr;
```
BenjamenMeyer, line 290: do we really need the preprocessor block here? Or can we do away with it and just initialize them to nullptr?

BenjamenMeyer (approved):

engine/src/gldrv/gl_init.cpp
stephengtuggy: no, as a matter of fact, I don't believe we do. Trying it without the preprocessor block now.

stephengtuggy: Superfluous preprocessor block removed. Seems to work fine.

Issue #783, opened 2023-05-25
GitHub Actions Migration - Node 12 -> 16


BenjamenMeyer:
We are going to have an issue with CI soon as the version of our GH Actions uses Node 12.

Issue #782, opened 2023-05-25
Windows: Upgrade Boost Library


BenjamenMeyer:
Upgrade the version of boost on Windows.

- [x] 0.8.x for 0.8.2
- [x] master for 0.9.x

PR #789, opened 2023-06-19
Fix Build/Update Boost on Windows (non-PR-754 version)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- #782 
- #783 

Purpose:
- What is this pull request trying to do? Update Boost on Windows, on the `master` branch.


engine/src/cmd/collide_map.cpp
```diff
@@ -36,27 +39,23 @@ void CollideArray::erase(iterator target) {
...
-            if (&*i == target) {
-                targlist->erase(i);
-                return;
-            }
-        }
+        CollidableBackref *targ = vega_dynamic_cast_ptr<CollidableBackref>(&*target);
+        std::list<CollidableBackref> *targlist = &toflattenhints.at(targ->toflattenhints_offset);
+        auto first_to_remove = std::stable_partition(targlist->begin(), targlist->end(), [target](CollidableBackref &backref) { return &*backref != target; });
+//        std::for_each(first_to_remove, targlist->end(), [](CollidableBackref &backref) { delete backref; });
+        targlist->erase(first_to_remove, targlist->end());
```
stephengtuggy, line 58: Please note that this change may affect game behavior, unlike most of the changes in this PR. However, I think this block of changes is almost mandatory for correctness.

stephengtuggy:
Please note also that this looks like a very large changed-line count, but probably at least 1,000 of those are just whitespace changes.

BenjamenMeyer:
@stephengtuggy Try https://github.com/boostorg/python/issues/233 and see if that will do it

stephengtuggy:
This was working, and everything was golden. Then nasm.us went down (https://github.com/microsoft/vcpkg/issues/32600), and all our Windows builds stopped working. Dang it!

stephengtuggy:
I think I'm going to go ahead and submit this PR for review anyway, based on the fact that all the builds were succeeding prior to nasm.us going down and breaking the openssl builds.

stephengtuggy:
@evertvorster , if you have time to play-test this PR a bit before it gets merged, that would be great. If not, no worries.

engine/src/cmd/collide_map.h
```diff
@@ -195,6 +199,8 @@ class CollideMap : public CollideArray {
     bool CheckCollisions(Unit *un, const Collidable &updated); //will be handed off to a templated function
     bool CheckUnitCollisions(Unit *un,
             const Collidable &updated); //DANGER must be used on lists that are only populated with Units, not bolts
+
+    // TODO: Add virtual destructor?
```
BenjamenMeyer, line 203: Yes, even if just empty and/or calling the parent

script/bootstrap.ps1
```diff
@@ -17,25 +19,36 @@
...
+)
+
+Set-Variable -Name CMAKE_VERSION -Value "3.25.1"
 
 New-Item "$VCKPG_PARENT_DIR" -ItemType Directory -Force
 Push-Location "$VCKPG_PARENT_DIR"
-git clone https://github.com/Microsoft/vcpkg.git
-.\vcpkg\bootstrap-vcpkg.bat -disableMetrics
+git clone https://github.com/stephengtuggy/vcpkg-local.git ./v
+.\v\bootstrap-vcpkg.bat -disableMetrics
```
BenjamenMeyer, line 34: I assume this is temporary while the service is down, no?

BenjamenMeyer (commented):
Couple questions before I approve

engine/src/cmd/collide_map.h
stephengtuggy: Agreed. That is most likely what needs to happen, sooner or later. But can it wait for another PR?

script/bootstrap.ps1
stephengtuggy: > I assume this is temporary while the service is down, no?

@BenjamenMeyer Actually, no. This is intended to be a long-term solution. For the last several months -- well before the nasm.us issues -- I have found it necessary to use a fork-of-a-fork of vcpkg in order to get Vega Strike to build successfully on Windows. I do intend to keep this fork updated semi-regularly.

stephengtuggy: @BenjamenMeyer working on that right now: https://github.com/vegastrike/vcpkg-local

engine/src/cmd/collide_map.h
BenjamenMeyer: yes, we probably need to do a PR with adding it to everything we can find; though adding:
```c++
virtual ~CollideMap() { ~ListMutableSet<Collidable>() }
```
here would do well too.

engine/src/cmd/collide_map.cpp
```diff
@@ -161,8 +160,8 @@ void CollideArray::flatten() {
             collideUpdate(*tmp, index);
         }
 
-        std::list<CollidableBackref>::iterator listend = toflattenhints[i].end();
-        for (std::list<CollidableBackref>::iterator j = toflattenhints[i].begin();
+        auto listend = toflattenhints[i].end();
+        for (auto j = toflattenhints[i].begin();
                 j != listend;
```
BenjamenMeyer, line 165: we should just test directly against `.end()` instead of creating a variable here.

BenjamenMeyer (changes requested):

engine/src/cmd/collide_map.cpp
BenjamenMeyer: NOTE: the `end` could potentially change which is why you're always supposed to test directly against `end()` instead of using a var

.github/workflows/Windows-CI.yml
```diff
@@ -22,80 +19,144 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os:             windows-2019
-            cmakeGenerator: VS16Win64
-          - os:             windows-2022
-            cmakeGenerator: Ninja
+          - os:              windows-2019
```
royfalk, line 22: Are we really testing all these variants? It may be a bit excessive IMO.

.github/workflows/gh-actions-pr.yml
```diff
@@ -6,183 +6,367 @@ jobs:
...
+          #  COMPILER: 'gcc'
+          #  OpenGL_GL_PREFERENCE: 'GLVND'
+          #  ENABLE_PIE: 'OFF'
+          #  allow_failure: false
+          #- FROM:     'ubuntu:lunar'
+          #  COMPILER: 'clang'
+          #  OpenGL_GL_PREFERENCE: 'GLVND'
+          #  ENABLE_PIE: 'OFF'
+          #  allow_failure: false
+          - FROM:     'ubuntu:kinetic'
```
royfalk, line 39: Again, I think we need to restrict our officially supported everything. We are not a large corporation and even there I don't recall supporting so many versions.

.github/workflows/gh-actions-pr.yml
```diff
@@ -6,183 +6,367 @@ jobs:
...
             COMPILER: 'clang'
             OpenGL_GL_PREFERENCE: 'GLVND'
+            ENABLE_PIE: 'ON'
+            allow_failure: false
           - FROM:     'rockylinux:8.5'
             COMPILER: 'clang'
             OpenGL_GL_PREFERENCE: 'LEGACY'
+            ENABLE_PIE: 'ON'
+            allow_failure: false
           - FROM:     'rockylinux:8.5'
```
royfalk, line 304: Also, this looks way more than just updating boost. It's better and easier to submit several smaller PRs than a big one.

engine/CMakeLists.txt
```diff
@@ -629,537 +678,561 @@ CONFIGURE_FILE(${Vega_Strike_SOURCE_DIR}/cmake-config.h.in ${Vega_Strike_BINARY_
...
-    src/gfx/pass.cpp
-    src/gfx/tex_transform.cpp
-    src/gfx/vdu.cpp
-    src/gfx/vid_file.cpp
-    src/ffmpeg_init.cpp
-    src/gfx/vsbox.cpp
-    src/gfx/vsimage.cpp
-    src/gfx/warptrail.cpp
-)
+        src/gfx/ani_texture.cpp
```
royfalk, line 929: Easier to have these cosmetic (tabs and copyright changes) in a separate PR. It's pretty much auto approve and removes a lot of noise from here.
Also, probably a good idea to squash the commits to keep the tree cleaner.

engine/src/cmd/base_util.cpp
```diff
@@ -16,19 +15,20 @@
...
 
 
-#include <Python.h>
+#define PY_SSIZE_T_CLEAN
+//#include <Python.h>
 
 #include <boost/version.hpp>
-#if BOOST_VERSION != 102800
 #include <boost/python.hpp>
+#if BOOST_VERSION != 102800
```
royfalk, line 31: Same as above.

engine/src/cmd/collide_map.cpp
```diff
@@ -36,27 +39,23 @@ void CollideArray::erase(iterator target) {
...
-        CollidableBackref *targ = static_cast< CollidableBackref * > (&*target);
-        std::list<CollidableBackref> *targlist = &toflattenhints[targ->toflattenhints_offset];
-        std::list<CollidableBackref>::iterator endlist = targlist->end();
-        for (std::list<CollidableBackref>::iterator i = targlist->begin(); i != endlist; ++i) {
-            if (&*i == target) {
-                targlist->erase(i);
-                return;
-            }
-        }
+        CollidableBackref *targ = vega_dynamic_cast_ptr<CollidableBackref>(&*target);
```
royfalk, line 54: I don't understand this code at all. Can we maybe split this into a separate PR?

engine/src/cmd/collide_map.cpp
```diff
@@ -161,8 +160,8 @@ void CollideArray::flatten() {
             collideUpdate(*tmp, index);
         }
 
-        std::list<CollidableBackref>::iterator listend = toflattenhints[i].end();
-        for (std::list<CollidableBackref>::iterator j = toflattenhints[i].begin();
+        auto listend = toflattenhints[i].end();
```
royfalk, line 163: I'm fine with type inference where it is trivial. Example from other languages is something like

`let someString = String('blah blah')`

However, I'm less comfortable with replacing _std::list<CollidableBackref>::iterator_ with _auto_. I prefer to clearly see my C++ abominations.

engine/src/cmd/collide_map.h
```diff
@@ -45,7 +43,7 @@ class Collidable {
         Unit *unit;
         unsigned int bolt_index;
     }
-            ref;
+            ref{};
```
royfalk, line 46: What does this do? I assume it's defined either way.
Also, can we fix the indentation. It's handing out there.

engine/src/python/unit_exports.cpp
```diff
@@ -13,16 +15,17 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
  * along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
  */
 
+#define PY_SSIZE_T_CLEAN
 #include <boost/version.hpp>
-#if BOOST_VERSION != 102800
```
royfalk, line 24: Here as well.

engine/src/python/unit_wrapper_class.h
```diff
@@ -162,15 +167,15 @@ class UnitWrapper : public UnitContainer {
     boost::python::tuple GetOrientation() {
         {
         CHECKME VS_BOOST_MAKE_TUPLE(VS_BOOST_MAKE_TUPLE(0, 0, 0), VS_BOOST_MAKE_TUPLE(0, 0, 0), VS_BOOST_MAKE_TUPLE(0,
-                    0,
```
royfalk, line 165: Probably better to just not break the line or break earlier for easier reading. Same for the rest of the file.

engine/src/universe_util_generic.cpp
```diff
@@ -513,525 +517,525 @@ int getNumUnits() {
...
+                            collision = true;
+                        } else {
+                            QVector r(.5, .5, .5);
+                            pos += (radial_size + un->rSize()) * r;
+                            collision = true;
+                        }
                     }
                 }
-            }
-            if (collision == false) {
```
royfalk, line 731: Did you move this code block to an inner brace intentionally?

engine/src/cmd/collide_map.h
stephengtuggy: I don't think virtual destructors need to call each other. IIRC, they automatically get called in the proper sequence.

Making the destructor virtual, for every class that is intended to be inheritable, IS something that needs to happen. But I think in this case, I'll wait until the next PR.

Thanks for the review!

script/bootstrap.ps1
stephengtuggy: Continuing to have problems with the Windows builds. Now it's Python that isn't compiling, via vcpkg. (Annoying that we even have to compile it at all. Maybe there's another way; not sure.)

engine/src/cmd/collide_map.cpp
stephengtuggy: @BenjamenMeyer you're probably right. I would like to point out that the `.end()` pre-assignment was there before; I just changed the variable type declarations to `auto`, based on CLion's suggestion to use `auto` for all iterator variables. Based on @royfalk 's comment further down, it sounds like we still need to discuss whether or not we want to do that as a general practice. Maybe I'll revert the changes to these lines for now.

.github/workflows/gh-actions-pr.yml
stephengtuggy: Normally, I would agree. In this case, I was just trying to hurry up and get everything building again.

Good point that this is about a lot more than updating Boost. I have updated the PR title to reflect that.

.github/workflows/Windows-CI.yml
stephengtuggy: Good point. We can probably cut back a bit.

engine/CMakeLists.txt
stephengtuggy: I can squash the commits when merging this PR. I am fine with that.

Do you want me to reverse these cosmetic changes to CMakeLists.txt in this PR, and then reformat later in a separate PR?

engine/src/cmd/base_init.cpp
```diff
@@ -17,24 +15,25 @@
...
  * You should have received a copy of the GNU General Public License
  * along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
  */
 
 
-#include <math.h>
+#define PY_SSIZE_T_CLEAN
 
 #include <boost/version.hpp>
-#if BOOST_VERSION != 102800
```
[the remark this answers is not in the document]
stephengtuggy: @royfalk Indeed, this is quite an ancient version of Boost. However... there is a possibility that this specific version is being used to distinguish between Python environments (embedded Python vs. ... doubly-embedded Python, or something like that). I'm not quite sure of the details or history of this. Would like to get some input from a long-term project member or two, if possible.

.github/workflows/gh-actions-pr.yml
stephengtuggy: I agree. The policy we decided on last was to support the LTS versions of distros such as Ubuntu that are still in mainstream support, as well as the latest. We may have to revisit that a bit; it's turning out to be more than I can handle lately.

engine/src/cmd/base_init.cpp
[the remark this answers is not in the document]
stephengtuggy: In the meantime, can we leave it as is?

engine/src/cmd/collide_map.cpp
stephengtuggy: We were doing a `static_cast` to pointer-to-CollidableBackref, without any type checking, where we should have been doing a `dynamic_cast` in order to verify that the actual type (at runtime) is correct.

Does that explain it sufficiently? Or do you want to go ahead and split this part out into its own PR?

engine/src/cmd/collide_map.cpp
stephengtuggy: Fair point LOL.

I was going by a CLion suggestion that all iterator variables should use `auto`. Sounds like we need to discuss that further though.

engine/src/cmd/collide_map.h
stephengtuggy: Adding the empty braces initializes the value of `ref`. It was uninitialized previously, as far as I can tell.

engine/src/universe_util_generic.cpp
stephengtuggy: Let me take another look. I think so, but now I'm not sure which way is correct.

engine/src/universe_util_generic.cpp
```diff
@@ -513,525 +517,525 @@ int getNumUnits() {
...
-        mission->player_autopilot = Mission::AUTO_NORMAL;
+    void LoadMissionScript(string missionscript) {
+        ::LoadMission("nothing.mission", missionscript, false);
     }
-}
 
-QVector SafeStarSystemEntrancePoint(StarSystem *sts, QVector pos, float radial_size) {
-    if (radial_size < 0) {
-        radial_size = game_options()->respawn_unit_size;
+    void SetAutoStatus(int global_auto, int player_auto) {
```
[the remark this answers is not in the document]
stephengtuggy: Yeah, I'm using CLion. I highly reco

engine/src/cmd/collide_map.cpp
royfalk: That's fine.

engine/CMakeLists.txt
royfalk: No. It's already in review and I'm not that strict about full separation. Everything in moderation.

engine/src/cmd/base_init.cpp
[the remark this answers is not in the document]
royfalk: Sure.

.github/workflows/gh-actions-pr.yml
BenjamenMeyer: It's generally best effort right now. If we need to drop something because we don't have the bandwidth then that's fine - let's just document it and why.

That said, most of our distros are highly related with some simple packaging changes between them so I don't think it's a huge effort for Linux. Mac/Windows, OTOH, are still in a larger need of support and we could certainly use some help there.

engine/src/cmd/collide_map.h
BenjamenMeyer: C++11 initialization method; we've generally stuck to constructor initialization but we've started getting more C++11 brace initialization in. We just have to make sure it's C++11 format and not C++20 feature in the process.

engine/src/cmd/collide_map.cpp
BenjamenMeyer: yeah, the code on the left certainly isn't right. I did some similar change in @royfalk 's SDL2 PR....https://github.com/royfalk/Vega-Strike-Engine-Source/pull/1/files#r1292449185

engine/src/cmd/collide_map.cpp
BenjamenMeyer: yeah,  I realize it was there before; this would be one improvement for the code changing that up to actually be correct. Use `auto` instead of specifying the full iterator type is also useful here (and one of few things I like to use `auto` for).

engine/src/cmd/collide_map.h
stephengtuggy: Filed an issue for this. Resolving conversation for now.

engine/src/cmd/collide_map.cpp
stephengtuggy: OK, I'll change this to use `.end()` directly. I'll also leave it as `auto` for now. Any objections?

.github/workflows/gh-actions-pr.yml
stephengtuggy: Definitely.

engine/src/cmd/base_init.cpp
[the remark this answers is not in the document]
stephengtuggy: Filed as #797

engine/src/cmd/base_util.cpp
stephengtuggy: Covered by #797

engine/src/cmd/collide_map.cpp
stephengtuggy: Started a discussion for this: https://github.com/orgs/vegastrike/discussions/800

engine/src/cmd/collide_map.h
royfalk: Intuitively, it seemed to me it was initialized. However, on closer inspection, this is a non-static variable, so it really is uninitialized. it would be clearer if we were to specify :

`Unit *unit = nullptr;
        unsigned int bolt_index;
 } ref;`

BenjamenMeyer: This initialization is for any kind of variable, not simply non-static. The C++ committee's idea is that the constructors don't need to do all the initialization of the class since they have this other means.
I think Strostroup's idea is that the constructor isn't necessary in 99% of cases and this lets them do that.

That said, I still kind of like having explicit constructors doing all the initialization.

engine/src/cmd/collide_map.cpp
BenjamenMeyer: I'm okay with `auto` for the loops with iterators because (a) the loops should be fairly short, (b) the iterator notation (`namespace::type::iterator`) can get quite long and lengthy and doesn't serve much purpose, and (c) the iterator type can change somewhat so it's helpful.

So this is actually one of the few places I think `auto` is okay. That said, I already mentioned that the `end` check needs to be done in the loop since `end` could  change or become invalid during the loop.

engine/src/cmd/collide_map.cpp
BenjamenMeyer: I think this is all that this PR is waiting on for approval

BenjamenMeyer:
@stephengtuggy please merge in master

engine/src/cmd/collide_map.cpp
stephengtuggy: Fixed.

stephengtuggy:
> @stephengtuggy please merge in master

Done, I believe.

.github/workflows/Windows-CI.yml
BenjamenMeyer: let's just stick to 1 Windows build for the time being; that should at least get the ball rolling.

stephengtuggy: How about 2 Windows builds? One on Windows (Server) 2019, and one on Windows (Server) 2022. I think that will be enough even to satisfy me, haha.

BenjamenMeyer:
> > @stephengtuggy please merge in master
> 
> Done, I believe.

Unfortunately not, or at least the merge didn't work right. It's failing at minimum because it can't find SDL2; and checking the bootstrap script it doesn't seem to be trying to install it so it doesn't have some of the required changes.
It's almost acting like it has the wrong Docker images too. Hopefully it's just the bootstrap scripts and some related CMake settings for SDL2.

.github/workflows/Windows-CI.yml
BenjamenMeyer: @stephengtuggy if we need to

stephengtuggy:
> > > @stephengtuggy please merge in master
> > 
> > 
> > Done, I believe.
> 
> Unfortunately not, or at least the merge didn't work right. It's failing at minimum because it can't find SDL2; and checking the bootstrap script it doesn't seem to be trying to install it so it doesn't have some of the required changes. It's almost acting like it has the wrong Docker images too. Hopefully it's just the bootstrap scripts and some related CMake settings for SDL2.

@BenjamenMeyer I merged your PR into mine. 😃 Hopefully it works now. Thanks.

BenjamenMeyer:
Well more successes than failures overall.
- Windows 2022 passed; 2019 did not
- most Mac builds failed, but 3 passed
- all Linux builds passed except Ubuntu Kinetic
So progress

stephengtuggy:
(Yes, I know that's a change in tune from what I usually say 😄 )

BenjamenMeyer:
@stephengtuggy I can live with that.

.github/workflows/Windows-CI.yml
BenjamenMeyer: well, VS2022 on Windows 2022 passed.

.github/workflows/gh-actions-pr.yml
BenjamenMeyer: So Kinetic is no longer supported since Manic is out. Will fix in master.

.github/workflows/macos-ci.yml
```diff
@@ -49,8 +49,8 @@ jobs:
         run: brew install boost-python3 gtk+3 gtkglext sdl
 
       # The following Apple-provided libraries are deprecated:
-      # * OpenGL as of MacOS 10.14
-      # * GLUT as of MacOS 10.9
+      # * OpenGL as of macOS 10.14
+      # * GLUT as of macOS 10.9
```
BenjamenMeyer, line 53: NOTE: All the MacOS 11 builds failed; MacOS 12 passed for all except including OpenAL with GCC if I understand the matrix right.

BenjamenMeyer (approved):
We'll fix the Mac/Windows builds and swap Ubuntu Kinetic for Mantic on master.

stephengtuggy:
Merged, finally!

engine/src/universe_util_generic.cpp
stephengtuggy: Update: Yes, this was intentional. I think a previous PR of mine messed up the structure of this file a bit. This is an attempt to fix it.

.github/workflows/macos-ci.yml
stephengtuggy: Filed https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/812 for the macOS build failures. They all seem to be related to which version of Python Boost is trying to use.

PR #792, opened 2023-08-07
Upgrade SDL to version 2

royfalk:
This upgrades SDL to version 2. 
This is needed for dear imGui and also a good thing in general.
This PR touches inputs and not much else. Things to test are keyboard, mouse and joystick.

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- Joystick not tested (need someone to help)
- Buy new ship fails with mesh error. Almost certainly #716.

PR #796, opened 2023-08-31
Task aug unit

royfalk:
This is a near complete refactoring of the cloaking in the game.

This is also a move from refactoring using subclasses to fields. 
Instead of energetic, we'll have reactor and capacitors.

Not sure why it is picking extra stuff for this PR. It should only have taken 8668683.

Code Changes:

Game Changes:
- The game now decides if a unit is cloaked if it exceeds its minimum. Before it sometimes used an arbitrary .8. At some point, we'll want to make radars that can pierce at short range, etc. But there are much more urgent tasks.

Issues:
- Game is still fragile. Can't buy a second ship and fly with it. 
- I was unable to make the graphics do what I wanted. I wanted cloaked ship to be black and somewhat transparent and glass cloaked ships to be nearly transparent. 
- Cloaking takes longer. If an issue, we'll need to adjust rate for ships and modules.
- Cloak can be damaged but can't be repaired (probably). I was unable to find the code to fix modules.


engine/src/cmd/cloak.h
```diff
@@ -0,0 +1,143 @@
...
+
+    // If this unit cloaks like glass or like fading.
+    // Glass is alpha only. Non glass affects rgb as well.
+    bool glass;
+
+    // Current cloak value. 0 is uncloaked. 1 is fully cloaked.
+    double current;
+
+    // The minimum cloaking value...
+    double minimum;
```
BenjamenMeyer, line 62: I like this - a lot. My only question is, can we keep this to integer math? Or do we have to do it within floating point? If we need to keep it as floating point, then fine; if we can move it to integer math that'd be better. Floating point math generally has a higher cost than integer math, but there's also the precision factor that can cause issues so it'd be better for us to use integer as much as possible. If this isn't immediately doable okay. Not going to block on this, just raising the conversation.

engine/src/cmd/unit_generic.h
```diff
@@ -613,16 +603,14 @@ class Unit : public Armed, public Audible, public Drawable, public Damageable, p
...
 
-//how visible the ship is from 0 to 1
+// how visible the ship is from 0 to 1
+// Need this function to expose it to python
     float CloakVisible() const {
-        if (cloaking < 0) {
-            return 1;
-        }
-        return ((float) cloaking) / 2147483647;
+        cloak.Visible();
```
BenjamenMeyer, line 609: linter raising an issue here since it does need to return a value or be refactored to be void.
Perhaps this should be `return clock.Visible();`?

BenjamenMeyer (changes requested):
Would be good to have to some instructions for testing this.
Aside from the linting issue this looks good; please fix that. I'll be happy to test it if given some guidance - I haven't had any ships with cloak yet.

engine/src/cmd/cloak.h
royfalk: I'll take a look. I think so.

royfalk:
I'll wait on this until #792 and #789 are both ready.

engine/src/cmd/cloak.h
royfalk: Sort of, but no. I implemented it but the problem is the simulation_atom_var. Unless you make the integers really large (say 10k or more), 
`rate` * simulation_atom_var < 0 
which is cast to 0 and then and cloak doesn't work.

As it is now, it works and is tested. Let's leave it at that.

engine/src/cmd/unit_generic.h
royfalk: Yep.

engine/src/cmd/cloak.h
BenjamenMeyer: Thanks for checking. Good information to know.

engine/src/cmd/unit_generic.h
royfalk: > Would be good to have to some instructions for testing this.
1. Visual - switch to view 6. Hit c and watch the ship slowly cloak itself and decloak itself.
2. Pick a fight and then cloak yourself and check the enemy doesn't shoot you. He may accidentally ram you...
3. Buy a cloaking device. Sell a cloaking device. The whole upgrade issue has not been tested properly.

BenjamenMeyer: @royfalk where can I buy a cloaking device? I have yet to see one.

stephengtuggy: @BenjamenMeyer Try buying a cloaking device at an Aera planet or base, maybe? I seem to recall someone saying at some point that Aera bases and planets tend to have cloaking devices, even if others don't.

stephengtuggy (approved):
Cool. This code is clean, well organized, well written, and seems quite straightforward. Well done, @royfalk .

I believe you also fixed the linter issue, correct?

engine/src/cmd/unit_generic.h
BenjamenMeyer: @stephengtuggy thanks for the suggestion...I'll have to try that out

BenjamenMeyer (approved):
agree with @stephengtuggy so approving; will still try to test it out once I can play long enough to find somewhere to buy a clock device from

PR #804, opened 2023-10-17
Implement radar code as a separate component

royfalk:
~~This PR introduces the concept of components - reactor, radar, cloaking device, etc.
This allows us to more elegantly move code from unit to components. 
It will also allow for the removal of unit superclasses (e.g. movable) and easier upgrade, buy, sell, damage and fix components.~~

_~~Let's use this to PR (and the cloaking one) discuss the concept of using components as a way forward.~~_

Implement radar as a component.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? both standard and specific to radar.

Issues:
- I was unable to damage and kill a Dgn Dodo at point blank. Other ships were affected. The guns locked just fine.
- ~~I think radar numbers are different from master.  Need to verify this.~~ Fixed.
- ~~The ship loses lock easy. This may have to do with the point above. If the lock cone is no longer 360 degrees, that would probably happen.~~ Fixed.
- ~~The game crashes on startup. #1021~~. aeraturretadvtorpedo (also navpoint.blank and eject.template) have max radar range of 3b and not 300m.
- Radar, like other components does not currently support repair via the standard button. Clicking it activates the old repair, which copies some stuff from the original upgrade. Clicking basic repair activates the proper repair function.


engine/src/components/component.cpp
```diff
@@ -0,0 +1,45 @@
...
+/*Component::Component()
+{
+
+}*/
+
+
+double random20()
+{
+    if(std::rand() < 0.2) {
+        return std::rand();
```
BenjamenMeyer, line 41: shouldn't this be more like:
```c++
double r = std::rand();
if (r < 0.2) {
    return r;
}
```
the code you have will retrieve a random number, compare it, and if it's less than 0.2 retrieve another random number to return. not sure that's what was intended.

BenjamenMeyer (changes requested):

BenjamenMeyer:
Yes, I like moving things to components. It makes a lot of sense.

engine/src/components/component.cpp
stephengtuggy: I agree with @BenjamenMeyer here.

engine/src/components/radar.h
```diff
@@ -0,0 +1,117 @@
...
+    float min_target_size;
+
+    // What kind of type and capability the radar supports
+    RadarType type;
+    unsigned int capabilities;
+    bool locked;
+    bool can_lock;
+    bool tracking_active;
+
+    std::unique_ptr<CRadar> original;
```
stephengtuggy, line 75: Huh! This is interesting. Can you explain this a bit more, Roy? I'm interested to hear what it's for.

engine/src/resource/resource.h
```diff
@@ -37,7 +37,7 @@ class Resource {
     T adjusted_max_value_;
     bool no_max_;
 public:
-    Resource(const T &value, const T &min_value = 0, const T &max_value = -1);
+    Resource(const T &value = 0, const T &min_value = 0, const T &max_value = -1);
```
stephengtuggy, line 40: If this constructor can be called with no arguments, or only one argument, it might be best to mark the constructor as `explicit`.

stephengtuggy (commented):
Nice! I'm excited about this new componentized approach. I think that this is what we've been needing for a long time. I look forward to seeing where it goes.

Some of the refactors I have in mind will also be considerably easier if Unit doesn't inherit from 50 million things LOL

engine/src/components/component.cpp
royfalk: Code has been removed.

engine/src/components/radar.h
royfalk: This was a first draft of keeping a copy of the original component in order to record damage.
We have for a while now been using `Resource` for this.

royfalk:
PR is ready for testing. If you intend to test it, please modify aeraturretadvtorpedo in units.json to reduce Max_Cone to 300000000 (300k km). I'll test it over the weekend.

@BenjamenMeyer, @stephengtuggy - WRT to the explicit keyword in the constructor above, I'd like more guidance. I've looked at https://stackoverflow.com/a/121163/5276890 and I think I understand. 

If I have `Resource<double>` , I can only instantiate it with a double, not an int. I'm not convinced this is in our interest. Resource really works with only ints and doubles. Do we not want the following to work?

```
int i = 8;
Resource<double> a = Resource<double>(i);
```

Also, my example was really of an implicit casting from int to double. Can you think of a scenario where an actual constructor is called? One where we don't want this to work. I'm not seeing the gotcha, that is surely there.

stephengtuggy:
> PR is ready for testing. If you intend to test it, please modify aeraturretadvtorpedo in units.json to reduce Max_Cone to 300000000 (300k km). I'll test it over the weekend.
> 
> @BenjamenMeyer, @stephengtuggy - WRT to the explicit keyword in the constructor above, I'd like more guidance. I've looked at https://stackoverflow.com/a/121163/5276890 and I think I understand.
> 
> If I have `Resource<double>` , I can only instantiate it with a double, not an int. I'm not convinced this is in our interest. Resource really works with only ints and doubles. Do we not want the following to work?
> 
> ```
> int i = 8;
> Resource<double> a = Resource<double>(i);
> ```
> 
> Also, my example was really of an implicit casting from int to double. Can you think of a scenario where an actual constructor is called? One where we don't want this to work. I'm not seeing the gotcha, that is surely there.

@royfalk the `explicit` keyword is mostly just good practice in general. In practice, it would likely come into play only if a constructor or method takes a parameter of type `Resource<int>` or `Resource<double>`, and we pass a straight `int` or `double` instead. I don't know that we do that anywhere in the codebase as it stands today. It's more just a good habit to get into, I think.

royfalk:
> @royfalk the explicit keyword is mostly just good practice in general. In practice, it would likely come into play only if a constructor or method takes a parameter of type Resource<int> or Resource<double>, and we pass a straight int or double instead. I don't know that we do that anywhere in the codebase as it stands today. It's more just a good habit to get into, I think.

I'm really on the fence here. On the one hand, it is considered good practice. On the other hand, the whole point of Resource is to elegantly wrap int/double. We have a whole bunch of overloaded operators for exactly that reason. I'll think about it some more.

BenjamenMeyer:
> > @royfalk the explicit keyword is mostly just good practice in general. In practice, it would likely come into play only if a constructor or method takes a parameter of type Resource or Resource, and we pass a straight int or double instead. I don't know that we do that anywhere in the codebase as it stands today. It's more just a good habit to get into, I think.
> 
> I'm really on the fence here. On the one hand, it is considered good practice. On the other hand, the whole point of Resource is to elegantly wrap int/double. We have a whole bunch of overloaded operators for exactly that reason. I'll think about it some more.

for now, let's skip the `explicit` keyword. Once we have more things as components we can revisit and see if it makes sense.

BenjamenMeyer (approved):
I can't play-test but things look okay to me

royfalk:
Pending #1016 merge.

royfalk:
I'm going to hold off on merging this until Sunday or so to give people a chance to play test.

@evertvorster , @kheckwrecker 
Please play test with this Assets PR: https://github.com/vegastrike/Assets-Production/pull/160

kheckwrecker:
Safe to merge on top of PR#1036 for play testing?

engine/src/cmd/upgradeable_unit.cpp
```diff
@@ -283,15 +283,18 @@ bool UpgradeableUnit::RepairUnit(const std::string upgrade_name) {
...
             break;
 
-        
+        case ComponentType::Radar:
+            if(unit->radar.Damaged()) {
+                unit->radar.Repair();
+                return true;
+            }
+            break;
+
```
BenjamenMeyer, line 292: NOTE: This is where Polymorphism helps out since you could just do:
```c++
if (unit->Damaged()) {
    unit->Repair();
    return true;
}
return false;
```
since of having all the various switch cases. Now the generic `Unit` class probably isn't the right level to have that, perhaps a `Component` class would be. Hopefully that's where we're eventually heading.

engine/src/resource/resource.cpp
```diff
@@ -31,6 +31,49 @@
...
+T _convert(const std::string input, const T modifier);
+
+template<>
+float _convert<float>(const std::string input, const float modifier) {
+    try {
+        return std::stod(input) * modifier;
+    } catch(...) {
+        return 0.0;
+    }
+}
```
BenjamenMeyer, line 45: this should return `float` types not `double` types. So `0.0f` instead of `0.0` and `std::stof()` instead of `std::stod()`

BenjamenMeyer (changes requested):
incorrect float implementation

royfalk:
> Safe to merge on top of PR #1036 for play testing?

No. This should come before. #1016 has made significant changes and I now have to modify this and #1036 to match.

engine/src/cmd/upgradeable_unit.cpp
royfalk: Sort of, but it's a lot more complicated. This code is for components, which sub-class from `Component`, not `Unit`. What we need is for `Component` itself to be a sub-class of `Cargo` for this to work. 
Except, even then, you'd probably need similar code to tie the GUI element to the `Component`.

kheckwrecker:
> Please play test with this Assets PR: [vegastrike/Assets-Production#160](https://github.com/vegastrike/Assets-Production/pull/160)

Done. Looks good to me. The only comment I have is that the sensor (radar?) loses focus on occasion. For example, if I hit "n" to navigate to Serenity and start autopiloting there, my VDU would stay on that destination. With this version, it will occasionally switch off to targeting ships that are nearby. It's easy enough to switch back to my planned destination, but it's new behavior, so I thought best to report it. Otherwise, from flying around Cephid-17, looks good. Also, repairing and buying components for my ship worked as expected, as did Basic Repair and Refuel.

engine/src/cmd/upgradeable_unit.cpp
stephengtuggy: Actually, I think this is an example of preferring composition over inheritance, which I have seen recommended in several places recently.

royfalk:
> With this version, it will occasionally switch off to targeting ships that are nearby. It's easy enough to switch back to my planned destination, but it's new behavior, so I thought best to report it. 

I've seen this as well. I'll have to think about it some more.

kheckwrecker:
Discovered a new problem while testing. Loaded my saved game, went to Ship Upgrade and my shields show damaged (at 99%) and the fix costs .70. Click on the fix button, and the credits are deducted, but the fix never happens. Can click on fix repeatedly and the credits go down each time, but the repair never happens. 

Saved game attached.
[Llama.begin.json](https://github.com/user-attachments/files/18915023/Llama.begin.json)

royfalk:
> With this version, it will occasionally switch off to targeting ships that are nearby. It's easy enough to switch back to my planned destination, but it's new behavior, so I thought best to report it.

I think the game can't really lock on serenity because it's too far away (as it should) and when something closer comes along, it auto locks to it. Open an issue so we can make an exception to navigable stuff.

> Discovered a new problem while testing. Loaded my saved game, went to Ship Upgrade and my shields show damaged (at 99%) and the fix costs .70. 

Using your save game, I show damage but no fix button. 

> Click on the fix button, and the credits are deducted, but the fix never happens. Can click on fix repeatedly and the credits go down each time, but the repair never happens.

I'm not sure how you were able to do this, but I think I know what happened. The shield is somehow saved as 149/150/150. That means that it is currently 149 but it is not damaged. The code incorrectly uses the 149/150 percent to check if damaged. I'll fix this and see if it works for you.

WRT to auto pilot, this stuff sometimes happens but is unrelated to the current changes I made to the code. I was unable to reproduce. However, I often overshot Atraxia and the auto pilot had to turn around to get to it. I'm not sure why.
Can you reliably reproduce this with the save game?

kheckwrecker:
> > With this version, it will occasionally switch off to targeting ships that are nearby. It's easy enough to switch back to my planned destination, but it's new behavior, so I thought best to report it.
> 
> I think the game can't really lock on serenity because it's too far away (as it should) and when something closer comes along, it auto locks to it. Open an issue so we can make an exception to navigable stuff.
> 

Will do.

> > Discovered a new problem while testing. Loaded my saved game, went to Ship Upgrade and my shields show damaged (at 99%) and the fix costs .70.
> 
> Using your save game, I show damage but no fix button.
> 
> > Click on the fix button, and the credits are deducted, but the fix never happens. Can click on fix repeatedly and the credits go down each time, but the repair never happens.
> 
> I'm not sure how you were able to do this, but I think I know what happened. The shield is somehow saved as 149/150/150. That means that it is currently 149 but it is not damaged. The code incorrectly uses the 149/150 percent to check if damaged. I'll fix this and see if it works for you.
> 

Okay. Every time it happens now (and it happens regularly), I'm able to fix my shield back to 100%. Please let me know when the fix is ready for testing.

> WRT to auto pilot, this stuff sometimes happens but is unrelated to the current changes I made to the code. I was unable to reproduce. However, I often overshot Atraxia and the auto pilot had to turn around to get to it. I'm not sure why. Can you reliably reproduce this with the save game?

Not reliably, unfortunately. I've seen it once more, but then it stopped in the middle of a flight and autopilot resumed as normal. I'll keep watching and see if I can spot a pattern.

royfalk:
OK, so I think I figured out what happened. When you dock your ship, it automatically refuels and maxes out your capacitors (including FTL) and shields. 
For some reason, the save game was saved with a value of 149/150/150 for shields. 
First value means the current shield strength - that shouldn't be different from the second when we dock.
Second value is the temporary maximum - if the shield generator is damaged, this is lower.
Third value is the maximum. When we repair, first and second are equal to this.

I fixed it so illegal values like this will not impact the game. Please test.

engine/src/resource/resource.cpp
royfalk: Good catch. I wonder if the compiler is smart enough to optimize this.

BenjamenMeyer: it might catch L43, but it can't do the conversion of L41 automatically; though it might treat it like L43. Still, better to just to the right thin to start with. Thanks for fixing!

engine/src/cmd/upgradeable_unit.cpp
BenjamenMeyer: > Sort of, but it's a lot more complicated. This code is for components, which sub-class from `Component`, not `Unit`. What we need is for `Component` itself to be a sub-class of `Cargo` for this to work. Except, even then, you'd probably need similar code to tie the GUI element to the `Component`.

yes, `Unit` certainly is the right place; it would be good if we could have this pointing at `Component` instead so we could do that. I'm sure we'll get there eventually.

@stephengtuggy for the language we work in (C/C++) inheritance is the right tool; we just need to get the codebase to where we can move away from everything using `Unit` and using more specific sub-classes appropriate to the different areas - in this case as @royfalk pointed out it's probably `Component`.

I do a lot of Go for other stuff; and Go uses Composition instead; at the end of the day it isn't really any different other than that the compiler has to enforce interface more. You can still end up with the same exact issues.

BenjamenMeyer (approved):

royfalk:
OK. Merging this now. If something else comes up, we'll deal with it separately.

Issue #801, opened 2023-09-20
cmake fails on non-Debian Linux


stamasd:
After updating the engine tree today, cmake fails to generate a makefile. It's because it can't figure out which version of Debian I'm on. I don't use Debian (Funtoo is my distribution, similar to Gentoo) and there is no file /etc/debian_version

Relevant output:

-- Configuring Packaging for Linux
-- Configuring Debian Packaging
CMake Error at CMakeLists.txt:1313 (FILE):
  FILE failed to open for reading (No such file or directory):

    /etc/debian_version

Looking for /home/silviu/games/vegastrike/Vega-Strike-Engine-Source/build/dependency.list
Still need to generate dependency list
SEND_WARNING!! Unsupported Debian version.

stamasd:
I overcame this by manually configuring with Debian packaging disabled. 
FWIW I did not have to do this in the past.

stamasd:
On further investigation, this may be because in the past month I have installed a version of dpkg on my system (was playing with installing some Debian packages) and cmake finds it and thinks incorrectly that I am on Debian. This issue may only apply to my very special case (a non-Debian system that has dpkg installed anyway) and I thing this issue can be closed.

BenjamenMeyer:
@stamasd So it primarily looks at `/etc/os-release` and the LSB Release information (`lsb_release`). It checks for `dpkg-buildpackage` and `rpmbuild` to determine if its able to build those packages regardless of the host - e.g cross platform package building. So what does your `/etc/os-release` file look like and do you have `lsb_release` installed?

FabioLolix:
On Arch Linux this isn't happening, there is `/etc/os-release`; even when building in clean chroot where lsb_release isn't installed

BenjamenMeyer:
@FabioLolix could you post a copy of what `/etc/os-release` looks like on Arch? That's part of the [FreeDesktop standards](https://www.freedesktop.org/software/systemd/man/os-release.html#:~:text=Description,like%20shell%2Dcompatible%20variable%20assignments) so I'd expect it even when LSB data isn't available.

FabioLolix:
I've got a message on AURWeb and instead I confirm the problem https://aur.archlinux.org/packages/vegastrike-engine?all_deps=1#comment-938453

Will update later this evening

evertvorster:
```
cat /etc/os-release

NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
[...]

PR #805, opened 2023-10-23
Bug Fix: Check for /etc/debian_version (0.8.x)

BenjamenMeyer:
Code Changes:
- [x] Build system change

Issues:
- #801 

Purpose:
Check that /etc/debian_version exists Before relying on it existing.

NOTE: The logic was originally setup for Debian distros; however, when run on a non-Debian distro with the Debian Packaging tools installed this file does not exist and in some cases it can cause a problem.


royfalk (approved):
LGTM

royfalk (approved):
LGTM

FabioLolix:
I tested this is working

BenjamenMeyer:
Regarding the failed CI issues:
- they're not related to the change here
- the Windows one is due to Compiler settings not being properly segrated, MSVC doesn't support two that got specified
- the Manjaro ones are due to a string and int concat via `str+int` which isn't legal.
Let's address those in separate PRs.

PR #807, opened 2023-10-29
Bug Fix: Windows Build on 0.8.x

BenjamenMeyer:
Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [ ] This is a documentation change only

Issues:
- CI build issue discovered in #805 

Purpose:
- What is this pull request trying to do? Fix the Windows CI builds for 0.8.x
- What release is this for? 0.8.x
- Is there a project or milestone we should apply this to? 0.8.x

PR #808, opened 2023-10-29
Bug Fix: Manjaro Compilation on 0.8.x

BenjamenMeyer:

Please answer the following:

Code Changes:

Issues:
-  CI build issue discovered in #805 

Purpose:
- What is this pull request trying to do? Compiler error on trying to concat a string and int.
- What release is this for? 0.8.x
- Is there a project or milestone we should apply this to? 0.8.x

### DIFF
--- a/engine/src/cmd/script/script_statement.cpp
+++ b/engine/src/cmd/script/script_statement.cpp
@@ -77,7 +77,7 @@ void Mission::doIf( missionNode *node, int mode )
 
         int nr_subnodes = node->subnodes.size();
         if (nr_subnodes != 3) {
-            fatalError( node, mode, "an if-statement needs exact three subnodes, not "+nr_subnodes );
+            fatalError( node, mode, sprintf("an if-statement needs exact three subnodes, not %d", nr_subnodes ));
             printf( "nr_of_subnodes: %d\n", nr_subnodes );
 
             assert( 0 );
--- a/engine/src/cmd/script/script_statement.cpp
+++ b/engine/src/cmd/script/script_statement.cpp
@@ -77,10 +77,8 @@ void Mission::doIf( missionNode *node, int mode )
 
         int nr_subnodes = node->subnodes.size();
         if (nr_subnodes != 3) {
-            char buffer[2048];
-            memset(buffer, 0, sizeof(char)*2048);
-            snprintf(buffer, sizeof(char)*2047, "an if-statement needs exact three subnodes, not %d", nr_subnodes );
-            fatalError( node, mode, buffer);
+            std::string error_message = (boost::format("an if-statement needs exactly three subnodes, not %1%") % nr_subnodes).str();
+            fatalError( node, mode, error_message.c_str() );
             printf( "nr_of_subnodes: %d\n", nr_subnodes );
 
             assert( 0 );
--- a/engine/src/cmd/script/script_statement.cpp
+++ b/engine/src/cmd/script/script_statement.cpp
@@ -77,7 +77,10 @@ void Mission::doIf( missionNode *node, int mode )
 
         int nr_subnodes = node->subnodes.size();
         if (nr_subnodes != 3) {
-            fatalError( node, mode, sprintf("an if-statement needs exact three subnodes, not %d", nr_subnodes ));
+            char buffer[2048];
+            memset(buffer, 0, sizeof(char)*2048);
+            snprintf(buffer, sizeof(char)*2047, "an if-statement needs exact three subnodes, not %d", nr_subnodes );
+            fatalError( node, mode, buffer);
             printf( "nr_of_subnodes: %d\n", nr_subnodes );
 
             assert( 0 );
--- a/engine/src/gfx/particle.cpp
+++ b/engine/src/gfx/particle.cpp
@@ -76,9 +76,9 @@ void ParticleTrail::ChangeMax( unsigned int max )
 
 
 
-static inline void UpdateColor( std::vector< QVector, aligned_allocator<QVector> > &vloc,
-                                const std::vector< Vector, aligned_allocator<Vector> > &vvel,
-                                std::vector< GFXColor, aligned_allocator<GFXColor> > &vcol,
+static inline void UpdateColor( std::vector< QVector > &vloc,
+                                const std::vector< Vector > &vvel,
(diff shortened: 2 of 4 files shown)


BenjamenMeyer:
NOTE: The Windows build will fail on this one. That will be fixed by #807.

BenjamenMeyer:
NOTE: I'm mainly working in Golang by day right now...getting lazy about `;`... will fix it.

engine/src/gfx/particle.cpp
```diff
@@ -76,9 +76,9 @@ void ParticleTrail::ChangeMax( unsigned int max )
 
 
 
-static inline void UpdateColor( std::vector< QVector, aligned_allocator<QVector> > &vloc,
-                                const std::vector< Vector, aligned_allocator<Vector> > &vvel,
-                                std::vector< GFXColor, aligned_allocator<GFXColor> > &vcol,
+static inline void UpdateColor( std::vector< QVector > &vloc,
+                                const std::vector< Vector > &vvel,
+                                std::vector< GFXColor > &vcol,
```
stephengtuggy, line 81: Um. I think there was a reason that `aligned_allocator`s were being used here. Possibly an important reason. I suggest that we not change this if we can help it. (Though changing its iterator references to `auto`, further down, is fine by me.)

engine/src/gfx/particle.h
```diff
@@ -50,10 +50,10 @@ struct ParticlePoint
...
 class ParticleTrail
 {
-    std::vector< Vector, aligned_allocator<Vector> > particleVel;
-    std::vector< QVector, aligned_allocator<QVector> > particleLoc;
-    std::vector< GFXColor, aligned_allocator<GFXColor> > particleColor;
-    std::vector< float, aligned_allocator<float> > particleSize;
+    std::vector< Vector > particleVel;
+    std::vector< QVector > particleLoc;
+    std::vector< GFXColor > particleColor;
+    std::vector< float > particleSize;
```
stephengtuggy, line 56: Can we continue to use the `aligned_allocator`s here as well?

engine/src/cmd/script/script_statement.cpp
```diff
@@ -77,7 +77,10 @@ void Mission::doIf( missionNode *node, int mode )
 
         int nr_subnodes = node->subnodes.size();
         if (nr_subnodes != 3) {
-            fatalError( node, mode, "an if-statement needs exact three subnodes, not "+nr_subnodes );
+            char buffer[2048];
+            memset(buffer, 0, sizeof(char)*2048);
+            snprintf(buffer, sizeof(char)*2047, "an if-statement needs exact three subnodes, not %d", nr_subnodes );
+            fatalError( node, mode, buffer);
```
stephengtuggy, line 83: How about just using `boost::format(...).str()`? Something along the lines of:

```suggestion
            std::string error_message = (boost::format("an if-statement needs exactly three subnodes, not %1%") % nr_subnodes).str();
            fatalError( node, mode, error_message.c_str() );
```

stephengtuggy: Although what you have here probably works too. It's just that Windows / Visual Studio might complain about it.

stephengtuggy (changes requested):

engine/src/gfx/particle.cpp
BenjamenMeyer: I removed them as the types being used were not compliant with the requirements for the `aligned_allocator`. I'm not familiar enough with building a type that would be. Happy to have a PR against my branch to make it work properly or we can address it again in another PR.

engine/src/gfx/particle.h
BenjamenMeyer: `float` isn't a valid type for using with `aligned_allocator`.

engine/src/cmd/script/script_statement.cpp
BenjamenMeyer: oh yeah - MSVC will probably complain about it since I didn't use the `_s` versions. Happy to use your suggestion.

engine/src/gfx/particle.h
stephengtuggy: @BenjamenMeyer Do you have a reference link for that? I find that surprising.

stephengtuggy: I believe that [this](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/aligned.h) is the aligned_allocator intended to be used here. Not the C++17 `aligned_alloc`, if that is what you were thinking of. (Though we might eventually switch to the latter, if we decide to start using C++17 features at some point.)

stephengtuggy: > @BenjamenMeyer Do you have a reference link for that? I find that surprising.

Or what was the specific error message(s)?

BenjamenMeyer: @stephengtuggy the compilers refused to even build it as the types did not have the correct support for that functionality.

stephengtuggy: OK. I think I found some of the error messages you were getting. Stuff like "Cannot use _Tp_Alloc here because it does not name a type" ? Among other things?

I got some of the same errors when I tried to build the project on Ubuntu Mantic for the first time. It looks like we're going to have to deal with this issue more and more as we move to the newer Linux versions.

I'm wondering if it's time to take a step back; rethink what the code was originally trying to do, using compiler-specific constructs and such; and see if there's a better way to accomplish that same goal, using C++11 standard constructs like some of those mentioned [here](https://en.cppreference.com/w/cpp/memory/align).

BenjamenMeyer: IMHO we need to stick to standard C++ and not do the compiler specific route. There's no need anymore. Yeah, when the project first started one needed to do so because so much was custom to compilers and what is now standardized across compilers wasn't.

More often than not the compiler does the same right thing without hints. So I think we'll be good if we just let it do its thing.

stephengtuggy: Definitely, let's stick to standard C++ as much as possible. However, I'm not convinced that we can get by without the alignment hints in this case.

I believe my latest code fixes the compile errors we were getting. Hopefully, those changes can be incorporated into this PR, for a solution that we can all be happy with.

stephengtuggy (approved):
LGTM!

PR #810, opened 2023-11-10
Draft - Lib Component

royfalk:
Lib component decouples the code base into discrete components - drive, life support, reactor, etc.

Code Changes:

Issues:
- TBD


engine/src/components/component.cpp
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+double random20()
+{
+    if(std::rand() < 0.2) {
+        return std::rand();
+    }
+
+    return 1.0;
```
BenjamenMeyer, line 44: this should probably be:
```c++
double random_value = std::rand();
if (random_value < 0.2) {
    return random_value;
}
return 1.0;
```
As it is, if L40 passes then a new random number will be used and it could still be greater than or equal to 0.2, thereby likely defeating the check.

BenjamenMeyer (changes requested):
one little change and we should be good.

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -1292,7 +1293,7 @@ static Unit *ChooseNavPoint(Unit *parent, Unit **otherdest, float *lurk_on_arriv
     int whichlist = 1;  //friendly
     std::string fgname = UnitUtil::getFlightgroupName(parent);
 
-    bool insys = (parent->GetJumpStatus().drive == -2) || fgname.find(insysString) != std::string::npos;
+    bool insys = (parent->jump.drive == -2) || fgname.find(insysString) != std::string::npos;
```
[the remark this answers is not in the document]
royfalk: I did that, and then I realised I don't actually know what these value mean and need to dig into it to figure it out/guess. And then I reverted and kept the original values.

engine/src/cmd/ai/firekeyboard.cpp
```diff
@@ -2178,9 +2176,11 @@ void FireKeyboard::Execute() {
...
     }
-    static bool actually_arrest = XMLSupport::parse_bool(vs_config->getVariable("AI", "arrest_energy_zero", "false"));
-    if (actually_arrest && parent->energyRechargeData() == 0) {
+
+    // TODO: something with this. It's very unlikely reactor will be 0
+    /*static bool actually_arrest = XMLSupport::parse_bool(vs_config->getVariable("AI", "arrest_energy_zero", "false"));
+    if (actually_arrest && parent->energy_manager.GetReactorCapacity() == 0) {
         Arrested(parent);
-    }
+    }*/
```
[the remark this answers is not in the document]
royfalk: Even if they did, if we run out of reactor before we get to a base, life support goes out and we die (to be implemented). This straight out doesn't make sense on any level.

engine/src/cmd/basecomputer.cpp
```diff
@@ -5738,39 +5753,39 @@
...
-        float maxshield = playerUnit->totalShieldEnergyCapacitance();
+        float maxshield = totalShieldEnergyCapacitance(playerUnit);
         if (shields_require_power) {
             maxshield = 0;
         }
         PRETTY_ADDU(statcolor + "Minimum time to reach full overthrust speed: #-c",
                 playerUnit->getMass() * uc.max_ab_speed() / playerUnit->limits.afterburn, 2, "seconds");
         //reactor
-        float avail = (playerUnit->maxEnergyData() * RSconverter - maxshield * VSDM);
+        float avail = (playerUnit->energy_manager.GetReactorCapacity() * RSconverter - maxshield * VSDM);
```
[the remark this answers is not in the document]
stephengtuggy: Agreed

engine/src/components/component.cpp
stephengtuggy: @BenjamenMeyer You are correct, AFAICT.

engine/src/cmd/ai/firekeyboard.cpp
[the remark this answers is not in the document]
stephengtuggy: Maybe this is one of those "fence" situations. I would like to ask some of the long-term project members whether they remember what this code was for.

stephengtuggy (commented):
A couple minor things to change, and I haven't finished looking through all of this code in detail yet. But overall, I really like the direction this is headed.

engine/src/components/component.h
```diff
@@ -0,0 +1,61 @@
...
+
+class Component
+{
+protected:
+    bool installed = true;
+    bool enabled = true;
+    bool damaged = false;
+public:
+    Component();
+    virtual ~Component() {}
```
stephengtuggy, line 40: Nice. I see you have the virtual destructor here, for a class that's intended to be inherited. Excellent.

engine/src/components/armor.cpp
```diff
@@ -0,0 +1,170 @@
...
+        double armor_value = UnitCSVFactory::GetVariable(unit_key, 
+                                               armor_facets[i], 0.0);
+        facets[i].health.Set(armor_value);
+    }
+}
+
+void Armor::SaveToCSV(std::map<std::string, std::string>& unit) const {
+    for(int i=0;i<8;i++) {
+        unit[armor_facets[i]] = std::to_string(facets[i].health.Value());
+    }
```
BenjamenMeyer, line 78: is this only CSV? or any format? Could we just name this `Save`?

engine/src/components/component.h
```diff
@@ -28,34 +28,64 @@
...
+    Component(std::string upgrade_name, double mass, 
+              double volume, bool integral);
 
-    void Install() { installed = true; }
-    void Uninstall() { installed = false; }
-    bool Installed() { return installed; }
+    // Load from units dictionary
+    virtual void Load(std::string upgrade_key, std::string unit_key);      
+    
+    virtual void SaveToCSV(std::map<std::string, std::string>& unit) const = 0;
```
BenjamenMeyer, line 67: is this doing something specific to CSV files? Or does it work for JSON and other formats too?
E.g could we just call it `Save`? Or does it need to only apply to CSV files?

royfalk:
This PR got out of hand. I'll wait until @BenjamenMeyer and I finish discussing something and then I'll close it.
I'll port the code here in manageable parts.

engine/src/components/armor.cpp
royfalk: All our saves are CSV only. A proper JSON save would be hierarchical. 
- Ship
   - Component (e.g. radar)
      - Damage to specific attribute. 

The current system saves everything as one big struct. When you repair or downgrade, it checks the stock attributes. It works but because everything is in one file, you can't do any unit tests.

engine/src/components/component.h
royfalk: All our saves are CSV only.

BenjamenMeyer: It'd be nice to change that up. I'd like to change the saved game format to use JSON too, perhaps also use a zip like format similar to ODF. But that's another refactor

stephengtuggy: Since CSV is an implementation detail that could change, it's probably better to describe it more generically. I think `Save` or `Saves` is great, personally.

stephengtuggy:
> This PR got out of hand. I'll wait until @BenjamenMeyer and I finish discussing something and then I'll close it.
> I'll port the code here in manageable parts.

Thanks. I was not looking forward to reviewing this. LOL

royfalk:
As noted, this PR has been for the most part implemented in multiple PR's.  Closing.

PR #817, opened 2023-11-24
Enhancement: Update the GH Actions

BenjamenMeyer:
- move checkout to latest (v4.1.1)
- move upload artifacts to latest (v3.1.3)
- move codeql-action to v2 (best version I can find)
- update the scorecard analysis and move version to latest


.github/workflows/Windows-CI.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
         # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - name: checkout
-        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 #v3.5.3
+        uses: actions/checkout@v4.1.1
```
stephengtuggy, line 110: Can we consider using the full SHA, together with a comment identifying which version it is? I.e.:

```suggestion
        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
```

BenjamenMeyer: SHA is harder to lookup. Is there a specific reason? Using the tag allows easier updates.

BenjamenMeyer: Ahh... https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash got it

stephengtuggy: @BenjamenMeyer Yep.

stephengtuggy: The key seems to be to find the source repository for the GitHub Action in question. Usually there will be a link to it in the sidebar of the GH Action's Marketplace page. (E.g., https://github.com/marketplace/actions/ossf-scorecard-action links to https://github.com/ossf/scorecard-action ) Then you can look at the Releases (or sometimes just Tags) for that repo; find the latest one; and copy/paste its hash.

If that makes sense.

BenjamenMeyer: updated the rest of them

stephengtuggy: Nice! Looks good!

stephengtuggy (approved):
Looks good to me!

.github/workflows/codeql.yml
```diff
@@ -21,20 +21,29 @@ jobs:
...
-        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 #v3.5.3
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
         with:
           fetch-depth: 2
 
+      - name: Download benchmark bundle
+        env:
+           GH_TOKEN: ${{ github.token }}
+        run: |
+           gh release download -R octo-org/codeql-bundle --pattern 'codeql-bundle.tar.gz'
```
stephengtuggy, line 32: Hmm. This line seems to be erring out.

I think `octo-org` might just be an example organization, or something. I think this was an example of how to add your own custom CodeQL rules to the CodeQL Standard Library. I'm still trying to figure out where one gets the latter. Is it this? https://github.com/github/codeql-coding-standards

stephengtuggy: Oh, OK. Looks like it's this: https://github.com/github/codeql-action/releases

BenjamenMeyer: I was copying their example....guess its more example than working functionality...

BenjamenMeyer: okay - updated to use the right things for those downloads.

.github/workflows/codeql.yml
```diff
@@ -21,20 +21,30 @@ jobs:
...
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
         with:
           fetch-depth: 2
 
+      - name: Download benchmark bundle
+        env:
+           GH_TOKEN: ${{ github.token }}
+        # download the latest version
+        run: |
+           gh release download -R github/codeql-action --pattern 'codeql-bundle.tar.gz'
```
BenjamenMeyer, line 33: We can specify a tag to lock to a version or leave out the tag and always get the latest release.
It's not clear from the command-line docs if `gh release download` will work with the SHA hashes of the release so skipping that and just letting it use the latest release.

stephengtuggy (approved):

PR #843, opened 2024-04-21
Bug Fix: Mac/Windows Release Builds

BenjamenMeyer:
- Rename the workflows to match reality (Release not CI)
- Remove the final publisher as it does not support Mac/Win, only Linux. A new publisher will be needed to support those platforms


stephengtuggy (approved):
LGTM!

PR #846, opened 2024-04-23
0.9.x Updates 2024-04-18 to 04-22 by SGT

### LINKED ISSUE #837: Game crashes in `GFXDeleteTextures(...)`
So I'm definitely seeing **a** crash with the 0.9.x code. Pretty consistently, too. It occurs in a different part of the code -- in `GFXDeleteTextures(...)`. So probably a different bug. But I'm pretty sure that the Textures crash was one that I fixed -- or at least tried to -- in the ill-fated PR #754 . I will see if I can fix it again (without all the other problems that #754 came with LOL).

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/716#issuecomment-2065622877_

### LINKED ISSUE #844: Our vcpkg setup is pulling the backdoored version of liblzma/xz!
We need to update the vcpkg baseline we're using, ASAP, so that we stop getting the backdoored v5.6.0 or v5.6.1 of liblzma and xz.

stephengtuggy:

Please answer the following:

Code Changes:
- [X] (Partially) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
Fixes #844 for 0.9.x
Attempts to fix #837 for 0.9.x

Purpose:
- What is this pull request trying to do? Primarily, fix the backdoored liblzma issue, ASAP. Secondarily, make various improvements to the CI workflows, and also try to fix the crashes in `GFXDeleteTextures(...)`.


BenjamenMeyer (approved):

engine/src/gfx/sprite.h
```diff
@@ -47,7 +51,7 @@ class VSSprite {
     bool isAnimation;
 
 //For private use only
-    VSSprite() : surface(0) {
+    VSSprite() : surface(nullptr), isAnimation(false), rotation(0.0F), maxs(0.0F), maxt(0.0F), heighto2(0.0F), widtho2(0.0F), xcenter(0.0F), ycenter(0.0F) {
```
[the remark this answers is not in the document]
stephengtuggy: OK. Thanks. I think I might just leave it like this for now, in this particular case, since the code is already written. But I'll keep that in mind for the future.

.github/dependabot.yml
stephengtuggy, line 1: I had a runaway find-and-replace process go nuts on me. Garbled the contents of a number of files before I realized what was going on. I'm still in the process of cleaning that up, which I plan to finish before merging this PR.

engine/src/vs_logging.cpp
```diff
@@ -32,7 +34,7 @@
 #include <boost/log/expressions.hpp>
 #include <boost/log/sources/severity_logger.hpp>
 #include <boost/log/sources/record_ostream.hpp>
-//#include <boost/log/sinks/sync_frontend.hpp>
+#include <boost/log/sinks/async_frontend.hpp>
```
stephengtuggy, line 37: Note to self: I need to go back to sync_frontend rather than async, at least for the time being.

engine/vcpkg.json
```diff
@@ -13,14 +14,15 @@
         "boost-chrono",
         "boost-atomic",
         "boost-assign",
+        "boost-format",
+        "egl-registry",
```
stephengtuggy, line 18: Not sure if we even need egl-registry

engine/src/gfx/mesh_gfx.cpp
```diff
@@ -458,6 +460,7 @@ Mesh::~Mesh() {
                         resetIter = true;
                         cachedHashItem = hashItem + 1;
                     }
+                    //delete hashItem;
```
stephengtuggy, line 463: Note to self: This does not need to be here

.github/dependabot.yml
stephengtuggy: There. It's fixed now.

engine/src/gfx/mesh_gfx.cpp
stephengtuggy: Removed

engine/src/vs_logging.h
```diff
@@ -13,27 +15,34 @@
...
-#include <boost/smart_ptr/shared_ptr.hpp>
-#include <boost/format.hpp>
-//#include <boost/log/sources/logger.hpp>
-#include <boost/log/sources/severity_logger.hpp>
-#include <boost/log/sources/record_ostream.hpp>
-#include <boost/log/sinks/sync_frontend.hpp>
-#include <boost/log/sinks/text_file_backend.hpp>
-#include <boost/log/sinks/text_ostream_backend.hpp>
-#include <boost/filesystem.hpp>
+#include "boost/move/utility_core.hpp"
```
BenjamenMeyer, line 31: headers should be in `<>` enclosures not `""` (double quote) enclosures.

stephengtuggy: Hmm. Really? That didn't compile correctly in the case of at least one source file.

engine/src/vs_logging.h
```diff
@@ -48,41 +57,63 @@ enum vega_log_level {
...
+    do {                                                                                                                                        \
+        VegaStrikeLogging::VegaStrikeLogger::instance().LogAndFlush(VegaStrikeLogging::vega_log_level::log_level, (log_message));               \
+    } while (false)
+#define VS_LOG_FLUSH_EXIT(log_level, log_message, exit_code)                                                                                    \
+    do {                                                                                                                                        \
+        VegaStrikeLogging::VegaStrikeLogger::instance().LogFlushExit(VegaStrikeLogging::vega_log_level::log_level, (log_message), (exit_code)); \
     } while (false)
 
-class VegaStrikeLogger {
+struct VegaStrikeLogger {
```
BenjamenMeyer, line 82: we should keep this as a `class`; the only real difference is default scoping.
`struct` -> default public
`class` -> default private

stephengtuggy: OK. Yeah, I'll change it back.

BenjamenMeyer (commented):
a few comments to keep in mind; but generally looking good

stephengtuggy:
> a few comments to keep in mind; but generally looking good

@BenjamenMeyer thanks.

engine/src/vs_logging.h
BenjamenMeyer: difference between them is that `<>` is stricter so the headers have to be in the path; where `""` is a little more liberal and includes the project paths and local directory. It's better to use the stricter set and configure the compiler properly to ensure the right sets of headers are easily found.

stephengtuggy: OK. I'll see if I can get that to work.

engine/src/vs_logging.h
stephengtuggy: Done ✔️

engine/src/vs_logging.cpp
stephengtuggy: Done ✔️

stephengtuggy:
Well, that last commit was a winner lol.

With regard to the logging system: A few days ago, I thought I would try switching to asynchronous logging, to see if that increased performance. Well, maybe it did and maybe it didn't. But what I primarily noticed was that the last message logged before program shutdown, never made it to the console. No matter what I tried.

So I am switching back to synchronous sinks. Maybe I'll try the async version again some other time.

stephengtuggy:
Also: Rocky Linux 8.9 keeps throwing one error after another. I think I'm going to allow it to fail for now. And maybe stop building it at all in the CI, until further notice.

engine/src/vs_logging.h
stephengtuggy: Seems to be working that way, with angle brackets instead of quotes. Resolving.

stephengtuggy:
> Also: Rocky Linux 8.9 keeps throwing one error after another. I think I'm going to allow it to fail for now. And maybe stop building it at all in the CI, until further notice.

Done ✔️

stephengtuggy:
I have switched back to using synchronous logging sinks. The logging system seems to be working again, more or less as before.

stephengtuggy:
... And #837 is not fixed. Damn. Oh well, I need to get this merged to fix the other issue (#844 ).

I'm going to go ahead and merge.

PR #860, opened 2024-05-07
Task core components

royfalk:
This is part of the lib component tasks.
This commit introduces Component, EnergyConsumer, EnergyContainer and Reactor.
These are not yet connected to the engine. Please focus on reviewing the classes themselves, as they are fairly major.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Minimal tests to check changes to UnitCsvFactory. The rest is not yet connected.

Issues:
- Because I'm pulling changes by files, there are a few things that are going to be used in future PRs.


engine/src/components/component.h
```diff
@@ -0,0 +1,92 @@
...
+enum class ComponentType {
+    Hull, 
+    Armor, 
+    Shield,
+    Drive
+};
+
+class Component
+{
+protected:
```
BenjamenMeyer, line 52: just to note, we should put `public` information first, then `protected`, then `private`. This helps with ABI compatibility, not that we're really concerned about that right now.

engine/src/components/energy_consumer.h
```diff
@@ -0,0 +1,52 @@
...
+// -*- mode: c++; c-basic-offset: 4; indent-tabs-mode: nil -*-
+
+#ifndef ENERGYCONSUMER_H
+#define ENERGYCONSUMER_H
+
+#include "energy_container.h"
+
+class EnergyConsumer {
+    EnergyContainer *source;
+    bool partial; // Can power consumer with less energy than requested
```
BenjamenMeyer, line 35: do we want/need these to be public? either way, we should make an explicit declaration of their visibility - better to be explicit than implicit with `class`es

engine/src/components/reactor.h
```diff
@@ -0,0 +1,80 @@
...
+#define REACTOR_H
+
+#include "component.h"
+#include "energy_container.h"
+#include "energy_consumer.h"
+#include "resource/resource.h"
+
+class EnergyManager;
+
+class Reactor: public Component, public EnergyConsumer
```
BenjamenMeyer, line 38: are all `EngeryConsumer` objects going to be `Component`s? I would think so. If so we might want to make `EnergyConsumer` a child of `Component` and then just have a single inheritance here.

BenjamenMeyer (commented):
few comments for consideration; overall I like where its going but I'll need to take another more detailed pass too

engine/src/components/component.h
royfalk: For some reason I assumed the reverse. I'll fix this here, though I can't promise for all of my work.

engine/src/components/energy_consumer.h
royfalk: These are implicitly private. I'll do that.

engine/src/components/reactor.h
royfalk: Not necessarily. A component is an upgrade-able and damageable module. One potential example of a consumer but not component is life support. 
I prefer to finish lib components and then do a cleanup pass for something like this. Constraining things now and finding out I need to revert is one of the reasons I've been working on this for six months.

engine/src/components/component.h
BenjamenMeyer: Changing things changes the offsets. Keeping public data first keeps the offsets to it the same, and new public goes at the end, pushing the protected and private data further away from the zero point.

If we really wanted a public binary ABI then we'd look at doing another method (what Qt does), but it's not really important just good habit for when it is.

royfalk: So, I dug deeper and while most agree on public/protected/private order, fields/constructors/methods is less so.
Since most of the code is fields/constructors/methods, I'll stick with that but reverse the visibility as I go along as these are already not consistent.

royfalk:
If no one has any more comments and my answers satisfy, it is ready for merge.

engine/src/components/energy_consumer.cpp
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+// -*- mode: c++; c-basic-offset: 4; indent-tabs-mode: nil -*-
+
+#include "energy_consumer.h"
+
+double EnergyConsumer::simulation_atom_var = 0.1;
```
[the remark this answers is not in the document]
stephengtuggy: I think this has the potential to cause a lot of confusion, having two (or more) different instantiations of `simulation_atom_var` in different scopes. But I suppose we can leave it for now. Approving.

stephengtuggy (approved):

stephengtuggy:
@BenjamenMeyer did you still want to review this PR in more detail? Or is it ready for merge?

BenjamenMeyer:
I'll try to take a look tonight

engine/src/components/energy_consumer.cpp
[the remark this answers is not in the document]
BenjamenMeyer: @royfalk thanks for the explanation; it'd probably be good to have some documentation about it in the code; but not going to require it for this PR to go through. Would you please file a follow-up PR with a small comment about it?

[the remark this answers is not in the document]
BenjamenMeyer: I do like your idea of the follow-on to have a core library that way too, BTW, and certainly agree we need to take a calculated, stepped approach to getting there.

engine/src/components/reactor.h
BenjamenMeyer: @royfalk that's fine - just thinking of the double inheritance and whether it's possible to avoid it. If not, okay, and we can always refactor it later too. Thanks for the explanation.

BenjamenMeyer (approved):

PR #873, opened 2024-09-02
Refactor python code

royfalk:
This PR touches some python code in the engine:
1. Create a unit test to call a python function as reference to future coders
2. Replace project specific macros with original boost python macros and functions

Code Changes:

Issues:
- The code is not portable and needs some polish before it can be merged.


BenjamenMeyer (approved):

engine/CMakeLists.txt
```diff
@@ -1743,6 +1744,8 @@ IF (USE_GTEST)
 
     TARGET_LINK_LIBRARIES(
         ${TEST_NAME}
+        ${Boost_LIBRARIES}
+        ${Python3_LIBRARIES}
```
royfalk, line 1748: I wasn't able to make this support python 2, which we still support.

I admit I'm not very good with CMake. We have some code higher up which determines whether we're using ver 2 or 3 but I wasn't able to make it work here.

This is really the show stopper IMO for this PR and where I need your help.

evertvorster:
Most distrobutions have dropped support for Python 2. Maybe it is time?

engine/CMakeLists.txt
BenjamenMeyer: 0.9.x does not support Python2. No need to be concerned about it.
0.8.x was the last version to support Python 2.

PR #875, opened 2024-09-11
Add support for ftl and jump drives 

royfalk:
This is part of the lib component refactor effort.
I mostly dealt with Jump Drive and a new component - FTL Drive.
As it is a new component, FTL  Drive does not have presence in the game and is currently _integrated_. This means you can't sell, buy, upgrade, repair, etc.

Code Changes:

Issues:
- Can't repair components in flight. This is an older issue I just identified. I'll open an issue and get to it later.
- FTL Drive can be damaged. If damage = 100, it is destroyed. This doesn't seem to do anything yet though.
- The whole damage aspect of these components has not been tested in the game. It's very hard to do and damage in unit tests does not test anything beyond the very narrow stuff defined in the component.


engine/src/cmd/ship.cpp
```diff
@@ -1,7 +1,7 @@
 /*
  * ship.cpp
  *
- * Copyright (C) 2001-2023 Daniel Horn, Benjaman Meyer, Roy Falk, Stephen G. Tuggy,
+ * Copyright (C) 2001-2023 Daniel Horn, Benjamen Meyer, Roy Falk, Stephen G. Tuggy,
```
BenjamenMeyer, line 4: :smiley_cat:

engine/src/cmd/unit_generic.cpp
```diff
@@ -1255,36 +1262,20 @@ void Unit::DamageRandSys(float dam, const Vector &vec, float randnum, float degr
...
-        } else if (randnum >= cargovolume_damage_prob) {
-            CargoVolume *= dam;
-        } else if (randnum >= upgradevolume_damage_prob) {
-            UpgradeVolume *= dam;
-        } else if (randnum >= cargo_damage_prob) {
+        std::random_device dev;
+        std::mt19937 rng(dev());
+        std::uniform_int_distribution<std::mt19937::result_type> dist20(0,19); // distribution in range [1, 6]
+
+        switch(dist20(rng)) {
```
BenjamenMeyer, line 1269: Qs:

1. Can we get some documentation on this?
2. Did you mean to skip the `break` and allow fall through on cases 1-7?
3. Everything seems to add some damage; what about the default case? I see it commented-out - we should have some documentation on why it's not needed but even then it'd probably be good to have it and log it just in case.

BenjamenMeyer (approved):

engine/src/cmd/unit_generic.cpp
royfalk: 1. It's a simple algorithm to replace the complex one before. Random 20 (for now). 1-8 damages something by calling component->Damage();
2. **Yikes! Definitely not.** This is why we review stuff. The issue here of course is we can't really test this in the game. We need to pull this out of unit_generic and add a unit test to do it properly. 
3. It's a simple mechanism to ensure components don't get damaged every time we damage the hull.

BenjamenMeyer: > 1. It's a simple algorithm to replace the complex one before. Random 20 (for now). 1-8 damages something by calling component->Damage();

more thinking of the next time we come to this code and wanting to understand it and why it's there; also per :two: and why the `break` should *not* be there so it doesn't get added accidentally by someone well-meaning in the future (e.g solving linter issues) so having some comments in the code to resolve that

PR #876, opened 2024-09-11
Add support for specifying monitor to run on

### LINKED ISSUE #863: VS in fullscreen mode pops up on the wrong screen.
With the 0.8.x version of the engine, VS in fullscreen mode pops up on Monitor 0 (Primary monitor)

With the git master version, it pops up on the leftmost screen, no matter which monitor is set to be the primary monitor.

I believe the default for SDL is to fullscreen on monitor 0, and I am not sure what changed between the two game versions to effect this change in behaviour. 

Pie in the sky forward looking best fix would be to read the display configuration while starting up, and compare it with the last display configuration that was set. If there is a difference, maybe have a popup window with the detected displays and a simple interface to pick one to be used. Then, save this configuration for the next time. 

With the quickest Google search, I found two interesting links on how to manipulate the SDL startup window:
https://discourse.libsdl.org/t/getting-display-information-before-creating-a-window/21146

and

https://stackoverflow.com/questions/41745492/sdl2-how-to-position-a-window-on-a-second-monitor

royfalk:
This should fix #863

To test with multiple monitors, edit vegastrike.config as follows:
<section name="general">
         _<var name="screen" value="1" />_
	<var name="times_to_show_help_screen" value="-1"/>

Code Changes:
- [x] Have the PR Validation Tests been run? Minimal tests to check game runs.


evertvorster:
Hi there!

I have applied this change to the master source, but cannot see where to set the preferred monitor. 
Vsettings did not have an extra option, and just running vs pops up on the leftmost monitor. 

How do I use this modification to specify which monitor to pop up on?

engine/src/configuration/configuration.cpp
```diff
@@ -42,6 +42,8 @@ Configuration::Configuration() {
 * behavior/range bounding is complicated/necessary-for-stability abd additional validation code is required for checking/adjusting user inputs
 */
 void Configuration::OverrideDefaultsWithUserConfiguration() {
+    general_config.screen = GetGameConfig().GetInt16("general.screen", general_config.screen);
```
BenjamenMeyer, line 45: we should have some kind of method for outputting which monitor is which value.
Perhaps a `vegastrike-engine --list-monitors` which just prints out something like:
```
[0] Primary Display (ASUS...)
[1] Built-in LCD
...
```
where the listing is dynamically generated.
Something like:
```c++
const char * SDL_GetDisplayName(int displayIndex);

for (int i=0; i < SDL_GetNumVideoDisplays(); ++i) {
     const char* name = SDL_GetDisplayName(i);
}
```
only it would have to give a warning about disconnecting/reconnecting displays potentially receiving different values

Looks like like SDL3 is solving some of that but does give that warning with their new `DisplayID` type:
https://wiki.libsdl.org/SDL3/SDL_DisplayID
https://wiki.libsdl.org/SDL3/SDL_GetDisplays

We might want to think through that a little. Should we use the display name (and then compare it to find the right display)? that probably shouldn't change...but it's hard to say.

BenjamenMeyer (approved):
We can start here, but we will want to think through identifying displays a little for consistency given the warnings from SDL3. (SDL2 doesn't list anything but also doesn't have an official Display ID type either.)

evertvorster:
I'm still at a loss as to what to modify in my vegastrike.conf file to get this to pop up on the proper monitor. I would really like to play test this one as well. :)

royfalk:
Sorry. I kept checking the issue and didn't think to also check the PR. 
In the config file, in the general section, add a line with "screen" and a value of 1.
Hopefully this should work:

```

   <section name="general">
        <var name="screen" value="1" />
	<var name="times_to_show_help_screen" value="-1"/>
	<var name="dockingtime" value="20"/>
	<var name="remember_savegame" value="false"/>
	<var name="load_last_savegame" value="false"/>
	<var name="write_savegame_on_exit" value="false"/>

```

engine/src/configuration/configuration.cpp
royfalk: We can, but this is a lot of work for no real reason. For people with 2 monitors it's either 0 or 1. If you're in the wrong monitor, change the number.

Yeah, we could do what you said or even add support to the game to switch monitor in the game menu. But it's overkill and there are more urgent tasks. 

This rebuke is mostly for me. I was halfway through the design in my head while opening issues for _upgrade to SDL 3_ and _this feature_ before I reigned myself in.

BenjamenMeyer: I'm just thinking through the help desk support side of it and the confusion that may happen and how can we make it a good user experience. At minimum, if we don't do this here, then it should be filed as its own GH Issue with a `help wanted` label for someone to pickup.

evertvorster:
Thanks for the instruction. 

After some testing, for my configuration I needed a screen value of 2, and then it pops up on the proper monitor. 

There is still a little room for improvement in screen handling, but I will put those in different issues so they can be dealt with separately. 
We should make the change self-documenting. Is there a way that we can add a comment and this line with the default monitor to the vegastrike.config file in Assets?

engine/src/configuration/configuration.cpp
evertvorster: Yes, outputting which monitors are available would be a nice to have. Maybe just display it in the console output when the game starts?

evertvorster: I have tested the changes here, and now it is possible to specify which monitor the game starts up on.

evertvorster (approved):
For this change, it does what it is intended to do. 
We should modify the vegastrike.config in Assets to have this line to guide users to specify the monitor.

PR #802, opened 2023-10-07
Refactor cmake

BenjamenMeyer:
Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [ ] This is a documentation change only
- [x] Build System Change

Issues:
- #56 

Purpose:
- refactor the CMake configuration to use the root of the repo as its base instead of the `engine` directory as its base; this will allow us to start factoring out the libraries easier without having to resort to any weird CMake-foo.

PR #889, opened 2024-10-04
Refactor: CMake Support

BenjamenMeyer:
- move the core of the CMake Support to the root
- move our CMake modules into a dedicated directory
- move all binary targets to a single build directory
- disable old CPU optimizations for CPUs that are generally too old to care about any more; compilers tend to do the right thing now
- remove unnecessary CMake directives sets; we no longer need to UNSET variables, the CMake Modules should take care of that
- Remove all the stuff previously commented out on this branch
- Update Windows GH Actions
- Update README

Replaces #802


engine/CMakeLists.txt
```diff
@@ -248,17 +113,6 @@ OPTION(INSTALL_GTEST "Should we download and install GTest?" ON)
...
-# Provide boolean options for enabling various cpu optimizations.
-OPTION(CPUAMD_k8  "Enable AMD K8 optimizations (Athlon through athlon64)" OFF )
-OPTION(CPUAMD_k9  "Enable AMD K9 (sse3) optimizations (Athlon64 AM3) " OFF )
-OPTION(CPUAMD_k10  "Enable AMD K10 optimizations (Athlon2)" OFF )
-OPTION(CPUAMD_vishera  "Enable AMD Vishera optimizations" OFF )
-OPTION(CPUINTEL_p4 "Enable Intel P4 optimizations (1st gen P4's)" OFF )
-OPTION(CPUINTEL_prescott  "Enable Intel P4 (sse3) optimizations" OFF )
-OPTION(CPUINTEL_nocona "Enable Intel P4 (64bit) optimizations" OFF )
-OPTION(CPUINTEL_core2 "Enable Intel Core2 optimizations" OFF )
-OPTION(CPUINTEL_native "Enable host-specific optimizations for this machine" OFF )
```
BenjamenMeyer, line 260: there are mainly older processors that I would be shocked if someone still had around and was using them to game; also compilers are generally smart enough now to figure out the right optimizations.

royfalk: I'm surprised we dropped Sparc and Alpha.

royfalk (approved):
I'm no CMake expert. I mostly hack changes as needed without understanding the big picture. However, I like the separation, if only because it started getting very big.
LGTM.

engine/CMakeLists.txt
BenjamenMeyer: > I'm surprised we dropped Sparc and Alpha.

probably never got migrated to CMake...but yeah...

script/docker-entrypoint.sh
```diff
@@ -45,5 +45,5 @@ else
 fi
 
 pushd build
-GTEST_OUTPUT=xml:test-results ctest -V
+GTEST_OUTPUT=xml:$(pwd)/test-results ctest -V
```
BenjamenMeyer, line 48: tests were getting dumped under the `engine` directory from what I can tell; this should enforce them being in where we expect them to be

BenjamenMeyer:
@royfalk so my goal with the CMake refactor is to start splitting up the source tree a little and enable the ability to separate out things into libraries and binaries in a more flat structure. This first one is just to refactor it into a more standard CMake approach where the CMakeLists.txt lives at the Project Root.

BenjamenMeyer:
We probably should figure out how to make it use `nmake` so things are consistent. `nmake` comes packaged with VS, and uses Makefiles just like GNU Make does; it's just Microsoft's variant of the family and works really well for build systems.

BenjamenMeyer:
Okay - looks like Windows is failing due to:
```
++ OpenGL found
CMake Error at D:/a/_temp/1826076666/cmake-3.29.2-windows-x86_64/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find GLUT (missing: GLUT_glut_LIBRARY GLUT_INCLUDE_DIR)
Call Stack (most recent call first):
  D:/a/_temp/1826076666/cmake-3.29.2-windows-x86_64/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  D:/a/_temp/1826076666/cmake-3.29.2-windows-x86_64/share/cmake-3.29/Modules/FindGLUT.cmake:168 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  v/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
  engine/CMakeLists.txt:318 (FIND_PACKAGE)
  ```

BenjamenMeyer:
Okay...I understand what is wrong with the Windows builds now...but I need a Windows system I can test on to figure out why it's broken - since I moved the PROJECT() to the root and the various Find*.cmake files to `cmake/` its no longer properly finding the VCPKG files. Just tried being explicit with where they are, but it's still failing on the VCPKG install - which is what manages all our dependencies. Might be time to pull out a temp Windows VM to test with.

BenjamenMeyer:
NOTE: The Mac and Linux builds are broken because of #938

stephengtuggy:
This looks good in general, although obviously the build breakage will need to be fixed. I like where this is heading. It makes a lot more intuitive sense.

BenjamenMeyer:
thanks @stephengtuggy for helping to update this

stephengtuggy:
> We probably should figure out how to make it use `nmake` so things are consistent. `nmake` comes packaged with VS, and uses Makefiles just like GNU Make does; it's just Microsoft's variant of the family and works really well for build systems.

What about using Ninja across the board instead? That's what we're using now, in many cases (though not all).

engine/CMakeLists.txt
stephengtuggy: Looking into it a bit more, it looks to me like the compilers could use some help with the optimizations a lot of times. Especially when our code is written in a way that isn't thread-safe. I would like to get the Draw and CalculatePhysics parts of the code revised to be more thread-safe/reentrant. Then we could maybe look at using something like OpenMP to parallelize things explicitly for better runtime performance. I anticipate that that could significantly improve the experience of playing the game.

stephengtuggy (approved):
Looks good to me!

BenjamenMeyer:
> > We probably should figure out how to make it use `nmake` so things are consistent. `nmake` comes packaged with VS, and uses Makefiles just like GNU Make does; it's just Microsoft's variant of the family and works really well for build systems.
> 
> What about using Ninja across the board instead? That's what we're using now, in many cases (though not all).

We only use ninja on Windows. We use make on Linux and Mac.

stephengtuggy:
> > > We probably should figure out how to make it use `nmake` so things are consistent. `nmake` comes packaged with VS, and uses Makefiles just like GNU Make does; it's just Microsoft's variant of the family and works really well for build systems.
> > 
> > 
> > What about using Ninja across the board instead? That's what we're using now, in many cases (though not all).
> 
> We only use ninja on Windows. We use make on Linux and Mac.

I use ninja at least part of the time on all three platforms.

So maybe I should have said, "That's what **I'm** using now." 😉

Issue #865, opened 2024-06-15
Save Games with the starting Llama are corrupt in 0.9.x


evertvorster:
This is a pretty serious bug, as one expects this game to be quite a long-running excersize. 
To show one aspect of this bug, start a new campaign. Don't take off, don't do anything but save your game, and then load it again. 
On my system, once you load that saved game, you are dropped almost 10000km from Atlantis, and your  yaw controls don't work, and you can't use SPEC to get closer to the planet to land.

If you exit out of VS, and then start it again, and then load the game you saved in the previous bit, you are on Atlantis again, but if you just launch you are again 10000km away from the planet, without yaw controls and a SPEC which misbehaves. 

If, on the other hand you start a new campaign, and don't save, and just take off, you are dropped 9222km from the planet, but now you can use your SPEC drive to get closer to the planet quickly, and your yaw control works.
You can also land and take off multiple times, and all the yaw controls work, but it is only when you load a saved game that the game starts misbehaving.

evertvorster:
Interestingly enough, if you load a saved game that was not make by this latest engine, it works properly. 
Also, I am running a different ship than the initial Llama, so the corruption could be sneaking in when saving and loading the initial Llama, which is a bit special.

evertvorster:
Doing a bit more testing here. 
I have a saved game that was made with a previous version of the engine, and here is it's Llama.csv file:
[Llama.begin.csv](https://github.com/user-attachments/files/15847767/Llama.begin.csv)


I loaded that game, and then saved it as something else, and this is the csv file that is produced:
[Llama.begin.csv](https://github.com/user-attachments/files/15847770/Llama.begin.csv)


With even a cursory glance, I can see that there are huge differences. 

However, if I load a saved game that was made by the previous version of the engine where I was piloting a different ship, and save that, there are almost no differences. 
Before:
[Franklin.hunter.csv](https://github.com/user-attachments/files/15847786/Franklin.hunter.csv)

And after:
[Franklin.hunter.csv](https://github.com/user-attachments/files/15847789/Franklin.hunter.csv)

So, as I mentioned before, the Llama you get in the beginning is special, and it looks like the save game routine does not take that into account.

evertvorster:
One other thing to note is that saved games with the 0.9.x engine is incompatible with the 0.8.x engine. My games saved in Sol in 0.9.x crash the 0.8.x engine when loading them.

evertvorster:
[...]

PR #895, opened 2024-10-19
Periodic Housekeeping PR 1

royfalk:
- Remove VS_PI and replace with standard M_PI
- Fuel tank is integrated (can't upgrade). Remove options in base_computer.
- Use standard ftl_energy.MaxLevel() instead of getWarpEnergy and warpCapData.
- Remove support for additive and multiplicative capacitors and reactor.
- Make unit_csv load reactor and capacitors the standard way using Load.
- Remove WCfuelhack in unit_csv. Make energy source a configuration item. Game doesn't need to compute this every loop.
- Remove vsfilesystem reference in unit_csv_factory. Breaks testing.
- Remove reference to Unit in Cloak. Improves encapsulation.
- Merge EnergyType enum into ComponentType.
- Move GetUpgradeType to component_utils.
- Remove Component::Describe. We'll do this in python along with the rest of the text in base_computer.
- Make component query manifest for price, volume and description.
- Add a whole bunch of getters to component.
- Add support for infinite energy source. Useful for WC drive and reactor.
- Move factors to configuration. You can now control utilization of drive and other components using the config file.
- Add resiliency to Graphics2 and Manifest loading. Does not crash game if missing.

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. 

Issues:
- ~~Jump causes damage. Unclear if this is new. There is code to do this in the engine.~~
- Python ship view doesn't include the factors. Specifically, FTL drive factor (0.1) isn't applied in python.


evertvorster:
I'm having a little trouble making this as a patch that applies to master. 
I might have to try a different way of getting this source to compile.

evertvorster:
OK, first mayor problem.
When starting a new campaign, the SPEC drive seems to draw energy from the drives, and depletes the drive capacitor in a second, and then disengages. 
This is true for both the initial Llama, and the Llama you get when you do the first repair and refuel. 
Suffice it to say, the game is unplayable without SPEC. 

When loading a saved game, the SPEC drive seems to perform normally. 

Also, if you start a new campaign, and immediately save your game and then exit VegaStrike, loading that saved game coredumps VegaStrike. 
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stod
/usr/bin/vs: line 28: 43001 Aborted                 (core dumped) vegastrike -d"${VEGASTRIKE_SHARE_DIR}" "$1"
```

evertvorster:
Some added information, I checked with master branch, and on that SPEC works normally.

royfalk:
The crash is because of #865 . It seems I didn't fix the issue. I had a recollection that I did.
I'll fold this into https://github.com/vegastrike/Assets-Production/pull/128.

engine/src/resource/product.h
```diff
@@ -32,6 +32,8 @@ class Product
     std::string name;
     // TODO: Can be a fraction for things such as fuel, water, etc. But not for now.
     Resource<int> quantity;
+
+    // TODO: move to int, x100 and add code to deal with cents.
```
BenjamenMeyer, line 36: we probably should just do whole credits and not concern ourselves with fractional credits
(e.g 1 credit = 1 penny = 1 shilling .... )

engine/src/vs_math.h
```diff
@@ -50,6 +50,5 @@
 #define FINITE(x) ( std::isfinite( x ) )
 #define ISNAN(x) ( std:: isnan( x ) )
 
-#define VS_PI 3.1415926535897931
```
BenjamenMeyer, line 53: So I like moving from `VS_PI` to `M_PI` but from what I'm reading `M_PI` isn't standard (ANSI/ISO C/C++); do we have something checking for it to exist at the CMake level and if not defining it? How we define `version.h` could probably be a good method.

royfalk: I checked and it is. You just need some kind of magic pre-processor for it to be included. I'll add that to the list of things to do and reasons why windows annoys me :)

evertvorster:
Just did a little cross-checking here. 

I'm pretty sure this PR introduces #898 as the issue is not present in the current master. 
Is this what that extra python script is about?

Also, please update this branch by merging in master.

royfalk:
> Also, please update this branch by merging in master.

I will. 

I find myself with a bit more time and to facilitate reviews by @BenjamenMeyer , I'm working on multiple branches in parallel. This is all fine and dandy, except when it isn't.

engine/src/resource/product.h
royfalk: Yeah. I figured that one out too. Still need a todo.

royfalk:
> When starting a new campaign, the SPEC drive seems to draw energy from the drives, and depletes the drive capacitor in a second, and then disengages.

I think I fixed this in https://github.com/vegastrike/Assets-Production/pull/129. You can test this by making the changes to ships.json.

engine/src/vs_math.h
royfalk: So I spent a few minutes trying to figure this out and then stopped. The game builds on windows.

evertvorster:
Hi there. 
Yes, the multiple branches are a little confusing to me too. What is most confusing is that there seems to be some inter dependency between the branches, and definitely between branches between Assets and Engine. 

Of course, a little testing can help to get things merged faster and then hopefully we will have less branches hanging around. 

I'm testing this branch now. It's running together with task_ship_view branch of Assets that has been patched with PR 129. 

So far, its pretty good. I am able to start a new campaign, and save the game. When starting a new instance of vs and loading the created saved game, you now have flight controls, which is great. 
Unfortunately you can't buy a jump drive and SPEC does not stay engaged, but now it does not show as a drain on the drives bar, it just kicks on and then disengages. This sounds awfully familiar, and I'll go check to see if I am missing a patch or something.

evertvorster:
Ah, I see what the issue is. In #896, the issue with SPEC and buying the jump drive is fixed, but not this branch. 
I'll go test as much as I can of this branch so that it can get merged, and then we'll concentrate on getting #896 with a view of getting that one merged next.

evertvorster:
Just saw my ship take some damage going through a jump as well, that's definitely something new. 
For reference, I was flying a Goddard in pristine condition, and just jumping once took out half of my right front armor. 
I'll go do a few jumps with a faster ship and see if there is any pattern to it.

evertvorster:
Just played a little more. Every jump now causes damage to the armor. Every jump I did some other part of the armor got damaged, so there is some randomness there. It does not seem to be confined to one ship type either, any ship that you are flying is taking damage through a jump. 
Could it be that your current stats are "saved"  when you exit a system, and "restored" in the next system, and that it is not completely restored?

evertvorster:
Yes, I noticed the comment at the end of the PR. 

If this PR gets merged as is, this will be an issue which needs to be solved. 
I am not certain that it is this PR that causes this issue, and will be testing git master to confirm.

royfalk:
OK. This issue got fixed as part of the Drive PR. I'll see if I can pull the specific commit here too.

evertvorster:
Tested the PR #901 against this branch, and there is no damage when you jump to a new system now. 
From my point of view, this one is ready to go. :)

evertvorster:
One thing I am noticing with this branch is that in a save game where I am flying around in a Goddard with a SPEC capacitor V, it says I don't have enough jump energy. In master branch the same jump drive with SPEC works flawlessly.

royfalk:
Can you post the following:
ship info - jump cost and ftl capacitor.
SPEC capacitor V capacity.

evertvorster:
Hi there. 
On the master branch, in ship info it gives the error:
Error: Ship description not found. 
And so it does not list a jump cost. 

SPEC capacitor V Spec:
FTL Capacitor states 2 500 000 M/J of capacity. 

With the task_drive_components branch:
The Ship description is read properly, and explains all the virtues of flying a military spec Goddard.
Jump Cost:
166 666 Mj/s

It does not appear to recognize the SPEC capacitor. 
Screenshot_20241104_072006

Here is the csv of the milspec Goddard that is displaying this issue:
[Goddard.milspec.csv](https://github.com/user-attachments/files/17614031/Goddard.milspec.csv)

royfalk:
This is odd. I looked at the csv and Goddard.milspec is 200 for warp and 200 for jump. It comes with a warp capacitor of 450 so it should work out of the box. 

Out of curiosity, can you start a new game with a Goddard.milspec and just fly it. I think the issue is the ftl capacitor overwrote the current number with 0 for some reason. I'll try and get to it today and see if I can figure this out.

evertvorster:
Hi there. 

I started a new campaign, and saved it. 
Then I edited the save file to have Goddard.milspec instead of Llama.begin, and when loading that game, in the HUD it displays the message: low jump energy. 

The thing with this saved game, though, is that I sold the milspec package some time ago, and installed my own components. I had a bit of spare cash lying around, and could afford more upgraded components than that of the milspec package. Now, if you remove the milspec package, the ship should just be the stock version, right?

evertvorster:
A bit more information. 

Started a new campaign, and went to the shipyards straight away. The SPEC capacitors listed for sale have the same error message as the screenshot I added in to this PR. To me, it looks like the game engine is not properly reading the SPEC capacitor part of Assets. The other components appear to be fine, with descriptions and all.

engine/src/vs_math.h
BenjamenMeyer: when I was reviewing this, I know I was seeing an error around `M_PI` for at least one of the various PRs out there....but yes, this PR does build on all the platforms. May be there was a hiccup in the GH workers?

evertvorster (commented):
Hi there!

While Play testing this PR, there are a few issues. If these are fixed in another branch, then this one is good to go. However, they do bear mentioning here:

SPEC capacitor details are not available, and Jump drive cannot be purchased. 
In the following screenshot, a new campaign was started, and in the upgrade ship area of Atlantis, you can see that the jump drive is red and unpurchase-able. Also, the SPEC capacitor specs are showing an error message. 

image

Saving and loading works, and you have all flight controls in the saved game, so that is a bonus. 

SPEC engages, but then a second later disengages. This might be due to the game not reading the proper values for the SPEC capacitor. 

Top speed in Llama begin is 120, and after you do your first repair and refuel it is 125, which is what is normal, which means that the game is reading Llama.begin properly.

When loading a saved game with a .milspec ship, you also cannot jump, and on the HUD is states that the jump energy is low. 
When loading a saved game with a .stock ship, you can jump, but still in the ship upgrades computer there are no stats on the SPEC capacitor or the jump drive. 

Fuel burn rates seem to be normal. 

Comparing the same tests against current master, SPEC capacitor has details and engages properly on a new campaign.
You still can't purchase a jump drive, but I do believe that this is fixed in another branch. 

In conclusion, it would appear that this PR does introduce the issue where the SPEC capacitor's info is not read properly, and that has some knock-on effects.

evertvorster:
Hi there!

I am testing just the branch, but I now know how to check out master and merge the branch to it. In my limited understanding of how git works, I thought it would not make a difference if I just checked out the branch, rather than merge it into master. 

Will double check that Assets is the latest master, and play test with latest engine with this branch merged into it.

evertvorster:
This branch has a merge conflict with engine/src/cmd/basecomputer.cpp, and so does not merge cleanly. I don't know how to fix this, unfortunately.

evertvorster:
Hi there!

Thanks for fixing that merging conflict. 
I uploaded a video here to show my building and testing of this pull request as it stands now. 
https://youtu.be/xQNEZ0m0sAc
Since we are getting different results, something is different between what it is that we are testing. 

The volume on the video is quite low, as there has been no processing done on it, so you will have to turn up the volume a bit to hear properly. Please see if you can see a problem in my testing setup.

royfalk:
Unfortunately, I discovered I was working on the wrong branch. I intended to work on the drive, which is why it's called 'task_drive_components'. When I actually worked on the drive, I had to name the branch 'task_real_drive_components'.

You can stop play testing for now. Everything reproduces.
So there are several issues here I need to fix:
1. With units now consistent, a ship with with reactor output of 50, warp capacitor of 45 and warp usage of 20 should work. Right now it doesn't. These are the stats of Goddard.milspec.
2. With the move to python, ship view doesn't take the factors from configuration. Need to do so.
3. Need to make sure to reapply factors in the opposite direction when saving the ship. Otherwise, two saves and a factor != 1and you'll see a change in how the game behaves.

evertvorster:
OK, good to know it was not something on my side.

Let me know when to test this branch again, please.

royfalk:
> Let me know when to test this branch again, please.

Go for it. It should now work even with Llama.begin with reactor02.

evertvorster:
Hi there!

Just took it for a spin, and SPEC now works in a new campaign. So, technically you can start a new game and be able to just play normally, which I suppose is a good thing for someone new to Vega Strike. 

However, there is still something that bothers me. 
When selecting a SPEC capacitor in the base computer, it still shows that error message about not being an upgrade. 
One other curiosity is that when starting a new campaign, your Llama weighs 250mt, and uses 12000Mj/s for SPEC flight. 
My saved Goddard weighs 4898 metric tonnes, and only needs 400Mj/s for sustained SPEC. It feels like something is off there. Mind you, this is only a small detail, and does not prevent you from flying and enjoying the game, it just looks weird.

royfalk:
>When selecting a SPEC capacitor in the base computer, it still shows that error message about not being an upgrade.

I was unable to reproduce that.

> One other curiosity is that when starting a new campaign, your Llama weighs 250mt, and uses 12000Mj/s for SPEC flight.
My saved Goddard weighs 4898 metric tonnes, and only needs 400Mj/s for sustained SPEC. 

Yeah. I am not a fan of that. At some point in the near future I plan on simplifying assets somewhat. The current thinking is to make warp cost and jump cost functions of mass. That would get rid of these inconsistencies and shorten the file.

I also want to remove support for eight and six facets and make armor four facets. Merge moment into mass. Save four facet shields and front, back, left and right (and not top-left-port whatever).
One problem with this is the fact armor support in the cockpit is for eight. Tricky.

evertvorster:
I fully agree, stuff that can be calculated should be. It does not really add to the game play apart from a nice description. It's not like a better SPEC makes you faster, or a better jump drive makes a jump faster. What I have noticed is that a bigger jump capacitor allows you to make successive jumps if you are in a particularly fast ship. Which is nice. 

That error message with the SPEC capacitor is still present. When you have it selected in the upgrades section, this is the description of the SPEC drive, and the error message is included in the description:
image

This is with this branch of Engine, and the task_ship_info_3 branch of Assets and starting a new campaign.

evertvorster (approved):
When merging in the other branch mentioned, the issue with the SPEC capacitor description does go away. 
With both these branches merged, I cannot find any fault with the game play. 
This one is good to go!

royfalk:
@BenjamenMeyer ?

engine/src/components/energy_container.cpp
```diff
@@ -96,45 +108,49 @@ void EnergyContainer::Refill() {
...
-    return std::string();
+        case ComponentType::Capacitor:
+        unit[FUEL_CAPACITY] = std::to_string(MaxLevel() / configuration()->fuel.energy_factor);
+        break;
+        
+        case ComponentType::FtlCapacitor:
+        unit[FUEL_CAPACITY] = std::to_string(MaxLevel() / configuration()->fuel.ftl_energy_factor);
+        break; 
+
+        default: // This really can't happen
```
BenjamenMeyer, line 147: Would be good to log why we're aborting

BenjamenMeyer (changes requested):
Please add a log message for the abort

BenjamenMeyer (approved):

PR #897, opened 2024-10-21
Fix bug 865

### LINKED ISSUE #865: Save Games with the starting Llama are corrupt in 0.9.x

royfalk:
This fixes #865 .

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Minimal. Checked the bug was fixed and the game works. Ship fire and lands. SPEC.


evertvorster (approved):
I did a play test of this change. 
Started a new campaign, and saved the game. 
The game loads normally, and when launched the ship has yaw controls. 
So, that indicates to me that the Llama is properly loaded from a saved game. 

From a playability point of view, this is a good step in the right direction.

I was able to complete the training mission, almost. 

Setting max speed in maneuver mode is still not enough to satisfy Oswald's requirements, and I had to set travel mode to gather enough speed to pass to the next step in the training mission. 

Oswald's last instruction was to purchase a jump drive and then get back to him. 
I was not able to purchase the jump drive, despite there being one available on Atlantis, me having enough money and upgrade space in the Llama. 

Another thing I noticed was that the fuel burn rate was quite high, and just doing the training mission used about a third of the fuel in the Llama. 

After doing a refuel and repair, I also noticed that the fuel was not, in fact, refueled. 

The issues noted here I am assuming that it is somewhere else in the code, and I have no problems approving this fix to go in ASAP.

BenjamenMeyer (approved):

PR #899, opened 2024-10-27
Move ship view to python (partial)

royfalk:
This moves the ship view to a python script. It provides the script with a dictionary generated in the same code as unit saves. It will allow modification by different games. e.g. antimatter as fuel as well as localization.
This PR covers most text up to armament section. Missing cloak.

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. **Pending**

Issues:
- None


engine/src/python/base_computer/ship_view.h
```diff
@@ -0,0 +1,36 @@
...
+
+// -*- mode: c++; c-basic-offset: 4; indent-tabs-mode: nil -*-
+
+#ifndef VEGA_STRIKE_ENGINE_PYTHON_BASE_COMPUTER_SHIP_VIEW_H
+#define VEGA_STRIKE_ENGINE_PYTHON_BASE_COMPUTER_SHIP_VIEW_H
+
+#include <string>
+#include <map>
+
+const std::string GetShipView(const std::map<std::string, std::string>ship_stats);
```
BenjamenMeyer, line 34: this should probably be a const reference:
```
const std::string GetShipView(const std::map<std::string, std::string>& ship_stats);
```
otherwise it's going to copy the map.

BenjamenMeyer (changes requested):
I'm mixed on moving more to Python... but aside from that, mostly concerned about the copying time for the map since it will probably hold all our ships and their variants, no?

engine/src/python/base_computer/ship_view.h
royfalk: Answering both your questions here:
1. This is for a single ship, not all ships. It will be used by ship view - that page in base computer where you see the ship specs. It will also be used by save ship as part of the save game. If you have a fleet of ships, it will several times.
2. It will copy. I'll fix it as it is good practice. However, it is negligible.

BenjamenMeyer (approved):

engine/src/python/base_computer/ship_view.h
BenjamenMeyer: Thanks!

BenjamenMeyer:
still a bit mixed on moving stuff to Python but my other concerns alleviated so approved

PR #901, opened 2024-11-02
Fix bug 900

royfalk:
This fixes the jump damage when jumping.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Minimal. This is a PR of an existing commit from another PR. It has been tested there and I'm fast tracking it.


evertvorster (approved):
Tested this PR against master, and there is no damage when you jump to a new system now. There does not seem to be any other side effects to this PR. 

Approved.

BenjamenMeyer (approved):
seems reasonable

PR #911, opened 2024-11-24
Fix bug 909

### LINKED ISSUE #909: Set speed craft exceeded while turning.
Hi there. 

This has been ongoing for a while now, and is definitely different from the way that Vega Strike used to play in the old days. 
Testing is now done on the latest master of Assets and Engine as of 2024-11-22. 

The problem is this. 
The Llama, once you applied the first fix, has a top speed of 125. (Of whatever units Vega Strike uses)
If you point the ship in a direction, and apply full thrust, it gets to that speed. 
However, if you now turn your ship, the speed increases. 
This is a problem, as the speed should stay at the speed limit, and not exceed it. 

This is the way that the game used to work, but somewhere in the last year or two this has changed. In theory, your ship's computer should use it's thrusters to change direction, and ensure that the speed limit is not exceeded. 

This may be why the autopilot overshoots it's targets, as the ship is doing something that is not expected. 

Now, very recently, this effect seems to have gotten worse. 
When accelerating to full speed, and then turning into a new direction still pushes the ship past its top maneuvering speed. 
When you cut thrust, the ship does not come to a stop at all, and continues on indefinitely at this higher than the speed limit speed.

Here is a youtube video that shows what I am talking about. Volume is a bit low, so you might have to increase the volume to hear my voice. 

https://youtu.be/bFDHvAJ1ze0

royfalk:
closes #909 

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation minimal. Check issue is resolved.


evertvorster:
Hi there!

This one is a little better, because usually after cutting thrust the speed returns to 0, but I am still able to far exceed the speed limit for the Llama. 

When starting a new campaign, did a repair and refuel and then pop off the planet. 
Aim at Serenity, and accelerate to full speed. (125)
Now, at a fairly slow rate of turn, turn 180 degrees so that serenity is behind you. If you see the same issue that I am seeing, your speed should touch roughly 300. 

In the classic versions of Vega Strike, this was simply impossible, and your maximum speed in MANEUVER mode was always pegged at 125 for the repaired Llama, and whatever other speed was set if you are flying another ship. 
For all ships, the maximum speed that you could get in a straight line was also the speed that you got through a turn. 

With this branch of the game engine, and the master branch of Assets there is a save file which I basically did the same thing. New campaign, and repair and refuel the Llama. In more than one instance, when loading this game, I was able to do this turn and cut thrust and the ship continued at a speed of 300+

To re-iterate this in maybe another way. When you have your flight computer in Maneuver mode, it decreases the thrust from your main thruster while you are maneuvering to ensure that you never exceed the maximum set speed. 
As I mentioned before, you can easily turn off this functionality by setting the flight computer to travel mode, and then your maximum speed is much higher.

evertvorster (approved):
This does indeed put back the Orthogonal acceleration. 

Not having the orthogonal acceleration is a different issue than being able to exceed the set speed. I'll continue that as a separate bug. 
This one is good to go. :)

PR #922, opened 2024-11-30
Switch save game to json

royfalk:
There is a utility to convert save games to json.

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation _minimal. Tested correctness of load/save and checked ship view._

Issues:
- None


evertvorster:
Not reviewing this yet, but I do have a question. 
Is loading older saved games still supported?
If so, will it only save in the new format?

Should this get merged, we need to let users know somehow that they need to back up their old saved games just in case something goes wrong.

royfalk:
> There is a utility to convert save games to json.

I just need to finish polishing it and push it to Assets-Production. I should also add some text to the readme, explaining how to migrate it. 

> Is loading older saved games still supported?

No
> If so, will it only save in the new format?
Yes

> Should this get merged, we need to let users know somehow that they need to back up their old saved games just in case something goes wrong.
The unit CSV is left alone. It isn't deleted.

BenjamenMeyer:
> > Is loading older saved games still supported?
> 
> No

we should add a script to convert older games to make transition easier; though most likely folks will be starting over

BenjamenMeyer (approved):
I agree it's time to make this happen; we should have some kind of transition though for the saved games; even if it's just a script to do a conversion.

Issue #929, opened 2024-12-08
Mac no longer works in the CI


royfalk:
It seems that MacOS 13 and newer are not supported by CMake.
With the deprecation of MacOS 12, I attempted to increase the version of the CI (#928 ) to 13 and the build broke.
It needs fixing both for CI and for people who want to play on Mac.

BenjamenMeyer:
Looking at my PR (#891) I noticed that some of the dependencies that were assumed have changed.  I'm going to build a bootstrap script for MacOS and try to detect stuff so install it.

BenjamenMeyer:
good progress on the PR; but need to figure out how to set some of the libs with homebrew

BenjamenMeyer:
Seeing something odd with Boost Python in Homebrew. It's supposedly built against Python 3.12; but Homebrew installs Python 3.13. Following the simplest route yields errors between the two versions, but I figured out how to get it set correctly so it's building against Python 3.12 only to have it not work as if it's built against Python 3.13.

royfalk:
That looks like good progress. Let me know if you need some help.

BenjamenMeyer:
So based on my PR it looks like we can now easily continue forward with MacOS 13.
But we're going to need some help making the ARM transition for MacOS 14+. If we don't get help then we may have to drop Mac support again.

royfalk:
What seems to be the issue?
I have an M1 Mac and can spend some time on this.

BenjamenMeyer:
@royfalk from [this build](https://github.com/vegastrike/Vega-Strike-Engine-Source/actions/runs/12486471619/job/34846582185) (L3395-L3409):
```
[ 19%] Linking CXX executable vegasettings
cd /Users/runner/work/Vega-Strike-Engine-Source/Vega-Strike-Engine-Source/build/setup && /opt/homebrew/bin/cmake -E cmake_link_script CMakeFiles/vegasettings.dir/link.txt --verbose=1
Undefined symbols for architecture arm64:
[...]

PR #942, opened 2024-12-28
More On Packaging - 0.9.x

stephengtuggy:
Code Changes:
- [X] CI change only

Issues:
Addresses #929 

Purpose:
- What is this pull request trying to do? Get the installer packaging to work for Mac


stephengtuggy:
I tested these changes locally, so the packaging step, at least, should be correct this time.

BenjamenMeyer (approved):

PR #947, opened 2024-12-31
Include Third-Party Dependencies In Installers - 0.9.x

stephengtuggy:
Code Changes:

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Change the Windows and Mac installers to include the required third-party dependencies, such as Boost


stephengtuggy:
Okay, I've tested the Windows 11 installer, built locally, and it works. The games installs and runs fine, from the EXE, although the shortcuts to it don't work yet. But those will probably be created by the vsUTCS installer anyway.

I don't have a good way to test the macOS installer. The only Mac I have available is running macOS 11, which is two major versions behind the macOS version we build the installer for. If we could get a volunteer or two to test on macOS, that would be great. (Intel-based Macs only, at this point.)

stephengtuggy:
I can probably test the installer on at least a couple of Linux distributions, such as Fedora 40/41 and Debian 12. We will probably want volunteers for the remaining Linux distributions and versions, also.

script/clean
```diff
@@ -24,7 +24,7 @@ SRC_DIR=$ROOT_DIR/engine
 
 # delete build directory
 echo "deleting build directory ($BUILD_DIR)"
-rm -r $BUILD_DIR
+rm -rf $BUILD_DIR
```
BenjamenMeyer, line 27: Should probably put a guard around that to make sure it's not empty or root

BenjamenMeyer (changes requested):

script/clean
stephengtuggy: There. How's that?

BenjamenMeyer (approved):

script/clean
BenjamenMeyer: Thanks

PR #964, opened 2025-01-07
unit_csv.cpp: Fix parsing of "Yaw_Governor_Right" and "Yaw_Governor_Left"

### LINKED ISSUE #962: Complete loss of yaw controls (zero values in save file)
Somewhere between saving and loading a savegame with the 0.9.x code, the Yaw Governor Left and Yaw Governor Right values get completely zeroed out. The result is that it is no longer possible to turn (yaw) your ship to the left or the right.

This issue presents similarly to #898 (and perhaps also to #924 ) but I guess the cause is different. Here, the values get zeroed out completely, rather than set to small values such as 0.01.

This bug is still very much present in 0.9.x, and it is a high priority to fix, IMO.

stephengtuggy:
Code Changes:

Issues:
Fixes #962 

Purpose:
- What is this pull request trying to do? Fix issue 962, complete loss of yaw controls due to zero values in savegame

### DIFF
--- a/engine/src/cmd/unit_csv.cpp
+++ b/engine/src/cmd/unit_csv.cpp
@@ -1,5 +1,7 @@
 /*
- * Copyright (C) 2001-2022 Daniel Horn, pyramid3d, Stephen G. Tuggy,
+ * unit_csv.cpp
+ *
+ * Copyright (C) 2001-2025 Daniel Horn, pyramid3d, Stephen G. Tuggy,
  * and other Vega Strike contributors.
  *
  * https://github.com/vegastrike/Vega-Strike-Engine-Source
@@ -13,7 +15,7 @@
  *
  * Vega Strike is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
@@ -842,8 +844,8 @@ void Unit::LoadRow(std::string unit_identifier, string modification, bool saved_
 
     YawPitchRollParser(unit_key,
             "Yaw_Governor",
-            "Yaw_Governor",
-            "Yaw_Governor",
+            "Yaw_Governor_Right",
+            "Yaw_Governor_Left",
             computer.max_yaw_right,
             computer.max_yaw_left);
     YawPitchRollParser(unit_key,


royfalk:
I don't get it. I looked at the blame for this and it goes back to a commit of yours from 2022. This one only formatted stuff and didn't change the text. As far as I can tell, this was always there.

I looked at the save code below and it hasn't changed materially in a while too. I touched in 2023 but this was a formatting change too. 

I can see why this fixes things but I'd like to know how everyone missed the fact that the ship has no yaw. I can tell you for a fact my save games work fine with this code, which makes no sense.

stephengtuggy:
> I don't get it. I looked at the blame for this and it goes back to a commit of yours from 2022. This one only formatted stuff and didn't change the text. As far as I can tell, this was always there.
> 
> I looked at the save code below and it hasn't changed materially in a while too. I touched in 2023 but this was a formatting change too.
> 
> I can see why this fixes things but I'd like to know how everyone missed the fact that the ship has no yaw. I can tell you for a fact my save games work fine with this code, which makes no sense.

I have no clue. I don't get it either. You are correct that I was the last one to touch these lines of code, in 2022. No idea how we missed this issue that whole time since then.

Anyway, this fixes it. I guess that's the important part.

royfalk:
> I was the last one to touch these lines of code, in 2022. 

No you didn't. You just formatted it. This code predates 2022.

> Anyway, this fixes it. I guess that's the important part.

I'd still like to know how we missed it all these years. Specifically, we didn't. I tested extensively with saved games and I would have noticed not having yaw. I rarely roll, so if that were the case, I can believe. But yaw, no way.

stephengtuggy:
> > I was the last one to touch these lines of code, in 2022.
> 
> No you didn't. You just formatted it. This code predates 2022.
> 
> > Anyway, this fixes it. I guess that's the important part.
> 
> I'd still like to know how we missed it all these years. Specifically, we didn't. I tested extensively with saved games and I would have noticed not having yaw. I rarely roll, so if that were the case, I can believe. But yaw, no way.

The only thing I can think of is that maybe some other part of the code was masking this bug somehow. And that other code section changed recently, uncovering this issue. But what would that have been?

stephengtuggy:
I'm flying the starting Llama. Llama.begin.csv.

stephengtuggy:
It might make a difference what screen one loads the savegame from. Are you loading it from the main menu? Or from the base_computer or similar?

royfalk (approved):
It's clear this change is correct and needed. Approved.

BenjamenMeyer:
To be fair, I've had this issue for quite while but never had the time to track it down. Looks like you both may have.

I could often roll, but not yaw so would get by with roll+pitch changes at times.

PR #973, opened 2025-01-11
Fix bug 969 - pressing m caused segfault

royfalk:
This is a regression from when I dropped 8 facet support. Player ship VDU (video display unit) draws player ship with damage. Current code was not changed and used an 8 sized array to generate the colors, causing a crash.

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. minimal. checked the game works, fought with Oswald and checked the bug is fixed.

Issues (both to be fixed by upcoming PR - Hull/Armor/Shield components):
- Armor upgrade isn't working.
- Ship view has some extra text (?)


engine/src/gfx/vdu.cpp
```diff
@@ -1185,21 +1185,18 @@ void VDU::DrawDamage(Unit *parent) {
...
     //float th;
     //char st[1024];
-    GFXColor4f(1, parent->GetHull() / (*maxhull), parent->GetHull() / (*maxhull), 1);
+    double health_percent = parent->layers[0].facets[0].Percent();
+    GFXColor4f(1, health_percent, health_percent, 1);
     GFXEnable(TEXTURE0);
     float armor[8];
     parent->ArmorData(armor);
     const bool draw_damage_sprite = configuration()->graphics_config.hud.draw_damage_sprite;
     DrawHUDSprite(this, draw_damage_sprite ? parent->getHudImage() : NULL, .6, x, y, w, h,
```
royfalk, line 1194: These lines are the bug. 
The code above was from a misreported line number, causing me to refactor GetHull.

engine/src/cmd/planet.cpp
```diff
@@ -573,7 +573,9 @@ void Planet::AddFog(const std::vector<AtmosphericFogMesh> &v, bool opticalillusi
     }
     fawg->setFaceCamera();
     getSubUnits().preinsert(fawg);
-    *fawg->current_hull /= fawg->GetHullPercent();
+
+    // TODO: this is very weird. Why are we doing this?
+    // fawg->hull /= fawg->GetHullPercent();
```
stephengtuggy, line 578: I don't know why the code was doing that. If you don't know, leave it. It's one of those "fences." In my opinion, anyway.

royfalk: Let me clarify, as I do understand this code. 
fawg (why not fog?!) is a unit and created with some unspecified hull value, several lines up.
Now fawg = fawg * percent_damage.
So, example:
```
fawg = 50, percent = 50
new_fawg = 100
```
Except, this was created several lines above. Hull is undamaged. So x /= 1.0 is x

royfalk: Let's be generous and assume this unit creation is different and hull_percent != 1.
Why not just do:
`hull = x * some_percent`
Why be so convoluted?!

royfalk: So, digging a bit deeper (why though?!) I found this:
```
if (opticalillusion) {
   fawg = new AtmosphereHalo(this->rSize(), fogs, 0);
} else {
   fawg = new Unit(fogs, true, 0);
}


AtmosphereHalo(float radiusOfPlanet, vector<Mesh *> &meshes, int faction) :
    Unit(meshes, true, faction) {
    planetRadius = radiusOfPlanet;
}

Unit::Unit(std::vector<Mesh *> &meshes, bool SubU, int fact)
        : Drawable(), Damageable(), Movable()
```

So, it doesn't look like it even gets initialized with anything other than the default 1 value.

stephengtuggy: So it looks like this is supposed to draw the atmosphere "halo" around a planet.

stephengtuggy: It's a "Unit" because just about everything that needs to be drawn is either a Unit or a Bolt, I believe. Something like that. In this case, the atmosphere is drawn separately from the surface of the planet itself. As a separate, transparent layer. Same thing with the city lights, in cases where a planet has city lights. They constitute another layer.

stephengtuggy: There are a lot of details that I still don't understand, but that seems to be the gist of it.

royfalk: Yep. You understood correctly

Also, we'd need @Loki1950 to comment on this but I'm reasonably certain none of these have really worked. Atmosphere, surface, nebula, etc.

Finally, this isn't how I'd implement this. I don't think firing a lot of drivers at the atmosphere is going to "boil" it.

stephengtuggy: Um, what? I don't know about nebulae, but atmosphere and planet surface definitely work.

royfalk: Moving this to a discussion. We started from why `fawg->hull /= fawg->GetHullPercent();` and this is beyond the scope of this PR.

BenjamenMeyer: @royfalk please link the discussion here

evertvorster:
I'm hitting issue #981 and so cannot test this PR at the moment.

engine/src/cmd/planet.cpp
royfalk: https://github.com/orgs/vegastrike/discussions/977

stephengtuggy (approved):
LGTM

PR #974, opened 2025-01-11
Get v0.10.x of the game to start up on macOS; port over many of my changes from v0.9.x

### LINKED ISSUE #857: Follow @daviewales ' suggestion to avoid using XQuartz at all, even for vegasettings
> You might be able to avoid using XQuartz by using `port install gtk3 +quartz`, which will attempt to use the native macOS backend rather than Xorg. I'll try to find some time to test your PR. I'm still on macOS 10.13. 

 _Originally posted by @daviewales in [#533](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/533#issuecomment-2091731698)_

### LINKED ISSUE #929: Mac no longer works in the CI

### LINKED ISSUE #933: Python 3.13. and newer do not work on Mac

### LINKED ISSUE #941: Python 3.13 not support
Hi there!

Something got updated on my system, and I am unable to build master, as well as branches that I have been able to build before. 

In all cases, the error message is this:
```
[ 10%] Building CXX object CMakeFiles/vegastrike_python.dir/src/python/config/python_utils.cpp.o
/mnt/int/system/aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/engine/src/python/config/python_utils.cpp: In function ‘const std::string GetPythonPath()’:
/mnt/int/system/aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/engine/src/python/config/python_utils.cpp:44:37: warning: ‘wchar_t* Py_GetPath()’ is deprecated [-Wdeprecated-declarations]
   44 |     wchar_t* w_path_ptr = Py_GetPath();
      |                           ~~~~~~~~~~^~
In file included from /usr/include/python3.13/Python.h:123,
                 from /usr/include/boost/python/detail/wrap_python.hpp:178,
                 from /usr/include/boost/python/detail/prefix.hpp:13,
                 from /usr/include/boost/python/args.hpp:8,
                 from /usr/include/boost/python.hpp:11,
                 from /mnt/int/system/aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/engine/src/python/config/python_utils.cpp:31:
/usr/include/python3.13/pylifecycle.h:46:43: note: declared here
   46 | Py_DEPRECATED(3.13) PyAPI_FUNC(wchar_t *) Py_GetPath(void);
      |                                           ^~~~~~~~~~
/mnt/int/system/aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/engine/src/python/config/python_utils.cpp: In function ‘PyObject* GetClassFromPython(std::string, std::string, std::string, std::string)’:
/mnt/int/system/aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/engine/src/python/config/python_utils.cpp:76:5: error: ‘Py_SetPath’ was not declared in this scope; did you mean ‘Py_GetPath’?
   76 |     Py_SetPath(python_path);
      |     ^~~~~~~~~~
      |     Py_GetPath
make[2]: *** [CMakeFiles/vegastrike_python.dir/build.make:79: CMakeFiles/vegastrike_python.dir/src/python/config/python_utils.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:224: CMakeFiles/vegastrike_python.dir/all] Error 2
make: *** [Makefile:156: all] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
```

Normally I have makeflags -j32, so this failure got jumbled among all the other jobs that are running, but for this build single jobs were specified.

stephengtuggy:

Please answer the following:

Code Changes:
- [X] (Partly) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Fixed:
Fixes #941 
Fixes #933 
Fixes #929 
Closes #857 

Issues Observed (Outstanding Issues): 
- Game keeps crashing and locking up my MacBook, shortly after I launch the game. I need to get that fixed if possible, but it will probably need to wait until another PR.

Purpose:
- What is this pull request trying to do? Achieve at least minimal game functionality with v0.10.x on macOS. Also, port over many of my changes from v0.9.x, such as lin_time refactors; logging-related changes; SDL and OpenGL-related changes; and vega_py_run-related changes. Fix the CI, and add workflows for Mac and Windows releases. Get vegasettings working too.


stephengtuggy:
I'm sorry, you guys! I had no idea how big this thing had gotten. 2,500 lines is a lot, no question about it.

royfalk:
> I'm sorry, you guys! I had no idea how big this thing had gotten. 2,500 lines is a lot, no question about it.

@BenjamenMeyer I'll take the source code stuff. Can you focus on the CI  side? You can, of course do everything if you want.

stephengtuggy:
The game starts, but crashes soon after startup. I was going to try to fix the crash(es), but it's turned into one thing after another. I think I will leave this PR as is, and we can address the crashes separately.

Thanks, guys.

stephengtuggy:
Hopefully, following the commit messages will tell the story pretty well and make reviewing this PR easier.

engine/setup/src/include/display.h
```diff
@@ -39,7 +39,7 @@ void ShowMain();
 
 #define GET_TITLE char title[1000]="Vega Strike Settings"; \
   if (strlen(CONFIG.program_name)+strlen(VERSION)<800) { \
-    sprintf(title, "Settings - %s - Version %s", \
+    snprintf(title, 1000, "Settings - %s - Version %s", \
```
BenjamenMeyer, line 42: should probably use 999 (`sizeof(title)-1`) so it's always zero terminated even if there's an error in the format string

BenjamenMeyer: Q: is there a reason we couldn't use `boost::format` here?

engine/src/cmd/script/script_variables.cpp
```diff
@@ -457,7 +457,7 @@ void Mission::doSetVar(missionNode *node, int mode) {
             }
             vi = global_var->script.varinst;
         }
-        if (vi->type != VAR_BOOL || vi->type != VAR_FLOAT || vi->type != VAR_INT || vi->type != VAR_OBJECT) {
+        if (vi->type != VAR_BOOL && vi->type != VAR_FLOAT && vi->type != VAR_INT && vi->type != VAR_OBJECT) {
```
BenjamenMeyer, line 460: this might be slightly more performant if a switch was used instead
```
swtich(vi->type) {
case VAR_BOOL:  // fall-thru
case VAR_FLOAT:  // fall-thru
case VAR_INT:  // fall-thru
case VAR_OBJECT:
    // do nothing
    break;
default:
    fatalError(node, node, "unsupported type in setvar");
    assert(0); // is there a better method of existing?
}

File: engine/src/precompiled.h
Comment (BenjamenMeyer) [inline engine/src/precompiled.h:1]: :+1:

File: engine/CMakeLists.txt [hunk 3]
```diff
@@ -32,6 +32,17 @@
...
+CMAKE_MINIMUM_REQUIRED(VERSION 3.21 FATAL_ERROR)
+
+IF (POLICY CMP0087)
+    cmake_policy(SET CMP0087 NEW)
+ENDIF ()
+SET(X_VCPKG_APPLOCAL_DEPS_INSTALL ON)
+
+IF (POLICY CMP0167)
+    CMAKE_POLICY (SET CMP0167 OLD)
+ENDIF (POLICY CMP0167)
```
BenjamenMeyer, line 44: This is all set in the root CMakeLists.txt so shouldn't be needed here.

engine/CMakeLists.txt
```diff
@@ -69,6 +80,8 @@ INCLUDE(CheckTypeSize)
 INCLUDE(CMakePushCheckState)
 SET(SIZEOF_VOID_P CMAKE_SIZEOF_VOID_P)
 
+SET(CMAKE_FIND_FRAMEWORK "LAST")
```
BenjamenMeyer, line 83: we probably should set this in the root CMakeLists.txt

engine/setup/CMakeLists.txt
```diff
@@ -22,10 +22,22 @@
...
+CMAKE_MINIMUM_REQUIRED(VERSION 3.21 FATAL_ERROR)
+
+IF (POLICY CMP0087)
+	cmake_policy(SET CMP0087 NEW)
+ENDIF ()
+SET(X_VCPKG_APPLOCAL_DEPS_INSTALL ON)
+
+IF (POLICY CMP0167)
+	CMAKE_POLICY (SET CMP0167 OLD)
+ENDIF (POLICY CMP0167)
```
BenjamenMeyer, line 34: This is set i the root Cmakelists.txt and shouldn't be needed here.

BenjamenMeyer (changes requested):
A few things to change in the CMakeLists.txt for stuff that is being duplicated from the root version.

With #889 I setup the root CMakeLists.txt which is going to become more and more pivotal going forward. Anything common should be done there and it can get inherited down the chain to the sub-projects. Eventually I'd like to have a folder for the different executables and libraries with each focused on what they specifically need to do and then the root will build the libraries first, then the executables, and then package them.

BenjamenMeyer:
@royfalk would very much appreciate your input on the reviews of the code.

engine/src/cmd/script/script_variables.cpp
stephengtuggy: Good idea

engine/setup/src/include/display.h
stephengtuggy: No reason. I was just in a hurry and didn't think of it. Thanks for the suggestion!

engine/CMakeLists.txt
stephengtuggy: OK. Fixed. What about the `CMAKE_MINIMUM_REQUIRED` clause? Do you consider that handled adequately at the root level as well?

stephengtuggy: Also, what about the `X_VCPKG_APPLOCAL_DEPS_INSTALL` setting? I'm not sure that that translates to subdirectories.

engine/CMakeLists.txt
stephengtuggy: I actually set `CMAKE_FIND_FRAMEWORK` a total of six times in this same file, flipping back and forth between `"LAST"` and `"FIRST"`, for reasons related to OpenGL on macOS. It's weird, but it seems to work.

engine/setup/CMakeLists.txt
stephengtuggy: Fixed

engine/src/cmd/script/script_variables.cpp
stephengtuggy: Done. Implemented it that way.

engine/setup/src/include/display.h
stephengtuggy: Oh, wait. Actually, this is in vegasettings. I'm not sure if we have Boost Format available there currently.

Also, from the [documentation of snprintf](https://en.cppreference.com/w/c/io/fprintf), I believe it stores one fewer characters than the length parameter you give it, to make room for the null terminator. So this should be fine as is, I believe.

engine/objconv/asteroidgen.cpp
```diff
@@ -30,7 +32,7 @@
 #define M_PI 3.14159265358979323846264338328
 #endif
 #include <float.h>
-#ifdef __APPLE__
+#if defined(__APPLE__) && defined (__MACH__)
```
stephengtuggy, line 35: To standardize how we detect macOS...

engine/setup/src/include/display.h
BenjamenMeyer: > Oh, wait. Actually, this is in vegasettings. I'm not sure if we have Boost Format available there currently.

That's fair

> Also, from the [documentation of snprintf](https://en.cppreference.com/w/c/io/fprintf), I believe it stores one fewer characters than the length parameter you give it, to make room for the null terminator. So this should be fine as is, I believe.

That would be highly dependent on the implementation of snprintf, and not guaranteed across C Libraries. I always just made it a habit to hand it one less than the buffer I created for it.

engine/CMakeLists.txt
BenjamenMeyer: > OK. Fixed. What about the `CMAKE_MINIMUM_REQUIRED` clause? Do you consider that handled adequately at the root level as well?

Yes, that only needs to happen once for the entire project. Let's not confuse ourselves.

> Also, what about the `X_VCPKG_APPLOCAL_DEPS_INSTALL` setting? I'm not sure that that translates to subdirectories.

This one might be more executable dependent - that is, it's specific to the Targets that make an executable. That would be appropriate here.

engine/CMakeLists.txt
BenjamenMeyer: yeah - I came across that when I was trying to figure out the MacOS stuff too. Fair enough.

engine/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@
 
 MESSAGE("PROCESSING DIRECTORY: ${CMAKE_CURRENT_SOURCE_DIR}")
 
+CMAKE_MINIMUM_REQUIRED(VERSION 3.21 FATAL_ERROR)
+
+SET(X_VCPKG_APPLOCAL_DEPS_INSTALL ON)
```
BenjamenMeyer, line 37: based on https://github.com/microsoft/vcpkg/blob/d7112d1a4fb50410d3639f5f586972591d848beb/scripts/buildsystems/vcpkg.cmake#L52 this would be target dependent; though not entirely sure we should depend on it given its an experimental option for VCPKG.

engine/CMakeLists.txt
```diff
@@ -32,6 +32,10 @@
 
 MESSAGE("PROCESSING DIRECTORY: ${CMAKE_CURRENT_SOURCE_DIR}")
 
+CMAKE_MINIMUM_REQUIRED(VERSION 3.21 FATAL_ERROR)
```
BenjamenMeyer, line 35: Yes this should be good at the root level only.

engine/setup/CMakeLists.txt
```diff
@@ -22,10 +22,15 @@
 # along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
 #
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.16 FATAL_ERROR)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.21 FATAL_ERROR)
+
+SET(X_VCPKG_APPLOCAL_DEPS_INSTALL ON)
```
BenjamenMeyer, line 27: See previous comments per L25-27

engine/setup/src/include/display.h
BenjamenMeyer: Okay - so looking a little closer - the behavior of snprintf using 1 less than specified is dependent on whether or not the implementation is C99 Compliant. I probably still would not assume C99 Compliance across all our platforms.

Mac - [kernel documentation](https://developer.apple.com/documentation/kernel/1441052-snprintf) doesn't say; though [an archived man page](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/snprintf.3.html) suggests its C99 compliant.

Linux - the [snprintf manpage](https://linux.die.net/man/3/snprintf) hints at it being C99 compliant but does not out right state it about the size parameter. Just not that it will `write at most size bytes (including the terminating null byte ('\0')) to str.`.

Windows - [MSVC 2019 snprintf](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-160) is not C99 compliant while [MSVC 2022 snprintf](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170) is.

So unless we're going to drop MSVC 2019 support then we should act as if it is not C99 compliant for the time being.

BenjamenMeyer (changes requested):
Looking good; please fix the other CMakeLists.txt directive for the minimum version.

engine/CMakeLists.txt
BenjamenMeyer: For the `X_VCPKG_APPLOCAL_DEPS_INSTALL` please continue conversation in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/974#discussion_r1912607454.

Per `CMAKE_MINIMUM_REQUIRED` - let's continue that here

engine/setup/src/include/display.h
stephengtuggy: Sure enough. The MSVC 2019 snprintf documentation you linked recommends passing the number of characters as one less than the length of the buffer, to avoid "squeezing out" the null terminator. OK. Good to know.

Maybe we should just pull in boost::format here, and use that across the board instead of snprintf. We know we have boost::format available anyway, since the main Vega Strike project depends on it.

stephengtuggy:
> Looking good; please fix the other CMakeLists.txt directive for the minimum version.

Done.

engine/setup/src/include/display.h
stephengtuggy: Fixed. Kept it using snprintf, at least for now.

engine/CMakeLists.txt
BenjamenMeyer: That's fine - let's just make sure we monitor the status of it in case it changes one way or the other.

BenjamenMeyer (approved):
@royfalk looks good from my end

Approved pending @royfalk 's approval for a more in-depth code side review

engine/setup/src/include/display.h
stephengtuggy: ... and eventually I did switch to using boost::format, after all 😄

engine/CMakeLists.txt
stephengtuggy: OK, I've removed the `CMAKE_MINIMUM_REQUIRED` clauses from everything except the root CMakeLists.txt.

engine/src/lin_time.cpp
```diff
@@ -144,7 +147,13 @@ void micro_sleep( unsigned int n )
 
 void micro_sleep( unsigned int n )
 {
-    (void) usleep( (useconds_t) n );
+    usleep(static_cast<useconds_t>(n));
+}
+
+#elif defined (__APPLE__) && defined (__MACH__)
```
royfalk, line 153: Probably better to look for all instances of micro_sleep and search and replace with sleep or nano_sleep.
https://stackoverflow.com/questions/42861913/why-use-usleep-and-not-sleep

engine/src/macosx_math.h
```diff
@@ -25,15 +27,15 @@
 /// works on *any* float... INF is 0x7ffffff and -INF is -(1<<31) NaN is 0
 extern int float_to_int(float);
 extern int double_to_int(float); //works on *any* float
-#if defined (__APPLE__) || defined (MACOSX)
```
royfalk, line 28: Is this really necessary today? Why wouldn't the same code work for Linux and Mac?
I have to check (again) and until 2006, Macs came with a power pc. Surely that was the reason for the discrepancy.

engine/src/star_system.cpp
```diff
@@ -918,8 +918,7 @@ void StarSystem::UpdateUnitsPhysics(bool firstframe) {
         try {
             UnitCollection col = physics_buffer[current_sim_location];
             un_iter iter = physics_buffer[current_sim_location].createIterator();
-            Unit *unit = nullptr;
-            for (; (unit = *iter); ++iter) {
+            for (Unit *unit = nullptr; (unit = *iter); ++iter) {
```
royfalk, line 921: Add a TODO here and create an issue please - this is a prime candidate for multi-threading. NM. Opened #992.

engine/src/gldrv/gl_init.cpp
```diff
@@ -281,14 +282,14 @@ void init_opengl_extensions() {
 #endif
         VS_LOG(trace, "OpenGL::GL_EXT_compiled_vertex_array supported");
     } else {
-#ifdef __APPLE__
+#if defined(__APPLE__) && defined (__MACH__)
```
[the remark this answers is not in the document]
stephengtuggy: My understanding is that the main cases where these would differ would be on other Apple operating systems, such as iOS, iPadOS, etc.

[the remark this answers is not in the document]
stephengtuggy: And I got this specific syntax from some Stack Overflow Q&A. I'd have to look to find which one it was again. I didn't just pull it out of a hat, in other words, as arbitrary as it may seem at first glance.

engine/src/gldrv/gl_init.h
```diff
@@ -56,19 +55,16 @@ extern "C"
  * count on that...  */
 
 #if defined (IRIX)
```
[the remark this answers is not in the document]
stephengtuggy: Good point about IRIX. Maybe I/we can start removing the IRIX-specific code in future PRs.

engine/src/gldrv/winsys.cpp
```diff
@@ -246,85 +247,105 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
-            VS_LOG(error, "GDI Generic software driver reported, trying to reset.");
+            VS_LOG_AND_FLUSH(error, "GDI Generic software driver reported, trying to reset.");
+            SDL_ClearError();
             SDL_Quit();
             game_options()->gl_accelerated_visual = false;
             return false;
         } else {
-            VS_LOG(error, "GDI Generic software driver reported, reset failed.\n");
-            VS_LOG(error, "Please make sure a graphics card driver is installed and functioning properly.\n");
+            VS_LOG(error, "GDI Generic software driver reported, reset failed.");
```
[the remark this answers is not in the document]
stephengtuggy: Software rendering still does happen. I've seen it on my own systems. It's usually not a hardware problem, but a driver issue or similar.

engine/src/lin_time.cpp
stephengtuggy: That's very useful info. I did not know about nano_sleep.

Maybe I can switch to that in a near-future PR?

engine/src/star_system.cpp
stephengtuggy: I emphatically agree. Though we will need to be careful in how we implement it. We'll probably have to have some sort of locking mechanism so that two threads don't do conflicting things with the same Unit at the same time.

But yes, this is definitely something I want to look at figuring out in the near future. The physics calculations are by far the most time-consuming part of the game. Making that multi-threaded would be a boon for performance.

engine/src/macosx_math.h
stephengtuggy: I don't know the reason for the discrepancy. Absent that knowledge, I prefer to leave the code mostly as-is.

Can we table this for now, and come back to it at a future date? Like next week, maybe?

engine/src/gfx/sprite.h
```diff
@@ -14,7 +16,7 @@
  *
  * Vega Strike is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
```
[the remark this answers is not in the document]
stephengtuggy: Yeah, you're right.

I'm working on a 4-part series of PRs to update the copyright notices on the 0.9.x branch. I will probably follow that up very soon with updates to the copyright notices on the master branch. Then we should be able to leave these headers alone until next year, hopefully.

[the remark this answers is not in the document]
stephengtuggy: Until then, I will try to avoid nit-picking the copyright headers like this.

engine/src/accountserver.cpp
```diff
@@ -31,7 +33,10 @@ using std::string;
 #include "networking/acctserver.h"
 #include "lin_time.h"
```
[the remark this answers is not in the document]
stephengtuggy: That makes sense. Shall I cut the file right now? Or can it wait a bit?

engine/setup/src/include/display.h
```diff
@@ -39,7 +32,7 @@ void ShowMain();
 
 #define GET_TITLE char title[1000]="Vega Strike Settings"; \
   if (strlen(CONFIG.program_name)+strlen(VERSION)<800) { \
```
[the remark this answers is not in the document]
stephengtuggy: Yeah, the C-style strings bother me too. The main reason I left some of it here is that this same header is used by two different .cpp implementation files -- display_gtk.cpp and display_dialog.cpp. In other words, there are some complicated interdependencies at play here.

But yes, I would like to get rid of the C-style strings in general. I'll file an issue for this particular section of code.

engine/setup/src/include/display_gtk.cpp
```diff
@@ -47,11 +51,13 @@ void exit_0(GtkWidget *w, void *arg) {
 
 void InitGraphics(int *argc, char ***argv) {
     gtk_init(argc, argv);
-    GET_TITLE; // sets title; uses sprintf, not snprintf -- //[MSVC-Warn]
+    GET_TITLE;
     GET_STATIC_TEXT;
     window = gtk_window_new(GTK_WINDOW_TOPLEVEL);
     gtk_window_set_default_size(GTK_WINDOW(window), 600, 400);
-    gtk_window_set_title(GTK_WINDOW(window), title);
+    gtk_window_set_title(
```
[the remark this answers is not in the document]
stephengtuggy: See my above comment.

PR #982, opened 2025-01-13
Enhancement: Minimize Download Size

### LINKED ISSUE #83: Git repository history might contain bloat

BenjamenMeyer:
Document how others can minimize downloads, specifically for maintainers of systems like Gentoo, Arch, etc.

Closes #83

Please answer the following:

Code Changes:

Issues:
- #83 

Purpose:
- What is this pull request trying to do? Provide a way to reduce download bloat
- What release is this for? N/A
- Is there a project or milestone we should apply this to? N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,21 @@ If you encounter any issues while playing, please create an issue with the Vega 
 
 # Compiling Vegastrike
 
+## Saving some space while downloading
+
+The Vega Strike Engine Source repository contains a lot of history not all of which is relevant to current developers if they want to quickly download and build something.
+By default, the full clone will download over 800 MB of data. There are several options to minimize this:
+
+1. Download a compressed copy of the source from the release
+2. Download a master.zip from GitHub
+3. Do a shallow clone using the following:
+
+```bash
+$ git clone git@github.com:vegastrike/Vega-Strike-Engine-Source.git --shallow-since=2023-09-27
+```
+
+This will produce a significantly smaller download - in the order of 22-30 MB; well over a 10x reduction.
+
 ## Compiling On Linux
 
 1. Install the development dependencies:


BenjamenMeyer:
@evertvorster please validate this with your AUR builds

stephengtuggy (approved):
That's good to know. Approving from my side.

evertvorster:
Hmm, I have the full depth git repos of Assets and Engine, and do not wipe them out between rebuilds, rather they get updated and the proper PR gets pulled for building. 

Also, I am now behind a very restrictive internet connection on a little boat in Ghana. This situation is likely to continue for eight weeks, so I don't particularly want to wipe my git repos.

However, doing shallow checkouts is a technique that I have used in the past to get a checkout that I can use. 

Don't let me hold you back on this one. Approved.

evertvorster (approved):
These instructions are good. 
Approved

Issue #969, opened 2025-01-10
Game coredumps when hitting m


royfalk:
It looks like the function GetHull refers to an improperly initialized current_hull pointer.

PR #987, opened 2025-01-13
Fix some crashes 2025-01-13

### LINKED ISSUE #981: Game crashes upon loading a new campaign.
Hi there!

Just compiled and installed git master of the engine and assets. 
The game starts up fine, but as soon as you start a new campaign the game exits. 
In the console:
```
Launching bases for Special/Empty
Adding news
conditioning
nonzeroing
terminate called after throwing an instance of 'std::invalid_argument'
  what():  stoi
/usr/bin/vs: line 28: 100896 Aborted                 (core dumped) vegastrike -d"${VEGASTRIKE_SHARE_DIR}" "$1"
```

Here is the latest log as well. 
[vegastrike_2025-01-13_05_32_02.127679.log](https://github.com/user-attachments/files/18393033/vegastrike_2025-01-13_05_32_02.127679.log)

More about my system:
```
Operating System: Arch Linux 
KDE Plasma Version: 6.2.5
KDE Frameworks Version: 6.9.0
Qt Version: 6.8.1
Kernel Version: 6.13.0-rc6-1-git-00262-gb62cef9a5c67-dirty (64-bit)
Graphics Platform: X11
Processors: 32 × AMD Ryzen 9 7945HX3D with Radeon Graphics
Memory: 62.0 GiB of RAM
Graphics Processor: AMD Radeon 610M
Manufacturer: ASUSTeK COMPUTER INC.
Product Name: ROG Strix G733PYV_G733PYV
System Version: 1.0
```

stephengtuggy:
Code Changes:

Issues:
- Addresses #981 

Purpose:
- What is this pull request trying to do? Fix #981 , along with several other crashes


stephengtuggy:
Note that this PR includes most of the changes from #974 , so should probably be tackled after #974 is out of the way.

evertvorster:
Hi there!

This does fix the crashing when loading a campaign #981

Pressing "m" in flight still causes a segfault. 

Also, it breaks ship info and purchasing repair and refuel's description, among other things, on a new campaign.

engine/setup/src/include/display.h
```diff
@@ -39,7 +39,7 @@ void ShowMain();
 
 #define GET_TITLE char title[1000]="Vega Strike Settings"; \
   if (strlen(CONFIG.program_name)+strlen(VERSION)<800) { \
-    sprintf(title, "Settings - %s - Version %s", \
+    snprintf(title, 1000, "Settings - %s - Version %s", \
```
BenjamenMeyer, line 42: per our other discussion - this should be `sizeof(title)-1`

Is this work on top of that branch?

BenjamenMeyer (changes requested):
Is this on top of the other work I reviewed? seems like a lot of duplication.

stephengtuggy:
@evertvorster yeah, I was encountering the `PyImport_ImportModule is null` error a lot yesterday or the day before, too. I'll see if I can straighten that out without adding too much to th e scope of this PR.

engine/setup/src/include/display.h
BenjamenMeyer: thanks!

evertvorster:
Just ran a little test now,  and the python errors have changed somewhat. 
When clicking on ship's information, it states that ship_view.py was not found, and when clicking on player info, it states that player_info.py was not found. 
In upgrades, it complains about upgrade_view.py not being found. 

So, you are on the right track, I believe.

stephengtuggy:
> Just ran a little test now, and the python errors have changed somewhat. When clicking on ship's information, it states that ship_view.py was not found, and when clicking on player info, it states that player_info.py was not found. In upgrades, it complains about upgrade_view.py not being found.
> 
> So, you are on the right track, I believe.

LOL that's good to hear.

In my very latest changes, I have succeeded in getting the .py files to load. Now it's just a matter of debugging the Python code! 😃

evertvorster:
Much, much improved!

Upgrades now work, as do player information. 
The only bit that still throws an error is ship info:

image


Once this is sorted, I'll test all the major functions of the game on my system. They all worked pretty much ok the last time I tested, but it is best to make sure.

stephengtuggy:
> Much, much improved!
> 
> Upgrades now work, as do player information. The only bit that still throws an error is ship info:
> 
> image
> 
> Once this is sorted, I'll test all the major functions of the game on my system. They all worked pretty much ok the last time I tested, but it is best to make sure.

That's really good to hear! Ship View is now working for me too. For that, I had to make some changes to ship_view.py in the assets. I have a PR open for that here: https://github.com/vegastrike/Assets-Production/pull/144

evertvorster:
Looking VERY good!

Only issue I see now is that in ship info it tells me that there is no shielding:

image


This may be because of the change in how many facets a shield has, and not related to this PR at all. Once this one is merged, I'll raise an issue on the engine for this one specifically.

evertvorster (approved):
Tested this for every major function that I can think of, and everything works. 
I did notice a few bugs from long ago, but not relevant to this PR.

kheckwrecker:
The only log I can find (in my .vegastrike/logs directory) has nothing in it related to the crash. And worse than that, when I tried it again, it works fine without any crash. I hate intermittent bugs.

stephengtuggy:
@evertvorster that's awesome!

@kheckwrecker I know, those are the worst. Unfortunately, the game has had intermittent crashes for quite a while. Probably not much we can do about that in this PR. Soon, though. Hopefully.

royfalk:
I'm going to hold off on reviewing this until we merge #974. There are too many changes otherwise.

engine/src/cmd/controls_factory.cpp
```diff
@@ -272,14 +271,18 @@ Control* getControl(std::map<std::string, std::string> attributes) {
     // Font
     if(attributes.count("font")) {
         std::vector<double> font_array = splitAndConvert(attributes["font"], ',');
-        Font font(font_array[0], font_array[1]);
-        c->setFont(font);
+        if (font_array.size() < 2) {
+            VS_LOG(error, "controls_factory getControl(): font_array doesn't have enough elements");
```
stephengtuggy, line 275: I don't know why this condition is happening. We will probably want to investigate further. But for now, this keeps the game from crashing at this point.

engine/src/cmd/unit_json_factory.cpp
```diff
@@ -39,7 +38,12 @@
 void UnitJSONFactory::ParseJSON(VSFileSystem::VSFile &file, bool player_ship) {
     const std::string json_text = file.ReadFull();
 
-    boost::json::value json_value = boost::json::parse(json_text);
+    boost::json::value json_value;
+    try {
+        json_value = boost::json::parse(json_text);
+    } catch (std::exception const& e) {
```
stephengtuggy, line 44: Let's try to be more consistent about putting try-catch blocks around code that could reasonably be expected to throw.

engine/src/python/init.cpp
```diff
@@ -69,6 +69,9 @@ class Unit;
 #define PATHSEP "/"
 #endif
 
+#define Q(x) #x
+#define QUOTE(x) Q(x)
+
 void Python::initpaths() {
```
stephengtuggy, line 75: As it turns out, Vega Strike already had a section of code for passing paths to Python. This was it. I added a few lines to this function, and suddenly the code was able to find the base_computer Python scripts.

royfalk: I originally set out to pass actual structs between C++ and python. I actually managed that but then realised that it's probably easier to just pass maps in most cases.

I did look at the original code but was unable to find an example that didn't start with something like 
_RunPython("some_python_code_snippet...");_
whereas I wanted
_RunPython("python_file", "python_function", args);_

engine/src/cmd/basecomputer.cpp
```diff
@@ -114,8 +116,8 @@ std::vector<std::string> getWeapFilterVec() {
 std::vector<std::string> weapfiltervec = getWeapFilterVec();
 
 bool upgradeNotAddedToCargo(std::string category) {
-    for (unsigned int i = 0; i < weapfiltervec.size(); ++i) {
-        if (weapfiltervec[i].find(category) == 0) {
+    for (const auto & i : weapfiltervec) {
```
royfalk, line 119: I personally don't have a problem with auto, but @BenjamenMeyer frowns upon them. 
Therefore, where I can deduce the type, I'll use it.

engine/src/cmd/basecomputer.cpp
```diff
@@ -527,7 +529,7 @@ void BaseComputer::constructControls(void) {
         return;
     }
 
-    if (m_displayModes.size() != 1 || m_displayModes[0] != NETWORK) {
+    if (m_displayModes.size() != 1 || m_displayModes.at(0) != NETWORK) {
```
royfalk, line 532: I assume these fail for some OS'es. I have the dangest errors in the CI, where my machine blissfully compiles and runs.

engine/src/cmd/basecomputer.cpp
```diff
@@ -3942,7 +3943,7 @@ string buildUpgradeDescription(Cargo &item) {
     const std::string key = item.GetName() + "__upgrades";
     PyObject* args = PyTuple_Pack(1, PyUnicode_FromString(key.c_str()));
     const std::string text = GetString("get_upgrade_info", "upgrade_view",
-        "python/base_computer/upgrade_view.py", args);
+        "upgrade_view.py", args);
```
royfalk, line 3946: Can you expand on why you did this?

engine/src/cmd/controls_factory.cpp
royfalk: So, this raises a bigger philosophical question of how much crap are we willing to let the engine take from the assets?

For JSON files, the file obviously needs to be formatted correctly or parsing will fail. Most JSON files are crucial to the game, so may as well crash it with an error.

But this is inside content and for a specific control so it points to an error in how I converted fonts to controls. Best open an issue/bug for this.

engine/src/cmd/damageable.cpp
```diff
@@ -524,7 +524,7 @@ float Damageable::RShieldData() const {
 void Damageable::ArmorData(float armor[8]) const {
     Damageable *damageable = const_cast<Damageable *>(this);
     DamageableLayer armor_layer = damageable->GetArmorLayer();
-    for (int i = 0; i < 8; i++) {
```
royfalk, line 527: Yeah. You caught a bug of mine here. Did this segfault or fill the array with junk?

engine/src/gfx/vdu.cpp
```diff
@@ -1185,21 +1185,18 @@ void VDU::DrawDamage(Unit *parent) {
...
-            (armor[0] + armor[2] + armor[4] + armor[6])
-                    / (float) (StartArmor[0] + StartArmor[2] + StartArmor[4] + StartArmor[6]),
-            (armor[0] + armor[1] + armor[4] + armor[5])
-                    / (float) (StartArmor[0] + StartArmor[1] + StartArmor[4] + StartArmor[5]),
-            (armor[2] + armor[3] + armor[6] + armor[7])
-                    / (float) (StartArmor[2] + StartArmor[3] + StartArmor[6] + StartArmor[7]),
-            (armor[1] + armor[3] + armor[5]
-                    + armor[7]) / (float) (StartArmor[1] + StartArmor[3] + StartArmor[5] + StartArmor[7]),
-            parent->GetHull() / (*maxhull), true, false);
+    DrawHUDSprite(this, draw_damage_sprite ? parent->getHudImage() : nullptr, .6, x, y, w, h,
```
royfalk, line 1194: A lot of the stuff for armor and 8 facets is stuff I'm waiting to submit. It would be interesting to compare and also a bit frustrating.

engine/src/python/infra/get_string.cpp
```diff
@@ -38,32 +37,36 @@ const std::string GetString(const std::string function_name,
                             const std::string module_name,
                             const std::string file_name,
                             PyObject* args) {
-    if(!boost::filesystem::exists(file_name)) {
```
royfalk, line 41: Why remove the check for the file? Is it because we no longer use absolute paths in the name?

royfalk (commented):
Still a long review but much better. Thanks.

@BenjamenMeyer we may need to think about syncing our efforts so we don't write the same code twice. At least some of the code is stuff I've written but was waiting for the cmake stuff to get merged and splitting other stuff from it (like #969 ) so the actual review of the PR will be shorter and easier.

engine/src/cmd/controls_factory.cpp
stephengtuggy: Filed as https://github.com/vegastrike/Assets-Masters/issues/56

engine/src/python/init.cpp
stephengtuggy: Maybe I should restate that. Vega Strike already had a function for initializing the Python paths. That's what this was.

I don't have a problem with you adding the `RunPython("python_file", "python_function", args);` function, @royfalk . I think that part makes sense.

engine/src/cmd/basecomputer.cpp
stephengtuggy: @BenjamenMeyer has said previously that he's OK with auto for iterators.

engine/src/cmd/basecomputer.cpp
stephengtuggy: These don't fail to compile on any OS particularly. This is just to add bounds checking, to make sure we're not reading or writing outside the bounds of the collection.

engine/src/cmd/basecomputer.cpp
stephengtuggy: You were specifying the subfolder (`"python/base_computer"`) twice. Once here, and once in the Python path initialization code. That's part of why VS was failing to locate these Python scripts.

Now the subfolder is specified in only one place: `initpaths()`. Which is more or less in keeping with the way existing code did things.

engine/src/cmd/damageable.cpp
stephengtuggy: I'm not sure if I actually hit this section of code during my testing. If I had, though, I think a segfault would have been most likely. The destination array was too small to write 8 elements to.

engine/src/gfx/vdu.cpp
stephengtuggy: Yeah. We would do well to communicate a bit better in real time about stuff we're working on. That isn't always feasible though, what with time zone differences and such.

engine/src/python/infra/get_string.cpp
stephengtuggy: Yes. For that reason, and also because this function doesn't have enough information to know the full path to the file. As far as it's concerned, the Python file could be in one of a number of locations.

engine/src/python/init.cpp
royfalk: That's absolutely fine.

engine/src/cmd/basecomputer.cpp
royfalk: I'm concerned that we could potentially have two utils.py somewhere with unexpected and hard to debug errors.

engine/src/python/infra/get_string.cpp
royfalk: ok

royfalk (approved):
LGTM

stephengtuggy:
> LGTM

Thanks!

PR #989, opened 2025-01-14
Enhancement: Clean up the Stdafx.h usage

### LINKED ISSUE #512: Do away with `Stdafx.h`?
So `StdAfx.h` is a Windows common header for precompiled headers, and honestly we should move away from using it.
1. Precompiled headers don't really save time in anything, and end up just complicating the build.
2. It's better to have individual files be explicit and limited in what they include, only including the absolutely minimal necessary headers to build. This does mean more work and research on Windows to find the right header set as one won't simply do a "#include <windows.h>" and but this alone does far more for speeding up builds than using precompiled headers.

I know...a big change; but it'd likely be worth it in the end. This is just my recommendation. I'll leave it to the group to decide if we head this direction, but that won't likely be until 0.9.x.

$0.02 for what it's worth.

_Originally posted by @BenjamenMeyer in https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/315#issuecomment-738578963_

BenjamenMeyer:
Do not use a precompiled header.

NOTE: This is the first step for a refactor of the headers. The next step would be to reduce what headers are used in any given implementation or header file.

Please answer the following:

Code Changes:

Issues:
- Closes #512

Purpose:
- What is this pull request trying to do? Remove Stdafx.h and get rid of precompiled headers entirely


BenjamenMeyer:
still trying to figure out the vegasettings crash issue; but this compiles and behaves the same. It's all header changes and just simplifying things. Some testing still needed.

engine/src/cmd/collide2/CSopcodecollider.cpp
```diff
@@ -27,12 +27,15 @@
...
  * Copyright (C) 2020-2022 Stephen G. Tuggy
+ * Copyright (C) 2025 Benjamen R. Meyer
  */
 
 
+#include "Opcode.h"
 #include "CSopcodecollider.h"
 #include "opcodeqsqrt.h"
 #include "opcodeqint.h"
+#include "vs_vector.h"
```
stephengtuggy, line 38: I actually just got rid of vs_vector a few days ago in one of my PRs. vs_vector was a mistake from the beginning. It inherits from std::vector, which is not intended to be inherited from. (std::vector doesn't have a virtual destructor.)

vs_vector was only ever used in one place (in this file). It can be replaced with std::vector again, I think. Then if we want to enforce bounds checking (which we probably should) we can change all occurrences of `operator[]` with `.at()`.

BenjamenMeyer: Cool.

BenjamenMeyer:
Looks like the expected builds failed.

engine/src/cmd/collide2/opcodetypes.h
```diff
@@ -22,8 +22,8 @@
 #define VEGA_STRIKE_ENGINE_CMD_COLLSION2_CS_CSTYPES_H
 
 // config.h is the VS ./configure determined header.
-#include <float.h>
-#include <wchar.h>
+#include <cfloat>
+#include <cwchar>
```
stephengtuggy, line 26: Nice to see these replacements being made

engine/src/cmd/collide2/CSopcodecollider.cpp
royfalk: This is #987, which is waiting on #974.

royfalk (approved):
I don't have anything to contribute here. It seems pretty straightforward.
I can try and squeeze a quick playthrough to test things, early next week.

stephengtuggy (approved):
LGTM

Issue #56, opened 2020-04-08
Restructure the source


BenjamenMeyer:
I started looking at the https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/CMakeLists.txt in part to add a C++ Linter and test it out. I did manage to get something, but it might have changed compilers on me since it was clang-tidy. Any how... I started looking at the file itself. There's a lot of cleanup that can be done to it and make it less work to maintain. For example:

```
FILE(GLOB LIB_NET_LOW_LEVEL_FILES
         RELATIVE_PATH src/networking/lowlevel
         *.cpp
)
ADD_LIBRARY(netlowlevel ${LIB_NET_LOW_LEVEL_FILES})
```
will do what https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/CMakeLists.txt#L51 through https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/CMakeLists.txt#L77 does and not require manual edits every time a new C++ file is added.

The problem, however, is that are several different targets pull files from that directory.

At a minimum, we should move the files around so that we can easily glob them like the above.
It would be good also if files were moved around based on sub-targets too.

Thoughts?

BenjamenMeyer:
I'm going to also suggest too that we change up the location of CMakeFiles.txt so that it's at the base or the tree; we can layer them in sub-directories as needed. This will clean up the structure and bring things to standard more.

BenjamenMeyer:
I'll take on the restructure if you all want to do it. If not; I'll leave things long.

BenjamenMeyer:
One goal here would be to enable making different compiler targets and even tooling too.

royfalk:
Given the fact that a lot of the code is about the multiplayer and server, I would strongly urge that any major refactoring and restructuring begins with a complete removal of these components. I've done this (and broke something) in my repo (https://github.com/royfalk/vegastrike-code). The result would be a much smaller code base without any impact on game play.
A followup issue can remove warnings of unused code and variables.
Of course, this is something I expect someone to do after a unanimous decision only.

nabaco:
@royfalk Why are are you offering to remove the multiplayer/server code? It doesn't work at all? Are you offering to re-write it from scratch once we bring the SP up to level?

stephengtuggy:
[...]

PR #991, opened 2025-01-15
License Updates for 0.9.x Part 2

stephengtuggy:

Please answer the following:

Code Changes:
- [X] (Mostly) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- N/A?

Purpose:
- What is this pull request trying to do? Update 155 more files with copyright-notice-related changes, including the 55 .h files without a blank line betwen the copyright header comment and the following comment, as well as the 105 collision detection files (both .cpp and .h). (There is some overlap between these two sets of files.)


engine/src/cmd/collide2/OPC_PlanesCollider.h
```diff
@@ -3,6 +3,8 @@
  *	OPCODE - Optimized Collision Detection
  *	Copyright (C) 2001 Pierre Terdiman
  *	Homepage: http://www.codercorner.com/Opcode.htm
+ *  Copyright (C) 2021, 2022, 2025 Stephen G. Tuggy
+ *  Copyright (C) 2023 Benjamen R. Meyer
```
BenjamenMeyer, line 7: we need to reference a license of some sort for this...
from http://www.codercorner.com/Opcode.htm:
```
You’re free to use them in any commercial or non-commercial program.
```
Which we can probably reference as an `UNLICENSE` or probably best to just say `Public Domain` (or https://wiki.creativecommons.org/wiki/public_domain) as that's effectively what it is.

engine/src/cmd/collide2/OPC_PlanesTriOverlap.h
```diff
@@ -1,9 +1,30 @@
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 /**
- * Updated by Stephen G. Tuggy 2021-07-03
- * Updated by Stephen G. Tuggy 2022-01-06
- * Updated by Benjamen R. Meyer 2023-05-27
+ * OPC_PlanesTriOverlap.h
```
BenjamenMeyer, line 3: All the files related to http://www.codercorner.com/Opcode.htm are public domain.
For 0.9.x let's hold off on adding GPL to them for the moment. Reviewing the Zip file from there for the 1.3 release it seems there's a couple text files (Readme, TemporalCoherence) that are missing from VS. I'll file a story about that library into its own library as part of #56 but we'll do that on master.

engine/src/vs_random.h
```diff
@@ -31,8 +31,6 @@
  *  Any feedback is very welcome.
  *  http://www.math.keio.ac.jp/matumoto/emt.html
  *  email: matumoto@math.keio.ac.jp
- */
-// NO HEADER GUARD
```
BenjamenMeyer, line 35: we need to keep the `NO HEADER GUARD` comment

BenjamenMeyer (changes requested):
1. We need to keep the `NO HEADER GUARD` comments - most were kept already; there were a couple that were dropped.
2. The Opcode and Ice code (but not the CS code) in the collider2 are essentially public domain licensed given how the author licensed them (see the website). After reading https://www.gnu.org/licenses/gpl-faq.en.html we need to craft a different licensing for them such that the original files are still essentially public domain and any modifications to them are GPL. I can't speak yet to the licensing on the CS files in there; they're probably from Crystal Space - and we need to research https://github.com/crystalspace/CS/tree/master on them - looks like CS is LGPL, but we'd have to go find it in the history and compare files in case CS changed from GPL to LGPL; but we need to make sure we have the right license on that code.

engine/objconv/basemaker/base_maker_texture.h
```diff
@@ -18,8 +18,6 @@
  *
  * You should have received a copy of the GNU General Public License
  * along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
- */
-// NO HEADER GUARD
```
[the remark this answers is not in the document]
stephengtuggy: This was dropped by mistake. Fixed.

engine/src/cmd/collide2/CSopcodecollider.cpp
```diff
@@ -1,5 +1,7 @@
 /*
 	Copyright (C) 2000 by Jorrit Tyberghein
+    Copyright (C) 2020 pyramid3d
+    Copyright (C) 2020-2025 Stephen G. Tuggy
```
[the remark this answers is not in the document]
stephengtuggy: Fixed.

engine/src/cmd/collide2/OPC_AABBTree.cpp
```diff
@@ -61,8 +63,7 @@
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 /**
- * Updated by Stephen G. Tuggy 2021-07-03
- * Updated by Stephen G. Tuggy 2022-01-06
+ * Copyright (C) 2021, 2022, 2025 Stephen G. Tuggy
```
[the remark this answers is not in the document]
stephengtuggy: Done. Looks like I missed this file.

engine/src/vs_random.h
stephengtuggy: Fixed

engine/src/cmd/collide2/OPC_PlanesCollider.h
stephengtuggy: Fixed

engine/src/cmd/collide2/OPC_PlanesTriOverlap.h
stephengtuggy: Put these under "Public Domain."

BenjamenMeyer (approved):
Thanks - we can license non-whitespace changes under GPL; but this is good for 0.9.x.

PR #998, opened 2025-01-17
Fix Issue #997

### LINKED ISSUE #997: Crash loading saved games on Windows due to "root" values in the ship JSON like "V:\Assets-Production" instead of "V:\\Assets-Production"

stephengtuggy:
Code Changes:

Issues:
- Fixes #997 

Purpose:
- What is this pull request trying to do? Fix a bug that affects saving and loading unit files on Windows


evertvorster:
You know what would be great? If the PR's or engine have some sort of description of which branch of Assets they need to be tested against.

kheckwrecker:
Having a problem with game crashing after launching new campaign. Here's the debugger output:
Exception thrown at 0x00007FFDEABEB699 in vegastrike-engine.exe: Microsoft C++ exception: boost::wrapexcept<boost::system::system_error> at memory location 0x00000006C056C2E0.
Error parsing JSON in UnitJSONFactory::ParseJSON(): syntax error [boost.json:1 at D:\VS_Master_Test\Vega-Strike-Engine-Source\build\VS2022Win64-pie-disabled-release\vcpkg_installed\vcpkg\blds\boost-json\src\ost-1.84.0-6591c5c930.clean\include\boost/json/basic_parser_impl.hpp:1270:19 in function 'parse_escaped']
The thread 10244 has exited with code 0 (0x0).
The thread 'alsoft-mixer' (9648) has exited with code 0 (0x0).
The thread 11380 has exited with code 0 (0x0).
The thread 11572 has exited with code 42 (0x2a).
The thread 9248 has exited with code 42 (0x2a).
The thread 11648 has exited with code 42 (0x2a).
The thread 12080 has exited with code 42 (0x2a).
The thread 2316 has exited with code 42 (0x2a).
The thread 12104 has exited with code 42 (0x2a).
The thread 9356 has exited with code 42 (0x2a).
The thread 11808 has exited with code 42 (0x2a).
The thread 12272 has exited with code 42 (0x2a).
The thread 9444 has exited with code 42 (0x2a).
The thread 11564 has exited with code 42 (0x2a).
The thread 10836 has exited with code 42 (0x2a).
The program '[12228] vegastrike-engine.exe' has exited with code 42 (0x2a).

Is this the same problem?

stephengtuggy:
@kheckwrecker yeah, I'm pretty sure that's the same problem that this PR addresses.

kheckwrecker:
Ok. I'll try pulling this patch and see if it works.

kheckwrecker:
GitHub is having trouble with the pull request. Something about not find your work? Anyway, I'll just wait until it's approved and merged then try again.

kheckwrecker:
I pulled this patch and re-compiled, but it's still crashing with the same error. It's entirely possible that I'm screwing up the pull and merge, so I'll wait and try again after this patch is officially merged with master. If the problem still occurs after that, I'll create a new issue report.

engine/src/cmd/csv.cpp
```diff
@@ -291,9 +291,9 @@ CSVRow::CSVRow(CSVTable *parent, const string &key) {
     iter = parent->rows[key] * parent->key.size();
 }
 
-CSVRow::CSVRow(CSVTable *parent, unsigned int i) {
+CSVRow::CSVRow(CSVTable *parent, unsigned int which) {
```
royfalk, line 294: I recently disabled CSV support in the game. Not sure which PR.
Tracing CSVRow, it's not actually used by anything.

engine/src/cmd/unit_csv.cpp
```diff
@@ -65,28 +65,9 @@ extern void pushMesh(std::vector<Mesh *> &mesh,
...
-        }
-        
-        i++;
-        json_string += new_line.str();
-    }
-
-    json_string += "\t}\n]\n";
-    
-    return json_string;
+    std::string return_value = boost::json::serialize(boost::json::value_from(unit));
```
royfalk, line 69: Good! However, it's only used once and is now a really trivial function. Please fold it into WriteUnit.

engine/src/cmd/unit_csv.cpp
```diff
@@ -1127,14 +1108,11 @@ void Unit::LoadRow(std::string unit_identifier, string modification, bool saved_
 
 CSVRow GetUnitRow(string filename, bool subu, int faction, bool readlast, bool &rread) {
```
royfalk, line 1109: This function is not called. You can delete it rather than refactor. It's a good practice in this project to check if the function is used. As I refactor, code gets orphaned (?).  Also, the original devs left quite a few instances of stuff in progress that wasn't actually complete. Obviously, it would be much harder for us to know where they were going with it today.

engine/src/cmd/unit_json_factory.cpp
```diff
@@ -44,24 +40,45 @@ void UnitJSONFactory::ParseJSON(VSFileSystem::VSFile &file, bool player_ship) {
...
+            unit_attributes["root"] = file.GetRoot();
+
+            if (player_ship) {
+                UnitCSVFactory::units["player_ship"] = unit_attributes;
+            } else {
+                UnitCSVFactory::units[unit_attributes["Key"]] = unit_attributes;
+            }
+        }
+    } else {
+        VS_LOG_FLUSH_EXIT(fatal, (boost::format("Saved game unit file '%1%' had an unexpected JSON structure. We don't know how to process it.") % file.GetFilename()), 42);
```
royfalk, line 81: I'm pretty sure this code also loads the ships from unit.json but not ships.json. That one is handled by UnitOptimizeFactory.

engine/src/cmd/unit_json_factory.cpp
```diff
@@ -24,16 +24,12 @@
...
-#include <sstream>
 #include <vector>
 #include <string>
 #include <map>
 #include <boost/json.hpp>
 
-#include "resource/json_utils.h"
-
 
 void UnitJSONFactory::ParseJSON(VSFileSystem::VSFile &file, bool player_ship) {
```
royfalk, line 34: My code (deficiencies and all) assumed the following:
```
[
   {"a": "1", "b": 2},
   {"a": "1", "b": 2},
   {"a": "1", "b": 2},
   ...
]
```

However, this looks like it supports other file formats such as `{"a": "1", "b": 2}` without an array. Why would we do this?

royfalk (commented):
Where is the fix for root? I assumed the offending line would be 
```unit_attributes["root"] = file.GetRoot();```
but you didn't actually change it.

File: engine/src/cmd/unit_csv.cpp (hunk 2 above)
Comment (stephengtuggy) [reply engine/src/cmd/unit_csv.cpp:69]: OK, I can do that.

This is the actual fix. The prior implementation of this function didn't escape backslashes properly.

File: engine/src/cmd/csv.cpp (hunk 1 above)
Comment (stephengtuggy) [reply engine/src/cmd/csv.cpp:294]: OK, I can take it out.

**Comment by stephengtuggy** [comment]
> Where is the fix for root? I assumed the offending line would be `unit_attributes["root"] = file.GetRoot();` but you didn't actually change it.

I thought it would be too, but as it turned out, that was not the problem.

File: engine/src/cmd/unit_json_factory.cpp (hunk 5 above)
Comment (stephengtuggy) [reply engine/src/cmd/unit_json_factory.cpp:34]: Because that's how the Boost JSON serialization function constructs it. So when we go to load the file, we need to support either format -- with or without an array at the top level.

File: engine/src/cmd/unit_csv.cpp (hunk 3 above)
Comment (stephengtuggy) [reply engine/src/cmd/unit_csv.cpp:1109]: OK, I can delete it.

File: engine/src/cmd/unit_json_factory.cpp (hunk 4 above)
Comment (stephengtuggy) [reply engine/src/cmd/unit_json_factory.cpp:81]: Based on my testing, this was the loading code that needed to change.

File: engine/src/cmd/unit_json_factory.cpp (hunk 5 above)
Comment (stephengtuggy) [reply engine/src/cmd/unit_json_factory.cpp:34]: To clarify further: Roy, your code saved an array with a single object element inside of it:

```json
[
    {
        ...
    }
]
```

My code simply removes the root array:

```json
{
    ...
}
```

engine/src/cmd/unit_csv.cpp
stephengtuggy: Done.

engine/src/cmd/unit_csv.cpp
stephengtuggy: Deleted.

engine/src/cmd/csv.cpp
stephengtuggy: Deleted.

engine/src/cmd/unit_json_factory.cpp
royfalk: I'm not arguing that this code didn't need to change. I'm saying it could try and load `unit.json` and fail and tell the player "Saved game unit file..." for an asset file.

engine/src/cmd/unit_json_factory.cpp
royfalk: Unfortunately, we actually want the array. Right now the game saves each ship in its own file. There's no reason to do this. We just need to figure the whole player_ship issue. Either, replace it with player_ship_1...2...3 or player_ship_<random 6 alphanumeric> or something like that and check for prefix.

This would also solve the issue of not supporting a fleet containing two ships of the same type.

engine/src/cmd/unit_json_factory.cpp
stephengtuggy: Oh, so you're saying that the error message could be erroneous? Referring to a "saved game unit file" when that might not be what it is?

engine/src/cmd/unit_json_factory.cpp
stephengtuggy: I see.

engine/src/cmd/unit_json_factory.cpp
royfalk: Yes. It's a unit file but it could be either units.json or <save_game>.json.

stephengtuggy: Corrected the error message.

royfalk (approved):
I'm still not wild about ParseJSON still supporting a root object.
It's extra code and I'm not a fan of that unless there's a roadmap to its use.
However, it's good enough for an approval. 
I'll leave it to you to keep or discard.

stephengtuggy:
> I'm still not wild about ParseJSON still supporting a root object. It's extra code and I'm not a fan of that unless there's a roadmap to its use. However, it's good enough for an approval. I'll leave it to you to keep or discard.

OK, cool. Thanks. I think I'll leave the extra code in there for now, and then if we don't use it within a certain period of time (six months?), we can delete it.

PR #999, opened 2025-01-18
Refactor vegastrike-engine and vegasettings to use boost::program_options -- 0.10.x

### LINKED ISSUE #581: If -d directory cannot be found, list actual path tried

### LINKED ISSUE #584: Make `-d` parameter more flexible in how it accepts input
Users often run into problems because they put a space or an equals sign between `-d` and the data path. We should probably make VS more flexible so that it can handle these variations.

### LINKED ISSUE #713: Better Command-line Argument Processing

### LINKED ISSUE #996: vegasettings: replace C-style strings and string operations with std::string and boost::format
If you're going to touch code like this, please convert it std::string. There's no reason to keep using C in C++ projects.

_Originally posted by @royfalk in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/974#discussion_r1917610902_

stephengtuggy:
Code Changes:
- [ ] (Not yet) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Addressed:
- Partly fixes #996 
- Closes #581 
- Closes #584 
- Closes #713 

Purpose:
- What is this pull request trying to do? Implement the boost::program_options refactor for the last time, hopefully, this time for 0.10.x / the 'master' branch


engine/src/main.cpp
```diff
@@ -716,162 +717,193 @@ void bootstrap_main_loop() {
...
+            std::cout << "Error reading coordinates for player location: " << inputFailure.what()
+                    << std::endl;
+            exit(EXIT_FAILURE);
+        }
+        SetPlayerLoc(PlayerLocation, true);
+    }
+
+    if (cmd_args.count("debug")) {
+        char vs_debug_level_temp = cmd_args["debug"].as<char>();
+        switch (vs_debug_level_temp) {
```
royfalk, line 843: I'm really tired and sleepy right now, but why don't you do
```
if temp >=0 && temp <=3 {
   g_game.vsdebug = temp;
} else {
...
}
```

engine/src/main.cpp
```diff
@@ -716,162 +717,193 @@ void bootstrap_main_loop() {
...
+                g_game.vsdebug = 1;
+                break;
+            case 2:
+                g_game.vsdebug = 2;
+                break;
+            case 3:
+                g_game.vsdebug = 3;
+                break;
+            case '0':
+                g_game.vsdebug = 0;
```
royfalk, line 857: ~Isn't this repeating~
Why are we supporting this? Do you really think someone would run
`vega-engine --debug="2"`
?

engine/src/main.cpp
```diff
@@ -716,162 +717,193 @@ void bootstrap_main_loop() {
...
+                break;
+            case '3':
+                g_game.vsdebug = 3;
+                break;
+            default:
+                VS_LOG_FLUSH_EXIT(fatal, "Invalid debug level specified", EXIT_FAILURE);
+        }
+    }
+
+    if (cmd_args.count("net")) {
```
royfalk, line 873: Delete please

engine/src/main.cpp
```diff
@@ -716,162 +717,193 @@ void bootstrap_main_loop() {
...
+    if (cmd_args.count("benchmark")) {
+        benchmark = cmd_args["benchmark"].as<double>();
+    }
+
+    if (cmd_args.count("test-audio")) {
+        // This is handled separately, since we have to undefine main first
+        // return Audio::Test::main(argc, argv);
+    }
+
+    if (cmd_args.count("a")) {
```
royfalk, line 887: Delete all three resolutions. We shouldn't support anything under 1920x1080.

engine/src/main.cpp
```diff
@@ -716,162 +717,193 @@ void bootstrap_main_loop() {
...
+    if (cmd_args.count("v")) {
+        g_game.x_resolution = 1280;
+        g_game.y_resolution = 1024;
+    }
+
+    if (cmd_args.count("mission_name")) {
+        mission_name = cmd_args["mission_name"].as<std::string>();
+    }
+
     if (false == legacy_data_dir_mode) {
```
royfalk, line 904: One day we should figure out this legacy data dir mode. Right now we support this and I'm not sure why.

royfalk (approved):
Feel free to be more aggressive in your changes (without breaking stuff). The game was written **25 years ago**. A lot of the code and features should go away or be rewritten from scratch.

engine/src/main.cpp
stephengtuggy: The short answer is, I tried it and it didn't work. 😄

engine/src/main.cpp
stephengtuggy: Good point. This is probably not necessary, I'll remove it.

engine/src/main.cpp
stephengtuggy: Agreed. Let's discuss with @BenjamenMeyer at some point.

engine/src/main.cpp
stephengtuggy: I disagree about not supporting anything under 1920x1080. There are times -- especially in a VM -- when running with a lower resolution is necessary.

With that said, 800x600 is ridiculously low. I think I will go ahead and remove that option.

engine/src/main.cpp
stephengtuggy: Removed.

engine/src/main.cpp
stephengtuggy: 800x600 option deleted.

evertvorster:
Just tried running this, and it fails instantly:
```
[evert@Evert vegastrike-engine-git]$ vs
Starting Vega Strike...

Registering codec ogg
.
Legacy Mode: TRUE
Saving current directory (/mnt/int/system/aur/vegastrike-engine-git) as DATA_DIR
 In path /mnt/int/system/aur/vegastrike-engine-git
Vega Strike  
See http://www.gnu.org/copyleft/gpl.html for license details.

Using target data directory specified on command line: '/usr/share/vegastrike'
Invalid debug level specified
[evert@Evert vegastrike-engine-git]$ 
```

stephengtuggy:
> Just tried running this, and it fails instantly:
> 
> ```
> [evert@Evert vegastrike-engine-git]$ vs
> Starting Vega Strike...
> 
> Registering codec ogg
> .
> Legacy Mode: TRUE
> Saving current directory (/mnt/int/system/aur/vegastrike-engine-git) as DATA_DIR
>  In path /mnt/int/system/aur/vegastrike-engine-git
> Vega Strike  
> See http://www.gnu.org/copyleft/gpl.html for license details.
> 
> Using target data directory specified on command line: '/usr/share/vegastrike'
> Invalid debug level specified
> [evert@Evert vegastrike-engine-git]$ 
> ```

@evertvorster Just fixed that 😄

engine/src/main.cpp
stephengtuggy: Deleted.

engine/src/main.cpp
stephengtuggy: As it turns out, we need the `case '0'` and similar cases, not the `case 0` ones. `case '2'`, for example, corresponds to `--debug=2` on the command line.

engine/src/main.cpp
royfalk: It's been 7 hours. I'm still tired and sleepy but I still don't understand this.
```
if temp >='0' && temp <='3' {
   g_game.vsdebug = temp - '0';
} else {
...
}
```
Are you telling me the above doesn't work?

engine/src/main.cpp
royfalk: If I had to guess, this is why the previous conversation had you stumped.

royfalk (approved):

engine/src/main.cpp
stephengtuggy: It's possible that the latter would work. Maybe I'll try it.

stephengtuggy: Sure enough, it works.

engine/src/main.cpp
stephengtuggy: Yes, more or less 😃

engine/src/main.cpp
BenjamenMeyer: @royfalk I don't think I've ever run at 1920x1080 - that would be full screen for me. I don't think I run as small as 800x600 often.

PR #1005, opened 2025-01-21
Misc. Refactoring of Components

royfalk:
As I'm working on moving hull, armor and shield code into lib_component, additional changes have also been made. I've separated these into another PR, so review will be easier.

The changes are many but very localised.
- Move graphicOptions.InWarp to ftl_drive.Enabled(). Also support enable, disable and toggle.
- Remove base_computer::showUnitStats. It's been replaced by python code.
- Remove two unused parameters in DamageRandSys.
- Remove unused parameter in Component::Load and subclasses
- Align Cloak with other components.
- unit_csv::UnitToMap no longer uses the original unit stats as a starting point.
- Rename Component::Percent to Component::PercentOperational. We will have namespace collisions when we implement hull, armor and shield, as there is a Percent function there as well.
- Standardise on Component::Load(unit_key) and Component::Upgrade(upgrade_key)
- Fix bug in AfterburnerUpgrade::Upgrade - added __upgrades suffix twice.
- Disabled Component::Load code related to MasterPartList. This needs a PR on its own to do right.
- Deleted Component::unit_key. This collides with parameters and doesn't actually add worth. Components shouldn't know about the ship.
- Fix issue in JumpDrive and other some other components, where the constructor switched two parameters around.
- Make Resource::Percent return 0 and not -1. The game uses this in cockpit and VDU and checking every cycle is costly.
- Resource now checks if it is destroyed (0) before repairing.

Code Changes:

Issues identified:
- Overdrive assets are missing the "Afterburner_Upgrade" key and are therefore not working when upgraded (https://github.com/vegastrike/Assets-Masters/issues/55).
- BaseComputer does not provide upgrade_view with the current ship status as a parameter. Currently, upgrade_view python script only uses units.json to generate the description. Therefore, any damage to the already installed component cannot be shown (#1006).
- Llama.begin shield02 randomly shows damage (#1007). 
- You can buy any shield you want (2/4) for your ship (#1008).
- The current upgrade management screen is showing the gaps between the old code and the new. Need to map these out and implement properly in python (#1011).
- The game doesn't list integrated components. This was OK before, as the engine treated the unit as a single struct and worked on it. Now, we **need** to list these missing components (#1009).
- Drive upgrades have not been implemented yet in assets.
- The game treats upgrades as cargo. We need to move away from this paradigm (#1010).
- The Component class needs to store the information about the upgrade (price, volume, etc.).


engine/src/components/afterburner.h
```diff
@@ -41,15 +41,14 @@ class Afterburner : public Component, public EnergyConsumer {
     Afterburner(EnergyContainer *source = nullptr);
     
     // Component Methods
-    virtual void Load(std::string upgrade_key, 
-                      std::string unit_key = "");      
+    virtual void Load(std::string unit_key);
```
stephengtuggy, line 44: If these are overloads, they should be marked with `override` at the end, rather than with `virtual`.

engine/src/components/tests/afterburner_tests.cpp
```diff
@@ -57,15 +57,15 @@ TEST(Afterburner, Sanity) {
...
+    afterburner.Load(afterburner_string + upgrades_suffix_string);
     upgrade.Load(upgrade_string + upgrades_suffix_string);
 
     EXPECT_EQ(afterburner.GetMass(), 0.0);
     EXPECT_EQ(afterburner.thrust.MaxValue(), 10.0);
     EXPECT_EQ(afterburner.speed.MaxValue(), 100.0);
     EXPECT_EQ(afterburner.GetConsumption(), 2.0);
 
-    EXPECT_EQ(upgrade.GetMass(), 5.0);
+    //EXPECT_EQ(upgrade.GetMass(), 5.0);
```
stephengtuggy, line 68: Why are these being commented out? I don't understand

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -260,16 +260,16 @@ void AggressiveAI::SetParent(Unit *parent1) {
     last_directive = "b";     //prevent escort race condition
 
     //INIT stored stuff
-    Fshield_prev = parent->graphicOptions.InWarp ? 1 : parent->FShieldData();
+    Fshield_prev = parent->ftl_drive.Enabled() ? 1 : parent->FShieldData();
```
stephengtuggy, line 263: I wonder if `ftl_drive.IsOn()`, `ftl_drive.TurnOn()`, and `ftl_drive.TurnOff()` might make better sense to the reader?

engine/src/components/tests/afterburner_tests.cpp
royfalk: If you look at Component.cpp, you'll see I've commented out the code to initialize name and mass of the upgrade. 
I have a general idea of where I want to go with this. Right now, the game uses cargo to represent upgrades and these (volume, price, name, etc.) are stored as part of the Master Part List (MPL).
Because these are part of Cargo and not Unit Asset, we need to process two separate elements. 
I'm not sure how to do that. I'm also not sure I need to. 
Right now, upgrade_view takes the key and pulls data directly from the Assets and MPL.

engine/src/cmd/ai/aggressive.cpp
royfalk: `IsOn`/... is more engine specific. `Enable`/`Disable`/`Enabled` is generally used in programming. 
The choice between the two isn't dramatic. 
Honestly, I like `Enabled` better because it's easier for me to scan quickly in code. I think `IsOn` is harder for me because the capital O breaks the reading flow. 
But, it's marginal and a simple search and replace. So I'll leave it to you to decide.

BenjamenMeyer: enabled/disabled is sufficient. We want a quick status.

engine/src/cmd/movable.h
```diff
@@ -106,27 +106,25 @@ class Movable {
...
+        unsigned SubUnit;
+        unsigned RecurseIntoSubUnitsOnCollision;
+        unsigned missilelock;
+        unsigned FaceCamera;
+        unsigned Animating;
+        
+        unsigned WarpRamping;
+        unsigned unused1;
+        unsigned NoDamageParticles;
+        unsigned specInterdictionOnline;
```
BenjamenMeyer, line 120: did you mean to leave these uninitialized? They were previously initialized to `1`, no?

royfalk: Nope. These are actually bit fields. https://en.cppreference.com/w/cpp/language/bit_field
As per https://stackoverflow.com/a/4240989/5276890 and common sense, it's actually worse for performance.
Since these variables are used extensively in every game cycle, this actually hurts performance in a small way.

engine/src/components/afterburner.h
stephengtuggy: Sorry, I misspoke. Not overloads -- overrides. These methods, along with many others in this PR, override methods inherited from parent classes. As such, they should be marked `override` rather than `virtual`. (`virtual` should be used only in the first base class that declares each method.)

engine/src/components/tests/afterburner_tests.cpp
stephengtuggy: So in other words, we're breaking functionality temporarily until we figure out how this should be done. I am not OK with that. At least not on `master`. The master branch is supposed to remain playable at all times, right?

engine/src/cmd/movable.h
stephengtuggy: The very same Stack Overflow answer you linked recommends against using bit fields in general. It says that among other things, they are non-portable and bad for performance.

Also: You can't just leave stuff uninitialized. Everything needs to be initialized at least once, unless you are using it as a source of entropy. Some languages don't even allow uninitialized variables or fields.

I would much rather our code be correct, and slightly less performant, than blazingly fast, and wrong.

stephengtuggy: Oh, maybe I misunderstood what you're saying, Roy. Are you saying that these used to be bit fields, and you intentionally changed them to not be?

stephengtuggy: Yeah, the ` : 1` syntax denotes a bit field, doesn't it? It's not an initialization value like I thought at first.

I can certainly get on board with eliminating the bit fields here. Let's just make sure we initialize these guys at some point in the code.

engine/src/components/afterburner_upgrade.h
```diff
@@ -14,7 +14,7 @@
  * (at your option) any later version.
  *
  * Vega Strike is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * but WITHOUT ANY WARRANTY override;without even the implied warranty of
```
stephengtuggy, line 17: Looks like our find and replace got a little too aggressive here 😄

engine/src/components/afterburner.h
stephengtuggy: Oh, now I see. This change has been implemented already, across quite a few files. Looking good. I don't know why I didn't see that before.

engine/src/cmd/movable.h
BenjamenMeyer: yeah.... it's been a while since I've dealt with that.
Typically, once uses a bit field with a union on a struct to make easy access to stuff - like wireline header fields for IP messages (essentially making it easier to decode binary protocols).

The methodology above isn't correct usage for them. Thanks for calling that out and fixing them.

BenjamenMeyer: we still should initialize them to a known state. Though perhaps we should use a bool type instead of an integer if we're swapping from a bit field; it'll keep the usage more or less the same.

engine/src/components/tests/afterburner_tests.cpp
evertvorster: I agree here, the master should be playable at all times, so each of the issues identified should be fixed before this PR can be merged. 

Saying that, this is a step in the right direction, and so I'll be play testing this branch, and as the fixes come in for the various issues identified I will continue play testing until we are sure that this branch does not introduce any new issues into master.

engine/src/cmd/movable.h
royfalk: I'll take another look and see about converting them to bools and about initialization.

engine/src/components/tests/afterburner_tests.cpp
royfalk: > So in other words, we're breaking functionality temporarily until we figure out how this should be done. I am not OK with that. At least not on master. The master branch is supposed to remain playable at all times, right?

Nope. Right now the game stores a list of `Cargo` items for each upgrade. I've made provisions in `Component` when I started development to store this information there. 

Commenting it out does not impact game play as this extra copy of Cargo attributes (mass, volume, name, description, etc.) is currently not used.

engine/src/cmd/movable.h
royfalk: They are initialized in the constructor, as per @BenjamenMeyer 's preferences.

Movable.cpp:73
```
Movable::graphic_options::graphic_options() {
    FaceCamera = Animating = missilelock = unused1 = WarpRamping = NoDamageParticles = 0;
    specInterdictionOnline = 1;
    NumAnimationPoints = 0;
    RampCounter = 0;
    MinWarpMultiplier = MaxWarpMultiplier = 1;

    // Added implementation to make var false
    // I don't like it, because it's true by default and false by default
    // TODO: figure out which it actually is
    RecurseIntoSubUnitsOnCollision = false;
    SubUnit = 0; // Also this
}
```

engine/src/components/tests/afterburner_tests.cpp
stephengtuggy: Oh. Okay.

stephengtuggy (approved):
Looks good to me

PR #1030, opened 2025-02-14
Finish updating copyleft headers - 0.9.x

stephengtuggy:
Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Finish the copyleft header updates for the game engine, 0.9.x branch. Included are 460 .cpp files and 2 .h files.
- What release is this for? 0.9.x
- Is there a project or milestone we should apply this to? 0.9.x

Issue #898, opened 2024-10-25
Loss of flight controls in newly created save file. 


evertvorster:
This bug was introduced by #893 

When running a new campaign and immediately saving the game, the saved game now contains three values separated by a slash in the governors for the flight controls. 

The values are also quite low, so the flight controls are almost non-existent. 

For reference here is a save file created by the 0.8.x version of the game engine:
[Llama.begin.csv](https://github.com/user-attachments/files/17518408/Llama.begin.csv)

And here is a save file created after that pull request was merged:
[Llama.begin.csv](https://github.com/user-attachments/files/17518453/Llama.begin.csv)

BenjamenMeyer:
Guessing this will apply to 0.9.x too

royfalk:
Correction - #893 did not introduce this. It introduce the capability to serialize a resource int the format of:

_<current>  / <max_damaged> / <max>_

However, it did not introduce any code that uses it.

The actual bug is probably in #896 (@evertvorster please confirm). The Yaw_Governor_Left/Yaw_Governor_Right save a value of _0.01/0.01/0.01_. It shouldn't be too hard to figure out why.

evertvorster:
This bug is fixed! Yay!

royfalk:
Sorry. I fixed it but didn't link the PR to it.

evertvorster:
Hmm, this bug was present in master before #986 was merged. I think Benjamin also noticed some loss of Yaw controls a while back, when loading a save file. 
However, it is fixed now.

BenjamenMeyer:
Yes, I've noticed loss of yaw on and off. Usually occurred when launching. Not sure I've seen it in master in a while...but I don't get to play often right now. Will keep an eye out for it.

stephengtuggy:
I am still seeing loss of yaw controls pretty consistently with the 0.9.x code. Is this bug supposed to be fixed there?

stephengtuggy:
> I don't know. What was the cutoff point for 0.9?
> 
> There were actually two bugs related to loss of yaw control:
> 
> 1. The CSV got mangled in saving or loading, and yaw values, which came fairly last, were moved one right and read 0.
> 2. This issue which saved some drive values as current/damaged_max/max, which somehow missed a multiplier and saved really low values. Yaw was not 0 but close to it.
> 
> Take a look at your llama.csv and check the yaw values. That should tell you which bug it is.

[...]

PR #1035, opened 2025-02-16
Fix Jumping - 0.9.x

### LINKED ISSUE #1032: 0.9.x now takes a lot of SPEC/FTL capacitors to provide enough energy to jump
It used to be that a single FTL Capacitor II, or two FTL Capacitor I's, were enough to allow the starting Llama to jump. This is no longer the case. I'm finding that I now need to buy an FTL Capacitor V, and maybe a III or IV in addition, in order to be able to jump through warp points.

These higher-capacity FTL capacitors are more expensive -- out of reach for players starting out with the default 13,500 credits. In order to be able to jump, players will have to either hack their savegames, or trade a bunch, or fly a bunch of missions in-system, before they will be able to afford enough FTL capacitance to jump out of the starting system. This is not a good experience for players of the game. And it's a change from how things used to be.

I think we should fix this before we release 0.9.0.

stephengtuggy:
Code Changes:

Issues:
- Fixes #1032 

Purpose:
- What is this pull request trying to do? Fix jumping, and possibly SPEC mode as well, for 0.9.x

### DIFF
--- a/engine/src/cmd/drawable.cpp
+++ b/engine/src/cmd/drawable.cpp
@@ -62,17 +62,17 @@ inline static float perspectiveFactor(float d) {
     }
 }
 
-Drawable::Drawable() :
-        halos(new HaloSystem()),
-        animatedMesh(true),
-        activeAnimation(0),
-        timeperframe(3.0),
-        done(true),
-        activeMesh(0),
-        nextactiveMesh(1),
-        infiniteLoop(true),
-        loopCount(0),
-        curtime(0.0) {
+Drawable::Drawable() : halos(new HaloSystem()),
+                       animatedMesh(true),
+                       activeAnimation(0),
+                       timeperframe(3.0),
+                       done(true),
+                       activeMesh(0),
+                       nextactiveMesh(1),
+                       infiniteLoop(true),
+                       loopCount(0),
+                       num_chunks(0),
+                       curtime(0.0) {
 }
 
 Drawable::~Drawable() {
--- a/engine/src/cmd/role_bitmask.cpp
+++ b/engine/src/cmd/role_bitmask.cpp
@@ -71,9 +71,9 @@ static vsUMap<string, int> rolemap;
 static vsUMap<int, string> irolemap;
 
 unsigned char InternalGetRole(const std::string &s) {
-    vsUMap<string, int>::const_iterator i = rolemap.find(strtoupper(s));
+    const vsUMap<string, int>::const_iterator i = rolemap.find(strtoupper(s));
     if (i != rolemap.end()) {
-        return (*i).second;
+        return i->second;
     }
     return 0;
 }
@@ -81,11 +81,11 @@ unsigned char InternalGetRole(const std::string &s) {
 const std::string &InternalGetStrRole(unsigned char c) {
     static const std::string empty;
 
-    vsUMap<int, string>::const_iterator i = irolemap.find(c);
+    const vsUMap<int, string>::const_iterator i = irolemap.find(c);
     if (i != irolemap.end()) {
-        return (*i).second;
+        return i->second;
     }
-    return rolemap.size() ? rolemap.begin()->first : empty;
+    return !rolemap.empty() ? rolemap.begin()->first : empty;
 }
 
 std::vector<std::vector<std::string> > buildscripts() {
@@ -290,4 +290,3 @@ unsigned int getCapitalRoles() {
     return retval;
 }
 }
-
--- a/engine/src/gfx/mesh.cpp
+++ b/engine/src/gfx/mesh.cpp
@@ -125,7 +125,7 @@ bool Mesh::LoadExistant(Mesh *oldmesh) {
     try {
         *this = *oldmesh;
     } catch (std::exception& ex) {
-        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCSIG__ % ex.what()));
+        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCTION__ % ex.what()));
         return false;
     }
     oldmesh->refcount++;
@@ -147,7 +147,7 @@ bool Mesh::LoadExistant(const string filehash, const Vector &scale, int faction)
             return LoadExistant(oldmesh);
         }
     } catch (std::exception& ex) {
-        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCSIG__ % ex.what()));
+        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCTION__ % ex.what()));
         return false;
     }
     return false;
--- a/engine/src/gfx/mesh.cpp
+++ b/engine/src/gfx/mesh.cpp
@@ -122,7 +122,12 @@ Mesh::Mesh() {
 }
 
 bool Mesh::LoadExistant(Mesh *oldmesh) {
-    *this = *oldmesh;
+    try {
+        *this = *oldmesh;
+    } catch (std::exception& ex) {
+        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCSIG__ % ex.what()));
+        return false;
+    }
     oldmesh->refcount++;
     orig = oldmesh;
     return true;
@@ -131,14 +136,19 @@ bool Mesh::LoadExistant(Mesh *oldmesh) {
 bool Mesh::LoadExistant(const string filehash, const Vector &scale, int faction) {
     Mesh *oldmesh;
 
-    hash_name = VSFileSystem::GetHashName(filehash, scale, faction);
-    oldmesh = meshHashTable.Get(hash_name);
-    if (oldmesh == nullptr) {
-        hash_name = VSFileSystem::GetSharedMeshHashName(filehash, scale, faction);
+    try {
+        hash_name = VSFileSystem::GetHashName(filehash, scale, faction);
         oldmesh = meshHashTable.Get(hash_name);
-    }
-    if (nullptr != oldmesh) {
-        return LoadExistant(oldmesh);
+        if (oldmesh == nullptr) {
+            hash_name = VSFileSystem::GetSharedMeshHashName(filehash, scale, faction);
+            oldmesh = meshHashTable.Get(hash_name);
+        }
+        if (nullptr != oldmesh) {
+            return LoadExistant(oldmesh);
+        }
+    } catch (std::exception& ex) {
+        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCSIG__ % ex.what()));
+        return false;
     }
     return false;
 }
--- a/engine/src/gfx/mesh.cpp
+++ b/engine/src/gfx/mesh.cpp
@@ -86,17 +86,17 @@ void Mesh::InitUnit() {
     numlods = 1;
     alphatest = 0;
     lodsize = FLT_MAX;
-    forcelogos = NULL;
-    squadlogos = NULL;
+    forcelogos = nullptr;
+    squadlogos = nullptr;
     local_pos = Vector(0, 0, 0);
     blendSrc = ONE;
     blendDst = ZERO;
-    vlist = NULL;
+    vlist = nullptr;
     mn = Vector(0, 0, 0);
     mx = Vector(0, 0, 0);
     radialSize = 0;
     if (Decal.empty()) {
-        Decal.push_back(NULL);
+        Decal.push_back(nullptr);
     }
 
     //texturename[0] = -1;
@@ -104,13 +104,13 @@ void Mesh::InitUnit() {
     myMatNum = 0;     //default material!
     //scale = Vector(1.0,1.0,1.0);
     refcount = 1;     //FIXME VEGASTRIKE  THIS _WAS_ zero...NOW ONE
-    orig = NULL;
+    orig = nullptr;
 
     envMapAndLit = 0x3;
     setEnvMap(GFXTRUE);
     setLighting(GFXTRUE);
-    detailTexture = NULL;
-    draw_queue = NULL;
+    detailTexture = nullptr;
+    draw_queue = nullptr;
     will_be_drawn = GFXFALSE;
     draw_sequence = 0;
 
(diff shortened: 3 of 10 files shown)


royfalk (approved):
I just reviewed the SPEC commit. LGTM.

BenjamenMeyer:
@stephengtuggy this is on top of #1030?

kheckwrecker:
Merged this PR and PR#1030 on top of the 0.9x engine source. Game compiled fine. Launched new game, bought a jump drive and an overdrive, selected a cargo mission and jumped to Everett. Jump went fine but used a lot of energy. Unfortunately, game crashed soon after jumping. Restarted game (in debugger), loaded saved game, accepted another cargo run to Everett and launched into space. Now I have JUMP: LOW ENERGY on the screen, and when I fly to a jump point, I'm unable to jump.

stephengtuggy:
> @stephengtuggy this is on top of #1030?

@BenjamenMeyer yes, it is.

stephengtuggy:
> Merged this PR and PR#1030 on top of the 0.9x engine source. Game compiled fine. Launched new game, bought a jump drive and an overdrive, selected a cargo mission and jumped to Everett. Jump went fine but used a lot of energy. Unfortunately, game crashed soon after jumping. Restarted game (in debugger), loaded saved game, accepted another cargo run to Everett and launched into space. Now I have JUMP: LOW ENERGY on the screen, and when I fly to a jump point, I'm unable to jump.

@kheckwrecker I am seeing pretty much the same thing. When loading a saved game, I no longer have enough FTL capacitance to jump. More work is needed.

BenjamenMeyer:
> > @stephengtuggy this is on top of #1030?
> 
> @BenjamenMeyer yes, it is.

Okay...thanks. I'll wait to review this until #1030 is merged and this reflects that.

kheckwrecker:
> Please upload your saved game ship json.

Here you go...
[Llama.begin.csv](https://github.com/user-attachments/files/18854369/Llama.begin.csv)

Except for some reason, it's CSV and not JSON?

royfalk:
Can you yaw? Turn right and left?
If you go into ship view in base computer, does it say you have 40,000 warp/spec capacitor?

@stephengtuggy this is almost certainly #865. It's a rather critical bug and fixed by #898. It should be pulled into 0.9.

I think we missed it because most people don't play test the Llama and it only affects this ship.

kheckwrecker:
> Can you yaw? Turn right and left?

Yes

> If you go into ship view in base computer, does it say you have 40,000 warp/spec capacitor?

No. Here's what it shows:
Capture

royfalk:
@stephengtuggy 
There's something wrong here. 166k MJ is the old number. 
833MJ is just weird. Llama begin has 325 for SPEC capacitor and the inconsistent spec_capacitor02 (400MJ).
My recommendation, find out where jump cost is defined and divide by 1000.

kheckwrecker:
> This is stupid. I keep trying to test and troubleshoot the jump functionality of this code, but the game crashes so often that it's almost impossible to test. My efforts basically amount to trying to put lipstick on a pig. I've had enough.
> 
> Let's get this 0.9.0 code out, FWIW, then follow up as quickly as possible with some code that actually works.

I feel your pain. My experience has been very similar when trying to test this PR. Happens so often I usually run this version in a debugger to see if anything useful shows up. 

I haven't had a chance to pull your latest changes to this PR. I'll try to do that tomorrow and see if I can make a couple of jumps.

royfalk:
@stephengtuggy 
Can you `coredumpctl -1 gdb` and tell me what's crashing the game?
We shouldn't release 0.9 if it's so unstable.

royfalk:
> I tried to test this latest PR. 

Please add your setup when writing these. I assume you started a new campaign and bought an afterburner. Given the difficulty you've had with 0.9, maybe move to something else. I'll try and take a look this week and see if I can't figure out what's going on.

@BenjamenMeyer , @stephengtuggy , can you tell me to cut-off point for 0.9?

kheckwrecker:
> > I tried to test this latest PR.
> 
> Please add your setup when writing these. I assume you started a new campaign and bought an afterburner. Given the difficulty you've had with 0.9, maybe move to something else. I'll try and take a look this week and see if I can't figure out what's going on.
> 
> @BenjamenMeyer , @stephengtuggy , can you tell me to cut-off point for 0.9?

Running in Windows, from a fresh build yesterday with the latest version of PR 1035. Yes, started a new campaign, and bought a new Overdrive II. Sorry to bother you with my posts. I can move on to some other project.

stephengtuggy:
@kheckwrecker you’re not bothering us at all. This is extremely valuable information to get from a user’s perspective.

The autopilot turning away at a certain distance from the target is another known issue. When it happens, I just turn off the autopilot and fly the rest of the way to the target manually. It’s not ideal, but it works.

stephengtuggy:
@royfalk I know where the crash is occurring. It’s in Mesh::loadExistant(Mesh &)

stephengtuggy:
I just haven’t figured out how to fix it yet

BenjamenMeyer:
> > I tried to test this latest PR.
> 
> Please add your setup when writing these. I assume you started a new campaign and bought an afterburner. Given the difficulty you've had with 0.9, maybe move to something else. I'll try and take a look this week and see if I can't figure out what's going on.
> 
> @BenjamenMeyer , @stephengtuggy , can you tell me to cut-off point for 0.9?

https://github.com/vegastrike/Vega-Strike-Engine-Source/network provides good insight there.

This seems to be the commit where we split of the 0.9.x branch at:
https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/adc62840b7a8065968794b1946131532aff5024a
Looks like part of #796.

Everything else seems to be:
- fixes for GitHub Action
- Installer work
- Licensing
- Pulling up some stuff that seemed to get applied only to 0.8.x so we lost it for 0.9.x and master (program options)
- Fixing the Python version stuff (eval.h)

IIRC there was some pretty significant changes about to happen to master (0.10.x) after #796 which is why we cut it there.

royfalk:
So, I think I have some serious doubts about this branch (fix/FTL_0.9.x).

There seems to be an issue with the Mesh code crashing the game. I'm not clear on whether this is due to this specific branch or something in 0.9 in general. However, the fix appears to be: 
a. to diverge significantly from master 
b. hide any issues via try/catch.

This code worked in 0.8 and works in master. If I had to guess, the issue is in the assets. The 0.9 assets are not aligned with the engine. Using try/catch will mean the game will run without meshes from some units.

If something is really broken in 0.9, the fix should be to pull the fix from master. I'll take a look and see if I can figure out what happened here.

royfalk:
OK. I play tested 0.9, pre fix/FTL and the copyright branch.

I'm not sure what's going on here, but this is before all my work on lib_component. I misled you when I said we've seen this issue, as we were having similar issues after I refactored the ftl_drive/jump_drive code.

The game runs great with no crashes or any funny stuff. The issue boils down to the fact that **Llama.begin is broken**. Specifically, **reactor 3** should generate 44 and not reactor 2 (30) and spec 1 should hold 200 and not 325. I didn't check but I think when we fixed something in the code, we started using the correct values.

I think the easiest fix would be to fix Llama.begin values to work properly - change reactor02 to reactor03 and change spec from 1 to 2.

stephengtuggy:
There is indeed an issue with the Mesh code crashing the game. There have been several such issues, right along. (Including in 0.8 and possibly prior.) We've narrowed it down to a single Mesh-related point of failure as of the code prior to this PR. This PR will (hopefully) fix that final Mesh-related crash.

I hear what you're saying about hiding the issues. However, I don't think that's really what we're doing here. Both `loadExistant` methods return a boolean -- `true` for success, `false` for failure. So I've actually fixed them to behave the way that their API contract implies. If an exception occurs, log it and return false. I think that's what these methods should have been doing in the first place.

It will be up to the calling code to check the return values of the `loadExistant` methods and respond accordingly. I can check on that as part of this PR. That's probably not a bad idea.

I think the fixes you're suggesting related to the starting Llama make a lot of sense. Unfortunately, those fixes won't help when loading a saved game. 0.9 does not correctly load saved games -- it is off by at least one column in the values it reads from the CSV. As a result, when loading a saved game, you end up with negative available upgrade space... and zero warp/FTL capacitance. That incorrect saved game logic will need to be fixed as part of this. Otherwise, the only way to jump will be to start a fresh campaign every time. ☹️ 

@royfalk what do you think the code should do if a Mesh can't be loaded? I'm open to suggestions.

stephengtuggy:
It's possible that the game is **unloading** Meshes when it shouldn't. I.e., when the use count of a particular ship Mesh or similar drops to zero, that Mesh seems to get removed from the Mesh hashtables. I don't think that is necessarily correct. The "original" of every Mesh should stay in memory until the game shuts down, seems to me.

Maybe fixing that would eliminate the errors loading Meshes altogether.

royfalk:
> There is indeed an issue with the Mesh code crashing the game. 

I don't recall seeing this recently. It's possible I didn't because I don't play the game beyond play testing. However, I also didn't get any feedback from our testers and they do play. I was under the impression something you did in 0.9 caused the game to be **more** unstable than usual.


> I hear what you're saying about hiding the issues. 

I was under the impression this was a lot more frequent in 0.9. I was concerned the game would actually fail to load a whole bunch of meshes. This does not seem to be the case and the changes to the code seem fine. 


> I think the fixes you're suggesting related to the starting Llama make a lot of sense. Unfortunately, those fixes won't help when loading a saved game. 0.9 does not correctly load saved games -- it is off by at least one column in the values it reads from the CSV. 

Bug #865. Fix: #897

> @royfalk what do you think the code should do if a Mesh can't be loaded? I'm open to suggestions.

I'm fine with your solution.

> It's possible that the game is unloading Meshes when it shouldn't. I.e., when the use count of a particular ship Mesh or similar drops to zero, that Mesh seems to get removed from the Mesh hashtables. I don't think that is necessarily correct. The "original" of every Mesh should stay in memory until the game shuts down, seems to me.

I agree.

stephengtuggy:
Cool, thanks. I have back-ported your #897 fix into this PR. Let's see if it works. (I expect it will.)

stephengtuggy:
OK, I implemented the changes we discussed, along with one or two more. I have never seen the game this stable! I've had it running for about 2.5 hours with barely even a stutter.

engine/src/gfx/mesh_gfx.cpp
```diff
@@ -455,11 +455,10 @@ Mesh::~Mesh() {
     } else {
         orig->refcount--;
         VS_LOG(debug, (boost::format("orig refcount: %1%") % refcount));
-        // TODO: Comment out?
```
vincele: Why should this code be commented ?
Is the refcounting not properly working, useless ?
Why not removing the code instead of commenting ?

engine/src/cmd/drawable.cpp
```diff
@@ -62,17 +62,17 @@ inline static float perspectiveFactor(float d) {
...
-        animatedMesh(true),
-        activeAnimation(0),
-        timeperframe(3.0),
-        done(true),
-        activeMesh(0),
-        nextactiveMesh(1),
-        infiniteLoop(true),
-        loopCount(0),
-        curtime(0.0) {
+Drawable::Drawable() : halos(new HaloSystem()),
```
vincele, line 65: This would have been easier (trivial) to review without the whitespace change.

engine/src/gfx/mesh.cpp
```diff
@@ -125,7 +125,7 @@ bool Mesh::LoadExistant(Mesh *oldmesh) {
     try {
         *this = *oldmesh;
     } catch (std::exception& ex) {
-        VS_LOG(error, (boost::format("Error in %1%: '%2%'") % __FUNCSIG__ % ex.what()));
```
vincele: It looks like you don't like rebases, but I'd squash/fixup this commit with the relevant commit, which is also in this same PR... But maybe that's just me, feel free to ignore.

engine/src/cmd/role_bitmask.cpp
```diff
@@ -71,21 +71,21 @@ static vsUMap<string, int> rolemap;
 static vsUMap<int, string> irolemap;
 
 unsigned char InternalGetRole(const std::string &s) {
-    vsUMap<string, int>::const_iterator i = rolemap.find(strtoupper(s));
```
vincele, line 74: This (all modifications to this file) should be in a separate commit, IMHO

vincele (commented):
I did not review the mesh.cpp/.h or energetic.cpp commits, I think they lack explanations in the commit message and mix cleanups with actual fixes or something, I cannot see what's the fix from the other churn. I won't request changes, it's your call, but I won't approve either. Sorry.

engine/src/gfx/mesh.cpp
stephengtuggy: You are correct. I don't like rebases. 😄

engine/src/cmd/drawable.cpp
stephengtuggy: It also would have been formatted incorrectly. 😉

engine/src/gfx/mesh_gfx.cpp
stephengtuggy: I can remove it instead of commenting it out.

The idea here is that the `original` copy of each Mesh should never be unloaded from memory until the game shuts down. If any Mesh originals are removed, then they might have to be reloaded from disk later. And then what is the point of pre-loading everything at startup?

vincele: That would make a nice explanation for the commit message. Can you please `git commit --amend` to add it ?

vincele:
> @vincele unfortunately, I can't entirely separate out the fixes from the other changes, either. So I think we should merge this PR as is, once testing and review is complete. That way we can be sure that all the fixes will be applied.

It's certainly doable, but I agree this can be a lot of time-consuming work, especially afterwards, so I won't ask you to do it for this PR, but give it a thought for the next one...

engine/src/gfx/mesh_gfx.cpp
vincele: but that would require a `rebase -i`, so I'll let you decide if you want to do it...

stephengtuggy: Deleted the commented-out code, and provided a more detailed explanation in the corresponding commit message.

vincele: Nice way out of a rebase, kudos !
Hey, commenting it was in fact actually useful... ;-)

vincele (approved):

engine/src/cmd/drawable.cpp
BenjamenMeyer: > This would have been easier (trivial) to review without the whitespace change.

you can disable whitespace changes in the PR by changing the PR settings - or adding IIRC `&wh=1` to the URL.

engine/src/gfx/mesh.cpp
BenjamenMeyer: yes - if you want to rebase do so before sharing. Once shared rebase is not an option.

engine/src/cmd/role_bitmask.cpp
BenjamenMeyer: we take a more pragmatic approach.

PR #1041, opened 2025-02-27
Refactor: move collide2 out as a library

BenjamenMeyer:
Add a new `libraries` directory in parallel to the `engine` for our various library components, and add the `collide2` code there as the first library component as it is the most independent of the various libraries.

NOTE: There is no real code change outside of header includes

Please answer the following:

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation -- I have done some play testing; there's not really any code changes outside of some header adaptions so this shouldn't play any different from master.

Issues:
- #56 

Purpose:
- What is this pull request trying to do? Breaking down the code structure into smaller, independent components


libraries/collide2/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+# along with Vega Strike. If not, see <https://www.gnu.org/licenses/>.
+
+MESSAGE("PROCESSING DIRECTORY: ${CMAKE_CURRENT_SOURCE_DIR}")
+
+LIST(APPEND
+    OPCOLLIDE_INCLUDES
+    # collide2 headers
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    # VS engine headers
+    #${CMAKE_SOURCE_DIR}/engine
```
royfalk, line 27: Please open an issue for refactoring VS out of collide library. It's no good separating it out if it refers to VS all over.

CMakeLists.txt
```diff
@@ -149,6 +149,39 @@ ELSE (ENABLE_PIE)
...
+INCLUDE(CheckIncludeFileCXX)
+INCLUDE(CheckTypeSize)
+INCLUDE(CMakePushCheckState)
+SET(SIZEOF_VOID_P CMAKE_SIZEOF_VOID_P)
+
+SET(CMAKE_FIND_FRAMEWORK "LAST")
+
+#Find Math
+INCLUDE(CheckSymbolExists)
+IF(NOT POW_FUNCTION_EXISTS AND NOT NEED_LINKING_AGAINST_LIBM)
```
royfalk, line 162: I was unable to find a lot of information about this. Is there really a chance modern systems don't have POW? (whatever that is)
I'm asking because a lot of stuff in our cmake was really legacy from 2001.

royfalk (commented):
Looks pretty straightforward. Not sure how meaningful it is without encapsulating it from the rest of the code though.

libraries/collide2/CMakeLists.txt
BenjamenMeyer: 1. collide2 is generally under a difference license (public domain from what I can tell... or rather the UNLICENSE) so at a minimum we gain that clarity.
2. agreed - there's a few things (gfx/quaternion.h) that seem to have gotten spread out. it'd be nice to make it more independent that way.

Filed #1042

CMakeLists.txt
BenjamenMeyer: I doubt it. Created #1043 to follow up on that.

libraries/collide2/CMakeLists.txt
BenjamenMeyer: Looking at the build failures (https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1041#issuecomment-2687035496) I'll have to address some of this somehow... but my goal is to be minimal right now. Not sure if I'll fix up #1041 immediately or just move some checks around in CMake for the moment so Mac/Win builds succeed - more likely just move stuff around b/c it'll probably make more sense long term with other libraries being split out any way.

royfalk (approved):
OK. Make it build and run and ship it.

CMakeLists.txt
stephengtuggy: We need `pow`. This block of code tells us where to find it (i.e., what library). I don't think this code should be removed.

engine/CMakeLists.txt
```diff
@@ -52,6 +53,9 @@ LIST(APPEND
     ${CMAKE_CURRENT_SOURCE_DIR}/src/resource
     ${CMAKE_CURRENT_SOURCE_DIR}/src/python/base_computer
     ${CMAKE_CURRENT_SOURCE_DIR}/src/python/config
+    # Library Headers
+    ${CMAKE_SOURCE_DIR}/libraries/collide2
```
vincele, line 57: Why not just `${CMAKE_SOURCE_DIR}/libraries` ? That would remove the need to modify a lot of the `#include`s

BenjamenMeyer: Consistency between the libraries. Honestly I'd like to get the headers paths down to a lot more limit set long term so this is the first step along a longer path trajectory.

The library itself should use its own directory as the import path so its self-contained. Similarly the code linking against it needs to use the same path so that they can pull in the headers for a library.

In my own code I split out headers as their own thing:
```
project
project\include
project\src
```
and then external sources linking link against the `project\include`. We haven't agreed to that here right now and given our state I'm not ready to push that question; we have a lot of other things that are more important; but that would also change this to:
```
${CMAKE_SOURCE_DIR}/libraries/collide2
```
That said, I'd also like to find a CMake way to do this automatically by searching for the directories under `${CMAKE_SOURCE_DIR}/libraries` and then adding them; however I haven't been able to find a directive that would allow that yet - it finds source files easily, but not directories or at least not filtered to a single level.

BenjamenMeyer:
Moved a chunk of the dependency searching to the root; this should fix Windows. Not sure about the endian one, but need a check and see how the CI system does with this change before debugging further.

Play tested a little to see how it was doing, and also brought up-to-date with master.

BenjamenMeyer:
@stephengtuggy I need some help figuring out the Windows side.

Mac is still misbehaving but I think I can figure that out for the moment.

BenjamenMeyer:
NOTE: for some reason MacOS wasn't falling into the BSD/FreeBSD logic for the Endianness header... added a `__APPLE__` case too.

libraries/collide2/CMakeLists.txt
```diff
@@ -0,0 +1,96 @@
...
+    ${CMAKE_SOURCE_DIR}/engine/src/python/base_computer
+    ${CMAKE_SOURCE_DIR}/engine/src/python/config
+    # CMake Artifacts
+    ${Vega_Strike_BINARY_DIR}
+    ${Vega_Strike_BINARY_DIR}/engine
+    ${Vega_Strike_BINARY_DIR}/src
+)
+MESSAGE("OpCollide Includes: ${OPCOLLIDE_INCLUDES}")
+
+ADD_LIBRARY(vegastrike-OPcollide STATIC
```
stephengtuggy, line 42: Looks like the `STATIC` clause is new here. That might be part of the problem. We link many, if not all, of the libraries dynamically on Windows. Possibly on macOS as well.

engine/src/vegastrike.h
```diff
@@ -75,14 +75,14 @@ extern float AUDIO_ATOM;
 #if !defined (NO_GFX)
 #   if defined(__APPLE__) && defined(__MACH__)
 #       define GL_GLEXT_FUNCTION_POINTERS 1
-#       include <gl.h>
-#       include <glu.h>
-#       include <glut.h>
+#       include <GL/gl.h>
+#       include <GL/glu.h>
+#       include <GL/glut.h>
```
stephengtuggy, line 80: Um. I know it's non-standard to include these headers without the directory prefix, but we're doing it this way on purpose. It's been discussed previously; I just don't remember where off the top of my head.

This might be part of the build problems, as well.

stephengtuggy, line 80: Um. I know it's non-standard to include these headers without the directory prefix, but we're doing it this way on purpose. It's been discussed previously; I just don't remember where off the top of my head.

This might be part of the build problems, as well.

libraries/collide2/opcodetypes.h
```diff
@@ -24,6 +24,7 @@
 // config.h is the VS ./configure determined header.
 #include <cfloat>
 #include <cwchar>
+#include <cstdint>
```
stephengtuggy, line 27: Why was this added?

stephengtuggy:
@BenjamenMeyer I've taken a crack at this. My draft PR is here: https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1069

libraries/collide2/CSopcodecollider.h
```diff
@@ -32,12 +32,12 @@
...
 #ifndef VEGA_STRIKE_ENGINE_CMD_COLLSION2_OPCODECOL_H
 #define VEGA_STRIKE_ENGINE_CMD_COLLSION2_OPCODECOL_H
-#include "cmd/collide2/Opcode.h"
+#include "Opcode.h"
 #include "csgeom2/opmatrix3.h"
 #include "csgeom2/opvector3.h"
 #include "csgeom2/opbox.h"
 #include "csgeom2/optransfrm.h"
-// #include "opcodegarray.h"
+#include "opcodegarray.h"
```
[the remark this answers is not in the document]
BenjamenMeyer: I might have commented it out as part of moving it around... might have found it provided some interface that the file needed to build.
May be I need to take a closer look and see if I can pull in a simpler set.

libraries/collide2/opcodetypes.h
BenjamenMeyer: the file needed the C-Types (`uint_8`, etc)

PR #1045, opened 2025-02-28
Fix README build instructions

vincele:
This is just fixing the README build instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,21 +184,27 @@ This will produce a significantly smaller download - in the order of 22-30 MB; w
    b. *OR* configure and compile VS manually, using the ncurses ccmake frontend:
 
    ```bash
-   mkdir build & cd build
+   mkdir build
+   cd build
    ccmake ..
    # (configure/edit options to taste in ccmake, press 'c' to save the selected options
    # and press 'g' to update the build configuration files used by the make build tool)
+   cd ..
    cmake --build ./build -j $(getconf _NPROCESSORS_ONLN) # (where the getconf clause returns the number of available CPU threads/cores on the system)
-   mkdir ../bin && cp vegastrike-engine ../bin/ && cp setup/vegasettings ../bin/ && cd ..
+   mkdir bin
+   cp build/vegastrike-engine build/setup/vegasettings bin
    ```
 
    c. *OR* configure and compile VS manually, using the command-line cmake frontend:
 
    ```bash
-   mkdir build & cd build
+   mkdir build
+   cd build
    cmake ..
+   cd ..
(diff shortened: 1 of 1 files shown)


stephengtuggy:
I think we should get rid of the manual building instructions entirely. People who want to build the code should be using the build scripts and/or the CMake Presets at this point, IMHO.

evertvorster (approved):
Looks good to go!

evertvorster:
> I think we should get rid of the manual building instructions entirely. People who want to build the code should be using the build scripts and/or the CMake Presets at this point, IMHO.

Why would you want to get rid of them? What is the value of not having manual build instructions? You'll just get someone in the future asking how do I build this from source...

vincele:
Yeah, I went having a look at the build script and decided to continue doing it manually. I think keeping it documented is still good to have.

README.md
```diff
@@ -184,21 +184,27 @@ This will produce a significantly smaller download - in the order of 22-30 MB; w
    b. *OR* configure and compile VS manually, using the ncurses ccmake frontend:
 
    ```bash
-   mkdir build & cd build
+   mkdir build
+   cd build
```
stephengtuggy, line 188: Why this change? What was wrong with having both commands on the same line, the way it was before?

README.md
```diff
@@ -184,21 +184,27 @@ This will produce a significantly smaller download - in the order of 22-30 MB; w
...
+   mkdir bin
+   cp build/vegastrike-engine build/setup/vegasettings bin
    ```
 
    c. *OR* configure and compile VS manually, using the command-line cmake frontend:
 
    ```bash
-   mkdir build & cd build
+   mkdir build
+   cd build
```
stephengtuggy, line 202: Same as above. Why is this change needed?

stephengtuggy (commented):
OK, maybe the manual build instructions do hold some value. I have more detailed questions about a couple of the changes though.

README.md
vincele: The problem was the wrong `&` instead of the intended `&&`
I don't understand the need to chain the commands with `&&` for that line but not the following ones.
I went for the simple solution of separating the commands

stephengtuggy: OK, I guess that makes sense.

royfalk:
@vincele go ahead and merge.

README.md
BenjamenMeyer: yeah it's fine

@vincele said:
> I don't understand the need to chain the commands with && for that line but not the following ones.

It basically moves it into the directory for that line and then the following lines just work in that directory.
It's fine to separate it out or use `pushd` to change into the directory instead.

Issue #1043, opened 2025-02-27
Do we really need to check if `pow` exists in CMake?


BenjamenMeyer:
While working on #1041 @royfalk [asked](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1041#discussion_r1972894971):
> I was unable to find a lot of information about this. Is there really a chance modern systems don't have POW? (whatever that is)
I'm asking because a lot of stuff in our cmake was really legacy from 2001.

Back in 2000 it was probably more common that `pow` and other math functions might not exist. But is that really true today? Could we get rid of these checks?

stephengtuggy:
We need `pow`. This block of code tells us where to find it (i.e., what library to link with). I don't think this block should be removed.

stephengtuggy:
I think the reason for the `IF` might be to avoid duplicating `pow`-detection logic if we've already done that work in a previous cmake run.

royfalk:
Honestly, I don't get this. According to cppreference.com, [pow](https://en.cppreference.com/w/cpp/numeric/math/pow) has been around since forever really. It's part of the standard library. Are you telling me there's a compiler in 2025 that doesn't support it?!
cppreference doesn't even mention when it was introduced. It just says pre-2011.

vincele:
I think I've read somewhere that there's libms that are automatically added by compilers, but not all compilers do/did that, but that was a long time ago... That may have been a reason for this kind of manual handling.

But  really don't know. I created the PR to see if CI caught something wrong, but it didn't.

So how can I dig deeper ?

Do we care about what is not covered by the CI ?
If there's something that break, and someone care, we'll get a report and fix it then...

Anyways, I can drop the PR if this is deemed important enough.

stephengtuggy:
@vincele Good questions. I don't know the answers, unfortunately.

Perhaps one thing we could check for is fallback code that uses a different method of calculating exponents if `pow()` is not found. If there isn't anything of that sort, then maybe we can just use `std::pow()` across the board; get rid of this CMake logic; and call it good. After making sure it passes CI and play-testing, of course.

BenjamenMeyer:
yes we probably should rely on `std::pow()` now and make it clear that that is what we are doing.

vincele:
OK, I'll try the full std::pow migration, and we'll see what the CI will tell us...

royfalk:
[...]

PR #1051, opened 2025-03-02
Cmake - Remove pow & some libm & math.h handling

vincele:
Fixes issue: #1043

Code Changes:

Issues:
- PR'ed to let CI run on different systems, was OK on void-linux, x86_64, musl libc

Purpose:
- What is this pull request trying to do? Removing old cruft from cmake config
- What release is this for? Next one
- Is there a project or milestone we should apply this to? Next one


stephengtuggy:
I'm not so sure this is a good idea. I suspect there was a good reason why this block was there, and the reason probably still applies.

vincele:
Still it built on all the CI platforms (BTW, there's no BeOS in there).

I think `pow()` is such a basic math function that missing it would be fairly obvious, I don't think this would induce subtle bugs. But with FP math, we never know what lurks in there...

royfalk:
Let's keep the discussion in #1043.

royfalk:
@BenjamenMeyer can you add your two cents here?
I'd prefer to reach some consensus here if possible.

In short, @vincele and I think pow checks can be removed.
The C++ documentation thinks this has been a part of the standard since a long time ago. Possibly C++03. Certainly Pre C++11.
@stephengtuggy advises caution.

BenjamenMeyer:
I advise caution for the time being while we confirm platforms.

Let me get #1041 finished & merged, and we can test and do some research on this in the meantime.

BenjamenMeyer (changes requested):
Blocking this per https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1051#issuecomment-2708502844 until I at least get that work done.

BenjamenMeyer:
@vincele now that a version of #1051 ( #1069) was merged, please update this PR and let's see where we are.

stephengtuggy:
I am leaning more and more towards this change probably being OK. This PR just needs to be updated to resolve the conflicts, and then we can take another look, @vincele .

vincele:
It was easier to do it all from scratch, so I force pushed over the branch... Let's see what the CI has to say about it now...

stephengtuggy:
> It was easier to do it all from scratch, so I force pushed over the branch... Let's see what the CI has to say about it now...

OK. So far, so good. Now the only other thing I want to check is whether `pow()` is called as `std::pow()` or not.

stephengtuggy:
> > It was easier to do it all from scratch, so I force pushed over the branch... Let's see what the CI has to say about it now...
> 
> OK. So far, so good. Now the only other thing I want to check is whether `pow()` is called as `std::pow()` or not.

And `pow` is defined everywhere, but it looks like it's pulling it from different places -- sometimes from `std::`, sometimes from the root namespace (`::pow()`), and sometimes from an inline function. If we can get all those places changed to use `std::pow()`, I think we'll be good. I count 31 occurrences.

BenjamenMeyer:
Yes, let's get this down to a single implementation so we can know for sure where's its coming from and whether or not the CMake is needed. Much prefer `std::pow` or a Boost equivalent.

vincele:
OK I'll add that on top

BenjamenMeyer (approved):
LGTM; approved pending play test confirmation

stephengtuggy:
Unfortunately, the Mac and Windows builds are now failing with `no member named 'pow' in namespace 'std'` errors. Should be an easy fix: just `#include <cmath>` instead of math.h in alphacurve.h.

stephengtuggy:
I just finished running through most of the play-test steps. Everything is looking good, about the same as before.

vincele:
> Should be an easy fix: just `#include <cmath>` instead of math.h in alphacurve.h.

DONE

stephengtuggy (approved):

stephengtuggy:
Thanks, @vincele ! Merged!

PR #1052, opened 2025-03-04
Auto-generate configuration based on config.json

royfalk:
This uses my https://github.com/royfalk/JsonStruct project to auto-generate configuration.h/cpp based on config.json.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Not yet. Just checked the game works

Issues:
- When launching, map opens automatically
- Crosshair is different


engine/src/components/component_utils.cpp
```diff
@@ -125,15 +125,19 @@ EnergyContainer* GetSource(ComponentType component_type, EnergyContainer* fuel,
...
+    if(source == "Infinite") {
+        return nullptr; 
+    } else if(source == "Fuel") {
+        return fuel;
+    } else if(source == "Energy") {
+        return energy;
+    } else if(source == "FTLEnergy") {
+        return ftl_energy;
+    } else {
+        return nullptr;
```
stephengtuggy, line 140: Why are we making this change? The `switch` block with enum values seems like a better way to do it than an `if` block comparing strings.

engine/src/configuration/configuration.cpp
```diff
@@ -1,8 +1,8 @@
 /*
- * configuration.cpp
+ * configuration.h
  *
- * Copyright (C) 2021-2022 Daniel Horn, Roy Falk, ministerofinformation,
- * David Wales, Stephen G. Tuggy, and other Vega Strike contributors
+ * Copyright (C) 2021-2023 Daniel Horn, Roy Falk, ministerofinformation,
```
stephengtuggy, line 4: Shouldn't the end year be 2025?

engine/src/configuration/configuration.cpp
```diff
@@ -22,467 +22,411 @@
...
-    fuel.variable_fuel_consumption = GetGameConfig().GetBool("physics.VariableFuelConsumption", fuel.variable_fuel_consumption);
-    // Also known as Lithium6constant
-    fuel.deuterium_relative_efficiency_lithium = GetGameConfig().GetFloat("physics.DeuteriumRelativeEfficiency_Lithium", fuel.deuterium_relative_efficiency_lithium);
-    //Fuel Mass in metric tons expended per generation of 100MJ
-    fuel.fmec_factor = GetGameConfig().GetFloat("physics.FMEC_factor", fuel.fmec_factor);
-    fuel.reactor_idle_efficiency = GetGameConfig().GetFloat("physics.reactor_idle_efficiency", fuel.reactor_idle_efficiency);
-    fuel.min_reactor_efficiency = GetGameConfig().GetFloat("physics.min_reactor_efficiency", fuel.min_reactor_efficiency);
-    fuel.ecm_energy_cost = GetGameConfig().GetFloat("physics.ecm_energy_cost", fuel.ecm_energy_cost);
+std::shared_ptr<Configuration> configuration() {
+    static std::shared_ptr<Configuration> kConfiguration = nullptr;
```
stephengtuggy, line 419: I think I was the one who introduced this form of singleton for holding the Configuration object. Unfortunately, I don't think my code was quite correct. At least, I don't think it's a true Meyers singleton if it uses a pointer instead of a reference for the static variable.

Hopefully this won't be too difficult to fix in your Json Struct generation code, @royfalk . The Vega Strike Logger class should demonstrate how to implement a Meyers singleton properly. (At least I think I have it right finally.)

engine/src/configuration/configuration.cpp
```diff
@@ -22,467 +22,411 @@
...
-vega_config::GameStart::GameStart(boost::json::object object) {
-    default_mission = JsonGetStringWithDefault(object, "game_start|default_mission", "");
-    introduction = JsonGetStringWithDefault(object, "game_start|introduction", "");
-}
-
-std::shared_ptr<Configuration> configuration() {
-    static const std::shared_ptr<Configuration> kConfiguration = std::make_shared<Configuration>();
     return kConfiguration;
-}
+}
```
stephengtuggy, line 432: Missing LF at end of file

engine/src/configuration/configuration.h
```diff
@@ -1,8 +1,8 @@
 /*
- * configuration.h
+ * configuration.cpp
```
stephengtuggy, line 2: Looks like the .cpp and .h files got their headers switched around.

engine/src/configuration/configuration.h
```diff
@@ -1,8 +1,8 @@
...
  *
- * Copyright (C) 2021-2023 Daniel Horn, Roy Falk, ministerofinformation,
- * David Wales, Stephen G. Tuggy, Benjamen R. Meyer, and other Vega Strike contributors
+ * Copyright (C) 2021-2022 Daniel Horn, Roy Falk, ministerofinformation,
```
stephengtuggy, line 4: I believe the end year here should also be 2025.

engine/src/configuration/configuration.h
```diff
@@ -22,476 +22,422 @@
...
+
+	struct {
+
+	} advanced;
 };
 
 extern std::shared_ptr<Configuration> configuration();
 
-#endif //VEGA_STRIKE_ENGINE_CONFIG_CONFIGURATION_H
+#endif //VEGA_STRIKE_ENGINE_CONFIG_CONFIGURATION_H
```
stephengtuggy, line 443: Missing LF at end of file

stephengtuggy:
Looks pretty good for the most part. Thanks, @royfalk

engine/src/components/component_utils.cpp
stephengtuggy: I see. I guess that makes sense. It's not ideal, but it will work.

engine/src/gfx/cockpit_gfx_utils.cpp
```diff
@@ -191,12 +191,12 @@
...
-    const float lock_line = configuration()->graphics_config.hud.lock_confirm_line_length;
-    const float diamond_size = configuration()->graphics_config.hud.diamond_size;
-    const float theta_speed = configuration()->graphics_config.hud.diamond_rotation_speed;
+    const float lock_line = configuration()->hud.lock_confirm_line_length;
+    const float diamond_size = configuration()->hud.diamond_size;
+    const float theta_speed = configuration()->hud.diamond_rotation_speed;
 
-    const float max = diamond_size * r_size * 0.75f * configuration()->graphics_config.hud.min_missile_bracket_size;
-    const float coord = configuration()->graphics_config.hud.min_missile_bracket_size + (configuration()->graphics_config.hud.max_missile_bracket_size - configuration()->graphics_config.hud.min_missile_bracket_size) * lock_percent;
+    const float max = diamond_size * r_size * 0.75f * configuration()->hud.min_missile_bracket_size;
```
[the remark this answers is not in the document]
BenjamenMeyer: please ensure this is fixed

engine/src/configuration/configuration.cpp
royfalk: > I think I was the one who introduced this 

Yes. 

> Hopefully this won't be too difficult to fix in your Json Struct generation code, @royfalk . The Vega Strike Logger class should demonstrate how to implement a Meyers singleton properly. (At least I think I have it right finally.)

I'll take a look.

engine/src/configuration/configuration.cpp
```diff
@@ -1,8 +1,8 @@
 /*
- * configuration.cpp
+ * configuration.h
```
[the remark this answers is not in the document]
stephengtuggy: @vincele Much of our copyleft header template is required under the GPL. The rest of it we have agreed upon after extensive discussion. I'd rather not make too many changes to what we include in the header of each file, unless there is a very good reason to.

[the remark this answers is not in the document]
vincele: I was only speaking about the file name in the header, but if that's already agreed upong, feel free to ignore my comment.

Rest assured that I won't try to suggest SPDX headers here.

royfalk:
Closing in favor of #1078

PR #1055, opened 2025-03-09
Remove unused function `showSplashProgress()`

vincele:
Usage from the python code in Assets-Production has been removed in:
https://github.com/vegastrike/Assets-Production/pull/164

Code Changes:
- [X] Have the PR Validation Tests been run? compiles & rune on linux/musl

Issues:
- Depends on the Assets PR being merged first

Purpose:
- What is this pull request trying to do? Remove unused code
- What release is this for? Next
- Is there a project or milestone we should apply this to? Next

### DIFF
--- a/engine/src/python/star_system_exports.h
+++ b/engine/src/python/star_system_exports.h
@@ -134,7 +134,6 @@ voidEXPORT_UTIL(receivedCustom)
 
 voidEXPORT_UTIL(showSplashScreen)
 voidEXPORT_UTIL(showSplashMessage)
-voidEXPORT_UTIL(showSplashProgress)
 voidEXPORT_UTIL(hideSplashScreen)
 EXPORT_UTIL(isSplashScreenShowing, true)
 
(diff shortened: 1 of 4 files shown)


BenjamenMeyer (changes requested):
We probably should figure out how to actually implement this as it seems like something that would give the user good feedback on what is going on.

vincele:
@BenjamenMeyer I think this has not been finished because what we already have seems good enough. That's certainly what I thought, when seeing that unused code. We have the changing text in lower-left of the screen telling us it's loading files... So I won't be the one finishing that progress bar.

BenjamenMeyer:
That's fine. we probably should file an issue to track it. Most use a bar not simply seeing the text like that so it'd probably be good so folks have an idea of progress as a whole. You see the text but you don't know how much there is to load so you just wait.

vincele:
OK, consensus looks established, closing.

PR #1056, opened 2025-03-09
Assorted cleanups for lin_time.cpp

vincele:
First easy batch of cleanups, there should be no issues with this fairly non-controversial lot.

Code Changes:
- [X] Have the PR Validation Tests been run? compiles & run on linux/musl

Issues:
- None ?

Purpose:
- What is this pull request trying to do? Cleanups & dead code removal
- What release is this for? Next
- Is there a project or milestone we should apply this to? Next

### DIFF
--- a/engine/src/lin_time.cpp
+++ b/engine/src/lin_time.cpp
@@ -130,12 +130,6 @@ bool toggle_pause() {
 
 #ifdef _WIN32
 
-#ifndef NOMINMAX
-#define NOMINMAX
-#endif //tells VCC not to generate min/max macros
-
-#include <windows.h>
-
 void micro_sleep( unsigned int n )
 {
     Sleep( n/1000 );
--- a/engine/src/lin_time.cpp
+++ b/engine/src/lin_time.cpp
@@ -97,7 +97,6 @@ void setTimeCompression(float tc) {
 
 bool toggle_pause() {
     static bool paused = false;
-    VS_LOG(debug, "toggle_pause() called in lin_time.cpp");
     if (paused) {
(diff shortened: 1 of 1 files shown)


royfalk (approved):
Looks pretty straightforward to me.

engine/src/lin_time.cpp
```diff
@@ -88,20 +90,6 @@ void reset_time_compression(const KBData &, KBSTATE a) {
...
-        if (paused == false) {
-            timecompression = .0000001;
-            timecount = 0;
-            paused = true;
-        } else {
-            paused = false;
-            reset_time_compression(s, a);
-        }
-    }
-}
```
BenjamenMeyer, line 103: Q: is this disabling the games ability to Pause? Or is that being handled somewhere else now?

BenjamenMeyer (changes requested):
1. Some of the changes only effect Win32 platform so let's make sure that is play tested.
2. let's make sure this doesn't break the game's ability to pause

engine/src/lin_time.cpp
vincele: This function is unused (AFAIK, I git grep'ed it in engine & assets), I think what is used is the `toggle_pause()` function, a few lines below, in `engine/src/cmd/ai/firekeyboard.cpp`.

BenjamenMeyer: @vincele thanks - I'll try to pull it down and test to validate. That's my main concern here.

vincele:
Closing this in favor of https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1076

PR #1061, opened 2025-03-14
vcpkg builtin-baseline, joystick, etc. 2025-03

### LINKED ISSUE #1062: Joystick not working
On the `master` branch, my joystick is not working at all. Even on Windows, where it works fine with 0.9.x. One of the changes to `master` must have broken joystick functionality. We need to get that fixed.

stephengtuggy:
Code Changes:

Issues Fixed:
- Fixes #1062 (joystick functionality)

Known Issues Remaining:
- The game crashed a couple more times. Nothing necessarily new.
- Several more times, certain letters went missing from the HUD, like `n`, `p`, and `z`. This is not a new issue either. It is annoying to see it again though, especially several times in a row.

Purpose:
- What is this pull request trying to do? Sync the master branch with many of the latest packaging, scripting, CMake, and vcpkg changes on the 0.9.x branch. Also update the vcpkg `builtin-baseline` value, in order to pull the latest version of each of our dependencies available from vcpkg as of this date. And finally, fix the long-standing joystick issues.

### DIFF
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -371,13 +371,6 @@ int main(int argc, char *argv[]) {
 #endif
 
     std::vector<std::vector<char> > temp = ROLES::getAllRolePriorities();
-#if defined(HAVE_SDL)
-#ifndef NO_SDL_JOYSTICK
-    if (SDL_InitSubSystem(SDL_INIT_JOYSTICK)) {
-        VS_LOG_FLUSH_EXIT(fatal, (boost::format("Couldn't initialize SDL: %1%") % SDL_GetError()), 1);
-    }
-#endif
-#endif
 
     InitTime();
     UpdateTime();
@@ -683,7 +676,7 @@ void bootstrap_main_loop() {
         {
             std::vector<std::string> intro_lines;
             boost::split(intro_lines, configuration()->game_start.introduction, boost::is_any_of("\n"));
-            
+
             for(const std::string& line : intro_lines) {
                 UniverseUtil::IOmessage(0, "game", "all", line);
             }
--- a/.github/workflows/Windows-CI.yml
+++ b/.github/workflows/Windows-CI.yml
@@ -64,9 +64,9 @@ jobs:
         shell: bash
 
       - name: install-cmake
-        uses: lukka/get-cmake@acb35cf920333f4dc3fc4f424f1b30d5e7d561b4 #v3.31.4
+        uses: lukka/get-cmake@56d043d188c3612951d8755da8f4b709ec951ad6 #v3.31.6
         with:
-          cmakeVersion: 3.31.4
+          cmakeVersion: 3.31.6
           ninjaVersion: 1.11.1
 
       # Restore vcpkg from the GitHub Action cache service. Note that packages are restored by vcpkg's binary caching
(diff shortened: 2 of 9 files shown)


engine/src/main.cpp
```diff
@@ -371,13 +371,6 @@ int main(int argc, char *argv[]) {
 #endif
 
     std::vector<std::vector<char> > temp = ROLES::getAllRolePriorities();
-#if defined(HAVE_SDL)
-#ifndef NO_SDL_JOYSTICK
-    if (SDL_InitSubSystem(SDL_INIT_JOYSTICK)) {
-        VS_LOG_FLUSH_EXIT(fatal, (boost::format("Couldn't initialize SDL: %1%") % SDL_GetError()), 1);
-    }
-#endif
-#endif
```
BenjamenMeyer, line 380: we always have SDL support so I can see removing outer layer; do Joysticks still work after this change? (I don't have one to test with.)

vincele: When reading this and having a look around, I asked myself if this `HAVE_SDL` ifdeffing is complete and/or necessary. You answered that here, so thanks. Would cleaning (removing) all those be wanted ? Worthwhile ?

vincele (commented):
Cannot approve, since I don't understand a lot of the changes (I suppose it's windows build stuff & scripts).

I think the joystick double-init removal looks sensible. You could have explained in the commit message why you choose to remove this one and keep the other though.

engine/src/main.cpp
stephengtuggy: @BenjamenMeyer My joystick only works **after** this change. 😄

stephengtuggy: @vincele Some of the ifdeffing might still be necessary. But only the inner ifdef, in this case, I believe. (The one that disables joystick functionality specifically.) After pushing up these changes last night, I realized that I might need to put back the inner ifdef after all.

stephengtuggy: I don't actually know the answer to the question about `HAVE_SDL` ifdeffing. That might still be needed as well. In theory, we can build the project without SDL; it's not marked as a required dependency in CMakeLists.txt. However, in practice, I don't know that the game does anything useful without SDL. So maybe we should make SDL required and take out the ifdefs.

royfalk: The game engine started out as something very flexible and open ended.
That's why you could run the game with SDL or some other alternative.
Today, this is a bad idea and it was a bad design decision back then.
We should limit options to reduce complexity. SDL is a good choice- we should remove the ifdefs and use it exclusively.

stephengtuggy: Makes sense to me. @BenjamenMeyer ?

BenjamenMeyer: I agree with @royfalk; we don't need a super amount of flexibility and could really stand the benefits of reducing the complexity so let's move SDL to a required dependency starting with 0.10.x (current master) and we can start cleaning up headers there.

@stephengtuggy thanks for filing #1064 was just about to myself.

BenjamenMeyer (approved):
approved pending a verified playtest

stephengtuggy:
> approved pending a verified playtest

Playtest complete ✅

PR #1076, opened 2025-04-03
Assorted cleanups for lin_time.cpp - take 2

vincele:
This one should only keep the *really* uncontroversial part of:
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1056
so that we can make some progress merging it piecemeal...

### DIFF
--- a/engine/src/lin_time.cpp
+++ b/engine/src/lin_time.cpp
@@ -97,7 +97,6 @@ void setTimeCompression(float tc) {
 
 bool toggle_pause() {
     static bool paused = false;
-    VS_LOG(debug, "toggle_pause() called in lin_time.cpp");
     if (paused) {
         VS_LOG(debug, "toggle_pause() in lin_time.cpp: Resuming (unpausing)");
         setTimeCompression(1);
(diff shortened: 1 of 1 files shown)


stephengtuggy (approved):

engine/src/lin_time.cpp
```diff
@@ -113,7 +97,6 @@ void setTimeCompression(float tc) {
 
 bool toggle_pause() {
     static bool paused = false;
-    VS_LOG(debug, "toggle_pause() called in lin_time.cpp");
```
BenjamenMeyer, line 116: Probably good to keep the log message in unless it's being too verbose.

vincele: I thought it is too verbose, because there are VS_LOG(...) calls in both branches of the following `if() then else`

The only thing missing from those is the "called" word, everything else is there...

WDYT ?

BenjamenMeyer: mostly want to make to make sure we know this is happening and can track it back to this code. If that is covered, then it can be removed.

BenjamenMeyer: yeah, L118 and L122.

BenjamenMeyer (approved):

PR #1088, opened 2025-04-10
Add all remaining settings from config.json. Shorter version

### LINKED ISSUE #921: config.json is not preserved through game updates.
We are now setting some configuration options in json.config, and the trend is that more options will be set in this file in the future. 

When installing the game, config.json ends up in a system folder: /usr/share/vegastrike
When the game is updated, this file is removed, and a new version is installed, erasing any changes the user makes. 

The old configuration file, vegastrike.config is also installed in the same location, and serves as a starting point for vsettings. 
With vsettings, any modifications are stored in ~/.vegastrike in Linux. 

When the engine binary is started, it has a hierarchy that it looks for the configuration files, and the system directory is at the bottom of that hierarchy. This means that if there is a configuration file in ~/.vegastrike, that will be used over the system installed one. 

To preserve changes made to json.config a similar strategy should be implemented for json.config. 
As a start, the game engine should look for json.config in the same hierarchy as vegastrike.config, and if it finds it there, that is the configuration that should be used for the game, rather than the system one. 

We can at a later stage implement vsettings copying over the config.json file. For now, I am happy to do this step manually.

stephengtuggy:
Code Changes:
- [X] (Mostly) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Fixed:
- Fixes #921 

Known Issues Remaining:
- ~When you launch into space for the first time, the Navigation Computer view comes up, obscuring your view of space.~ Fixed
- One can dock at any station from pretty much anywhere in the system.
- The targeting crosshairs look different.
- Guns don't work. At all.
- Sometimes when buying and selling ship upgrades, multiple copies of each of the Integral components will appear. I think I saw this previously, as well.

Purpose:
- What is this pull request trying to do? Import all the settings from the newly-expanded config.json, without making too many changes to the rest of the code

### DIFF
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -547,9 +547,10 @@ void bootstrap_first_loop() {
 }
 
 void SetStartupView(Cockpit *cp) {
-    cp->SetView(game_options()->startup_cockpit_view
-            == "view_target" ? CP_TARGET : (game_options()->startup_cockpit_view
-            == "back" ? CP_BACK : (game_options()->startup_cockpit_view
+    VS_LOG(debug, (boost::format("%1%: Setting cockpit startup view to: %2%") % __FUNCTION__ % configuration()->graphics.startup_cockpit_view));
+    cp->SetView(configuration()->graphics.startup_cockpit_view
+            == "view_target" ? CP_TARGET : (configuration()->graphics.startup_cockpit_view
+            == "back" ? CP_BACK : (configuration()->graphics.startup_cockpit_view
             == "chase" ? CP_CHASE
             : CP_FRONT)));
 }
--- a/engine/src/configuration/configuration.h
+++ b/engine/src/configuration/configuration.h
@@ -554,7 +554,7 @@ namespace vega_config {
         double roll = 0.0;
         int screen = 0;
         double simulation_atom = 0.03;
-        int times_to_show_help_screen = 3;
+        int times_to_show_help_screen = 0;
         bool trade_interface_tracks_prices = true;
         int trade_interface_tracks_prices_top_rank = 10;
         int verbose_output = 0;
--- a/engine/src/gfx/nav/navscreen.cpp
+++ b/engine/src/gfx/nav/navscreen.cpp
@@ -141,10 +141,8 @@ void NavigationSystem::Setup() {
     center_z = 0.0;     //updated after a pass
 
     path_view = PATH_ON;
-    static bool start_sys_ortho =
-            XMLSupport::parse_bool(vs_config->getVariable("graphics", "system_map_ortho_view", "false"));
-    static bool start_sec_ortho =
-            XMLSupport::parse_bool(vs_config->getVariable("graphics", "sector_map_ortho_view", "false"));
+    const bool start_sys_ortho = configuration()->graphics.system_map_ortho_view;
+    const bool start_sec_ortho = configuration()->graphics.sector_map_ortho_view;
     system_view = start_sys_ortho ? VIEW_ORTHO : VIEW_2D;
     galaxy_view = start_sec_ortho ? VIEW_ORTHO : VIEW_2D;
     system_multi_dimensional = 1;


engine/src/cmd/music.cpp
```diff
@@ -514,7 +514,7 @@ void Music::GotoSong(int whichlist, int whichsong, bool skip, int layer) {
         if (whichsong != NOLIST && whichlist != NOLIST && whichlist < (int) playlist.size() && whichsong
                 < (int) playlist[whichlist].size()) {
             if (muzak[(layer >= 0) ? layer : 0].lastlist != whichlist) {
-                if (configuration()->audio_config.shuffle_songs.clear_history_on_list_change) {
+                if (configuration()->audio.shuffle_songs_section.clear_history_on_list_change) {
```
royfalk, line 517: If you're going to remove the config suffix, why not the section suffix as well.
Brevity is the soul of wit.

engine/src/components/component_utils.cpp
```diff
@@ -135,6 +137,23 @@ EnergyContainer* GetSourceFromConfiguration(const EnergyConsumerSource source, E
     }
 }
 
+EnergyContainer* GetSourceFromConfiguration(const std::string& energy_source, EnergyContainer* fuel,
+    EnergyContainer* energy, EnergyContainer* ftl_energy)
+{
+    if (energy_source == "Infinite" || energy_source == "infinite") {
```
royfalk, line 143: Boost includes a handy algorithm for case insensitive. 

```
#include <boost/algorithm/string.hpp>
// Or, for fewer header dependencies:
//#include <boost/algorithm/string/predicate.hpp>

std::string str1 = "hello, world!";
std::string str2 = "HELLO, WORLD!";

if (boost::iequals(str1, str2))
{
    // Strings are identical
}

```
Source: https://stackoverflow.com/a/315463/5276890

engine/src/main.cpp
```diff
@@ -305,27 +303,23 @@ int main(int argc, char *argv[]) {
...
-    setup_game_data();
-    //loads the configuration file .vegastrike/vegastrike.config from home dir if such exists
+    setup_game_data();  // TODO: Combine with vega_config::config settings object
     {
         std::pair<std::string, std::string> pair = ParseCommandLine(argc, argv);
         std::string subdir = pair.first;
         mission_name = pair.second;
 
         VS_LOG(info, (boost::format("GOT SUBDIR ARG = %1%") % subdir));
-        if (CONFIGFILE == 0) {
```
royfalk, line 317: If you're going to touch this C code, why not convert it to C++?

`const std::string config_file_name = "vegastrike.config";`

should work just fine. No?

engine/src/configuration/configuration.cpp
```diff
@@ -1,7 +1,7 @@
 /*
  * configuration.cpp
  *
- * Copyright (C) 2021-2022 Daniel Horn, Roy Falk, ministerofinformation,
```
royfalk, line 4: This really should be an auto-generated file. That's 6759 lines of code. I salute you.

royfalk (approved):
Ship it.

engine/src/cmd/music.cpp
stephengtuggy: There was another setting simply called `audio.shuffle_songs`. To distinguish between that single setting and this setting section, I added the `_section` at the end.

engine/src/components/component_utils.cpp
stephengtuggy: That's right, I remember seeing that before. Thanks for the suggestion. Maybe I can apply it in the next pass.

engine/src/main.cpp
stephengtuggy: Yeah, probably. I'll leave that for another day though :-)

engine/src/configuration/configuration.cpp
stephengtuggy: It **was** auto-generated, if that wasn't clear. I used your JsonStruct tool, with a few tweaks, and some changes to template.h and template.cpp. I've opened a couple of PRs with your changes.

Your tool turned out to be very handy. I certainly would not have wanted to write all that code manually.

stephengtuggy:
Found out why the nav screen has started opening when player first launches into space! It's the general.times_to_show_help_screen setting. Should be 0 by default to not show the nav screen on startup. That is now fixed.

Assuming the CI passes, I think I will go ahead and merge this.

PR #1090, opened 2025-04-11
Fix "Nav 8" issue (crashing into invisible planets) -- 0.10.x

### LINKED ISSUE #516: PWCU: Unexpectedly, Nav 8 in Troy is an object that can be collided with

stephengtuggy:
Code Changes:
- [X] (Largely) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Fixed:
- Fixes #516 

Known Issues Remaining:
- Can still dock with Nav Points. Need to fix this in a follow-up PR.

Purpose:
- What is this pull request trying to do? Fix the Nav 8 case for the 0.10.x and following releases. Specifically, make it so that nav points cannot be collided with.


engine/CMakeLists.txt
```diff
@@ -1264,5 +1264,5 @@ IF (USE_GTEST)
     )
 
     INCLUDE(GoogleTest)
-    gtest_discover_tests(${TEST_NAME})
+    gtest_discover_tests(${TEST_NAME} PROPERTIES DISCOVERY_TIMEOUT 30)
```
stephengtuggy, line 1267: Builds have started failing more and more often on my local machine with a timeout error discovering tests. Turns out that the default timeout is only 5 seconds. This increases it to 30. No more timeouts, at least not that I've seen.

libraries/cmd/collision.cpp
```diff
@@ -106,6 +106,12 @@ void Collision::shouldApplyForceAndDealDamage(Unit *other_unit) {
 
             // Units (ships) should calculate actual damage
         case Vega_UnitType::unit:
+            // Handle the "Nav 8" case
+            if ((other_unit->invisible & Unit::INVISUNIT)
```
vincele, line 110: Shouldn't these conditions test be abstracted away in something like `other_unit->is_invisible()` ?

If I read this properly, does it mean I should rename all my units with "invisible" somewhere in their names ?

stephengtuggy: Perhaps abstracting these conditions away would be a good idea. There's only two conditions though, IIRC, so I wasn't sure if it was worth it to create a whole new method. Especially since this would be the only place where it gets used.

vincele: That's not the number of conditions that made me think it should be abstracted away (from my sight) ;-)

It's more related to my second question. It's the search for the "invisible" string in the unit name that I think is questionable.

So my question is in fact: why is that other_unit not having the invisible status and is using name tricks to cater for it ?

stephengtuggy: So the search for the `"invisible"` text may not actually be required. That's the way I handled this issue previously, before I knew about the `invisible` property. For all I know, the `invisible` property by itself may be enough.

This needs to be tested. With PWCU. Unfortunately, PWCU's game assets are woefully out of date with respect to the latest Vega Strike Game Engine versions. So that will take some doing.

royfalk: > This needs to be tested. With PWCU. Unfortunately, PWCU's game assets are woefully out of date with respect to the latest Vega Strike Game Engine versions. So that will take some doing.

I think it would be easier to port the nav point to VS and add it to Cephid.

royfalk: I don't like any of this. It was bad code in the graphics. It's even worse here.
Better to check for something like `mass == 0` IMO.
But that might trigger some obscure bug somewhere.

royfalk (approved):
Conditional approval:
1. Test in VS with a nav point.
2. Figure out whether you really need to search for invisib in the name. That could easily bite someone later.

libraries/cmd/collision.cpp
stephengtuggy: > > This needs to be tested. With PWCU. Unfortunately, PWCU's game assets are woefully out of date with respect to the latest Vega Strike Game Engine versions. So that will take some doing.
> 
> I think it would be easier to port the nav point to VS and add it to Cephid.

That makes sense. I'll try that.

vincele:
@royfalk why should mass have to do with the (in-)visibility tests ? Photons don't have mass, but I can still see (some) of those... And dark matter is invisible but has mass...

royfalk:
> @royfalk why should mass have to do with the (in-)visibility tests ? Photons don't have mass, but I can still see (some) of those... And dark matter is invisible but has mass...

We are talking about nav points here. These are not actual objects but artificial places on the map so you could navigate there and the quest side of things could fulfill conditions. These are non-collidable objects. We could add extra notation to the unit like `"object type": "nav_point"` but `mass=0` seemed simpler.

AFAIK, no other game object should have `mass=0`. 

As for unintended consequences, I'm reasonably fine right now with not drawing or handling collisions of objects with no mass.

vincele:
OK, if that's about being the simpler fix, I can understand.

libraries/cmd/collision.cpp
stephengtuggy: OK. So I took a look at the Nav_8 object and similar objects in PWCU, and imported Nav_8 into Cephid 17, as @royfalk suggested. I noticed that all the Nav_* objects had gravity == 0 and day == 240. The only other objects that had those same property values were jump points. It should not be possible to collide with any of the above, so I think those are the criteria I can use to determine whether an object can be collided with or not. No more string comparisons, which should be a lot better for performance.

stephengtuggy: OK, guys, I think I have a solution. I implemented it as listed in the last comment above. It's efficient; it's encapsulated; and it seems to work reliably. Let me know what you guys think.

stephengtuggy: To test it, I added the following to sectors/Crucible/Cephid_17.system, under "planet" Cephid 17 A:

```xml
	<planet name="Nav_8" file="invisible.png" alpha="ONE ONE" radius="256" gravity="0" x="4000" y="4000" day="240"/>
```

This was copied verbatim from the PWCU repo.

PR #1099, opened 2025-04-14
Misc. fixes to Drawable and other classes -- 0.9.x

stephengtuggy:
Code Changes:
- [X] (Partly) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues Fixed:
- N/A?

Known Issues Remaining:
- ...

Purpose:
- What is this pull request trying to do? Apply some CLion-suggested fixes to the 0.9.x code


engine/src/system_factory.cpp
```diff
@@ -331,7 +331,7 @@ Planet *SystemFactory::processPlanet(Star_XML *xml, Object &object, Planet *owne
         stream >> index >> local;
 
         Light light = lights[index];
-        GFXLight gfx_light(true, Vector(0, 0, 0));
+        GFXLight gfx_light(true, GFXColor(Vector(0, 0, 0), 1));
```
royfalk, line 334: This is an abomination.
Recycling Vector because it's a three value struct.
We should remove this from GFXColor and make the code here:

`GFXLight gfx_light(true, GFXColor(0, 0, 0, 1));`

royfalk (changes requested):
I was almost ready to approve and then I saw the monster and woke up. 
Sorry.

engine/src/system_factory.cpp
stephengtuggy: You're right. I should have caught that.

stephengtuggy: OK. It took some doing, but I think I have now fixed this. (I had to add overloads of several methods in different classes so that they could take either a Vector& or a GFXColor&.)

stephengtuggy: And that in itself is not ideal, but there doesn't seem to be a way around it at the moment.

engine/src/cmd/planet.cpp
```diff
@@ -709,7 +709,7 @@ void Planet::Draw(const Transformation &quat, const Matrix &m) {
     }
     GFXLoadIdentity(MODEL);
     for (unsigned int i = 0; i < lights.size(); i++) {
-        GFXSetLight(lights[i], POSITION, GFXColor(cumulative_transformation.position.Cast()));
```
royfalk, line 712: I don't think there is anything to be done here. I haven't delved into `cumulative_transformation.position.Cast()` but they're clearly using an actual vector here, not as a convenience. I would simply roll with it and do something like:

```
Vector v = cumulative_transformation.position.Cast();
GFXColor c = GFXColor(v.i,v.j,v.k);
...
```

engine/src/gldrv/gl_light.cpp
```diff
@@ -178,6 +178,31 @@ void /*GFXDRVAPI*/ GFXLight::SetProperties(enum LIGHT_TARGET lighttarg, const GF
     apply_attenuate(attenuated());
 }
 
+void /*GFXDRVAPI*/ GFXLight::SetProperties(const enum LIGHT_TARGET light_target, const Vector& vector) {
```
royfalk, line 181: Where did this code come from?
I think you took my comments in the wrong direction. The old devs were using `Vector` as a convenience struct for storing `GFXColor` (without alpha).
I wanted to get rid of the convenience as it was wrong. That's it.

I noted one place where it was clear they were doing more. There we should locally extract the tuple (i,j,k) and create GFXColor. But, looking deeper, I'm not even sure I understand what's going on at all.
`Called wrong overload for this property, with Vector& instead of GFXColor&`
but the parameter is `Vector`?!

Maybe you should write a brief annotation of where you're going here because I got lost.

royfalk (commented):
Sorry. I thought this was a simple ask but it looks like it was either misunderstood or the code wasn't as simple as I thought.

engine/src/cmd/planet.cpp
stephengtuggy: They are setting the POSITION property of the light... with a GFXColor... constructed from a Vector. This is doubly wrong LOL. We should just add another overload of the GFXSetLight method to set properties of lights -- an overload that takes Vector instead of GFXColor. Which is what I have done.

stephengtuggy: Another overload is needed for GFXLight::SetProperties as well, which GFXSetLight calls under the hood.

stephengtuggy: If it's a color-related property, set it using a GFXColor; if it's coordinates-related or similar, set it using a Vector. At least, that's how I think it should generally be done.

engine/src/gldrv/gl_light.cpp
stephengtuggy: See note above.

stephengtuggy: I mean, it's possible.

royfalk:
I hate to ask you to roll back stuff, but in this case I'm going to insist.
I don't think GFXColor should have methods for accepting vector. 
Instead, the right way would be to contain the wrong code to a single location.
Like I said, the code should be:

`
Vector v = cumulative_transformation.position.Cast();
GFXSetLight(lights[i], POSITION, GFXColor(v.i,v.j,v.k));
`

and be done with it. Think of the next gen of programmers who come upon GFXColor with support for Vector and don't know why. At least this way they'll know.

stephengtuggy:
> I hate to ask you to roll back stuff, but in this case I'm going to insist. I don't think GFXColor should have methods for accepting vector. Instead, the right way would be to contain the wrong code to a single location. Like I said, the code should be:
> 
> `Vector v = cumulative_transformation.position.Cast(); GFXSetLight(lights[i], POSITION, GFXColor(v.i,v.j,v.k));`
> 
> and be done with it. Think of the next gen of programmers who come upon GFXColor with support for Vector and don't know why. At least this way they'll know.

What on earth are you talking about?? GFXColor **doesn't** accept Vector any more. That's exactly what I went to all this trouble to avoid. Probably the majority of this PR is now about removing the GFXColor constructor that took a Vector as an argument. Where are you seeing something different?

stephengtuggy:
@royfalk are you confusing GFXLight with GFXColor? I suppose that would be an easy mistake to make.

GFXLight is a light source in the game's 3D space. It can have several properties, including color, location, direction, brightness/intensity, and more. GFXLight is not the same as GFXColor. GFXColor is just a color -- R, G, B, and A components.

engine/src/cmd/planet.cpp
royfalk: Yeah. I assumed it was a color they were setting there. Silly me.
I agree.

engine/src/gldrv/gl_light_state.cpp
```diff
@@ -464,6 +464,61 @@ void gfx_light::ResetProperties(const enum LIGHT_TARGET light_targ, const GFXCol
     }
 }
 
+void gfx_light::ResetProperties(const enum LIGHT_TARGET light_target, const Vector& vector)
```
royfalk, line 467: So, following your comment from above about using vector for position and color for well color, I assume the right way would be to split this function to two.
For vector, only support position. For color support all others. Except attenuation seems to use position for something using target. 
Enough, I'm  going to assume you've done your due diligence.

royfalk (approved):
Looks bad to me ;)
Approved.

PR #1126, opened 2025-05-08
Move computer to components

royfalk:
Also, add ComponentManager class and move all components from unit_generic to it.

Code Changes:

Issues:
- None related to this. Opened issues for everything else.

This is part of the effort to move components to lib component.
We are getting to the point where we'll be able to refactor stuff in unit code.
Additional comments in code.

- What release is this for? next
- Is there a project or milestone we should apply this to? no.


engine/src/components/computer.h
```diff
@@ -23,48 +21,83 @@
...
+    // the "correct" "subclass". This is Unit for ComponentsManager and
+    // Vector for CoreVector.
+
     //The nav point the unit may be heading for
-    Vector NavPoint;
+    CoreVector nav_point;
+
     //The target that the unit has in computer
-    UnitContainer target;
+    ComponentsManager* target;
```
royfalk, line 54: As part of encapsulation, we can't use Unit, as it is connected to everything in the game.
Instead, we're using a subclass to hold a pointer to it.
It requires casting to work.

engine/src/gfx/cockpit_gfx.cpp
```diff
@@ -846,8 +849,9 @@ void DrawTurretTargetBoxes(const Radar::Sensor& sensor)
         Radar::Track track = sensor.CreateTrack(target, Loc);
         if (configuration()->graphics.hud.draw_nav_symbol) {
             GFXColor4f( 1, 1, 1, 1 );
-            DrawNavigationSymbol( subunit->GetComputerData().NavPoint, CamP, CamQ,
-                                 CamR.Cast().Dot( (subunit->GetComputerData().NavPoint).Cast()
+            Vector nav_point = subunit->GetNavPoint();
```
royfalk, line 852: Vector is not available in lower libraries. It is currently in lib_gfx.
CoreVector is available in lib_resource. We use it and a convenience function GetNavPoint to convert.

libraries/cmd/planetary_orbit.cpp
```diff
@@ -28,6 +28,7 @@
 
 #include "cmd/unit_generic.h"
 #include "src/vs_logging.h"
+#include "root_generic/configxml.h"
```
royfalk, line 31: I don't know why, but I couldn't compile without explicitly importing configxml.h.

libraries/cmd/unit_generic.h
```diff
@@ -767,9 +739,19 @@ class Unit : public Armed, public Audible, public Drawable, public Damageable, p
...
  */
-
 public:
+    Vector GetNavPoint();
+    Unit *Target();
+    const Unit *Target() const;
+    void SetTarget(Unit *target);
     Unit *Threat();
+    const Unit *Threat() const;
+    void SetThreat(Unit *target);
```
royfalk, line 749: It's tempting to name this `Threat(Unit *target)` to be consistent with the rest.
However, `unit_generic.cpp:1577` has another function with the same name that does a whole lot more.
Ask me how I know.

engine/src/components/computer.h
stephengtuggy: I don't understand the logic here. Tell me again why we can't use Unit?

libraries/cmd/unit_generic.cpp
```diff
@@ -170,24 +172,65 @@ bool Unit::InRange(const Unit *target, double &mm, bool cone, bool cap, bool loc
     return true;
 }
 
+
+
+// two convenience functions
+Unit* CheckNullAndCastToUnit(ComponentsManager* manager) {
```
stephengtuggy, line 178: IIRC, vega_dynamic_cast_ptr (and the const variant) were intended to handle nulls, as well. If they aren't handling nulls adequately, maybe we should look at how to fix that, instead of introducing two new functions that barely do anything.

engine/src/gfx/cockpit_gfx.cpp
stephengtuggy: CoreVector? Where's that?

stephengtuggy:
Why are there three conflicting files already?

libraries/cmd/unit_generic.cpp
royfalk: Huh. I was actually automatically using your functions without thinking. 
I've had crashes when trying to cast a null *_ComponentManager_ to *_Unit_.
I think _dynamic_cast_ is enough here.

royfalk:
> Why are there three conflicting files already?

I was working on this for longer than I thought and time moved on.

stephengtuggy:
> > Why are there three conflicting files already?
> 
> I was working on this for longer than I thought and time moved on.

I see.

libraries/cmd/unit_generic.cpp
stephengtuggy: Perhaps `dynamic_cast` is enough. However, if we can encapsulate some common logic surrounding the cast, such as throwing an exception/aborting the game if the cast fails, then I think that might be worthwhile. That was the idea behind `vega_dynamic_cast_ptr` initially.

engine/src/components/computer.h
BenjamenMeyer: @stephengtuggy its too generic. This is one of the  big things we need to do with the code base - break some things down to subclasses where we work more specifically instead of having to cast everything.

libraries/cmd/planetary_orbit.cpp
BenjamenMeyer: probably one of the headers needing it and failing to include it; that'd probably be worth investigating

BenjamenMeyer (commented):
LGTM; interested in the questions that @stephengtuggy raised

engine/src/components/computer.h
stephengtuggy: I don't follow. If we don't want to cast everything, then we should pass around a pointer-to-Unit everywhere, as that's what the underlying data type is: Unit.

stephengtuggy: Right?

royfalk: So, the issue here is one of encapsulating certain lower level libraries.
Unit and a lot of other classes from the original devs reference everything and are referred to by everything.
That means that you can't add them to unit testing without adding the entire project.
It also takes forever to compile changes to such a file, because a lot of other stuff gets recompiled.
Finally, it's just bad practice.

Also, see below core vector.

engine/src/gfx/cockpit_gfx.cpp
royfalk: It's a very simple implementation of a struct with i,j,k.
I used it originally to refrain from including Vector in lib_damage, which would necessitate pulling all of gfx with it.
It's in lib_resource, which should really be renamed lib_core, as it houses some core stuff.

libraries/cmd/planetary_orbit.cpp
royfalk: So I've had this issue before. The game would stop compiling after I would make a change to some unrelated file. 
I think the issue here is that some of the includes are nested. 
_a.h -> b.h -> c.h_

When you delete
`#include "b.h" `
from _a.h_, stuff starts breaking down.

libraries/cmd/unit_generic.cpp
royfalk: Except I don't want to throw an exception and I never will.
The function checks for null and returns null in that case.
I should be able to return null Unit*.
The issue here is that if you try and dynamic cast null it will fail.

engine/src/components/computer.h
BenjamenMeyer: > I don't follow. If we don't want to cast everything, then we should pass around a pointer-to-Unit everywhere, as that's what the underlying data type is: Unit.

the original code passed around `Unit*` everywhere and then did casting in specific cases which meant everything had to check that the cast worked - or it assumed it did and might crash if it didn't.  This is generally bad practice.

Now we can manage a core list of Unit*, but then when something knows that it is a specific type, then it should pass it down to callers as that type so that lower levels do not need to cast and check, they can just rely on the interface of a specific type.

That is to say - a generic list to manage memory might be Unit*; but most everything else should be working in specific unit types. We shouldn't accidentally treat a weapon as a ship or planet, and we should make it easy to keep from doing that. Even the physics engine should probably work on a subclass of Unit instead of Unit; but even then it'll probably still be pretty high up in the inheritance tree.

There's overhead to always casting it.

stephengtuggy: You guys seem to be forgetting how the code was when we inherited it (0.5.3). There was only Unit, and GameUnit, and UnitContainer. All these other related classes didn't exist. So there was no need to cast between them. Right?

Drawable and Collideable and the rest were Roy's invention.

libraries/cmd/unit_generic.cpp
stephengtuggy: Ultimately, a failing cast is a "shouldn't happen." If it happens, we should abort the game as gracefully as possible, offering the user a way to submit a bug report. That requires some common logic. How else are you going to do it?

engine/src/components/computer.h
BenjamenMeyer: @stephengtuggy true...there was a bunch of casting but yes. Wow we've come a long ways. :smiley_cat:

libraries/cmd/planetary_orbit.cpp
BenjamenMeyer: As a general guide:

- any header file should directly include header files it needs to compile correctly and no more
- any implementation file should directly include header files it needs to compile correctly and no more

We should winnow away at these issues by fixing the includes of the various header/implementation files. It may be a bit painful to do but it'll be worth it in the long run.

engine/src/components/computer.h
stephengtuggy: Indeed we have!

libraries/cmd/unit_generic.cpp
royfalk: Looking at it with fresh eyes and rereading your comments, I think my original code was actually the best.
There are three separate scenarios here:
1. `manager` refers to a `Unit` and gets cast successfully.
2. `manager` is **null** and if cast, would crash the game. We should check this and return **null**.
3. `manager` isn't **null** but isn't an instance of `Unit`. This would be caught by `vega_dynamic_cast_ptr`.

stephengtuggy: OK. I can go with your original code.

stephengtuggy: I think the three scenarios you identified are correct and make sense. Thanks.

PR #1132, opened 2025-05-11
Attempted Fix For Issue 1112 (game slowness)

### LINKED ISSUE #1112: Game is no longer playable in 2540x1440
As it says. The game is now too slow. This happened sometime in the last month or so. 
Need to figure out what changed. 
I assume this slow down is also relevant though less so in lower resolutions.

stephengtuggy:
Code Changes:

Issues:
- Fixes #1112 

Purpose:
- What is this pull request trying to do? Fix #1112 , the issue with the game running too slowly (especially at higher resolutions) over the last several weeks


libraries/cmd/unit_generic.h
```diff
@@ -165,6 +165,7 @@ class Unit : public Armed, public Audible, public Drawable, public Damageable, p
     // - model (stock) - should be variant, but only if actually different without upgrades
     // TODO: name is duplicated down below in some memory saving measure (StringPool::Reference)
     std::string unit_key;
+    bool is_planet{};
```
royfalk, line 168: Better to make it

`enum Vega_UnitType unit_type`

and then set it at every constructor and be done. That would also mean you don't need `isUnit` anymore.

libraries/cmd/unit_generic.h
```diff
@@ -990,13 +991,23 @@ class Unit : public Armed, public Audible, public Drawable, public Damageable, p
         return filename.get();
     }
 
-//Is this class a unit
+private:
+    void setUnitKey(const std::string& key) {
+        unit_key = key;
+        // Cache this calculation so that we don't have to compare the strings over and over again
+        is_planet = boost::algorithm::ends_with(unit_key, "__planets");
```
royfalk, line 998: This is much more fragile. Better to use the constructor.
Of course, my suggestion only works if we use `Planet` class for planets.

libraries/root_generic/macosx_math.cpp
```diff
@@ -206,7 +206,7 @@ float tanf( float v )
 
 float powf( float v, float p )
 {
-    return (float) pow( (double) v, (double) p );
+    return (float) std::pow( (double) v, (double) p );
```
royfalk, line 209: Do we really still need macos specific files?

royfalk (commented):
So I play tested the PR.

Performance is better. However, auto-pilot was broken for me. It started and stopped continuously. It also kept aligning to the target and then moving away.

It is possible I was in a bad place in the system. This sometimes happens in the game and it confuses the auto-pilot. But I started a new game in Atlantis and I don't recall ever having it happen there. 

It's not clear to me what about this PR fixed the performance so I just reviewed it normally. With the auto pilot down, I was not able to verify one test of smooth shield power down and up (visually).

I can't approve this until we figure out if auto-pilot is a regression here, earlier or just a fluke in my test.

libraries/cmd/unit_generic.h
stephengtuggy: Yeah. I'm not sure that we do use the `Planet` class for planets.

I verified that all of the places in the code that set unit_key use this method now. So `is_planet` will always be set correctly. (Or its replacement, `unit_type`, as you noted above.)

libraries/cmd/unit_generic.h
stephengtuggy: That's a good point. `unit_type` makes more sense. Although `isUnit` might be part of the public, Python-facing interface of the Unit class, so I don't think I will remove or rename that method now. But the field underneath can be `unit_type`.

libraries/root_generic/macosx_math.cpp
stephengtuggy: No, no we do not. This should not really be made macOS-specific. Including <cmath> and then using `std::pow` is the way to do it, everywhere, across the board. Let's just do that.

stephengtuggy: So this whole function would go away, wouldn't it?

libraries/cmd/unit_generic.h
stephengtuggy: We have only two choices with our implementation: either it can be fragile, or it can be unnecessarily computation-heavy. Seems to me

stephengtuggy:
I've had those types of issues with the autopilot forever, practically. I doubt that's anything new, based on what you're describing.

engine/src/gfx/cockpit.cpp
```diff
@@ -2008,7 +2008,7 @@ void GameCockpit::Draw() {
             const float deadband = configuration()->joystick.mouse_deadband;
             const int reverse_spr = configuration()->joystick.reverse_mouse_spr;
             const string blah = configuration()->joystick.mouse_crosshair;
-            VSSprite MouseVSSprite(blah.c_str(), BILINEAR, GFXTRUE);
+            static VSSprite MouseVSSprite(blah.c_str(), BILINEAR, GFXTRUE);
```
stephengtuggy, line 2011: This is what fixed the performance issue, I believe. When I made this variable non-static, all of a sudden we had to construct a new `VSSprite` every time this code was hit. Very computationally expensive.

stephengtuggy: I guess the lesson here is to leave computation-heavy and I/O-heavy object variables as `static`, even if their value depends on another variable that is now `const` and not `static`. At least, wherever that is feasible.

libraries/cmd/unit_generic.h
stephengtuggy: OK. So in the base `Unit` class, I have a field named `unit_type`. It's private. It is only used in the base class. Every subclass hard-codes their unit type. In the new, virtual method named `getUnitType()`. (In the base `Unit` class, of course, `getUnitType()` returns the value of `unit_type`.)

Note that `getUnitType()` and `isUnit()` are functionally identical. They always return the same thing. It's just a matter of which one makes the most logical sense to use in a particular context. In almost all cases, getUnitType seems to me to make more sense, so I have converted most of the 100 or so references to `isUnit` to `getUnitType` instead.

libraries/cmd/unit_generic.h
BenjamenMeyer: > This is much more fragile. Better to use the constructor. Of course, my suggestion only works if we use `Planet` class for planets.

if we don't, we probably should.

libraries/root_generic/macosx_math.cpp
royfalk: Sure looks that way.

libraries/cmd/unit_generic.h
stephengtuggy: > > This is much more fragile. Better to use the constructor. Of course, my suggestion only works if we use `Planet` class for planets.
> 
> if we don't, we probably should.

That's a bigger change that I'm not prepared to make in this PR.

libraries/root_generic/macosx_math.cpp
stephengtuggy: OK. Can that be done in a separate PR @royfalk ?

BenjamenMeyer (changes requested):
Please add a comment about the static requirement for the one variable.

I expect we'll have some similar issues as we go forward and clean things up that were overly one way or another. We just need to document when we find them.

engine/src/gfx/cockpit.cpp
stephengtuggy: @BenjamenMeyer done.

libraries/root_generic/macosx_math.cpp
royfalk: Of course.

libraries/cmd/movable.cpp
```diff
@@ -784,16 +784,16 @@ void Movable::Thrust(const Vector &amt1, bool afterburn) {
     }
 }
 
-// If in Travel mode (non-combat), speed is limited to x1000
+// If in Travel mode (non-combat), speed limit is multiplied by a power of 10.
```
royfalk, line 787: There's a lot to unpack here.
1. The engine refers to combat_mode_multiplier but shouldn't it be non_combat_mode_multiplier everywhere?
2. I assume I'll find a separate PR where you explain the change. In any case, this isn't correct. At best, the comment should say it defaults to x10.

royfalk (approved):
LGTM. 
I opened two issues:
#1139 
#1141

libraries/cmd/movable.cpp
stephengtuggy: I'm trying to keep track of both the old way things were, and the new way things are now, until we figure out how they're supposed to be. That's why I added back the old code, commented out. It's also why I said more generically that the speed limit is "multiplied by a power of 10." It could be 100, or it could be 1000. For now, that is left unspecified.

The old code said combat_mode_multiplier. The new code says non_combat_mode_multiplier. It's pretty confusing, and I don't know which is correct.

Obviously, this situation isn't ideal. We need to figure out which it's supposed to be, on all of these counts, and change the code and comments accordingly. I just don't know enough to do that yet.

royfalk: Sorry. I misunderstood your comment. But is there something stopping us from using x25?!

stephengtuggy: Not particularly. We could use x25 if we wanted lol

BenjamenMeyer (approved):

Issue #169, opened 2020-07-09
Change the Saved Game Format


BenjamenMeyer:
Presently the game uses a CSV format, and the former maintainers tried to also split the CSV data into multiple CSV files which we've found to be problematic (see https://github.com/vegastrike/Assets-Production/pull/7 for details).

Let's move to a different save game format.

Currently JSON, TOML, and YML are all currently excepted formats. However, only JSON has native support in Python (even Py3) without requiring additional libraries so if we need to open it from within the Python code, then we should limit it to the JSON format. (NOTE: JSON and TOML are both related to YML; I don't know why Python doesn't have a built-in YML parser.)

Tasks:
- [ ] Decide on the format to use
- [ ] Make the saved game file functionality a specific function of the game engine that is used by the assets so the assets don't have to be concerned about the file format, but can instead just do a load/save with the game engine.
- [ ] Make any conversion tools required to upgrade old file formats to the new file format.
- [ ] Provide a saved game editor tool that works with both the old and new formats, perhaps even providing a delta between the two so folks can ensure that the conversion worked properly.

pyramid3d:
My recommendation:

**Serialization Formats**
-------------------------------------

There are many formats available for serialization, the most commonly used being CSV, XML, JSON, YAML.

Our requirement for data serialization is that it would be:

- Editable with a text editor
- Human readable
- Machine readable
- Compact even for large data sets
- Libraries available in popular programming languages

Following the analysis and recommendations in this article, we equally **recommend** **JSON** as the select serialization data format for fulfilling all of the above criteria better than other formats.

[1]  [CSV vs XML vs JSON – Which is the Best Response Data Format?](https://digitalhospital.com.sg/csv-vs-xml-vs-json-which-is-the-best-response-data-format/). Digital Hospital. 2016. Retrieved on 2020-02-23.

Source: https://github.com/LibreGamingManifest/libre-gaming-manifest/blob/master/libre-gaming-manifest.md#serialization-formats

stephengtuggy:
Cool. Sounds like JSON might be the way to go then.

stephengtuggy:
Question: Are we making the correct distinction between the CSV format of the **assets** (which don't change) vs. the **saved game** format?

nabaco:
I think @BenjamenMeyer meant the CSV assets table, since the savefile doesn't seem to be a CSV file at all.
[...]

PR #1136, opened 2025-05-14
Enhancement: Disk Object Library

BenjamenMeyer:
Code Changes:

** has not yet been tested, not fully hooked up yet. Filling for the team to monitor and be able to give feedback **

Issues:
- #169 

Purpose:
    - Add a new diskobject library to start building out a new saved file format.
    - Link in the libarchive library which provides support for a variety of compressed file formats and containers types.
    - Add support for listing the internal files
    - Add support for reading one of the internal files.
    - Record warnings and errors to the VS Log
- What release is this for?
    - 0.10.x or later
    - 0.10.x or later


stephengtuggy:
I just want to note that there's already something along the lines of a `diskobject` library in the existing code. It's in VSFilesystem, and it supports reading from .pk3 archives, as well as plain files on disk, IIRC. Whatever solution we end up should probably either integrate with what's there today, or replace it completely.

stephengtuggy:
@BenjamenMeyer the Linux builds are passing, now that the base images have been updated. My guess is that next, you need to add libarchive to the Homebrew dependencies and MacPorts dependencies on macOS, and to vcpkg.json on Windows.

BenjamenMeyer:
> @BenjamenMeyer the Linux builds are passing, now that the base images have been updated. My guess is that next, you need to add libarchive to the Homebrew dependencies and MacPorts dependencies on macOS, and to vcpkg.json on Windows.

CodeQL doesn't have LibArchive either. Will have to look into that part to. I was already aware of the macOS and Windows portion needing updated too. Thanks for pointing out where I need to with Windows.

BenjamenMeyer:
> I just want to note that there's already something along the lines of a `diskobject` library in the existing code. It's in VSFilesystem, and it supports reading from .pk3 archives, as well as plain files on disk, IIRC. Whatever solution we end up should probably either integrate with what's there today, or replace it completely.

It really only needs to read the PK3 files since those are static during game play; where what I'm doing with `diskobject` will need to both read _and_ write. Now, looking up PK3 files they're supposedly just ZIP files so we may be able to use `diskobject` to support them too.

Right now I'm just focusing on getting some infrastructure together for the files. It'll probably need to get integrated into VSFilesystem to be useful.

Consider this the proposal stage of this. It'll be quite a while before I'll be ready to merge it in.

royfalk (changes requested):
Looks real nice. 

Not wild about using fopen but a quick check shows it's part of C++. [https://stackoverflow.com/q/6399822/5276890](stack overflow) don't have any firm conclusions. I couldn't find a boost way, so I guess it's fine.

I am missing unit tests and that's my only request here.

libraries/diskobject/diskobject.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+#ifndef VEGA_STRIKE_ENGINE_DISK_OBJECT_H
+#define VEGA_STRIKE_ENGINE_DISK_OBJECT_H
+
+#include <deque>
+#include <string>
+
+#include <archive.h>
+
+namespace vegastrike {
```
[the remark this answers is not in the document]
stephengtuggy: Yeah, maybe use namespace `vega_disk` or something?

stephengtuggy:
I'm not wild about using fopen either. Can we use something closer to C++ streams syntax? Similar to `cin >> var`, or in this case, `file >> var`?

BenjamenMeyer:
@royfalk yes, I'd like to get tests setup for this too. Just posting this now for early feedback.

libraries/diskobject/diskobject.h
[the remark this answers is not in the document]
BenjamenMeyer: @stephengtuggy I like that - using `vega_disk` for the C++ namespace and `vegadisk` as the library name instead of `disk_object`

BenjamenMeyer:
Also, noticed that `posh` is really a 3rd party library kind of like the collider is, so moving it to its own library for now.

BenjamenMeyer:
> I wish I knew what was wrong with the macOS build. Not sure what's going on there.

Possible we are running into https://github.com/Homebrew/homebrew-core/issues/120526 - apparently its PKG CONFIG is broken. :( I'll have to see if I can get the workarounds to function.

stephengtuggy:
> > I wish I knew what was wrong with the macOS build. Not sure what's going on there.
> 
> Possible we are running into [Homebrew/homebrew-core#120526](https://github.com/Homebrew/homebrew-core/issues/120526) - apparently its PKG CONFIG is broken. :( I'll have to see if I can get the workarounds to function.

Maybe try `brew reinstall libarchive` and see if that helps.

script/cibuild
```diff
@@ -97,13 +97,20 @@ then
     echo "   Homebrew Python CFlags: ${HOMEBREW_PYTHON3_CLFLAGS}"
     echo "  Homebrew Python LDFlags: ${HOMEBREW_PYTHON3_LDFLAGS}"
 
+    export HOMEBREW_LIBARCHIVE_PREFIX="$(brew --prefix libarchive)"
+    export PKG_CONFIG_PATH="${HOMEBREW_LIBARCHIVE_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH}"
+    export HOMEBREW_LIBARCHIVE_CFLAGS="-I${HOMEBREW_LIBARCHIVE_PREFIX}/include"
+    export HOMEBREW_LIBARCHIVE_LDFLAGS="-L${HOMEBREW_LIBARCHIVE_PREFIX}/lib"
```
stephengtuggy, line 103: This is a far-less-than-optimal way to do this, and I doubt it's necessary. Just set these environment variables in the proper macOS base preset in CMakePresets.json. Then make sure that we use those presets in the CI scripts and stuff. (I think we already do on macOS, as well as on Windows.)

stephengtuggy: Maybe the way you tried it isn't so bad after all. I'm proposing to make the necessary adjustments in CMakePresets.json, but in there, I don't think I have access to the `brew --prefix` command. I'll have to hard-code the result of that command, which is not ideal either.

stephengtuggy: > The result changes between Arm and Intel Mac releases too

Indeed it does. For that and other reasons, I think we should go ahead and run the `brew --prefix` command and use its result to help locate LibArchive.

My latest code does that, but it's still not working for some reason.

stephengtuggy: I got it to work! See https://github.com/BenjamenMeyer/Vega-Strike-Engine-Source/pull/4

BenjamenMeyer:
NOTE: The value in this PR now might just be moving stuff to the `vegadisk` library. @royfalk is working on a JSONification of the saved game (see #1325) which might overtake using libarchive/diskobject - not sure yet.

@royfalk can we merge this AS-IS and then I'll come back with unit tests for diskobject afterwards? This would allow the work to be easily merged with #1325 without holding up on the diskobject portion which might get outdated by #1325 to some degree but I think the vegadisk library and some of the other refactors here still have merit. Thoughts?

Refactors:
- posh.[h|cpp] :arrow_forward: posh library
- pk3.[h|cpp], savedgame.[h|cpp], vsfilesystem.[h|cpp] :arrow_forward: vegadisk library

New Additions:
- diskobject.[h|cpp] to vegadisk library; creates an interface for working with compressed files using libarchive.

script/bootstrap-mac.sh
```diff
@@ -56,7 +56,7 @@ echo "Detected Mac OS X ${DETECT_MAC_OS_VERSION} - ${MAC_OS_NAME}"
 
 # Install the stuff we know needs to get installed all the time
 brew install \
-    cmake \
+    gcc \
```
stephengtuggy: What is the reason for this change? Isn't CMake a hard requirement? And I doubt that the average macOS system will have a suitable version installed by default.

stephengtuggy: Overall, this seems like a good and necessary set of changes. I just have that one question about installing CMake.

BenjamenMeyer: @stephengtuggy it was a change on master that I pulled into this branch - it drops `cmake` from L59 but then adds it in another line a few lines below.

stephengtuggy: Oh, interesting. OK. That works.

BenjamenMeyer:
@royfalk @stephengtuggy I added the extremely basic tests to the vegadisk for diskobject so that there's something there for now; documented what parts still need tests which would be better for another PR right now if we can.

stephengtuggy:
@royfalk just FYI, this PR is waiting on you at the moment

BenjamenMeyer:
Posh deals with file format for the assets.

PR #1138, opened 2025-05-21
0.9.x: Don't install any Start Menu shortcuts. Leave that to the Assets installer

### LINKED ISSUE #1130: 0.9.1 beta 2 installer creates invalid Start menu shortcut on Windows
After installing 0.9.1 beta 2 on Windows 11 (build 26100.3775) along with the latest 0.9.1 beta 1 assets, attempting to launch the game from the Start menu shortcut created during installation fails.

Some quick examination of the shortcut properties reveals that the shortcut attempts to launch `<InstallRoot>\bin\vegastrike.exe.exe`, which obviously is incorrect due to the double .exe extension. So this is the first thing that needs fixing.

Now, even fixing that alone would not resolve the issue. The way the engine and asset installers combine right now would require further changes to the shortcut to make it work. The engine installer places its stuff in `<InstallRoot>\bin`, while the assets installer place its own stuff in `<InstallRoot>\share\vegastrike`. In order for the vegastrike.exe to work,, the `bin` folder should also be under `<InstallRoot>\share\vegastrike` or, in general, wherever all the assets are located. Fixing this implies either changing the default installation folder locations or swapping out vegastrike.exe with vegastrike-engine.exe in the shortcut properties and passing as command line parameter (-d) the location of the data files. So ideally, the assets installer should be run first and the engine installer should be able to detect that an assets installation is already present or ask the user to specify its location in order to be able to create the correct -d parameter for the shortcut.

Or maybe there is another, simpler way which I may have missed. In any case, I think that resolving the shortcut issue would greatly help new players get into the game with as little hassle as possible, especially since they have to perform a manual installation of Python too to get it to run. Windows users generally prefer a one-click-does-it approach and if it's not there they tend to quickly give up and move on.

stephengtuggy:
Code Changes:
- [X] This is an installer change only

Issues Addressed:
- Fixes #1130 

Known Issues Remaining:
- Nothing new that I'm aware of

Purpose:
- What is this pull request trying to do? Remove any Start Menu or desktop shortcuts created by this installer. Leave shortcut creation to the vsUTCS installer
- What release is this for? 0.9.1

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1344,7 +1344,8 @@ IF (CMAKE_SYSTEM_NAME STREQUAL Windows)
     SET(CPACK_NSIS_HELP_LINK "https://vega-strike.org")
     SET(CPACK_NSIS_URL_INFO_ABOUT "https://vega-strike.org/about")
     SET(CPACK_NSIS_CONTACT "maintainers@vega-strike.org")
-    SET(CPACK_NSIS_MODIFY_PATH ON)
+    SET(CPACK_NSIS_MODIFY_PATH OFF)
+    SET(CPACK_NSIS_MENU_LINKS "")
 
     # Windows installer format(s) to output
     SET(CPACK_GENERATOR "NSIS64")


stephengtuggy:
See also https://github.com/vegastrike/Assets-Production/pull/182

royfalk (approved):
Approved.

Issue #1110, opened 2025-04-21
Unable to jump for no apparent reason


kheckwrecker:
Playing the latest version of 0.10x in Windows, and took a cargo run from Cephid-17 to War, flying in a Llama. Got as far as Aega and now I can't jump. My HUD just says Jump: Low Energy. All the power bars on my HUD show full power.  I've landed at the Commerce center, replaced the jump drive, capacitor and reactor. Purchased Basic Repair and Refuel. Nothing seems to make any difference. I'm attaching the saved game files if that helps. 

[Llama.begin.json](https://github.com/user-attachments/files/19837739/Llama.begin.json)

I'm also attaching a screenshot from the game where I'm stuck.

![Image](https://github.com/user-attachments/assets/512dfb2f-98c6-453b-9fb9-96647d560eb9)

stephengtuggy:
Yes, the required number of MJ for jumping has gone up significantly in 0.10.x. However, FTL Capacitor capacity has increased significantly as well. A single FTL Capacitor 3 should be enough to let you jump any number of times. In fact, an FTL Capacitor 2 might be enough. Or, two to three FTL Capacitor 1’s.

I recommend that you stock up on FTL capacitors at Atlantis, as early in the game as possible; they’re difficult to find elsewhere.

@royfalk  and @evertvorster  and @BenjamenMeyer  and I are having an ongoing discussion about these units of measure. Feel free to chime in. I will see if I can find you the link.

stephengtuggy:
I don’t have as much personal experience with the afterburner issue, but I believe that that’s been reported as well.

kheckwrecker:
> Yes, the required number of MJ for jumping has gone up significantly in 0.10.x. However, FTL Capacitor capacity has increased significantly as well. A single FTL Capacitor 3 should be enough to let you jump any number of times. In fact, an FTL Capacitor 2 might be enough. Or, two to three FTL Capacitor 1’s.
> 
> I recommend that you stock up on FTL capacitors at Atlantis, as early in the game as possible; they’re difficult to find elsewhere.
> 
> [@royfalk](https://github.com/royfalk) and [@evertvorster](https://github.com/evertvorster) and [@BenjamenMeyer](https://github.com/BenjamenMeyer) and I are having an ongoing discussion about these units of measure. Feel free to chime in. I will see if I can find you the link.

[...]

PR #1155, opened 2025-05-28
Refactor damage VDU

royfalk:
- Instead of generating the text every cycle, we generate the text when damaged.
- As lib_component has no visibility to lib_gfx and colors, we provide the GetDamageColor function as a callback parameter.
- Deleted shadowing components from unit_generic.h.

Code Changes:
    Partially. There seems to be an issue with ships losing capacitor and not being able to jump. #1110.
    I have tested the VDU functionality.
    I did not test real time damage reflected in VDU. It should work but it's hard to generate.

Issues:
- #1110

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -1234,6 +1234,8 @@ IF (USE_GTEST)
         src/components/tests/drive_tests.cpp
         src/components/tests/afterburner_tests.cpp
         src/components/tests/jump_drive_tests.cpp
+        src/components/tests/utils_tests.cpp
+        src/components/tests/operational_tests.cpp
     )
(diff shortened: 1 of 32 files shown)


BenjamenMeyer (approved):

BenjamenMeyer:
@royfalk 100% agreed. We setup the play test as a basic means if testing, fully intending it to eventually be replaced by unit and integration testing. For a long time we didn't have the understanding of the codebase to do it, and the code was structured such that it made it hard as well. I think we're at the turning point for that now. There's still a lot of untestable code, but there's also a lot that is testable too. So yes, unless there's some reason we can't test it - which should then be well documented - we should start including at least unit tests in the code now. I certainly intend to with my file format PR.

stephengtuggy:
I started a separate discussion: https://github.com/orgs/vegastrike/discussions/1163

PR #1157, opened 2025-05-29
Convert Vega_UnitType to enum class

### LINKED ISSUE #1156: Game crash due to unknown track type
Following jump to one of the neighboring systems (Enyo?!), the game crashed due to unknown track.

Root cause:
_track.cpp:184_ has 
`assert(false);`
which would crash the game intentionally.
I've added some cout to see the value provided.

royfalk:
Fix #1156 

- Move it to a separate file
- Fix bug? in sensor.cpp where isPlanet result was compared to planet
- Fix issue where game crashed when track result was unknown. This was intentional and would display this for the Rlaan stealth vessel.

Code Changes:
- [-] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
    Minimal. Checked you can jump to Enyo and not crash which consistently happened before with my saved game.


stephengtuggy:
Let's finish running the PR Validation Tests on this PR, please. Other than that, it looks reasonable.

stephengtuggy (approved):
LGTM.

libraries/cmd/unit_type.cpp
```diff
@@ -0,0 +1,70 @@
...
+Vega_UnitType DeserializeUnitType(std::string type) {
+    if(type == "asteroid") {
+        return Vega_UnitType::asteroid;
+    } else if(type == "planet") {
+        return Vega_UnitType::planet;
+    } else if(type == "nebula") {
+        return Vega_UnitType::nebula;
+    } else {
+        return Vega_UnitType::unit;
+    }
```
BenjamenMeyer, line 69: We should use a map for this. Using `if` blocks will have limits and each entry requires its own test. Early ones are faster than later - essentially O(N).

I generally use a static map for this kind of thing - something like the following:
```C++
typedef std::map<std::string, Vega_UnitType> dataMap; 
typedef std::pair<std::string, Vega_UnitType> dataEntry;
static dataMap deserializer;
static bool initialized = false;
if (!initialized) {
      // temporary convenience to DRY the code for inserting all the values
     #define ADD_ENTRY(name, value) \
         {  \
         deserializer[name] = value;\
         }
     ADD_ENTRY("asteroid", Vega_UnitType::asteriod);
     ADD_ENTRY("planet", Vega_UnitType::planet);
     ADD_ENTRY("nebula", Vega_UnitType::nebula);
     #undef ADD_ENTRY
     initialized = true;
}

if (auto search = deserializer.find(type);  search != deserializer.end()) {
    return search;
} 

return Vega_UnitType::unit;
```

I understand this is a small thing right now. Don't know if we'll add more in the future but this is a good pattern to use and ends up with essentially O(log(N)) after the first call as std::map::find is pretty efficient.

Given the size I'm not going to block on this.

PR #1211, opened 2025-06-17
Convert graphics settings to `configuration()->graphics.` ...

### LINKED ISSUE #1183: Migrate `graphics` settings section to new mechanism

stephengtuggy:
Code Changes:

Issues Resolved:
- Closes #1183 

Known Issues Remaining:
- On bases/planets, it's harder to tell where to click on the screen to go to various other "rooms"
- Font size is tiny
- Startup music is different. This is an assets issue, and I will address it soon.

Purpose:
- What is this pull request trying to do? Migrate the graphics settings section to the latest configuration mechanism we are using, configuration()


BenjamenMeyer (commented):
Absolutely love this. Cleaning up the XML stuff so that we can manage the file format within the configuration tooling is going to pay off in multiple dividends.

Also the change from the `static` to the `const` will be very beneficial too. Though, let's please get some extra play testing on this since we've already seen a case where removing the static caused a performance issue. So let's test this like crazy to make sure that's not the case here - that should be something rare so it shouldn't be an issue but let's make sure.

All-in-all I love where this is going.

stephengtuggy:
> Absolutely love this. Cleaning up the XML stuff so that we can manage the file format within the configuration tooling is going to pay off in multiple dividends.
> 
> Also the change from the `static` to the `const` will be very beneficial too. Though, let's please get some extra play testing on this since we've already seen a case where removing the static caused a performance issue. So let's test this like crazy to make sure that's not the case here - that should be something rare so it shouldn't be an issue but let's make sure.
> 
> All-in-all I love where this is going.

Testing complete. There were a few differences, but in general, everything passed.

Yes, I too am happy with this progress.

stephengtuggy:
Game performance seemed fine. No issues noticed there. In fact, if anything it was faster on some operations.

engine/src/cmd/ai/firekeyboard.cpp
```diff
@@ -1017,7 +1017,7 @@ bool TargMission(Unit *me, Unit *target) {
 }
 
 bool TargAll(Unit *me, Unit *target) {
-    static bool can_target_sun = XMLSupport::parse_bool(vs_config->getVariable("graphics", "can_target_sun", "false"));
+    const bool can_target_sun = configuration()->graphics.can_target_sun;
```
royfalk, line 1020: why not just do?

`|| me->InRange(target, true, true)) && (configuration()->graphics.can_target_sun || !UnitUtil::isSun(target)) `

engine/src/cmd/ai/firekeyboard.cpp
```diff
@@ -1026,8 +1026,7 @@ bool TargAll(Unit *me, Unit *target) {
 }
 
 bool TargSig(Unit *me, Unit *target) {
-    static bool can_target_asteroid =
```
royfalk, line 1029: Also, I wonder if you can't write a quick script to go over all cpp files and search and replace. 
It would be foolhardy to do all but you can give it keys to search for.
Something to think about.

royfalk (approved):

engine/src/cmd/ai/firekeyboard.cpp
stephengtuggy: I did not manually touch this code. Just ran search and replace on it, for the most part. I figure we can perform individual optimizations / tweaks like this later.

engine/src/cmd/ai/firekeyboard.cpp
stephengtuggy: I did write a search-and-replace script. 😄 Checked in as `script/convert_config_section.py`.

engine/src/cmd/ai/firekeyboard.cpp
stephengtuggy: Oh wait. This is more of a case of inlining the setting, isn't it? I do plan to add that functionality to the convert_config_section.py script; just haven't done it yet.

PR #1224, opened 2025-06-25
Graphics settings continued

stephengtuggy:
Code Changes:

Issues Addressed:
- Contributes to #1211 

Known Issues Remaining:
- ...

Purpose:
- What is this pull request trying to do? Convert more graphics settings to the latest settings mechanism, `configuration()->graphics....`


engine/src/cmd/unit_functions.cpp
```diff
@@ -88,28 +88,28 @@ void update_ani_cache() {
 }
 
 std::string getRandomCachedAniString() {
-    if (cached_ani.size()) {
+    if (!cached_ani.empty()) {
```
royfalk: Better to reverse the checks and remove excess indentation. e.g.
```

if(cached_ani.emptry()) {
  return "";
}

unsigned int rn = rand() % cached_ani.size();
vsUMap<std::string, Animation *>::iterator j = cached_ani.begin();
for (unsigned int i = 0; i < rn; i++) {
...
```

engine/src/cmd/unit_functions.cpp
```diff
@@ -88,28 +88,28 @@ void update_ani_cache() {
...
         unsigned int rn = rand() % cached_ani.size();
-        vsUMap<std::string, Animation *>::iterator j = cached_ani.begin();
-        for (unsigned int i = 0; i < rn; i++) {
-            j++;
+        auto j = cached_ani.begin();
+        for (unsigned int i = 0; i < rn; ++i) {
+            ++j;
```
royfalk: stack overflow has the somewhat more elegant:
```
auto it = m.begin();
std::advance(it, rand() % m.size());
K random_key = it->first;
```

https://stackoverflow.com/a/15425535/5276890

engine/src/cmd/unit_functions.cpp
```diff
@@ -88,28 +88,28 @@ void update_ani_cache() {
...
     } else {
         return "";
     }
 }
 
 Animation *getRandomCachedAni() {
-    if (cached_ani.size()) {
+    if (!cached_ani.empty()) {
         unsigned int rn = rand() % cached_ani.size();
-        vsUMap<std::string, Animation *>::iterator j = cached_ani.begin();
```
royfalk, line 106: Same here, if you decide to. Also, why not merge the two functions. They practically do the same.

```
getRandomCachedAni_() {
 The heavy lifting
}
getRandomCachedAniString { one liner}
getRandomCachedAni { one liner}
```

engine/src/gldrv/gl_texture.cpp
```diff
@@ -937,10 +938,13 @@ GFXBOOL /*GFXDRVAPI*/ GFXTransferTexture(unsigned char *buffer,
         internalformat = GetTextureFormat(internformat);
         if (((textures.at(handle).mipmapped & (TRILINEAR | MIPMAP)) && gl_options.mipmap >= 2) || detail_texture) {
             if (detail_texture) {
-                static FILTER fil = game_options()->detail_texture_trilinear ? TRILINEAR : MIPMAP;
-                textures.at(handle).mipmapped = fil;
+                static boost::optional<FILTER> fil;
```
royfalk: Why make it an optional? The underlying configuration is enum/int. How can it be null?
Also, you first declare it, then immediately check if it's 0 and then...
Looking at FILTER, it's an enum so I think its value is going to be a random int?!
This will rarely fail and default to NEAREST, at least for now.

engine/src/star_system_jump.cpp
```diff
@@ -47,10 +47,12 @@
 #include "root_generic/options.h"
 
 void CacheJumpStar(bool destroy) {
-    static Animation *cachedani = new Animation(game_options()->jumpgate.c_str(), true, .1, MIPMAP, false);
+    static boost::optional<Animation> cachedani;
```
royfalk: I don't like this.
According to Google AI (unreliable, I know) and uninitialized boost::optional has a value of boost::none.
Are you sure the two are equivalent?

Looking at this some more, I fail to see the point of the whole exercise. We're creating an `Animation` instance that can't be accessed anywhere. There's only one call to the function and it creates and ~~immediately destroys the animation~~ leaks the animation.
Unless there's some evil code inside `Animation` that stores stuff in a singelton/static, the whole thing appears to be an evil exercise in futility.

engine/src/universe_util.cpp
```diff
@@ -185,17 +185,20 @@ void saveGame(const string &savename) {
...
             delete curSplash;
             curSplash = nullptr;
         }
-        curSplash = new Animation(filename.c_str(), 0);
+        curSplash = new Animation(filename.c_str(), false);
     } else if (!curSplash && !GetSplashScreen()) {
-        static std::vector<std::string> s = ParseDestinations(game_options()->splash_screen);
-        int snum = time(NULL) % s.size();
-        curSplash = new Animation(s[snum].c_str(), 0);
+        boost::optional<std::vector<std::string>> s;
```
royfalk, line 196: Am I missing something here?
s isn't static anymore and is going to be destroyed and recreated as boost::none in the next run.
I don't understand the original code or this one.

**Original**
1. Create a static vector
2. Initialize static Animation with a value

**New**
1. Create an optional
2. Check it isn't 0. This should be defined behavior but I'm not sure what it will be. Should be is_initialized?!
3. Create a random animation.

Both codes appear to leak curSplash and store stuff on the stack that has no business being there. We should move away from storing stuff on the stack.

libraries/cmd/unit_generic.cpp
```diff
@@ -1664,13 +1664,16 @@ bool Unit::Explode(bool drawit, float timeit) {
             FactionUtil::GetRandExplosionAnimation(this->faction, bleh);
         }
         if (bleh.empty()) {
-            static Animation cache(game_options()->explosion_animation.c_str(), false, .1, BILINEAR, false);
+            static boost::optional<Animation> cache;
+            if (!cache) {
```
royfalk: Same question regarding this check here. 

Also, this whole code is originally bleh. 
Check after check after check to make sure there is an animation. 
This is of course not on you, but I'd add a TODO somewhere to nuke this.

Nope. Took another look and this can't stay.
Animation isn't a pointer, it's a whole instance on the stack. 

My suggestion: 
1. Make cache a pointer. I assume it needs creating to initialize some internal mechanism in Animation.
2. Open an issue to investigate what animation does inside that needs this. That's twice in one PR we see these static Animation instances created without someone accessing them.
3. I guess that's it really.

libraries/cmd/unit_generic.cpp
```diff
@@ -1701,16 +1704,19 @@ bool Unit::Explode(bool drawit, float timeit) {
         if (!sub) {
             un = _Universe->AccessCockpit()->GetParent();
             if (this->getUnitType() == Vega_UnitType::unit) {
-                if (rand() < RAND_MAX * game_options()->percent_shockwave && (!this->isSubUnit())) {
-                    static string shockani(game_options()->shockwave_animation);
```
royfalk, line 1705: And here's another static instance that doesn't get accessed.
This time, it sets SetFaceCam.

royfalk (commented):
I'd like to discuss this some more. There are some open questions here that I don't understand.

engine/src/cmd/unit_functions.cpp
stephengtuggy: Good point. I will make that change.

engine/src/cmd/unit_functions.cpp
stephengtuggy: I like that much better. I will use that.

engine/src/cmd/unit_functions.cpp
stephengtuggy: Yes, the two functions could probably be merged. I will consider doing that.

engine/src/gldrv/gl_texture.cpp
stephengtuggy: In this PR in general, I was trying to solve the problem of a complex (class) type variable being declared as `static`, and then initialized before the settings from the config files have a chance to load. Googling what to do, it was suggested to use `std::optional` in C++17, or `boost::optional` as an alternative. The latter is what I adopted.

The thing is, though, `FILTER` is an enum. It's not a class. It doesn't even need this treatment, most likely.

engine/src/star_system_jump.cpp
stephengtuggy: Again, I was trying to solve the problem of a complex (class) type variable being declared as static, and then initialized before the settings from the config files have a chance to load. There were a number of these scattered through the code. If you have a better way to handle them, I'm all ears.

With `Animation` in particular, I think one of two things is happening: Either A) the cached Animations are getting accessed directly in a non-obvious manner from some other part of the code; or B) the cached Animations are not getting accessed directly, and the goal is just to keep them fresh in the operating system's in-memory cache of the files. In other words, preloading the animation(s) for faster performance.

I could be wrong here, but that is what I think is happening.

engine/src/universe_util.cpp
stephengtuggy: The original code:

1. Parse the splash screen settings to get the list of possible splash screens to show during startup.
2. Pick one at random.
3. Load, return, and display it.

The new code should do pretty nearly the same thing, except with the addition of a check to see whether the splash screen settings have been loaded from the config files first. So in other words:

1. Check if the splash screen settings have been loaded yet.
2. If they haven't, return without doing anything.
3. If they have been loaded, then parse the splash screen settings to get the list of possible splash screens to show during startup.
4. Pick one at random.
5. Load, return, and display it.

At least that's what I think it will do. Does that make sense?

stephengtuggy: I can double-check to make sure I'm not creating class instances on the stack (except for std::string, perhaps).

As for curAnimation getting leaked -- it's a singleton, for the life of the program. It will get destructed at shutdown, one way or another. So what does it matter?

libraries/cmd/unit_generic.cpp
stephengtuggy: 1. I don't understand what you're referring to. `cache` is static, isn't it? So it isn't created on the stack. Unless I'm missing something.
2. Will do.
3. OK 😄 

Also, yes, that is a lot of checks to make sure we actually have an animation. I will add a TODO, as you requested.

stephengtuggy: https://www.reddit.com/r/cpp/comments/do5v2m/lazy_initialisation_in_c/ is what my googling turned up. Looking at the subreddit discussion again, it sounds like `cache` might or might not be created on the stack. But for large objects, perhaps this optional mechanism is less appropriate, since it is more likely that the large object will be created on the stack.

I will rethink my approach here. You're welcome to chime in as well. I think I'll start a discussion.

engine/src/universe_util.cpp
royfalk: > As for curAnimation getting leaked -- it's a singleton, for the life of the program.

Reading through it again, I see each call first deletes the previous animation.

libraries/cmd/unit_generic.cpp
royfalk: So, I've googled quite a bit and then asked chatgpt and it's neither. It's created in the static area.
Unlike the stack, the static area is quite large and we can probably get away with creating stuff there.
You can mark this resolved.

royfalk:
Please fix the stuff we agree on and open that issue.
I'll do another review with fresh eyes.

libraries/cmd/unit_generic.cpp
stephengtuggy: Okay, thanks.

stephengtuggy:
> Please fix the stuff we agree on and open that issue. I'll do another review with fresh eyes.

Opened the issue as #1227 . (At least, I think that's the right one you were referring to.)

The other items should be fixed, at least for the most part. I'm not sure about the FILTER enum, one way or the other. Perhaps it should be converted to a `class enum`? Or should we leave that till later?

engine/src/cmd/unit_functions.cpp
stephengtuggy: Both functions are now so short that there doesn't seem to be much point combining them. They're like 6 lines each.

engine/src/cmd/unit_functions.cpp
stephengtuggy: Done.

engine/src/cmd/unit_functions.cpp
stephengtuggy: Done.

evertvorster:
What am I supposed to test on this one? The game loads and plays OK with PR. 
There is this one thing, though. There are two animations that usually play. The first one when first opening the game, and then the second one once you load a game or start a new game. 

In git master version of the engine, and this PR, the first animation does not play. Sometimes I get a pure white screen, and on other times I get a pure black screen. 
In either case it is not a big deal, on my computer this one flashes through in a second or so. I know on older, slower machines this first animation did take some time, though.

stephengtuggy:
@evertvorster Try testing with this version of the assets: https://github.com/vegastrike/Assets-Production/pull/195

Just make sure that the general functionality of the game works as expected, without too many unexpected differences in settings or behavior.

stephengtuggy:
That startup animation difference does matter. I noticed it as well. I think I've got it fixed on the assets side, in https://github.com/vegastrike/Assets-Production/pull/195

evertvorster:
~~Hmm, with this PR, my shaders don't look so nice anymore... ~~
~~However, I tried with git master, and it looks the same! ~~

~~So, I don't know what's going on anymore. Will investigate tonight.~~ 

image

I figured out what was going on. When travelling, my computer switches to its iGPU, and that GPU does not render the transparent textures. 
Once I forced to use the dGPU, all the textures are rendered properly.

kheckwrecker:
I'm not sure if this is from this PR, but when I set Geometry to Extreme, I see the following in my terminal window output:

on_geometry details:_change: <kivy.uix.spinner.Spinner object at 0x000001A3E298BAF0> Extreme
Modifying graphics/fog to True
Modifying graphics/model_detail to 10
Modifying graphics/mipmap_detail to 3
Modifying graphics/reflection to True
Modifying graphics/draw_star_body to True
Modifying graphics/draw_star_glow to True
Modifying graphics/high_quality_sprites to True
Modifying graphics/blend_panels to True
Modifying graphics/smooth_lines to True
Modifying graphics/per_pixel_lighting to True
Modifying graphics/specmap_with_reflection to True
Modifying graphics/znear to 8
Modifying graphics/zfar to 65536
Modifying graphics/star_blend to True
Modifying graphics/point_sparkles to False
on_geometry details:_change error. graphics/point_sparkles not in game_config.
Modifying graphics/background to True
Modifying graphics/cockpit to True
Modifying graphics/planet_detail_level to 96
Modifying graphics/jump_detail to 16
on_geometry details:_change error. graphics/jump_detail not in game_config.

A couple of settings -  "point_sparkles" and "jump_detail" show errors.

Otherwise, LGTM.

stephengtuggy:
I think "jump_detail" should be "jump_detail_level". I think that's a problem with the new settings UI. "point_sparkles" is probably a similar situation. I'm not sure what should replace that one though.

kheckwrecker:
> I think "jump_detail" should be "jump_detail_level". I think that's a problem with the new settings UI. "point_sparkles" is probably a similar situation. I'm not sure what should replace that one though.

Looking at vegastrike.config, I found this:

                  
			                <var name="starblend" value="true"/>
			                 <var name="point_sparkles" value="false"/>

and this:

           "

So, the names are the same in vegastrike.config. Any idea why they wouldn't work in the new settings UI?

stephengtuggy:
While migrating the settings, I renamed a few of them where I thought that made sense. `planet_detail` got renamed to `planet_detail_level`, and `jump_detail` to `jump_detail_level`. Not sure about `point_sparkles`; will look into that one some more.

royfalk:
I was aware of this issue (I saw the same errors) but tried to and failed to identify these values. 
Open an issue in Assets Master for this and I'll fix it.

evertvorster:
I figured out my shader issue, and it is not the PR, but the GPU I used to run the game. 

@royfalk Should I make an issue with Assets for the changed variable names, or would you like @stephengtuggy to file the issue?

royfalk:
Feel free to do it and if you figure out what the sparkle thing was, add that too.
Assign me to it and I'll fix it this weekend.

engine/src/gldrv/gl_texture.cpp
BenjamenMeyer: @stephengtuggy typically for static container variables I use a bool to control its initialization state:
```c++
void myFunc(const std::string& key) std::string {
    static bool initialized = false;
    static std::map<std::string, std::string> myMap;
    if (!initialized) {
         #define SET_VALUE(a, b) { myMap[a] = b; }
         SET_VALUE("foo", "bar")
        ...
        #undef SET_VALUE
         initialized = true;
    }
    ...
}
```
Should probably use a lock for multi-threaded code, but the pattern has worked well - and yes, I use C Macros to help configure it; I also keep them scope limited so they're defined and undefined within that block.

engine/src/star_system_jump.cpp
BenjamenMeyer: Remember when we made the animation non-static and the performance ground to a stand still?

I've used the static pattern many times, but I always couple it with a 2nd static boolean variable to track initialization state. probably because I was always building std::maps for various purposes that couldn't rely on the state of the container itself to determine that it needed to be configured; but even then it still seems like a good pattern to use. See https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1224/files#r2216598429 for my example

engine/src/gldrv/gl_texture.cpp
stephengtuggy: I have no idea why you need that C Macro here. Why not just assign the value using the straight map syntax?

I will consider using the `static bool initialized` mechanism.

engine/src/star_system_jump.cpp
stephengtuggy: @BenjamenMeyer Yes, I remember.

I suppose the static boolean variable makes sense. I will give that a try.

engine/src/gldrv/winsys.cpp
```diff
@@ -375,7 +375,7 @@ void winsys_init(int *argc, char **argv, char const *window_title, char const *i
     constexpr Uint32 sdl_flags = SDL_INIT_VIDEO | SDL_INIT_JOYSTICK;
 #endif
 
-    gl_options.fullscreen = configuration()->graphics.full_screen;
+    configuration()->graphics.full_screen = configuration()->graphics.full_screen;
```
stephengtuggy: Say what? This line can be removed

engine/src/main_loop.cpp
```diff
@@ -792,22 +791,25 @@ void createObjects(std::vector<std::string> &fighter0name,
     vector<std::string> fighter0mods;
     vector<int> fighter0indices;
 
-    static Vector TerrainScale(game_options()->xscale, game_options()->yscale, game_options()->zscale);
+    static boost::optional<Vector> TerrainScale;
```
stephengtuggy: Here's another spot where I could use a `static bool initialized = false`

libraries/cmd/unit_generic.cpp
stephengtuggy: I can actually use a `static bool initialized = false` here too.

libraries/cmd/unit_generic.cpp
```diff
@@ -1701,16 +1704,19 @@ bool Unit::Explode(bool drawit, float timeit) {
...
-                    static Animation *__shock__ani = new Animation(shockani.c_str(), true, .1, MIPMAP, false);
+                if (rand() < RAND_MAX * configuration()->graphics.percent_shockwave && (!this->isSubUnit())) {
+                    const std::string shockani(configuration()->graphics.shockwave_animation);
+                    boost::optional<Animation> shock_ani;
```
stephengtuggy: And here.

engine/src/gldrv/gl_texture.cpp
stephengtuggy: In this particular case, since `fil` is an enum rather than a collection, I think I actually need to use `optional`. Unless you guys have another suggestion.

engine/src/gldrv/winsys.cpp
stephengtuggy: Done

engine/src/main_loop.cpp
stephengtuggy: Done

libraries/cmd/unit_generic.cpp
stephengtuggy: Done

libraries/cmd/unit_generic.cpp
stephengtuggy: > I can actually use a `static bool initialized = false` here too.

Done, I believe.

engine/src/star_system_jump.cpp
stephengtuggy: > @BenjamenMeyer Yes, I remember.
> 
> I suppose the static boolean variable makes sense. I will give that a try.

Done.

engine/src/cmd/unit_functions.cpp
```diff
@@ -88,28 +90,20 @@ void update_ani_cache() {
...
-        for (unsigned int i = 0; i < rn; i++) {
-            j++;
-        }
-        return (*j).first;
-    } else {
+    if (cached_ani.empty()) {
         return "";
     }
+    auto it = cached_ani.begin();
```
royfalk, line 96: Nice.

engine/src/gldrv/gl_texture.cpp
```diff
@@ -937,7 +938,12 @@ GFXBOOL /*GFXDRVAPI*/ GFXTransferTexture(unsigned char *buffer,
         internalformat = GetTextureFormat(internformat);
         if (((textures.at(handle).mipmapped & (TRILINEAR | MIPMAP)) && gl_options.mipmap >= 2) || detail_texture) {
             if (detail_texture) {
-                static FILTER fil = game_options()->detail_texture_trilinear ? TRILINEAR : MIPMAP;
+                static FILTER fil;
```
royfalk: I know `static sometype x` is faster than `sometype x` when initializing a complex variable but is it really faster than just doing

`textures.at(handle).mipmapped = configuration()->graphics.detail_texture_filter `

where filter is an int. It maps directly into the `FILTER struct` in `gfxlib_struct.h`.

engine/src/gldrv/winsys.cpp
```diff
@@ -215,14 +215,14 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
-        game_options()->rgb_pixel_format = ((bpp == 16) ? "555" : "888");
+    if (configuration()->graphics.rgb_pixel_format == "undefined") {
+        configuration()->graphics.rgb_pixel_format = ((bpp == 16) ? "555" : "888");
     }
-    if ((game_options()->rgb_pixel_format.length() == 3) && isdigit(game_options()->rgb_pixel_format[0])
-            && isdigit(game_options()->rgb_pixel_format[1]) && isdigit(game_options()->rgb_pixel_format[2])) {
-        rs = game_options()->rgb_pixel_format[0] - '0';
-        gs = game_options()->rgb_pixel_format[1] - '0';
-        bs = game_options()->rgb_pixel_format[2] - '0';
+    if ((configuration()->graphics.rgb_pixel_format.length() == 3) && isdigit(configuration()->graphics.rgb_pixel_format[0])
```
royfalk, line 221: I think we can deprecate this whole code. This isn't the 90's and we're not supporting 16 bit colors.

engine/src/main_loop.cpp
```diff
@@ -792,22 +791,27 @@ void createObjects(std::vector<std::string> &fighter0name,
     vector<std::string> fighter0mods;
     vector<int> fighter0indices;
 
-    static Vector TerrainScale(game_options()->xscale, game_options()->yscale, game_options()->zscale);
+    static Vector * terrain_scale;
```
royfalk: Why is this code better? Vector isn't a heavy class. It's really three floats.

engine/src/resource/random_utils.cpp
```diff
@@ -42,6 +42,6 @@ double randomDouble() {
     std::random_device dev;
     std::mt19937 rng(dev());
     std::uniform_int_distribution<std::mt19937::result_type> int_dist(0,precision);
-    int random_int = int_dist(rng);
-    return (double)random_int/precision;
-}
\ No newline at end of file
+    const unsigned int random_int = int_dist(rng);
```
royfalk, line 45: I'm pretty sure that with `precision = 10000`, you can't get negative `random_int`.

engine/src/star_system_jump.cpp
```diff
@@ -47,10 +47,15 @@
 #include "root_generic/options.h"
 
 void CacheJumpStar(bool destroy) {
-    static Animation *cachedani = new Animation(game_options()->jumpgate.c_str(), true, .1, MIPMAP, false);
+    static Animation* cached_ani;
```
royfalk: I'm pretty sure this is slower, so why is it better?
Looking through the code, we only call `CacheJumpStar` once, in `Universe::Init`.
It does not look as if we are ever getting to the cleanup section.

Looking further, `cachedani` isn't accessible. I wonder if `Animation` is doing something weird inside or we can just assume they wrote something sane for once and just delete the whole thing. 

If deleting this breaks something, we can just delete the function and move 

`static Animation *cachedani = new Animation(game_options()->jumpgate.c_str(), true, .1, MIPMAP, false);`

to Universe::Init.

engine/src/universe_util.cpp
```diff
@@ -185,17 +185,20 @@ void saveGame(const string &savename) {
 }
 
 void showSplashScreen(const string &filename) {
```
royfalk, line 187: I spent a bit of time looking at this and `showSplashScreen` is called several times:
1. Twice in `main.cpp` with `filename=""`.
2. Once in `universe.cpp` with `filename=""`.
3. Once in `basecomputer.cpp` with `filename=""`.
4. Once in activateProgressScreen in showProgress.py with some data from VS.vsConfig.

The last one will almost surely break because it's reading from the old config with 

`return vs_config->getVariable(category, option, def);`

It is also called several times in seven places with a non-empty file.

engine/src/universe_util.cpp
```diff
@@ -185,17 +185,20 @@ void saveGame(const string &savename) {
...
+    static Animation *curSplash = nullptr;
     if (!filename.empty()) {
         if (curSplash != nullptr) {
             delete curSplash;
             curSplash = nullptr;
         }
-        curSplash = new Animation(filename.c_str(), 0);
+        curSplash = new Animation(filename.c_str(), false);
     } else if (!curSplash && !GetSplashScreen()) {
-        static std::vector<std::string> s = ParseDestinations(game_options()->splash_screen);
```
royfalk, line 196: We can replace parse destination with an array in the JSON.

libraries/cmd/unit_generic.cpp
```diff
@@ -1664,13 +1664,18 @@ bool Unit::Explode(bool drawit, float timeit) {
             FactionUtil::GetRandExplosionAnimation(this->faction, bleh);
         }
         if (bleh.empty()) {
-            static Animation cache(game_options()->explosion_animation.c_str(), false, .1, BILINEAR, false);
+            static Animation * cache;
```
royfalk: I'd drop this code and see if the game still works. 
This whole philosophy of a resilient engine in the face of deficient assets is silly.
If we force Unit.pImage.explosion_type to have a valid value every time, we can drop all of this code:

```
        if (bleh.empty()) {
            FactionUtil::GetRandExplosionAnimation(this->faction, bleh);
        }
        if (bleh.empty()) {
            static Animation cache(game_options()->explosion_animation.c_str(), false, .1, BILINEAR, false);
            bleh = getRandomCachedAniString();
            if (bleh.size() == 0) {
                bleh = game_options()->explosion_animation;
            }
        }
```

royfalk (approved):
I've marked several places that need refactoring. I'll open issues for them.
Just be aware of `showSplashScreen` and the fact it's a bit of a ticking time bomb.
Just take another look at main_loop and figure out if it's really worth it.

Ship it.

engine/src/main_loop.cpp
stephengtuggy: I made that change to avoid using `optional<Vector>`. `terrain_scale` needs to be `static`, and it needs to be initialized exactly once, after the game config is loaded. This implies that it needs to start out uninitialized. Two options for that: either it can be a pointer, with a value of `nullptr`, or a `boost::optional`. Take your pick.

engine/src/gldrv/winsys.cpp
stephengtuggy: Maybe later. I don't see that as a high priority right now.

engine/src/gldrv/gl_texture.cpp
stephengtuggy: Good point! I have made that change.

engine/src/main_loop.cpp
stephengtuggy: I suppose there are other options as well, but those seem like the two best options. Feel free to suggest something else if you like.

engine/src/star_system_jump.cpp
stephengtuggy: Again, I made these changes to allow `cachedani` / `cached_ani` to be static, and initialized only once, after the game config is loaded. (I also renamed the variable to be easier to read.)

I think `Animation` must be doing something weird inside. At any rate, I'm not prepared to delete this whole section in this PR. Maybe later.

stephengtuggy: I don't see how is this more than marginally slower than the old way. Can you elaborate? Evaluating bools should go pretty quickly, seems to me.

libraries/cmd/unit_generic.cpp
stephengtuggy: Let's check the assets for deficiencies first, and make the game engine less resilient later. At least, that would be my preference. We already know that both the assets and the game engine have deficiencies.

engine/src/gldrv/gl_texture.cpp
stephengtuggy: I ended up inlining `fil` and getting rid of it entirely, at Roy's suggestion. Let's see if that works. If it compiles, it's probably good.

engine/src/universe_util.cpp
stephengtuggy: Good catch. We will definitely want to fix the reference to vs_config.

engine/src/gldrv/winsys.cpp
royfalk: Opened #1240.

royfalk:
Well, I've approved already, so you can merge.

engine/src/gldrv/gl_texture.cpp
BenjamenMeyer: > I have no idea why you need that C Macro here. Why not just assign the value using the straight map syntax?

I simplified it because I can't remember off hand all the syntax. IIRC there was something about creating an std::pair and then doing the insertion with that or something - it wasn't something quite so trivial so the C Macro would create a code block, its own variables, and then do the work. C++ inlines don't work well for that particular situation, though now with lambdas may be that would work better.

engine/src/main_loop.cpp
BenjamenMeyer: please ensure this is set to `nullptr`; even though we're using `initialized` too.

engine/src/star_system_jump.cpp
BenjamenMeyer: same here

libraries/cmd/unit_generic.cpp
BenjamenMeyer: and here

BenjamenMeyer (changes requested):
one minor change - let's make sure those static pointers are set to nullptr when they're created before we check `initialized`. It should get quickly updated but this would be helpful for tracking some potential bugs.

engine/src/main_loop.cpp
stephengtuggy: Done

engine/src/star_system_jump.cpp
stephengtuggy: Done

libraries/cmd/unit_generic.cpp
stephengtuggy: Done

libraries/cmd/unit_generic.cpp
```diff
@@ -1701,16 +1706,21 @@ bool Unit::Explode(bool drawit, float timeit) {
         if (!sub) {
             un = _Universe->AccessCockpit()->GetParent();
             if (this->getUnitType() == Vega_UnitType::unit) {
-                if (rand() < RAND_MAX * game_options()->percent_shockwave && (!this->isSubUnit())) {
-                    static string shockani(game_options()->shockwave_animation);
-                    static Animation *__shock__ani = new Animation(shockani.c_str(), true, .1, MIPMAP, false);
+                if (rand() < RAND_MAX * configuration()->graphics.percent_shockwave && (!this->isSubUnit())) {
+                    const std::string shockani(configuration()->graphics.shockwave_animation);
+                    static Animation * shock_ani;
```
BenjamenMeyer: this should be set to nullptr

BenjamenMeyer (changes requested):
one more I think

libraries/cmd/unit_generic.cpp
stephengtuggy: Sure enough. Fixed.

BenjamenMeyer (approved):

PR #1228, opened 2025-06-27
Windows-CI.yml, Windows-Release.yml: Remove Windows 2019…

stephengtuggy:
… since GitHub Actions will no longer support it in a few days

Code Changes:
- [x] This is a CI / build system change only (well, more or less)

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Remove Windows 2019 from the set of operating systems we build on, since GitHub Actions will no longer support it in a few days
- What release is this for? All of them, going forward
- Is there a project or milestone we should apply this to? 0.10.x?

### DIFF
--- a/.github/workflows/Windows-CI.yml
+++ b/.github/workflows/Windows-CI.yml
@@ -19,14 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os:              windows-2019
-            triplet:         x64-win10
-            preset-name:     VS2019Win64-pie-enabled-release
-            cmake-generator: VS2019Win64
-            enable-pie:      1
-            build-type:      Release
-            python-version:  3.12.10
-            is-release:      0
           - os:              windows-2022
             triplet:         x64-win10
             preset-name:     VS2022Win64-pie-enabled-debug
--- a/.github/workflows/Windows-Release.yml
+++ b/.github/workflows/Windows-Release.yml
@@ -21,14 +21,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os:              windows-2019
-            triplet:         x64-win10
-            preset-name:     VS2019Win64-pie-enabled-release
-            cmake-generator: VS2019Win64
-            enable-pie:      1
-            build-type:      RelWithDebInfo
-            python-version:  3.12.10
-            is-release:      1
           - os:              windows-2022
             triplet:         x64-win10
             preset-name:     VS2022Win64-pie-enabled-debug


royfalk (commented):
I'm sorry but I'm not following what it is you're doing here.
This is because I've rarely looked into this area of the code and when I did, it was to solve very specific issues.

You can either wait for @BenjamenMeyer to review or approve yourself. I'm fine with either.

stephengtuggy:
> I'm sorry but I'm not following what it is you're doing here. This is because I've rarely looked into this area of the code and when I did, it was to solve very specific issues.
> 
> You can either wait for @BenjamenMeyer to review or approve yourself. I'm fine with either.

OK, fair enough. I'll wait for @BenjamenMeyer then.

script/build
```diff
@@ -51,32 +74,25 @@ BIN_DIR="${ROOT_DIR}/bin"
 COMMAND=""
 
 # -p creates if the target doesn't exist, noop otherwise
-mkdir -pv "${BUILD_DIR}" && cd "${BUILD_DIR}"
```
BenjamenMeyer, line 54: we still need to move into the directory

.github/workflows/gh-actions-release.yml
```diff
@@ -22,34 +22,28 @@ jobs:
...
             COMPILER: 'clang'
-            OpenGL_GL_PREFERENCE: 'GLVND'
-            ENABLE_PIE: 'OFF'
+            preset-name: 'linux-unix-makefiles-pie-enabled-glvnd-RelWithDebInfo'
             ARTIFACT_EXT: 'deb'
           - FROM:     'debian:bookworm'
             COMPILER: 'clang'
-            OpenGL_GL_PREFERENCE: 'GLVND'
-            ENABLE_PIE: 'OFF'
+            preset-name: 'linux-unix-makefiles-pie-enabled-glvnd-RelWithDebInfo'
```
BenjamenMeyer: are we no longer building with PIE disabled? I see we have them in the Presets but all the ones that are marked with PIE disabled here seem to now be getting PIE enabled instead.

.github/workflows/gh-actions-pr.yml
```diff
@@ -18,60 +18,28 @@ jobs:
...
             COMPILER: 'clang'
-            OpenGL_GL_PREFERENCE: 'GLVND'
-            ENABLE_PIE: 'OFF'
+            preset-name: 'linux-unix-makefiles-pie-enabled-glvnd-RelWithDebInfo'
             allow_failure: false
           - FROM:     'debian:bookworm'
             COMPILER: 'clang'
-            OpenGL_GL_PREFERENCE: 'GLVND'
-            ENABLE_PIE: 'OFF'
+            preset-name: 'linux-unix-makefiles-pie-enabled-glvnd-RelWithDebInfo'
```
BenjamenMeyer: are we no longer building with PIE disabled? I see we have them in the Presets but all the ones that are marked with PIE disabled here seem to now be getting PIE enabled instead.

BenjamenMeyer (changes requested):
1. fixing the build directory so we're doing out-of-tree builds properly
2. builds with PIE disabled

script/build
stephengtuggy: That is handled by the presets now. No need to do it explicitly in the script -- at least not when a CMake Preset is in use.

.github/workflows/gh-actions-release.yml
stephengtuggy: Good catch. Fixed that

.github/workflows/gh-actions-pr.yml
stephengtuggy: This should now be fixed also

stephengtuggy:
It's worth noting that this PR in its current form will require specifying a preset any time someone builds the project using the build scripts. Is that what we want? Or should we leave the old way of doing things as an alternative?

script/build
BenjamenMeyer: Okay - as long as it does not cause an issue for the other scripts for packaging. I see it's using:
```
$(sourceDir)/build/$(presetName)
```
for the binaryDir path. As long as we're using `cpack` that probably isn't an issue.

Still learning the CMake Presets functionality. Overall it's probably a good improvement, just takes some getting use to.

.github/workflows/gh-actions-release.yml
BenjamenMeyer: Thanks!

.github/workflows/gh-actions-pr.yml
BenjamenMeyer: Thanks, confirmed

BenjamenMeyer:
> It's worth noting that this PR in its current form will require specifying a preset any time someone builds the project using the build scripts. Is that what we want? Or should we leave the old way of doing things as an alternative?

As long as its consistent.

I generally just do:
```bash
$ mkdir build
$ cd build
$ cmake ..
$ make -j 10
```

So I use a lot of the defaults. So long as that works reasonably well - since that's a basic flow for most projects - then I'm okay with our CI system and scripts presenting a standard for how they should work as long as they use the common tools on each platform which I think this is helping us get to.

stephengtuggy:
> > It's worth noting that this PR in its current form will require specifying a preset any time someone builds the project using the build scripts. Is that what we want? Or should we leave the old way of doing things as an alternative?
> 
> As long as its consistent.
> 
> I generally just do:
> 
> ```shell
> $ mkdir build
> $ cd build
> $ cmake ..
> $ make -j 10
> ```

Yes, the above should still work fine.

> So I use a lot of the defaults. So long as that works reasonably well - since that's a basic flow for most projects - then I'm okay with our CI system and scripts presenting a standard for how they should work as long as they use the common tools on each platform which I think this is helping us get to.

Yep. Okay, sounds good.

Issue #1231, opened 2025-06-29
Segfault trying to tractor in some cargo


stephengtuggy:
Just now, I got a segfault (SIGSEGV) while trying to tractor in some cargo floating in space into my upgraded starting Llama.

stephengtuggy:
It's reproducible; I have now gotten that same segfault several times when tractoring cargo, and once in another part of the code.

PR #1232, opened 2025-06-30
Convert Physics Settings

### LINKED ISSUE #1200: Migrate `physics` settings section to new mechanism

### LINKED ISSUE #1246: Game keeps freezing up for 89 seconds at a time
This happens at different times, apparently in different parts of the code. I haven't quite finished figuring out the cause(s) yet.

stephengtuggy:
Code Changes:

Issues Addressed:
- Closes #1200 
- Closes #1246 

Known Issues Remaining:
- Still getting a crash when trying to tractor in some cargo (#1231 ), except that now it seems to be an invalid cast exception, rather than a segfault. At any rate, this isn't new.
- ~The game is freezing up at times, seemingly at random, for up to a minute or more (#1246 ).~ I think I have fixed this.

Purpose:
- What is this pull request trying to do? Migrate the `physics` settings to the latest configuration mechanism, `configuration()->physics.` ...

### DIFF
--- a/engine/src/cmd/unit_functions.cpp
+++ b/engine/src/cmd/unit_functions.cpp
@@ -94,7 +94,7 @@ std::string getRandomCachedAniString() {
         return "";
     }
     auto it = cached_ani.begin();
-    std::advance(it, randomInt(cached_ani.size()));
+    std::advance(it, randomInt(cached_ani.size() - 1));
     return it->first;
 }
 
@@ -103,7 +103,7 @@ Animation *getRandomCachedAni() {
         return nullptr;
     }
     auto it = cached_ani.begin();
-    std::advance(it, randomInt(cached_ani.size()));
+    std::advance(it, randomInt(cached_ani.size() - 1));
     return it->second;
 }
 


royfalk:
Should I review or hold off for now?

stephengtuggy:
> Should I review or hold off for now?

Probably hold off for now. Thanks

stephengtuggy:
@royfalk Feel free to review this now. Thanks for your patience.

engine/src/star_system.cpp
```diff
@@ -902,14 +902,20 @@ void StarSystem::RequestPhysics(Unit *un, unsigned int queue) {
 //randomization on priority changes, so we're fine.
 void StarSystem::UpdateUnitsPhysics(bool firstframe) {
     static int batchcount = SIM_QUEUE_SIZE - 1;
-    double collidetime = 0.0;
-    double bolttime = 0.0;
+#if defined(LOG_TIME_TAKEN_DETAILS)
+    double collide_time = 0.0;
+    double bolt_time = 0.0;
+    double time_on_two_arg_UpdateUnitPhysics = 0.0;
+#endif
```
BenjamenMeyer, line 909: could we make this a slightly more generic C++ object that has 1 of 2 implementations?
(a) actual implementation when we want to log stuff
(b) NOP implementation when we don't

This would clean up the code a little and give us a tool for this across debug tasks.
Just a thought, not going to block on it.

BenjamenMeyer (approved):

engine/src/star_system.cpp
stephengtuggy: I'm not sure I understand what you mean. Can you give an example?

BenjamenMeyer: @stephengtuggy define an abstract class with an interface that does the tracking, have one implementation that actually tracks and records it which is linked when this is enabled and have another implementation that just does NOP and gets optimized out when its not enabled.

stephengtuggy: Seems like that would introduce a lot of unnecessary boilerplate code. And the tracking class would know too much about the innards of the class it was tracking.

Another possibility I have considered is to use the [C++ OpenTelemetry library](https://opentelemetry.io/docs/languages/cpp/) to measure the performance of the code. It would replace all these `#ifdef`s, and would do so in an industry-standard way that is suited for just this type of job. What do you think of that option?

BenjamenMeyer: @stephengtuggy that's a fair option as well assuming we can capture the output correctly. OpenTelemetry gets used a lot for enterprise stuff; I know it's easy to shunt to things to do metrics, etc. If we can make it work without having to have all the infrastructure for Grafana that'd be an awesome solution too.

> Seems like that would introduce a lot of unnecessary boilerplate code. And the tracking class would know too much about the innards of the class it was tracking.

There might be; I was hoping it could be abstract enough that the tracking class wouldn't know about the details - perhaps something as simple as registering some variables in a std::map can having a time tracker. May be as a template class if necessary. That's where I was thinking at least.

The OpenTelemetry solution might be better if we could make it work. Haven't tried doing that without having the normal enterprise-type infrastructure.

royfalk: For my sins, I also play modern games. These come with fps, which can be displayed by apps such as Nvidia's GeForce Experience or AMD's Radeon Software. They also track temperature, CPU/GPU usage, low 1%, etc.

We should definitely consider dropping this in favor of that. In fact, I'll test it and see if it works.

**EDIT:** mangohud didn't work. I'll edit the open issue with my suggestions.

PR #1233, opened 2025-07-01
Add support in Cargo class for the following:

royfalk:
- component
- weapon
- passenger
- slave

Also:
- Improve Cargo class serialization and de-serialization from string.
- Refactor Enslave function
- Call GenerateHudText after repair. Fix bug where in flight status of components did not reflect repairs.
- Replace hold/upgrade space/hidden upgrade code with a dedicated function - UpAndDownCargoHoldAndUpgradeSpace
- Delete unused upgrade code dealing with Warp_Min_Multiplier, Warp_Max_Multiplier
- Remove redundant MyPercentMin
- Remove legacy upgrade code handling shields at end of Unit::RepairUpgrade
- Remove macros STDUPGRADECLAMP, STDUPGRADE, STDUPGRADE_SPECIFY_DEFAULTS
- Remove upgrade needs_redemption. No longer used. Remove IsWeapons function
- Remove *Serializer functions - not called

This PR is matched by https://github.com/vegastrike/Assets-Production/pull/198.

**Testing**
    
Also tested the following:
- Enslave/Free functions. To test in gameplay, add the following to vegastrike.config at around line 280:
```
<bind key="f" modifier="ctrl" command="FreeSlaves"/>
    <bind key="e" modifier="ctrl" command="Enslave"/>
```
- Enslave/Free functions unit tests.
- Tested and fixed all issues with cargo/upgrade space upgrades. Did not test hidden cargo as it is not supported in VS, but it should work.
- Tinkered with upgrading and selling upgrades.
- Tinkered with repairing items and checking hud to see it reflected.

Issues:
- Halos from engine exhaust see blocky and low res. Probably something to do with the settings I've changed for the settings app.
- Weird purple lines in asterisk form across the screen during flight. Unlikely to be related to this PR, as gfx code not touched.


engine/CMakeLists.txt
```diff
@@ -1220,6 +1220,7 @@ IF (USE_GTEST)
         src/configuration/tests/configuration_tests.cpp
         src/damage/tests/layer_tests.cpp
         src/damage/tests/object_tests.cpp
+        src/resource/tests/cargo_tests.cpp
```
BenjamenMeyer, line 1228: :100:

engine/src/resource/tests/cargo_tests.cpp
```diff
@@ -0,0 +1,176 @@
...
...;1.000000;1.000000;;false;true|mult_overdrive06;upgrades/Overdrive;120685.796875;1;18.000000;18.000000;1.000000;1.000000;;false;true|reactor08;upgrades/Reactors/Standard;60342.917969;1;8.000000;100.000000;1.000000;1.000000;;false;true|add_spec_capacitor02;upgrades/SPEC_Capacitors;16971.439453;1;10.000000;6.000000;1.000000;1.000000;;false;true|hawkeye4;upgrades/Sensors/Confed;150857.171875;1;0.010000;21.000000;1.000000;1.000000;;false;true|quadshield10;upgrades/Shield_Systems/Standard_Quad_Shields;241371.500000;1;10.000000;61.000000;1.000000;1.000000;;false;true|afterburner;upgrades/integral;2000.000000;1;0.100000;0.100000;1.000000;1.000000;;false;true|drive;upgrades/integral;6000.000000;1;0.100000;0.100000;1.000000;1.000000;;false;true|ftl_drive;upgrades/integral;4500.000000;1;0.100000;0.100000;1.000000;1.000000;;false;true|hull;upgrades/integral;12000.000000;1;0.100000;0.100000;1.000000;1.000000;;false;true|Slaves;Contraband;100;17;32;23;1;1;slaves to sell;false;false;false;false;false;true;true|Hitchhiker;Passengers;50;27;12;17;1;1;some people;false;false;false;false;false;true;false|passengers;Passengers;500;7;10;27;1;1;some more people;false;false;false;false;false;true;false";
```
BenjamenMeyer, line 87: would probably be better to have something that could programmatically make this or at least validate that its correct so we can find bugs quickly when these strings change

stephengtuggy:
Can we merge in the latest changes from the `master` branch, so that we aren't trying to build on Windows 2019 still?

engine/src/resource/tests/cargo_tests.cpp
royfalk: This is legacy code. It's present in two places:
1. When saving a game.
2. In the JSON files themselves.

The right fix would be to migrate the whole thing into proper JSON. However, because right now we are reading the JSON files into an `std::map<std::string, std::string>`, we have no support for nested JSON or anything other than string.

BenjamenMeyer: Understood. I do think we should aim for that proper fix of in the long run. For now, I just want to make sure that we can easily find issues and understand them when something breaks.

stephengtuggy:
I see the CI passed this time. Much better.

engine/src/resource/cargo.cpp
```diff
@@ -28,6 +28,20 @@
...
+
+#include <numeric>
+#include <iostream>
+#include <boost/format.hpp>
+#include <boost/algorithm/string/split.hpp>
+#include <boost/algorithm/string/classification.hpp>
+
+
+// A simple utility function to parse a boolean value
+static bool _parse_bool(const std::string& value) {
```
stephengtuggy, line 41: There's a function like this in the `XMLSupport` namespace (`parse_bool`). Could you use that here?

engine/src/python/unit_wrapper.cpp
```diff
@@ -119,7 +119,7 @@ BOOST_PYTHON_MODULE(VS) {
         PYTHON_DEFINE_METHOD(Class, &Cargo::SetCategory, "SetCategory");
         PYTHON_DEFINE_METHOD(Class, &Cargo::GetCategoryPython, "GetCategory");
         PYTHON_DEFINE_METHOD(Class, &Cargo::SetMissionFlag, "SetMissionFlag");
-        PYTHON_DEFINE_METHOD(Class, &Cargo::GetMissionFlag, "GetMissionFlag");
+        PYTHON_DEFINE_METHOD(Class, &Cargo::IsMissionFlag, "GetMissionFlag");
```
stephengtuggy, line 122: Cool.

libraries/cmd/unit_csv.cpp
```diff
@@ -199,26 +199,7 @@ static string nextElement(string &inp) {
     return ret;
 }
 
-static bool stob(const string &inp, bool defaul) {
```
stephengtuggy, line 202: Nice that you were able to get rid of these.

libraries/cmd/unit_generic.cpp
```diff
@@ -2687,71 +2687,41 @@ static bool cell_has_recursive_data(const string &name, unsigned int fac, const
     return retval;
 }
 
-#define STDUPGRADE_SPECIFY_DEFAULTS(my, oth, temp, noth, dgradelimer, dgradelimerdefault, clamp, value_to_lookat) \
```
stephengtuggy, line 2690: Nice! We did away with a big, ugly macro!

stephengtuggy (approved):
Approved! Looks good to me!

PR #1234, opened 2025-07-10
Task cargo 2

royfalk:
This is a major refactor of the cargo subsystem. Ideally, I'd like to extract as much code as I can into smaller PRs before submitting this.

Here are the major changes:
- Add the class CargoHold in components. This isn't a true component, as it can't be upgraded (yet) or sold. However, it provides an implementation (abstraction?) for the concept of holding cargo items.
- Split the current cargo items in carrier into three categories - cargo hold, upgrade space and hidden hold. This means a lot of the cargo code that iterates over the vector to filter out upgrades or cargo items goes away.
- Provides an API for cargo subsystem. This allows for cleaner code in engine. See comm_ai.cpp:188 for resulting refactored code.
- Credits and mass are now part of ComponentsManager. This allows us to handle sale/purchase/upgrade in an encapsulated and clean way.
- Add additional constructors for Cargo - JSON and std::string. 
- Manifest is superclass of CargoHold. Now is a read-only class for use as manifest and read-write functions are in CargoHold.

Some additional minor changes:
- ComponentsManager has ShipDamaged and AllowedUpgrade.
- Remove incrementCargo/decrementCargo. Should use proper AddCargo, ForceAddCargo or RemoveCargo.
- Merge Product into Cargo. 
- Delete Store.

Code Changes:
- [ ] Have the PR Validation Tests been run? Not yet.

Issues:
- ~~Mass is probably not yet taken from ComponentsManager but from Movable.~~ **Done**
- basecomputer should mostly go away after this PR. GUI code to go to python. A lot of the current code is about generating a list of items from the relevant list of items. 
- ~~pilot.cpp and pilot_server.cpp has identical functions. Probably need to merge this somehow later.~~ **Done**
- script_call_unit_generic should be refactored.


engine/src/components/components_manager.cpp
```diff
@@ -105,12 +164,50 @@ void ComponentsManager::DamageRandomSystem() {
...
+bool ComponentsManager::AllowedUpgrade(const Cargo& upgrade) const {
+    for(const std::pair<const std::string, const int> prohibited_upgrade : prohibited_upgrades) {
+        if(prohibited_upgrade.first != upgrade.GetCategory()) {
+            continue;
+        }
+
+        Manifest category_upgrades = upgrade_space.GetCategoryManifest(prohibited_upgrade.first);
+
+        if(category_upgrades.Size() > prohibited_upgrade.second) {
+            std::cerr << "ComponentsManager::AllowedUpgrade: " << category_upgrades.Size() <<
```
BenjamenMeyer: this should go to the log except under rare circumstances

stephengtuggy:
Can we merge `master` into this as well?

engine/src/components/components_manager.cpp
royfalk: This is one of those rare circumstances. That's why the next line is `assert(0);`
It means that somehow a ship has an upgrade in the prohibited list or that it has more of the upgrade than it is allowed.

stephengtuggy: Then please use `VS_LOG_FLUSH_EXIT`. It's intended for exactly this type of situation (fatal errors that should cause the game to exit immediately).

BenjamenMeyer: @royfalk understood; let's use `VS_LOG_FLUSH_EXIT` as @stephengtuggy suggested. Just using `assert()` could mean loss of log data and loss of context.

royfalk:
The PR is ready. 
I'll test it today.

evertvorster:
There are two recent bugs in git master related to cargo: #1268 and #1267 
Testing whether this PR breaks anything else is not feasible while those bugs are open, as they directly influence how the game deals with cargo.

engine/src/components/components_manager.cpp
```diff
@@ -185,16 +186,16 @@ bool ComponentsManager::ShipDamaged() const {
  *  The previous implementation also worked by specific upgrades.
  */
 bool ComponentsManager::AllowedUpgrade(const Cargo& upgrade) const {
-    for(const std::pair<const std::string, const int> prohibited_upgrade : prohibited_upgrades) {
+    for(const std::pair<const std::string, const int>& prohibited_upgrade : prohibited_upgrades) {
```
stephengtuggy, line 189: Nice

engine/src/components/components_manager.cpp
```diff
@@ -185,16 +186,16 @@ bool ComponentsManager::ShipDamaged() const {
...
         }
 
         Manifest category_upgrades = upgrade_space.GetCategoryManifest(prohibited_upgrade.first);
 
         if(category_upgrades.Size() > prohibited_upgrade.second) {
-            std::cerr << "ComponentsManager::AllowedUpgrade: " << category_upgrades.Size() << 
-                      " is greater than " << prohibited_upgrade.second << std::endl << std::flush;
+            VS_LOG_FLUSH_EXIT(fatal,(boost::format("ComponentsManager::AllowedUpgrade: %1% is greater than %2%\n") % category_upgrades.Size() 
+                      % prohibited_upgrade.second), EXIT_FAILURE);
             assert(0);
```
stephengtuggy: This line is now unreachable code. That's okay, though; we can come back and fix this in another PR.

Issue #1236, opened 2025-07-12
Impossible to switch to antialiased fonts


evertvorster:
Hi there!
This is on assets and engine master. 

When launching vegastrike, the fonts are all bitmapped. I have set high_quality fonts to false. 
There is no ~/.vegastrike/vegastrike.config. 

This is the vegastrike.config that Assets installed, with the only edit to disable high quality fonts, and enable AA fonts:
[vegastrike.config.zip](https://github.com/user-attachments/files/21192336/vegastrike.config.zip)

This is my config.json in ~/.vegastrike:
[config.json](https://github.com/user-attachments/files/21192339/config.json)

When launching from the terminal, I see this error message that might point to the issue:
```
load_config: Exception loading config: 'value is not boolean [boost.json:29 at /usr/include/boost/json/detail/value_to.hpp:155 in function 'value_to_impl']'
```

Here is the full output of my console. It is rather lengthy:
[console.txt](https://github.com/user-attachments/files/21192345/console.txt)

royfalk:
> Sigh. That took way longer than expected to debug

The issue isn't with antialiased fonts. It's with this line:

`"force_lighting": false,`

In the old config, `0` was treated as `false` and `1` as `true`. However, the new config is in JSON format which uses actual boolean values true/false only. You can see that the original JSON has the value as true.

Did you edit the value manually based on vegastrike.config or is this an issue in the new settings app?

evertvorster:
I did not edit the config.json manually, instead letting the config tool do it. 
So, this might be an issue in the config app?

evertvorster:
Hmm, did a little troubleshooting. 
When opening the settings app, force_lighting was a text field. 
Removed config.json out of my ~/.vegastrike directory. 
Re-ran the config app, and now force_lighting is a tick box. 

Also, settings now work as expected. 
This may have been set with the many different versions we had of the settings app recently. 

Thanks for looking into this. 

This may be an opportunity to build in some robustness into the settings app, and have it verify that the settings it reads from config.json in the user directory are valid options? This may head off similar issues in the future.

royfalk:
> This may be an opportunity to build in some robustness into the settings app, and have it verify that the settings it reads from config.json in the user directory are valid options? This may head off similar issues in the future.

You're right, but the question is how. 
Originally, I planned to create a new template file that specifies type, min value, max value, etc. But doing this for the hundreds or thousands of entries is hard.

[...]

PR #1247, opened 2025-07-30
Convert joystick settings to latest configuration mechanism

### LINKED ISSUE #1164: "reverse_mouse_spr": false limits mouse cursor displayed to horizontal movement only.
This setting is now in config.json, and when it is set to true, the mouse cursor moves normally across all the screen in a vertically inverted way when piloting your ship. 

However, I don't like to play in this inverted way, and find it more natural to have it act like a normal mouse cursor. 

When setting this reverse_mouse_spr to false the displayed cursor in your HUD seems to be locked to the horizontal plane. The ship does indeed move up or down in the expected way, this seems to be a display issue of the mouse cursor only. 

This issue is present on the git master, and so affects every PR that hangs off of it.

### LINKED ISSUE #1196: Migrate `joystick` settings section to new mechanism

### LINKED ISSUE #1221: Mouse cursor missing in flight
Hi there!

I'm trying to fly with the mouse on my computer, and no matter which settings I try, the mouse cursor seems to be missing from the HUD. 

This is just an initial stab in the dark, just to find out if this has been removed, or whether this issue needs further investigation.

stephengtuggy:
Code Changes:

Issues Addressed:
- Closes #1196 
- May fix #1221 
- May fix #1164 

Purpose:
- What is this pull request trying to do? Migrate the settings in the `"Joystick"` setting to the latest configuration settings mechanism

### DIFF
--- a/engine/src/cmd/ai/flyjoystick.cpp
+++ b/engine/src/cmd/ai/flyjoystick.cpp
@@ -84,25 +84,17 @@ void FlyByJoystick::JDecelKey( KBSTATE k, float, float, int )
 #endif
 
 void FlyByJoystick::Execute() {
-    static bool clamp_joystick_axes = XMLSupport::parse_bool(vs_config->getVariable("joystick", "clamp_axes", "true"));
-    static bool nonlinear_throttle_nav =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "nonlinear_throttle_nav", "true"));
-    static bool nonlinear_throttle_combat =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "nonlinear_throttle_combat", "false"));
-    static float
-            expfactorn = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expfactor_nav", "6.0"));
-    static float pfactorn = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pfactor_nav", "2.0"));
-    static float
-            expamountn = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expamount_nav", "1.0"));
-    static float pamountn = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pamount_nav", "0.0"));
-    static float expfactorc =
-            XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expfactor_combat", "6.0"));
-    static float
-            pfactorc = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pfactor_combat", "2.0"));
-    static float expamountc =
-            XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expamount_combat", "1.0"));
-    static float
-            pamountc = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pamount_combat", "0.0"));
+    const bool clamp_joystick_axes = configuration()->joystick.clamp_axes;
+    const bool nonlinear_throttle_nav = configuration()->joystick.nonlinear_throttle_nav;
+    const bool nonlinear_throttle_combat = configuration()->joystick.nonlinear_throttle_combat;
+    const float expfactorn = configuration()->joystick.nonlinear_expfactor_nav;
+    const float pfactorn = configuration()->joystick.nonlinear_pfactor_nav;
+    const float expamountn = configuration()->joystick.nonlinear_expamount_nav;
+    const float pamountn = configuration()->joystick.nonlinear_pamount_nav;
+    const float expfactorc = configuration()->joystick.nonlinear_expfactor_combat;
+    const float pfactorc = configuration()->joystick.nonlinear_pfactor_combat;
+    const float expamountc = configuration()->joystick.nonlinear_expamount_combat;
+    const float pamountc = configuration()->joystick.nonlinear_pamount_combat;
     desired_ang_velocity = Vector(0, 0, 0);
     for (unsigned int i = 0; i < this->whichjoystick.size(); i++) {
         int which_joystick = this->whichjoystick[i];
--- a/engine/src/cmd/base_interface.cpp
+++ b/engine/src/cmd/base_interface.cpp
@@ -73,9 +73,8 @@ static unsigned int &getMouseButtonMask() {
 
 static void biModifyMouseSensitivity(int &x, int &y, bool invert) {
     int xrez = configuration()->graphics.resolution_x;
-    static int
-            whentodouble = XMLSupport::parse_int(vs_config->getVariable("joystick", "double_mouse_position", "1280"));
-    static float factor = XMLSupport::parse_float(vs_config->getVariable("joystick", "double_mouse_factor", "2"));
+    const int whentodouble = configuration()->joystick.double_mouse_position;
+    const float factor = configuration()->joystick.double_mouse_factor;
     if (xrez >= whentodouble) {
         x -= configuration()->graphics.resolution_x / 2;
         y -= configuration()->graphics.resolution_y / 2;
--- a/engine/src/cmd/ai/flykeyboard.cpp
+++ b/engine/src/cmd/ai/flykeyboard.cpp
@@ -140,7 +140,7 @@ FlyByKeyboard::FlyByKeyboard(unsigned int whichplayer) : FlyByWire(), axis_key(0
     //Initial Joystick Mode
     //NOTE: Perhaps it should be handled by FlyByJoystick, but it was cumbersome to do that
     //since it handled mainly keystrokes - Any ideas?
-    static string initialJoyMode = vs_config->getVariable("joystick", "initial_mode", "normal");
+    static string initialJoyMode = configuration()->joystick.initial_mode;;
     joy_mode = 0;
     if (initialJoyMode == "inertialxy") {
         joy_mode = joyModeInertialXY;
--- a/engine/src/gfx/cockpit.cpp
+++ b/engine/src/gfx/cockpit.cpp
@@ -1914,13 +1914,10 @@ void GameCockpit::Draw() {
     //CommandInterpretor.renderconsole();
     //}
     GFXAlphaTest(ALWAYS, 0);
-    static bool mouseCursor = XMLSupport::parse_bool(vs_config->getVariable("joystick", "mouse_cursor", "false"));
-    static bool mousecursor_pancam =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "mouse_cursor_pancam", "false"));
-    static bool mousecursor_pantgt =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "mouse_cursor_pantgt", "false"));
-    static bool mousecursor_chasecam =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "mouse_cursor_chasecam", "true"));
+    const bool mouseCursor = configuration()->joystick.mouse_cursor;
+    const bool mousecursor_pancam = configuration()->joystick.mouse_cursor_pancam;
+    const bool mousecursor_pantgt = configuration()->joystick.mouse_cursor_pantgt;
+    const bool mousecursor_chasecam = configuration()->joystick.mouse_cursor_chasecam;
     if (mouseCursor && screenshotkey == false) {
         if ((view == CP_PAN
                 && !mousecursor_pancam)
--- a/engine/src/gfx/nav/navscreen.cpp
+++ b/engine/src/gfx/nav/navscreen.cpp
@@ -353,15 +353,15 @@ void NavigationSystem::Setup() {
     ScreenToCoord(buttonskipby4_7[2]);
     ScreenToCoord(buttonskipby4_7[3]);
 
-//reverse = XMLSupport::parse_bool (vs_config->getVariable ("joystick","reverse_mouse_spr","true"))?1:-1;
+    reverse = configuration()->joystick.reverse_mouse_spr ? 1 : -1;
 
-    reverse = -1;
+    // reverse = -1;
     if ((screenskipby4[1] - screenskipby4[0]) < (screenskipby4[3] - screenskipby4[2])) {
         system_item_scale *= (screenskipby4[1] - screenskipby4[0]);            //is actually over 1, which is itself
     } else {
         system_item_scale *= (screenskipby4[3] - screenskipby4[2]);
     }
-    screenoccupation = new navscreenoccupied(screenskipby4[0], screenskipby4[1], screenskipby4[2], screenskipby4[3], 1);
+    screenoccupation = new navscreenoccupied(screenskipby4[0], screenskipby4[1], screenskipby4[2], screenskipby4[3], true);
 
     //Get special colors from the config
     currentcol = vs_config->getColor("nav", "current_system",
--- a/libraries/root_generic/options.cpp
+++ b/libraries/root_generic/options.cpp
@@ -211,21 +211,19 @@ void vs_options::init() {
     /* Player Options */
 
     /* Joystick Options */
-    joystick_exponent = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "joystick_exponent", "1.0"));
-    polling_rate = XMLSupport::parse_int(vs_config->getVariable("joystick", "polling_rate", "0"));
-    force_use_of_joystick =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "force_use_of_joystick", "false"));
-    debug_digital_hatswitch =
-            XMLSupport::parse_bool(vs_config->getVariable("joystick", "debug_digital_hatswitch", "false"));
-    deadband = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "deadband", "0.05"));
-    mouse_deadband = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "mouse_deadband", "0.025"));
-    warp_mouse = XMLSupport::parse_bool(vs_config->getVariable("joystick", "warp_mouse", "false"));
-    mouse_sensitivity = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "mouse_sensitivity", "50.0"));
-    mouse_exponent = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "mouse_exponent", "3.0"));
-    mouse_blur = XMLSupport::parse_floatf(vs_config->getVariable("joystick", "mouse_blur", ".025"));
-    force_feedback = XMLSupport::parse_bool(vs_config->getVariable("joystick", "force_feedback", "false"));
-    ff_device = XMLSupport::parse_int(vs_config->getVariable("joystick", "ff_device", "0"));
-    warp_mouse_zone = XMLSupport::parse_int(vs_config->getVariable("joystick", "warp_mouse_zone", "100"));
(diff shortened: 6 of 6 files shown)


engine/src/cmd/base_interface.cpp
```diff
@@ -1315,7 +1314,7 @@ inline QVector randyVector(float min, float max) {
 void BaseInterface::Room::Eject::Click(BaseInterface *base, float x, float y, int button, int state) {
     if (state == WS_MOUSE_UP) {
         Link::Click(base, x, y, button, state);
-        XMLSupport::parse_bool(vs_config->getVariable("physics", "AutomaticUnDock", "true"));
+        configuration()->physics.automatic_undock;  // TODO: Omit this line?
```
BenjamenMeyer: yeah this looks like something that shouldn't be there as it basically doesn't do anything for the block and the value isn't saved.

engine/src/cmd/bolt.cpp
```diff
@@ -183,11 +183,11 @@
 
     BlendTrans(drawmat, cur_position, prev_position);
     Matrix drawmat(this->drawmat);
-    if (game_options()->StretchBolts > 0) {
+    if (configuration()->graphics.stretch_bolts > 0) {
         ScaleMatrix(drawmat,
                 Vector(1,
                         1,
-                        type->speed * BoltDrawManager::elapsed_time * game_options()->StretchBolts / type->length));
+                        type->speed * BoltDrawManager::elapsed_time * configuration()->graphics.stretch_bolts / type->length));
```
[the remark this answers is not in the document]
BenjamenMeyer: we should ensure the correct typing here so prevent overflows

engine/src/cmd/base_interface.cpp
stephengtuggy: Removed

engine/src/star_system_jump.cpp
```diff
@@ -129,7 +129,7 @@
                         ->SetPosition(
                                 un->Position() + r.Cast() * un->rSize() * (pendingjump[kk]->delay + .25));
                 JumpAnimations[k].a->SetOrientation(p, q, r);
-                float dd = un->rSize() * configuration()->graphics.jump_gate_size
+                float dd = un->rSize() * configuration()->graphics.jump_gate_size_flt
```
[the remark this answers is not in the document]
stephengtuggy: Fixed

stephengtuggy:
See also #1252 
See also https://github.com/royfalk/JsonStruct/pull/4
See also https://github.com/vegastrike/Assets-Masters/pull/79

stephengtuggy:
What do you guys think? Should we go ahead and merge this, even though the game's performance is slower? Migrating the settings is a fairly important goal. Thoughts?

evertvorster:
My thought is that just reading a config should not be adjusting the game performance. While it is still fresh in your mind of what you did to change where the engine reads the config values from, it would be the easiest to debug.

Sorry for being absent for so long, I am a bit busy on a different project.

stephengtuggy:
> My thought is that just reading a config should not be adjusting the game performance. While it is still fresh in your mind of what you did to change where the engine reads the config values from, it would be the easiest to debug.
> 
> Sorry for being absent for so long, I am a bit busy on a different project.

That makes sense.

No worries.

engine/src/cmd/ai/flyjoystick.cpp
```diff
@@ -84,25 +84,17 @@ void FlyByJoystick::JDecelKey( KBSTATE k, float, float, int )
...
-    static float pamountn = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pamount_nav", "0.0"));
-    static float expfactorc =
-            XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expfactor_combat", "6.0"));
-    static float
-            pfactorc = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pfactor_combat", "2.0"));
-    static float expamountc =
-            XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_expamount_combat", "1.0"));
-    static float
-            pamountc = XMLSupport::parse_float(vs_config->getVariable("joystick", "nonlinear_pamount_combat", "0.0"));
+    const bool clamp_joystick_axes = configuration()->joystick.clamp_axes;
```
royfalk: This is almost certainly the reason for the slowdown. I'm pretty sure execute is called once per cycle.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -594,11 +594,11 @@ void AutoLongHaul::Execute() {
         return;
     }
     const bool compensate_for_interdiction = configuration()->physics.auto_pilot_compensate_for_interdiction;
-    const float enough_warp_for_cruise = configuration()->physics.enough_warp_for_cruise;
-    const float go_perpendicular_speed = configuration()->physics.warp_perpendicular;
-    const float min_warp_orbit_radius = configuration()->physics.min_warp_orbit_radius;
-    const float warp_orbit_multiplier = configuration()->physics.warp_orbit_multiplier;
-    const float warp_behind_angle = cos(3.1415926536 * configuration()->physics.warp_behind_angle / 180.0);
+    const float enough_warp_for_cruise = configuration()->physics.enough_warp_for_cruise_flt;
```
royfalk: Try making these static and see if that improves performance as well.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -684,7 +684,7 @@ void AutoLongHaul::Execute() {
         if (speed > .01) {
             cfacing = cfacing * (1. / speed);
         }
-        const float dotLimit = cos(3.1415926536 * configuration()->physics.auto_pilot_spec_lining_up_angle / 180.0);
+        const float dotLimit = cos(3.1415926536F * configuration()->physics.auto_pilot_spec_lining_up_angle_flt / 180.0F);
```
royfalk, line 714: Also here. In general, check all references to configuration in Execute functions. There's a good chance that's part of the slowdown.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -700,7 +700,7 @@ void AutoLongHaul::Execute() {
     float time_to_destination = dis / maxspeed;
 
     const bool rampdown = configuration()->physics.auto_pilot_ramp_warp_down;
-    const float warprampdowntime = configuration()->physics.warp_ramp_down_time;
+    const float warprampdowntime = configuration()->physics.warp_ramp_down_time_flt;
```
royfalk: Also here.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -723,8 +723,8 @@ void AutoLongHaul::Execute() {
     if (!finish) {
         ResetDone();
     }
-    const float distance_to_stop = configuration()->physics.auto_pilot_termination_distance;
-    const float enemy_distance_to_stop = configuration()->physics.auto_pilot_termination_distance_enemy;
+    const float distance_to_stop = configuration()->physics.auto_pilot_termination_distance_flt;
+    const float enemy_distance_to_stop = configuration()->physics.auto_pilot_termination_distance_enemy_flt;
```
royfalk: And here.

engine/src/cmd/ai/warpto.cpp
```diff
@@ -53,7 +53,7 @@ bool DistanceWarrantsWarpTo(Unit *parent, float dist, bool following) {
         float mytime = SIMULATION_ATOM * 1.5;
         const bool rampdown = configuration()->physics.auto_pilot_ramp_warp_down;
         if (rampdown == false) {
-            const float warprampdowntime = configuration()->physics.warp_ramp_down_time;
```
royfalk, line 56: Called from Navigation::Execute so also here.

engine/src/cmd/base_interface.cpp
```diff
@@ -391,17 +390,16 @@ void BaseInterface::Room::Draw(BaseInterface *base) {
     const bool enable_markers = configuration()->graphics.bases.enable_location_markers;
     const bool draw_text = configuration()->graphics.bases.draw_location_text;
     const bool draw_always = configuration()->graphics.bases.location_marker_draw_always;
-    static float y_lower =
-            -0.9;           //shows the offset on the lower edge of the screen (for the textline there) -> TODO: Should be defined globally somewhere
-    const float base_text_background_alpha = configuration()->graphics.bases.text_background_alpha;
+    static float y_lower = kYLower;           //shows the offset on the lower edge of the screen (for the textline there)
```
royfalk: Why have it here if you can just access kYLower from below?

engine/src/cmd/base_interface.cpp
```diff
@@ -1273,8 +1271,7 @@ extern void PlayDockingSound(int dock);
 void BaseInterface::Room::Launch::Click(BaseInterface *base, float x, float y, int button, int state) {
     if (state == WS_MOUSE_UP) {
         Link::Click(base, x, y, button, state);
-        static bool
-                auto_undock_var = XMLSupport::parse_bool(vs_config->getVariable("physics", "AutomaticUnDock", "true"));
+        const bool auto_undock_var = configuration()->physics.automatic_undock;
```
royfalk: Why do we have this? Why not just reference configuration()->physics.automatic_undock directly from _if ((playa && bas) && (auto_undock || ..._

engine/src/cmd/basecomputer.cpp
```diff
@@ -2121,13 +2123,15 @@ bool UpgradeAllowed(const Cargo &item, Unit *playerUnit) {
...
                 color_prohibited_upgrade_flag = true;
+                VS_LOG(debug, "Upgrade not allowed because quantity is zero");
                 return false;
             }
             unsigned int i = 0;
             Cargo *numUpgrades = playerUnit->GetCargo(item.GetName(), i);
             if (numUpgrades) {
                 if (numUpgrades->GetQuantity() >= quantity) {
                     color_prohibited_upgrade_flag = true;
+                    VS_LOG(debug, (boost::format("Upgrade not allowed because numUpgrades->GetQuantity(), %1%, is >= quantity, %2%") % numUpgrades->GetQuantity() % quantity));
```
royfalk: While technically correct, this message is meaningless.
A more accurate message would be:
_"Ship already has max number of upgrades of this type."_ or similar.

engine/src/cmd/basecomputer.cpp
```diff
@@ -2141,6 +2145,7 @@ bool UpgradeAllowed(const Cargo &item, Unit *playerUnit) {
             }
             if (totalquant >= quantity) {
                 color_prohibited_upgrade_flag = true;
+                VS_LOG(debug, (boost::format("Upgrade not allowed because totalquant, %1%, is >= quantity, %2%") % totalquant % quantity));
```
royfalk: Same here. Also, take a look at ComponentsManager::AllowedUpgrade in #1234 where I've refactored this code to something nicer.

engine/src/cmd/basecomputer.cpp
```diff
@@ -2492,6 +2497,14 @@ void BaseComputer::loadMasterList(Unit *un,
                     col.cargo.SetCategory("#c.5:1:.3#Uncategorized Cargo");
                 }
                 items->push_back(col);
+            } else {
+                if (un->GetCargo(i).GetCategory().find_first_of("Weapon") != string::npos) {
+                    VS_LOG(debug, (boost::format("%1%: Not adding weapon %2% to Master List because its quantity is 0") % __FUNCTION__ % un->GetCargo(i).GetName()));
```
royfalk: cargo.h already has weapon field. Also, I have no idea what this function does and therefore, can't figure out why you are "Not adding weapon %2% to Master List because its quantity is 0".
Once #1234 is merged, we won't bother checking for such stuff, as there won't be any option for 0 quantity cargo items.

royfalk (commented):
I've only reached beam.cpp before time ran out. 
I have some comments.

engine/src/cmd/base_interface.cpp
stephengtuggy: Good point

engine/src/cmd/ai/flyjoystick.cpp
stephengtuggy: Do you really think accessing a `configuration()->` item takes that long? I guess we can measure it in the unit tests and see.

stephengtuggy: The thing is, if we make it a static, then its value gets assigned only once, **before the game configuration is loaded.** That's why I'm making them consts and not statics.

engine/src/cmd/base_interface.cpp
stephengtuggy: Yeah, we could probably just do that.

engine/src/cmd/ai/flyjoystick.cpp
stephengtuggy: There are other alternatives. Like making each of these variables a `static optional<bool>`, etc. Then check whether it has a value before using it, and if it doesn't, then assign its value from `configuration()`.

stephengtuggy: (Or a `static optional<double>` ,and so forth.)

libraries/cmd/movable.h
```diff
@@ -45,22 +45,6 @@ class Movable {
...
-    }
-
-    float getMass() const {
-        return Mass;
-    }
-
-    void setMass(float mass) {
-        Mass = mass;
-    }
-
```
BenjamenMeyer: is this why we're getting warnings about no `getMass` method?

engine/src/cmd/ai/flyjoystick.cpp
stephengtuggy: OK, the results are in. I added the following to configuration_tests.cpp:

```c++
    {
        VS_LOG_AND_FLUSH(important_info, "Starting configuration()->..._dbl test");
        const auto start_time = std::chrono::system_clock::now();
        for (int i = 0; i < 1000000; ++i) {
            // ReSharper disable CppDFAUnreadVariable
            const double asteroid_difficulty = configuration()->physics.asteroid_difficulty_dbl;
            // ReSharper restore CppDFAUnreadVariable
        }
        const auto end_time = std::chrono::system_clock::now();
        const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end_time - start_time);
        VS_LOG_AND_FLUSH(important_info, (boost::format("Finished configuration()->..._dbl test. Took %1% milliseconds") % duration.count()));
    }

    {
        VS_LOG_AND_FLUSH(important_info, "Starting test of static optional setting variable");
        const auto start_time = std::chrono::system_clock::now();
        static boost::optional<double> setting;
        for (int i = 0; i < 1000000; ++i) {
            if (setting == boost::none) {
                setting = configuration()->physics.asteroid_difficulty_dbl;
            }
        }
        const auto end_time = std::chrono::system_clock::now();
        const auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end_time - start_time);
        VS_LOG_AND_FLUSH(important_info, (boost::format("Finished test of static optional setting variable. Took %1% milliseconds") % duration.count()));
    }
```

The first block takes roughly 22 milliseconds. The second, more like 3 ms. One is clearly faster than the other.

stephengtuggy: If we can use the second form, and still read game settings after they have been populated from the config file -- which I think will work, in general -- then it seems clear which is the better choice.

engine/src/cmd/base_interface.cpp
stephengtuggy: Fixed.

engine/src/cmd/base_interface.cpp
stephengtuggy: Done.

engine/src/cmd/basecomputer.cpp
stephengtuggy: Fixed.

engine/src/cmd/basecomputer.cpp
stephengtuggy: Changed to match the previous one.

engine/src/cmd/basecomputer.cpp
stephengtuggy: Changed to call IsWeapon(). Hopefully that is acceptable for now. It is certainly more efficient than what was there before.

libraries/cmd/movable.h
stephengtuggy: I don't think so. There was a line where I had capitalized `GetMass()` wrong. Or failed to fix a spot. It should now be fixed.

engine/src/cmd/ai/flyjoystick.cpp
stephengtuggy: I added a third test block, calling `.get()` on the optionals every time through the loop. The performance hit was negligible -- almost zero. Very impressive.

engine/src/cmd/ai/navigation.cpp
stephengtuggy: Should be fixed now

engine/src/cmd/ai/navigation.cpp
stephengtuggy: Should be fixed now

engine/src/cmd/ai/navigation.cpp
stephengtuggy: Should be fixed now

engine/src/cmd/ai/warpto.cpp
stephengtuggy: Should be fixed now

stephengtuggy:
Well, the game seems to be running faster. But now it's crashing almost as soon as I get out into space. That won't do.

stephengtuggy:
It reliably crashes when I adjust the throttle on my joystick. (Often at other times too, but definitely under those circumstances.)

Seems bizarre. How could the changes I made have caused this?

engine/src/cmd/ai/flyjoystick.cpp
BenjamenMeyer: it'd be good if especially for control settings and a few others we had the ability to change it at run-time more dynamically so that we could integrate the configuration into the game instead of having a separate application - the user might after that adjust things during the game run. Not sure what the answer is; but a long term goal.

stephengtuggy: Perhaps at some point in the future, we can start to differentiate between "reloadable" and "non-reloadable" settings. Some of the settings need to just be set once, for performance reasons. But others aren't as performance-critical, and perhaps could be read from the settings object every time, instead.

stephengtuggy:
Okay, you guys. I think I have finally gotten performance and stability back up to acceptable levels. I think this PR is finally ready for review again.

Sorry it has taken so long, and gotten so big.

stephengtuggy:
The frame rate is averaging around 25 fps on my Windows 11 VM with GPU passthrough. Although jumping to another system really threw it for a loop. We may need to come back to that.

And I dealt with the crashes in the GL-light-related code by logging errors instead of asserting and hard-crashing the program when expected conditions are not met. That, too, we might need to revisit. I think it will work for now though.

stephengtuggy:
@evertvorster there's a possibility that I've fixed #1236 , the antialiased fonts issue. Can you try it with this code and see?

stephengtuggy:
Update: When I compile and run the game in RelWithDebInfo configuration, it handles jumping just fine. It only has trouble in Debug mode, apparently.

stephengtuggy:
Okay, hold up. Serenity Mining Base is nowhere to be found! That's not going to work.

stephengtuggy:
Closing in favor of #1261

PR #1254, opened 2025-08-12
open-telemetry-cpp (Draft/WIP)

stephengtuggy:
Code Changes:

Issues:

Purpose:
- What release is this for?


engine/src/cmd/ai/pythonai.cpp
```diff
@@ -33,6 +33,7 @@
 #include <boost/python/class.hpp>
 
 #include <Python.h>
+#undef _CRT_NONSTDC_NO_DEPRECATE
```
BenjamenMeyer, line 36: we only need this on WIN32 platforms

CMakeLists.txt
```diff
@@ -113,18 +113,21 @@ MESSAGE("== Vega Strike Version: ${VEGASTRIKE_VERSION_LONG_STR}")
 
 SET(CMAKE_CXX_STANDARD 14)
 SET(CMAKE_CXX_STANDARD_REQUIRED TRUE)
-SET(CMAKE_CXX_EXTENSIONS ON)
+SET(CMAKE_CXX_EXTENSIONS OFF)
 SET(CMAKE_C_STANDARD 11)
 SET(CMAKE_C_STANDARD_REQUIRED TRUE)
-SET(CMAKE_C_EXTENSIONS ON)
+SET(CMAKE_C_EXTENSIONS OFF)
```
BenjamenMeyer, line 119: do we really need to turn the extensions off for this?

CMakeLists.txt
```diff
@@ -113,18 +113,21 @@ MESSAGE("== Vega Strike Version: ${VEGASTRIKE_VERSION_LONG_STR}")
...
+
+SET(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
 
 ADD_COMPILE_DEFINITIONS("NV_CUBE_MAP"
         "BOOST_PYTHON_NO_PY_SIGNATURES"
-        "BOOST_ALL_DYN_LINK"
+        "BOOST_ALL_NO_LIB"
+        "LIBTYPE_STATIC"
+        "AL_LIBTYPE_STATIC"
```
BenjamenMeyer, line 127: i'm mixed on this. If we do dynamic linking then users can more easily upgrade the libraries, and fix bugs. If we do static the binary is probably a bit nicer - larger b/c more code, but probably better overall and slightly more portable - but at the cost of the flexibility and having to recompile for any library change.

So I see pros and cons; though I lean towards keeping dynamic linking as much as we can.

engine/objconv/mesher/to_OgreMesh.cpp
```diff
@@ -792,7 +792,7 @@ static Ogre::VertexData * CreateVertexBuffers( const vector< GFXVertex > &vertic
             Vector3 P( p[0], p[1], p[2] );
             min.makeFloor( P );
             max.makeCeil( P );
-            maxsqr = std::max( P.squaredLength(), maxsqr );
+            maxsqr = (std::max)( P.squaredLength(), maxsqr );
```
BenjamenMeyer, line 795: this is an interesting change... I'd like some background on it...

CMakeLists.txt
stephengtuggy: I'm not sure. I'm still just trying lots of different things to see if I can get it to work.

opentelemetry-cpp and OpenAL do not seem to be playing nice. One wants static linking, the other dynamic linking. And there are other gotchas with each, as well.

CMakeLists.txt
stephengtuggy: The only reason I started changing to static linking was that opentelemetry-cpp seemed to want it.

engine/objconv/mesher/to_OgreMesh.cpp
stephengtuggy: Something I did caused the `min` and `max` macros to be defined again, by minwindef.h. This shouldn't happen; we have all sorts of safeguards against it. But it happened nonetheless.

Putting the function name in parentheses is a workaround to avoid the macro grabbing the `max(` out of `std::max(...` and substituting some other text in its place. Really an ugly hack, and I am not happy about it.

engine/src/cmd/ai/pythonai.cpp
stephengtuggy: Oh, perhaps you're right

BenjamenMeyer: if we need it we could probably add it to the CMake stuff when building on WIN32.

stephengtuggy:
I've had the most ridiculous amount of trouble with this. I've spent like 3 days on it, tried dozens of different variations of ways to incorporate opentelemetry-cpp into our codebase, and still haven't succeeded. I am going to officially recommend that we forget about using this library in Vega Strike.

PR #1256, opened 2025-08-12
Remove pilot_server.cpp

royfalk:
It is practically a duplicate of pilot.cpp.

Code Changes:

Issues (none appear related to PR):
- Game runs slowly. This does not affect frame rate though.
- Physics seems off as well. Fast, maneuverable ships do not respond.
- During combat, my ship got damaged when colliding with another and the front shield maxed at some fraction. However, component view still reported no damage.
- Autopilot is off. It jitters on/off near Atlantis and other places.


stephengtuggy:
I believe the `_server` versions of everything were the versions for multiplayer mode. Do we want to delete these, or not? If we're going to try to bring multiplayer back at some point, deleting these files now might be a waste of time and effort.

royfalk:
> I believe the `_server` versions of everything were the versions for multiplayer mode. Do we want to delete these, or not? If we're going to try to bring multiplayer back at some point, deleting these files now might be a waste of time and effort.

The server code in VS was IMO not the best way to implement such a solution. The right way is to decouple the GUI, Audio and Interface files from the engine. Also, after seven years and the progress we've made, I think we need to set realistic goals. We are not going to get multiplayer support without a couple more developers.

Seven years and we're still working on technical debt. We're preserving the game and making it available for people. But I don't see us releasing three games in the Vegastrike universe.

stephengtuggy:
> > I believe the `_server` versions of everything were the versions for multiplayer mode. Do we want to delete these, or not? If we're going to try to bring multiplayer back at some point, deleting these files now might be a waste of time and effort.
> 
> The server code in VS was IMO not the best way to implement such a solution. The right way is to decouple the GUI, Audio and Interface files from the engine. Also, after seven years and the progress we've made, I think we need to set realistic goals. We are not going to get multiplayer support without a couple more developers.
> 
> Seven years and we're still working on technical debt. We're preserving the game and making it available for people. But I don't see us releasing three games in the Vegastrike universe.

Fair enough. Has it really been seven years? Wow, time flies.

stephengtuggy (approved):
LGTM

BenjamenMeyer:
> The server code in VS was IMO not the best way to implement such a solution. The right way is to decouple the GUI, Audio and Interface files from the engine. Also, after seven years and the progress we've made, I think we need to set realistic goals. We are not going to get multiplayer support without a couple more developers.
> 
> Seven years and we're still working on technical debt. We're preserving the game and making it available for people. But I don't see us releasing three games in the Vegastrike universe.

yes; we need a different approach to multi-player/server mode. I don't see us even beginning to tackle it until after we hit the "1.0" target, and then we'll need to develop a proper protocol for it. IIRC the original multi-player kind of emulated the two player mode that WC Armada and a few others had, typical when one did direct dial-up (over a modem) to a friend. Today's multi-player expectations are significantly different than that; and yes I agree with @royfalk that we'd likely need (a) more devs and (b) funding towards it to properly tackle it but before we can do that we need to get some of these core bugs out of the way and resolve some things like having a stable universe between multiple systems (how we seed the universe).

Issue #1251, opened 2025-08-08
Document LibComponent by adding an MD or RST file in it.


royfalk:
From [198](https://github.com/vegastrike/Assets-Production/pull/198)

>  If I found this difficult to understand, I suspect that other people will, too. We'd better document this carefully, if we do make this change.

>>  I'll add a comment to the JSON.

Can we add a MD or RST file about this?
We need some kind of architectural/design document on this, preferably close to the code.

PR #1259, opened 2025-08-14
Create readme.rst for LibComponents

royfalk:
Fix: #1251

### DIFF
--- a/engine/src/components/readme.rst
+++ b/engine/src/components/readme.rst
@@ -1,0 +1,62 @@
+============
+LibComponent
+============
+************
+Introduction
+************
+LibComponent is a Vegastrike library dealing with ship components or modules. While it provides some functionality dealing with upgrade/downgrade and damage/repair, it is mostly a place to store the attributes of various components. As such, it does not deal with things such as GUI, 3d graphics, audio, etc.
+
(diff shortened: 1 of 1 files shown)


engine/src/components/readme.rst
```diff
@@ -0,0 +1,62 @@
+============
+LibComponent
+============
+************
+Introduction
+************
+LibComponent is a Vegastrike library dealing with ship components or modules. While it provides some functionality dealing with upgrade/downgrade and damage/repair, it is mostly a place to store the attributes of various components. As such, it does not deal with things such as GUI, 3d graphics, audio, etc.
+
+It is intended to be reasonably encapsulated in order to support unit testing.
```
BenjamenMeyer, line 9: Q: given this, should we split libcomponent out to the `libraries`? instead of keeping it under `engine/src`?

BenjamenMeyer (approved):
seems reasonable; one minor question (not blocking, but may trigger a follow-on PR)

engine/src/components/readme.rst
royfalk: I don't have a problem with that suggestion. That's true for logging, damage and resource libraries.

stephengtuggy (approved):

PR #1260, opened 2025-08-16
Specify permissions for GitHub workflows

stephengtuggy:
Code Changes:
- [x] This is a CI change only

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Follow best practices for GitHub Actions CI workflows by explicitly specifying the permissions granted to each
- What release is this for? The next one
- Is there a project or milestone we should apply this to? Not sure

### DIFF
--- a/.github/workflows/macos-release.yml
+++ b/.github/workflows/macos-release.yml
@@ -1,6 +1,6 @@
 name: 'MacOS-Release'
 permissions:
-    contents: read
+    contents: write
     pull-requests: read
 
 on:


.github/workflows/Windows-Release.yml
```diff
@@ -1,4 +1,7 @@
 name: Windows-Release
+permissions:
+    contents: read
+    pull-requests: read
```
BenjamenMeyer, line 4: We probably need `contents: write` on the Release ones so that we can write the release files back to the repository

https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

stephengtuggy: Looks like you're right. Done.

We should probably run our first alpha release of 0.10.0 once we get this merged, just to verify that things are working as far as the release process. What do you say?

BenjamenMeyer: @stephengtuggy sounds like a plan. it'd be nice to verify it's all working correctly; Hopefully we won't need to do any admin merges to fix stuff.

.github/workflows/fortify-on-demand-scan.yml
```diff
@@ -1,4 +1,7 @@
 name: Fortify on Demand Scan
+permissions:
+    contents: read
+    pull-requests: read
```
BenjamenMeyer, line 4: @stephengtuggy do we need the security event on this one too?

stephengtuggy: I'm not sure. I don't think we've used this one in quite a while. Maybe we should just delete it?

BenjamenMeyer: answer: we're not using it so we're going to clean it up in another PR

Issue #1174, opened 2025-06-09
Configuring Vega Strike to run now extremely hard


evertvorster:
Right now I am unable to play the master version of Vega Strike. No play testing, nothing. The game is unplayable. 
This comes down to me being unable to configure the game to run on my system. 
Mouse control is broken, and I can't enable or configure my joystick. 

Vegasettings does not work anymore, and most of it's functionality has been ported to /usr/share/vegastrike/config.json. 
I have no clue as to which options have been ported. 

What this means for me, and anyone that would like to play VegaStrike that is not a developer it is now extremely difficult to set options for the game. After some time in trying to configure the game, I give up.

Honestly, do a clean install of Vega Strike and try to set up the game for your computer. You'll see what I mean. 

The first area of confusion is whether the option has been ported to config.json.
Some of the settings have been ported, but the game does not use the variables from config.json, and prefers the variable in vegastrike.config. 
You can imagine the process of changing a setting first you have to edit the setting in config.json, then if the game ignores it, start looking in vegastrike.config, find the variable, then see if it is active, what the options are, etc.. etc. 

The second pain point I have with config.json is that it is installed in an area of the system where you need root access to modify it. There is no corresponding config.json in the user's home directory. 
This of course means that every time the engine is re-installed, the default config.json is written over the modified version. It is possible to make a backup of the config.json file, but then if there are any new variables in the new config file, they have to be manually set. 

Honestly, the process of changing configuration files could have been handled a little more elegantly enabling one to still configure and play the game while this process is ongoing.  
Some planning on how the transition between configuring the game would be handled would have gone a long way here. It would probably have been more prudent to first make a configuration utility that reads the config.json file in the system directory and compares it with a config.json file in the user's home directory as a start. The game should only use the config.json in the user's home directory to run. 

What is sorely needed is a process to remove every setting variable from the vegastrike.config file as it is moved over to config.json. In this way there is no confusion about in which file to modify variables. 

There really, really needs to be a UI for setting variables in the config.json. 
With information on what each setting does. Even the config.json text file has no explanation for each variable. 

Also, the game absolutely needs to have a copy of config.json that resides in the user's home directory that is used for starting the game.  These two go hand in hand, as when a new variable pops up in the system copy of config.json the configuration UI needs to highlight the new variable so the user can set the option in their own copy of config.json. 

TL;DR
[...]

PR #1263, opened 2025-08-18
More on settings and performance

### LINKED ISSUE #1278: Finish converting pi literals (3.1415926536F) to M_PI or M_PI_FLT
Why not use PI here. Is this what @BenjamenMeyer also noted?

_Originally posted by @royfalk in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1263#discussion_r2306117079_

stephengtuggy:
Code Changes:

Issues Addressed:
- Contributes to #1174 
- Closes #1278 

Known Issues Remaining:
- This PR doesn't convert any more sections of settings to the configuration() way of doing things. It more lays the groundwork for future progress in that direction. So, for example, the Joystick settings aren't fully converted yet.

Purpose:
- What is this pull request trying to do? Make further progress on converting settings, while maintaining game performance, and not changing too many lines of code at once. Specifically, this PR mostly introduces a `float` version and a `double` version of every floating-point setting.

### DIFF
--- a/engine/src/components/components_manager.cpp
+++ b/engine/src/components/components_manager.cpp
@@ -368,7 +368,7 @@ double ComponentsManager::PriceCargo(const std::string &cargo_name) {
         return cargo.GetPrice();
     }
 
-    return configuration().cargo.space_junk_price;
+    return configuration().cargo.space_junk_price_dbl;
 }
 
 bool ComponentsManager::_Sell(CargoHold *hold, ComponentsManager *buyer, Cargo *item, int quantity) {


evertvorster:
I'm sorry, but there seems to be issues with git master that prevents me from properly testing this now. I'll run a full suite of tests against git master, and once we get that running properly, I'll come back to this.

kheckwrecker (approved):
This PR looks good to me.

.gitignore
```diff
@@ -37,3 +37,6 @@ test_assets/python_tests.py
 test_assets/units.json
 test_assets/vegastrike.config
 CMakeUserPresets.json
+vcpkg_installed/
+share/
+nlohmann_json.natvis
```
BenjamenMeyer: does this last one need to be in here? Or is that something from your own environment?

engine/CMakeLists.txt
```diff
@@ -1238,6 +1238,7 @@ IF (USE_GTEST)
         src/components/tests/jump_drive_tests.cpp
         src/components/tests/utils_tests.cpp
         src/components/tests/operational_tests.cpp
+        ../libraries/gfx_generic/tvector.cpp
```
BenjamenMeyer: this shouldn't be in this file. It's provided by one of the libraries and should get linked. If something isn't building as a result then we need to fix the linker issue not add another compilation of it.

libraries/cmd/carrier.cpp
```diff
@@ -245,12 +245,12 @@ void Carrier::EjectCargo(unsigned int index) {
             }
             float arot = 0;
             const float grot =
-                    configuration().graphics.generic_cargo_rotation_speed * 3.1415926536 / 180;
+                    configuration().graphics.generic_cargo_rotation_speed_flt * 3.1415926536F / 180.0F;
```
BenjamenMeyer: these should get changed over to `M_PI` which we're already using elsewhere in the code, such as in `libraries/cmd/unit_generic.cpp`

BenjamenMeyer (changes requested):

.gitignore
stephengtuggy: Oh, it probably doesn't. Good point.

stephengtuggy: Removed.

libraries/cmd/carrier.cpp
stephengtuggy: OK. That makes sense.

stephengtuggy: The only issue I see with this is that `M_PI` is a `double`, not a `float`. If I use it, the whole calculation on each of these lines will become a `double` calculation. Not the end of the world, but worth pointing out, I think.

engine/CMakeLists.txt
stephengtuggy: I didn't want to pull in an entire additional library when all I needed was that one file.

BenjamenMeyer: You're not pulling a library in. It's should already be linking it. It's one of our own libraries anyway

libraries/cmd/carrier.cpp
BenjamenMeyer: You could downcast it to a float.
C++20 has a better solution.

engine/CMakeLists.txt
stephengtuggy: But I will try it your way and at least see if it works.

stephengtuggy: The fact is that it is a separate library. It's one of our own, but up to this point, Vega_Strike_Tests has been linked to very few of our libraries.  Only one, in fact, I think.

libraries/cmd/carrier.cpp
stephengtuggy: > You could downcast it to a float. C++20 has a better solution.

Are you thinking of `constexpr`? That's available in C++14 as well. And I am making use of it in my latest code here.

engine/CMakeLists.txt
stephengtuggy: OK. Vega_Strike_Tests seems to be working fine this way. I'll go ahead and leave it linked to the gfx_generic library, as you requested.

libraries/cmd/carrier.cpp
stephengtuggy: I believe this is also taken care of.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -594,11 +594,11 @@ void AutoLongHaul::Execute() {
         return;
     }
     const bool compensate_for_interdiction = configuration().physics.auto_pilot_compensate_for_interdiction;
-    const float enough_warp_for_cruise = configuration().physics.enough_warp_for_cruise;
-    const float go_perpendicular_speed = configuration().physics.warp_perpendicular;
-    const float min_warp_orbit_radius = configuration().physics.min_warp_orbit_radius;
-    const float warp_orbit_multiplier = configuration().physics.warp_orbit_multiplier;
-    const float warp_behind_angle = cos(3.1415926536 * configuration().physics.warp_behind_angle / 180.0);
```
royfalk, line 601: Why not use PI here. Is this what @BenjamenMeyer also noted?

engine/src/configuration/tests/configuration_tests.cpp
```diff
@@ -35,52 +35,131 @@
 #include <chrono>
 
 #include "gfxlib_struct.h"
+#include "vega_cast_utils.h"
 #include "configuration/configuration.h"
 #include "gfx/quadsquare.h"
 #include "resource/random_utils.h"
 #include "root_generic/lin_time.h"
-#include "src/resizable.h"
```
royfalk, line 42: Please document what you're trying to do here and summarise results for the rest of the class. I can totally see someone, five years from now, cursing us, just as we sometimes curse the original dev team for their obscure code and comments.

engine/src/star_system.cpp
```diff
@@ -301,55 +302,103 @@ class UnitDrawer {
 #define UPDATEDEBUG  //for hard to track down bugs
 
 void StarSystem::Draw(bool DrawCockpit) {
+#if defined(LOG_TIME_TAKEN_DETAILS)
+    const double start_time = realTime();
```
royfalk, line 306: This does not belong in a PR about settings. 
It breaks the flow of going over config settings. 
Reviewing it when there are so much other changes is a pain. I have to look for where LOG_TIME_TAKEN_DETAILS is defined and hope it's not one of those big files that are not displayed by default.

engine/src/star_system.cpp
```diff
@@ -301,55 +302,103 @@ class UnitDrawer {
...
+#endif
     }
-    double setupdrawtime = realTime();
+#if defined(LOG_TIME_TAKEN_DETAILS)
+    double setup_draw_time = realTime();
+#endif
     {
         cam_setup_phase = true;
 
-        Unit *saveparent = _Universe->AccessCockpit()->GetSaveParent();
```
royfalk, line 335: This edit is of marginal value. I was not even aware this was part of the great debate.
See https://stackoverflow.com/q/2660633/5276890.
Also, note that while C++ prefers this style, in the following, j is not a pointer:

`int* i,j;`

https://stackoverflow.com/a/398414/5276890

engine/src/universe.cpp
```diff
@@ -456,26 +472,34 @@ void Universe::StartDraw() {
...
-    VS_LOG(trace, (boost::format("%1%: Time taken by UpdateTimeCompressionSounds(): %2%") % __FUNCTION__ % (update_time_compression_sounds_end_time - update_time_end_time)));
+    VS_LOG(trace,
+           (boost::format("%1%: Time taken by UpdateTimeCompressionSounds(): %2%") % __FUNCTION__ % (
+               update_time_compression_sounds_end_time - update_time_end_time)));
 #endif
     _Universe->SetActiveCockpit(randomInt(_cockpits.size() - 1, 0));
 #if defined(LOG_TIME_TAKEN_DETAILS)
     const double universe_set_active_cockpit_end_time = realTime();
-    VS_LOG(trace, (boost::format("%1%: Time taken by _Universe->SetActiveCockpit(...): %2%") % __FUNCTION__ % (universe_set_active_cockpit_end_time - update_time_compression_sounds_end_time)));
+    VS_LOG(trace,
```
royfalk, line 489: I assume these changes are new lines. Note that when reviewing a several thousand lines PR, stuff like this should go in its own PR. By this I mean moving asterisk to the right, removing whitespace, etc.
It would shorten this PR considerably and that PR would be an auto-approve from me.

libraries/cmd/carrier.cpp
```diff
@@ -348,16 +345,16 @@ void Carrier::EjectCargo(unsigned int index) {
...
                             cargo->owner = unit->owner;
                         } else {
                             cargo->owner = unit;
                         }
-                    }                                            //switching NULL gives "dead" ai to the unit I ejected from, by the way.
+                    }                                            //switching NULL gives "dead" AI to the unit I ejected from, by the way.
                 }
                 _Universe->activeStarSystem()->AddUnit(cargo);
-                if ((unsigned int) index != ((unsigned int) -1) && (unsigned int) index != ((unsigned int) -2)) {
+                if (index != -1U && index != -2U) {
```
royfalk, line 357: Can you do this? Is this the equivalent of ((unsigned int) -1)?

libraries/cmd/movable.cpp
```diff
@@ -369,9 +393,9 @@ Vector Movable::ResolveForces(const Transformation &trans, const Matrix &transma
     lateral_air_res_coef = XMLSupport::parse_floatf(active_missions[0]->getVariable("lateral_air_resistance", "0"));
 
     if (air_res_coef != 0.0F || lateral_air_res_coef != 0.0F) {
-        float velmag = Velocity.Magnitude();
+        double velmag = Velocity.Magnitude();
```
royfalk, line 396: All these changes from float to double in movable should be in their own PR. This is exactly the stuff that comes up as "Hey, the auto pilot can't go into SPEC" comments.

libraries/cmd/unit_generic.cpp
```diff
@@ -250,45 +250,29 @@ void Unit::Ref() {
     ++ucref;
 }
 
-#define INVERSEFORCEDISTANCE 5400
+#define kInverseForceDistance 5400
```
royfalk, line 253: Why the k prefix?

libraries/cmd/unit_generic.cpp
```diff
@@ -855,9 +806,9 @@ static std::string NearestSystem(std::string currentsystem, QVector pos) {
     if (pos.i == 0 && pos.j == 0 && pos.k == 0) {
         return "";
     }
-    QVector posnorm = pos.Normalize();
-    posnorm.Normalize();
-    QVector cur = SystemLocation(currentsystem);
+    QVector pos_normalized = pos.Normalize();
```
royfalk: normalized_position?

libraries/cmd/unit_generic.cpp
```diff
@@ -1389,7 +1351,7 @@ void Unit::Kill(bool erasefromsave, bool quitting) {
     if (ucref == 0) {
         VS_LOG(trace, (boost::format("UNIT DELETION QUEUED: %1$s %2$s (file %3$s, addr 0x%4$08x)")
                 % name.get().c_str() % fullname.c_str() % filename.get().c_str() % this));
-        Unitdeletequeue.push_back(this);
+        Unit_delete_queue.push_back(this);
```
royfalk: Why leave the first letter as upper? Is this not a variable?

royfalk (approved):
**Well, this was an hour of my life ;)**
No, seriously and I'm still in the middle of unit_generic.cpp.

Update: this took 75 minutes to review. 

Is each commit a standalone one and can run? If so, the fact the PR is so big is less of an issue because we can bisect a bug. If not, this could really hurt us.

I did not review the changes you've made to profiling. I'm concerned you've added logs there. If it's a high frequency function, that could introduce delays.

Also not a fan of burying conversions of float to double in such a big PR. I've noted it in one instance specifically. 

One the plus side, this is certainly a lot of work done. Please don't take from this that I don't appreciate the effort.

evertvorster:
I can't seem to merge both this PR and the bugfix for the mass, as there is a conflict in cmd/unit_generic.cpp.

engine/src/cmd/ai/navigation.cpp
stephengtuggy: Hmm. Looks like I missed a few of these somehow. I will fix them in a followup PR.

engine/src/star_system.cpp
stephengtuggy: I was trying to at least maintain the same level of performance while converting settings. But point taken. Perhaps it would have been better to break these out.

On the positive side, they won't even get compiled if `LOG_TIME_TAKEN_DETAILS` isn't defined. As such, these blocks of code won't have any performance impact most of the time.

engine/src/configuration/tests/configuration_tests.cpp
stephengtuggy: IIRC, I was trying to add some unit tests for the Resizable class. The tests were failing, however, and I didn't have time to fix them in this PR. So I disabled them for now.

Really, tests like these should be in a different test file. But I haven't figured out how to get tests in a different source file to actually run.

engine/src/star_system.cpp
stephengtuggy: These were formatting changes that my IDE introduced. I wasn't aware that they had any significance whatsoever. I will keep this in mind. and probably adjust my IDE settings to lean toward keeping pointer declarations as they are.

engine/src/universe.cpp
stephengtuggy: Yes, they are newlines and indentation changes. Point taken. I will try to remember to put formatting changes like these in their own PRs moving forward.

libraries/cmd/carrier.cpp
stephengtuggy: As far as I know, you can. The `U` or `u` suffix indicates an `unsigned` literal.

libraries/cmd/unit_generic.cpp
stephengtuggy: `k` indicates a constant (like in calculus). @BenjamenMeyer wants us to use that notation for constants.

libraries/cmd/movable.cpp
stephengtuggy: This is, in fact, part of the meat of this PR. I am trying to improve execution speed by minimizing `float`-to-`double` and `double`-to-`float` conversions.

libraries/cmd/unit_generic.cpp
stephengtuggy: Oh, good point.

libraries/cmd/unit_generic.cpp
stephengtuggy: Good catch. I will fix that in a followup PR.

libraries/cmd/unit_generic.cpp
stephengtuggy: Fixed.

libraries/cmd/unit_generic.cpp
stephengtuggy: Actually just fixed it.

stephengtuggy:
> I can't seem to merge both this PR and the bugfix for the mass, as there is a conflict in cmd/unit_generic.cpp.

@evertvorster the conflicts should now be resolved. Thanks for your patience.

engine/src/cmd/ai/navigation.cpp
stephengtuggy: Fixed.

evertvorster:
The game plays normally with this. 
Which settings are we supposed to test for?

stephengtuggy:
> The game plays normally with this. Which settings are we supposed to test for?

No specific settings have been converted in this PR. It's more laying the gr0undwork for future PRs.

1. Make sure the game isn't broken in any new and obvious ways.
2. Verify that game performance is at least as smooth and efficient as it was before, if not more so.

stephengtuggy:
I'm glad to hear that the game plays normally. Thanks for verifying that.

stephengtuggy:
> From a playability perspective, this plays well.
> 
> The issues I have noticed may not be contributable to this PR, as it has been some time since I was able to play test master.
> 
> I was able to fly from Serenity to Atlantis with a load of cargo with no issues. Then I was able to complete Janek's first mission to Jacobs with no issues. Game performance was reasonable, with me getting my usual 120 - 60fps. Janek's second mission started off OK, but I experienced a major slowdown just as I was pulling away from Atlantis. FPS dropped to 0 and 1 for a little while, and then jumped back to the normal values. I am assuming the game engine was doing something in the background here. Again, I am not sure if this is due to this PR or not.
> 
> Then, being bored I started messing with the displays in the HUD. I pressed V a couple of times to change the view of the target. Then, the game crashed with this error:
> 
> ```
> /usr/include/c++/15.2.1/bits/uniform_int_dist.h:108: std::uniform_int_distribution<_IntType>::param_type::param_type(_IntType, _IntType) [with _IntType = long unsigned int]: Assertion '_M_a <= _M_b' failed.
> fish: Job 1, 'vegastrike' terminated by signal SIGABRT (Abort)
> ```
> 
> Does that seem like an area of the code you touched with this PR? Be that as it may, I am still good for merging this, and troubleshooting that a bit later.

@evertvorster I may have touched the HUD code. Not sure. I guess I will test this and see if I can reproduce the crash.

Thanks for reviewing and approving.

stephengtuggy:
> > From a playability perspective, this plays well.
> > The issues I have noticed may not be contributable to this PR, as it has been some time since I was able to play test master.
> > I was able to fly from Serenity to Atlantis with a load of cargo with no issues. Then I was able to complete Janek's first mission to Jacobs with no issues. Game performance was reasonable, with me getting my usual 120 - 60fps. Janek's second mission started off OK, but I experienced a major slowdown just as I was pulling away from Atlantis. FPS dropped to 0 and 1 for a little while, and then jumped back to the normal values. I am assuming the game engine was doing something in the background here. Again, I am not sure if this is due to this PR or not.
> > Then, being bored I started messing with the displays in the HUD. I pressed V a couple of times to change the view of the target. Then, the game crashed with this error:
> > ```
> > /usr/include/c++/15.2.1/bits/uniform_int_dist.h:108: std::uniform_int_distribution<_IntType>::param_type::param_type(_IntType, _IntType) [with _IntType = long unsigned int]: Assertion '_M_a <= _M_b' failed.
> > fish: Job 1, 'vegastrike' terminated by signal SIGABRT (Abort)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Does that seem like an area of the code you touched with this PR? Be that as it may, I am still good for merging this, and troubleshooting that a bit later.
> 
> @evertvorster I may have touched the HUD code. Not sure. I guess I will test this and see if I can reproduce the crash.
> 
> Thanks for reviewing and approving.

@evertvorster I did touch a line of code that may be relevant: libraries/cmd/carrier.cpp, line 405. Try changing it from:

```c++
i += static_cast<char>(randomInt(1, -1));
```

to:

```c++
i += static_cast<char>(randomInt(2) - 1);
```

Then see if you still get that error.

evertvorster:
> > > From a playability perspective, this plays well.
> > > The issues I have noticed may not be contributable to this PR, as it has been some time since I was able to play test master.
> > > I was able to fly from Serenity to Atlantis with a load of cargo with no issues. Then I was able to complete Janek's first mission to Jacobs with no issues. Game performance was reasonable, with me getting my usual 120 - 60fps. Janek's second mission started off OK, but I experienced a major slowdown just as I was pulling away from Atlantis. FPS dropped to 0 and 1 for a little while, and then jumped back to the normal values. I am assuming the game engine was doing something in the background here. Again, I am not sure if this is due to this PR or not.
> > > Then, being bored I started messing with the displays in the HUD. I pressed V a couple of times to change the view of the target. Then, the game crashed with this error:
> > > ```
> > > /usr/include/c++/15.2.1/bits/uniform_int_dist.h:108: std::uniform_int_distribution<_IntType>::param_type::param_type(_IntType, _IntType) [with _IntType = long unsigned int]: Assertion '_M_a <= _M_b' failed.
> > > fish: Job 1, 'vegastrike' terminated by signal SIGABRT (Abort)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Does that seem like an area of the code you touched with this PR? Be that as it may, I am still good for merging this, and troubleshooting that a bit later.
> > 
> > 
> > @evertvorster I may have touched the HUD code. Not sure. I guess I will test this and see if I can reproduce the crash.
> > Thanks for reviewing and approving.
> 
> @evertvorster I did touch a line of code that may be relevant: libraries/cmd/carrier.cpp, line 405. Try changing it from:
> 
> ```c++
> i += static_cast<char>(randomInt(1, -1));
> ```
> 
> to:
> 
> ```c++
> i += static_cast<char>(randomInt(2) - 1);
> ```
> 
> Then see if you still get that error.

I have played it a bit more, and did not see that error again, even without the fix. It may not even be related to this PR at all.

stephengtuggy:
> I have played it a bit more, and did not see that error again, even without the fix. It may not even be related to this PR at all.

Hmm. Interesting.

engine/src/gfx/nav/drawgalaxy.cpp
```diff
@@ -68,6 +68,8 @@
 using std::string;
 using std::vector;
 
+constexpr double kPi = 3.1415926536;
```
BenjamenMeyer: Could we just use `M_PI`?

stephengtuggy: We could. But I'm trying to follow a coding convention you requested several years ago: kPascalCase with a lowercase k at the front for constants.

BenjamenMeyer: since we don't control `M_PI` I think it'd be exempt from that.

stephengtuggy: OK. I'll use `M_PI` then.

stephengtuggy: Done.

evertvorster:
FYI, I'm actively playing on this branch now, and so far no issues.

evertvorster:
From my side this is good to go. 
I have had the same crash again, but I am not sure that it is even related to this PR. 
It happens when you have visited a few systems. In my case, flying from Atlantis to Jacob to deliver Janek's first cargo. 
It all goes well, but then when I am back in Atlantis, there is a general slowdown of the game, and I think this is due to the "previous systems" setting that I have set to 8. I see a general slow down of the game as I fly through more and more systems, with the slowdown getting generally worse. 
I do also mess with the displays all the time, and the crash happens when I set the lower-right VDU target display to video by pressing "n", but only when the system has already slowed down quite a bit. I have a suspicion that the number of ships in memory from all these systems are getting to be too much, and some buffer overflows...

libraries/cmd/carrier.cpp
BenjamenMeyer: > > You could downcast it to a float. C++20 has a better solution.
> 
> Are you thinking of `constexpr`? That's available in C++14 as well. And I am making use of it in my latest code here.

I'm referring to http://en.cppreference.com/w/cpp/numeric/constants.html where they outright define `pi` and introduce the template `pi_v` - `pi` is defined as `pi_v<double>` and we could make a `pi_v<float>` as well.

BenjamenMeyer: > I believe this is also taken care of.

I did see another spot where `kPi` was defined... but not going to hold it up. We can clean this all up in another PR.

stephengtuggy:
OK, I fixed the remaining kPi instances and (hopefully) the crash, in my next PR, #1296 .

royfalk:
I'm fine with adopting C++20, but note that I tried using modules in a test case and despite the hype, it just wouldn't work for me. Very disappointing.

stephengtuggy:
I am also fine with adopting C++20. Not modules though. From everything I've read, different compilers' support for them is very spotty so far.

PR #1274, opened 2025-08-26
Fix mass calculations

### LINKED ISSUE #1268: Git master - saved game incorrectly handles ship mass.
For this test, the engine and assets were build from the latest git master as of 2025-08-20.

To trigger this bug:
- Start a new campaign. 
- Buy some stuff, note your ship's mass in percentage. 
- Save the game.
- Exit.
- Start VegaStrike again, and load your game. 
- Go to the trading station, and look at your ship's mass in percentage again. 

It should be whatever value it was before, but now for me it reads 100%, which is incorrect. 

This has many knock-on effects. 

Your ship seems to get incrementally heavier as you trade, with the weight not really going away when you sell stuff. So, after running a load of gold from Serenity to Atlantis, and selling the cargo, when I take off the ship feels just as sluggish as if I had massively overloaded it with a heavy mineral. Which I just sold. 

Sometimes you can't sell things, as your mass gets into negative numbers. 

I have seen profits from trades disappear... 

In short, git master is currently broken as far as trade is concerned, as the mass of the ship is not handled correctly in save files. 

#### How it is supposed to work:
Your hull has a mass, which you can't change. For the standard layout of the ship, this is the 100%, or base, mass that all the other calculations hang off of. 

Then, you add upgrades, which also weighs a bit, so typically your ship will weigh in the range of 100% to 130% when it has no cargo. You could possibly sell something that came as standard with your ship, and then theoretically you could have a mass of slightly lower than 100% 
This is, or should be, calculated from the mass of the installed equipment, vs, the mass of the standard equipment. 

Then, the weight of the cargo is added. 

Your total weight changes dynamically, and needs re-calculation whenever you load a game, buy stuff, sell stuff or modify your ship. 

This weight is the value you use to calculate acceleration. 
The weight of the cargo is the only bit that can be sold at trading commodities. 

There are no short cuts here, this is the way it has to be to be realistic.

royfalk:
- basecomputer.cpp:sellUpgrade calling UnitUtil::RecomputeUnitUpgrades was the culprit. It re-added cargo items several times.
- Fix bug in AllowedUpgrade which returned the reverse.
- Fix bug in ComponentsManager parsing of prohibited upgrades to correctly parse entries without explicit limit number.
- Fix bug in cargo, where upgrade and weapon fields were not correctly read. This resulted in inability to choose mount to upgrade.
- Implement PlayerShip and SetPlayerShip in ComponentsManager
- Polish GetTitle to provide base_mass and mass
- Remove some unused functions or comment out in case used in API

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation _no yet_

What I tested:
- Mass is correctly loaded from save game. 
- Mass is displayed correctly in ship computer.
- Mass is changed correctly when buying and selling cargo/upgrade.
- Capacity is changed correctly when buying selling cargo/upgrade.
- Upgrades behave as expected in base computer. Can buy and sell stuff. Can buy weapons correctly, etc.

Issues:
- I think old save games save mass **with** upgrades and cargo. As such you would need to manually adjust this or figure out another solution.

fix #1268

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -229,7 +229,6 @@ static const char *const NEWS_NAME_LABEL = "news";
 //Some upgrade declarations.
 //These should probably be in a header file somewhere.
 
-extern const Unit *makeFinalBlankUpgrade(string name, int faction);
 extern int GetModeFromName(const char *);  //1=add, 2=mult, 0=neither.
 extern Unit &GetUnitMasterPartList();
 static const string LOAD_FAILED = "LOAD_FAILED";
@@ -399,7 +398,6 @@ static float SellPrice(float operational, float price) {
     return usedValue(price);// - RepairPrice(operational, price);
 }
 
-extern const Unit *makeTemplateUpgrade(string name, int faction);
 
 //Ported from old code.  Not sure what it does.
 const Unit *getUnitFromUpgradeName(const string &upgradeName, int myUnitFaction = 0);
@@ -2048,7 +2046,7 @@ bool BaseComputer::pickerChangedSelection(const EventCommandId &command, Control
 }
 
 bool UpgradeAllowed(const Cargo &item, Unit *playerUnit) {
-    if(playerUnit->AllowedUpgrade(item)) {
+    if(!playerUnit->AllowedUpgrade(item)) {
         color_prohibited_upgrade_flag = true;
         return false;
     }
@@ -3013,18 +3011,10 @@ bool BaseComputer::changeToUpgradeMode(const EventCommandId &command, Control *c
     return true;
 }
 
-//Actually do a repair operation.
+
 static void BasicRepair(Unit *parent) {
-    if (parent) {
-        int repairmultiplier = parent->RepairCost();
-        if (UnitUtil::getCredits(parent) < basicRepairPrice() * repairmultiplier) {
-            showAlert("You don't have enough credits to repair your ship.");
-        } else if ((repairmultiplier = parent->RepairUpgrade())) {
-            UnitUtil::addCredits(parent, -basicRepairPrice() * repairmultiplier);
-        } else {
-            showAlert("Your ship has no damage.  No charge.");
-        }
-    }
+    // This function does nothing. Kept for compatibility with python API.
+    // TODO: remove
 }
 
 //The "Operation" classes deal with upgrades.
@@ -3072,7 +3062,7 @@ class BaseComputer::BuyUpgradeOperation : public BaseComputer::UpgradeOperation 
 public:
     void start(void);             //Start the operation.
 
-    BuyUpgradeOperation(BaseComputer &p) : UpgradeOperation(p), m_theTemplate(NULL), m_addMultMode(0) {
+    BuyUpgradeOperation(BaseComputer &p) : UpgradeOperation(p) {
     };
 protected:
     virtual bool checkTransaction(void);              //Check, and verify user wants transaction.
@@ -3081,9 +3071,6 @@ protected:
 
     virtual ~BuyUpgradeOperation(void) {
     }
-
-    const Unit *m_theTemplate;
-    int m_addMultMode;
 };
 
 //Sell an upgrade from our ship.
@@ -3228,10 +3215,6 @@ void BaseComputer::BuyUpgradeOperation::start(void) {
         finish();
         return;
     }
-    m_theTemplate = makeTemplateUpgrade(playerUnit->name.get(), playerUnit->faction);
-
-    m_addMultMode =
-            GetModeFromName(m_selectedItem.GetName().c_str());     //Whether the price is linear or geometric.
 
     int index = baseUnit->cargo_hold.GetIndex(m_selectedItem);
     if( index == -1) {
@@ -3298,13 +3281,7 @@ void BaseComputer::BuyUpgradeOperation::selectMount(void) {
     for (int i = 0; i < playerUnit->getNumMounts(); i++) {
         //Mount is selectable if we can upgrade with the new part using that mount.
         double percent;             //Temp.  Not used.
-        const bool selectable = playerUnit->canUpgrade(m_newPart,
-                i,
-                m_selectedTurret,
-                m_addMultMode,
-                false,
-                percent,
-                m_theTemplate);
+        const bool selectable = playerUnit->canUpgrade(m_newPart, i, m_selectedTurret, 0, false, percent);
 
         //Figure color and label based on weapon that is in the slot.
         GFXColor mountColor = MOUNT_POINT_NO_SELECT();
@@ -3340,8 +3317,7 @@ bool BaseComputer::BuyUpgradeOperation::checkTransaction(void) {
         return false;         //We want the window to die to avoid accessing of deleted memory.
     }
     double percent;         //Temp.  Not used.
-    if (playerUnit->canUpgrade(m_newPart, m_selectedMount, m_selectedTurret, m_addMultMode, false, percent,
-            m_theTemplate)) {
+    if (playerUnit->canUpgrade(m_newPart, m_selectedMount, m_selectedTurret, 0, false, percent)) {
         //We can buy the upgrade.
         concludeTransaction();
         return false;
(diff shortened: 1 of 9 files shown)


evertvorster:
Thanks for making this fix. 
However, now when I start a new campaign and go to the ship upgrade, I see a much reduced list of standard equipment:
image


A big part of the game was that your armor and hull could get damaged, and needed repairs. 
I don't see armor, shields, skyscope, reactor and probably a few other items that I am used to seeing in this display. 

The mass off the vessel is displayed correctly, and I do like the distinction between ship and cargo that now shows up. You could even extend that a little further, as the ship without upgrades is supposed to be the base mass, upgrades have a mass, and then cargo has more mass. 

I'll continue testing, but from a preliminary point of view, there are a few things missing in your ship display that need to be there.

royfalk:
Interesting. My saved game loaded these correctly I think.
I'll start a new game and see what changed.

evertvorster:
Yeah, I was not loading a saved game, I was starting a new campaign each time.

royfalk:
This is a separate assets bug - https://github.com/vegastrike/Assets-Masters/issues/81.
I'll push a PR momentarily to fix it. I wonder if this also fixes Oswald.

evertvorster (approved):
Starting a new campaign works. 
Fixing equipment works, and you can see this in the HUD with damaged vs fixed equipment. 

Most importantly, mass now is tracked properly. Buying stuff increases the mass of your ship. 
Selling stuff decreases the mass of your ship. 
This mass is preserved with save files. 

Looks all good!

engine/src/cmd/basecomputer.cpp
```diff
@@ -3298,13 +3281,7 @@ void BaseComputer::BuyUpgradeOperation::selectMount(void) {
...
         //Mount is selectable if we can upgrade with the new part using that mount.
         double percent;             //Temp.  Not used.
-        const bool selectable = playerUnit->canUpgrade(m_newPart,
-                i,
-                m_selectedTurret,
-                m_addMultMode,
-                false,
-                percent,
-                m_theTemplate);
+        const bool selectable = playerUnit->canUpgrade(m_newPart, i, m_selectedTurret, 0, false, percent);
```
stephengtuggy, line 3284: Why are we hard-coding `0` here instead of `m_addMultMode`? I'm a bit confused.

engine/src/cmd/basecomputer.cpp
```diff
@@ -3340,8 +3317,7 @@ bool BaseComputer::BuyUpgradeOperation::checkTransaction(void) {
         return false;         //We want the window to die to avoid accessing of deleted memory.
     }
     double percent;         //Temp.  Not used.
-    if (playerUnit->canUpgrade(m_newPart, m_selectedMount, m_selectedTurret, m_addMultMode, false, percent,
-            m_theTemplate)) {
+    if (playerUnit->canUpgrade(m_newPart, m_selectedMount, m_selectedTurret, 0, false, percent)) {
```
stephengtuggy, line 3320: Same here

engine/src/components/components_manager.cpp
```diff
@@ -91,6 +94,14 @@ void ComponentsManager::SetMass(double mass) {
     this->mass = mass;
 }
 
+void ComponentsManager::SetPlayerShip() {
+    player_ship = true;
+}
+
+bool ComponentsManager::PlayerShip() {
```
stephengtuggy, line 101:
```suggestion
bool ComponentsManager::IsPlayerShip() {
```

engine/src/components/components_manager.cpp
```diff
@@ -274,24 +285,23 @@ std::string ComponentsManager::GetTitle(bool show_cargo, bool show_star_date, st
     
     // Cargo mass renders your ship harder to manoeuver. Display it.
     double mass_percent = mass / base_mass * 100;
+    const std::string mass_string = (boost::format("base %1%/ current %2% (%3$.0f%%)") % base_mass % mass % mass_percent).str();
```
stephengtuggy, line 288: Nice refactor

libraries/cmd/unit_generic.cpp
```diff
@@ -2953,9 +2838,11 @@ bool Unit::RepairUpgradeCargo(Cargo *item, Unit *baseUnit) {
                 if (price <= ComponentsManager::credits) {
                     ComponentsManager::credits -= price;
                     
-                    if (notadditive) {
-                        this->Upgrade(un, 0, 0, 0, true, percentage, makeTemplateUpgrade(this->name, this->faction));
-                    }
+                    // TODO: look closer at this
+                    //if (notadditive) {
+                    //    this->Upgrade(un, 0, 0, 0, true, percentage, makeTemplateUpgrade(this->name, this->faction));
+                    //}
```
stephengtuggy, line 2844: I think this should be left in, not commented out. It looks to me like, overall, this PR is going to break multiplicative upgrades.

stephengtuggy: (if left in its current form)

royfalk: I've deprecated both additive and multiplicative upgrades as generalized upgrades several months ago.
Multiplicative exists as a special case for engine_upgrade and afterburner_upgrade.

Looking at what was supported by the game, additive was only supported for SPEC capacitors and cargo expansions. 
Multiplicative was supported for afterburner upgrade and various shady.
Both cargo upgrades (mult_shady_moreupgrade__upgrades and add_cargo_expansion__upgrades) are currently supported. As for the rest, I can open an issue and figure out a way to support them.

stephengtuggy: Well, OK, but... why break this functionality in the meantime if you don't have to?

royfalk: This is a somewhat (much) longer discussion about the engine in general and upgrades in particular, but I'll try to summarize some relevant points:
The old system generated an **upgrade_struct** to modify the **ship_struct**. This had all kinds of implications such as:
1. Hard to unit test a specific upgrade
2. You could easily generate upgrade combinations such as shady improve x and damage y.
3. You did not store the original. This was a real issue. For every upgrade you had complex calculations of revert_to_blank and generate_template, which conveniently were spread all over the engine. This mass bug is a result of such a calculation applying cargo mass again.
4. It was convenient to introduce macros in order to do the actual upgrade of a variable.
5. The whole asset system used prefixes and suffixes to convey meaning to the engine. __upgrades is an upgrade. _ammo is an ammo. add prefix is for additive. mult is for multiplicative. basecomputer.cpp had many many lines of codes just to display what these did.
6. Upgrades are a special case of cargo. It still is. I haven't figured out an elegant way to change this. Weapons are a special case, period. Take a look at my reload PR. You need a Cargo for things like mass and price. A unit representing a weapon. In it is a mount which stores max ammo and weapon type. You take the weapon type from weapons.json. It wouldn't be a bad idea to deprecate unit entries for upgrades and move everything to cargo,but that's not a small update.

A lot of the complexity of the code comes from this architecture. A lot of the existing checks are there because it is easy to get an empty cargo entry for example or get an weapon upgrade (units.json) that doesn't have a corresponding weapons entry in weapons.json.

stephengtuggy: Thanks for the explanation. That makes sense. Approving.

engine/src/cmd/basecomputer.cpp
```diff
@@ -3013,18 +3011,10 @@ bool BaseComputer::changeToUpgradeMode(const EventCommandId &command, Control *c
...
-        if (UnitUtil::getCredits(parent) < basicRepairPrice() * repairmultiplier) {
-            showAlert("You don't have enough credits to repair your ship.");
-        } else if ((repairmultiplier = parent->RepairUpgrade())) {
-            UnitUtil::addCredits(parent, -basicRepairPrice() * repairmultiplier);
-        } else {
-            showAlert("Your ship has no damage.  No charge.");
-        }
-    }
+    // This function does nothing. Kept for compatibility with python API.
+    // TODO: remove
```
BenjamenMeyer, line 3017: @royfalk  You can remove in 0.10.x; just record it in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1287 and https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Vega-Strike-Engine-API; though we'll probably just move to what we have internally once that is merged.

We really need to keep track of the changes we've made to the Python API to help the game asset authors.

royfalk: I really can't. Some of these functions are used in the current game or in privateer and I need to figure out a way to remove them while keeping the functionality.

BenjamenMeyer: that's fair - we probably should mark them as DEPRECATED so we can start removing them - we could start by using https://docs.python.org/3/library/warnings.html#warnings.deprecated in the Python code directly; not sure how we integrate that with the C++ side.

PR #1276, opened 2025-08-27
Add ability to reload guns 

royfalk:
- Add description of ammo for guns with ammo
- Add reload button when supported and needed
- Display reload button only if gun isn't damaged. Otherwise display fix instead (not ideal)
- Code is by weapon type. So if one microdriver is damaged, you can't reload the other.
- Reload happens one gun at a time of same type. If you have 3 microdrivers with 88%/33%/2%, you'd have to reload the first two first.

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Mostly tinkered with the upgrade computer. Basic sanity in flight.

Issues:
- See above
- Did not delete units such as Hephaestus_Mini_ammo__upgrades yet. Once approved, we could delete real ammo but I'm not sure about missiles.

### DIFF
--- a/engine/src/cmd/basecomputer.h
+++ b/engine/src/cmd/basecomputer.h
@@ -149,6 +149,8 @@ protected:
     bool sellUpgrade(const EventCommandId &command, Control *control);
 //Fix an upgrade on your ship.
     bool fixUpgrade(const EventCommandId &command, Control *control);
+//Reload a gun on your ship.
+    bool reloadUpgrade(const EventCommandId &command, Control *control);
 //Buy ship from the base.
     bool buyShip(const EventCommandId &command, Control *control);
 //Sell ship from your stock
--- a/engine/src/components/components_manager.cpp
+++ b/engine/src/components/components_manager.cpp
@@ -55,7 +55,10 @@ void ComponentsManager::Load(std::string unit_key) {
     for (const std::string& upgrade : upgrades) {
         std::vector<std::string> parts;
         boost::split(parts, upgrade, boost::is_any_of(":"));
-        if (parts.size() == 2) {
+        if (parts.size() == 1) {
+            const std::string category = parts[0];
+            prohibited_upgrades.emplace_back(category, 0);
+        } else if (parts.size() == 2) {
             const std::string category = parts[0];
             const int limit = std::stoi(parts[1]);
             //const std::pair<const std::string, const int> pair(category, limit);
@@ -89,6 +92,14 @@ double ComponentsManager::GetMass() const {
 
 void ComponentsManager::SetMass(double mass) {
     this->mass = mass;
+}
+
+void ComponentsManager::SetPlayerShip() {
+    player_ship = true;
+}
+
+bool ComponentsManager::PlayerShip() {
+    return player_ship;
 }
 
 void ComponentsManager::DamageRandomSystem() {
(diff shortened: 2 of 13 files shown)


stephengtuggy:
Why are we tinkering with stuff like this when we haven't finished converting settings yet? I would really appreciate a review or two on #1263 , so I can move forward to converting the next chunk of settings. I am currently blocked.

I know #1263 looks huge, but most of the lines changed are in configuration.h and configuration.cpp (the code auto-generated by JsonStruct). My hope is that you can just skim through those.

royfalk:
> Why are we tinkering with stuff like this when we haven't finished converting settings yet? 

1. Because this is fun and sometimes I like to write code that gives me joy.
2. Because it annoyed me that you would need to find a place stocking metal balls in order to buy a reload.
3. Because it really annoyed me to see all those extra entries, that were basically duplicates.

> I would really appreciate a review or two on https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1263 , so I can move forward to converting the next chunk of settings. I am currently blocked.

Sorry. I'll get on it tomorrow morning.

stephengtuggy:
> > Why are we tinkering with stuff like this when we haven't finished converting settings yet?
> 
>     1. Because this is fun and sometimes I like to write code that gives me joy.
> 
>     2. Because it annoyed me that you would need to find a place stocking metal balls in order to buy a reload.
> 
>     3. Because it really annoyed me to see all those extra entries, that were basically duplicates.

Makes sense. Sorry for being grouchy about it.

> > I would really appreciate a review or two on #1263 , so I can move forward to converting the next chunk of settings. I am currently blocked.
> 
> Sorry. I'll get on it tomorrow morning.

Thanks!

engine/src/cmd/basecomputer.h
```diff
@@ -149,6 +149,8 @@ class BaseComputer : public WctlBase<BaseComputer> {
     bool sellUpgrade(const EventCommandId &command, Control *control);
 //Fix an upgrade on your ship.
     bool fixUpgrade(const EventCommandId &command, Control *control);
+//Reload a gun on your ship.
+    bool reloadUpgrade(const EventCommandId &command, Control *control);
```
stephengtuggy, line 153: What do you think of calling it `reloadGun`?

engine/src/components/components_manager.cpp
```diff
@@ -91,6 +94,14 @@ void ComponentsManager::SetMass(double mass) {
     this->mass = mass;
 }
 
+void ComponentsManager::SetPlayerShip() {
+    player_ship = true;
+}
+
+bool ComponentsManager::PlayerShip() {
```
stephengtuggy, line 101: Maybe call it `IsPlayerShip`?

libraries/cmd/unit_generic.cpp
```diff
@@ -2825,72 +2763,19 @@ Vector Unit::MountPercentOperational(int whichmount) {
                     == Mount::INACTIVE) ? 0.0 : (mounts[whichmount].status == Mount::UNCHOSEN ? 2.0 : 1.0)));
 }
 
+// TODO: remove function
+// We no longer do repair through basic repair.
```
stephengtuggy, line 2767: So we're getting rid of Basic Repairs entirely? I'm not sure how I feel about that from a gameplay perspective. I guess we can merge this as is, and then start a discussion about which way we want to go with this.

engine/src/cmd/basecomputer.h
BenjamenMeyer: or `reloadWeapon`?

PR #1277, opened 2025-08-28
Fix cargo purchase bug 1267

### LINKED ISSUE #1267: Bug in game's economics model
In Cephid 17, I flew to Ataraxia, and in the trade center, I was able to buy 1 unit of Gold under Precious Metals (that was all the base had). I then flew to Serenity, where they had 21 units of Gold in the trade center. However, when I purchased the 21 units my credits decreased by the cost of the 21 units, the Gold disappeared from the Buy window, but my inventory did not change from the 1 unit I already had. In other words, I paid credits and did not receive what I purchased. This is a bit of a problem.

It appears that if you already have a product in your ship's inventory, you cannot buy it anywhere else until you sell what you have, even though you will be charged the cost of it.

royfalk:
Closes #1267 

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. No. Only checked behavior of cargo purchase and sale.

Issues:
- Had to disable space checks on the base side. For items smaller than 1 the value of available space in units of item exceeded max_int and became negative.

@evertvorster, I'd appreciate a play test of this PR.

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -2482,20 +2482,17 @@ void BaseComputer::draw() {
 bool BaseComputer::buySelectedCargo(int requestedQuantity) {
     Unit *playerUnit = m_player.GetUnit();
     Unit *baseUnit = m_base.GetUnit();
-    if (!(playerUnit && baseUnit)) {
+    Cargo *item = selectedItem();
+
+    if (!(playerUnit && baseUnit && item)) {
         return true;
     }
-    Cargo *item = selectedItem();
-    if (item) {
-        int quantity = (requestedQuantity <= 0 ? item->GetQuantity() : requestedQuantity);
-        quantity = maxQuantityForPlayer(*item, quantity);
-        int index = baseUnit->cargo_hold.GetIndex(*item);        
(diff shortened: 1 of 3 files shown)


evertvorster (approved):
This fixes the cargo buying bug. 
I did not test beyond buying a few units, selling a few units and filling up my hold with all the units. 
One thing that might be an issue later, but definitely not blocking merging this in, is what happens if a location already has 9999 units of whatever you are selling?

Is 9999 some sort of shorthand for infinite? What happens if you already have 9999 units of whatever it is that you are buying? Is the sale refused? Does the unit just disappear? Is another digit added?

Are we assuming this could just never happen?

For instance, Atlantis has 9999 units of water. If I then try to sell another unit of water picked up in a previous save game, this is certainly a situation that can happen. 

We can assume that a location has near enough infinite storage space, so selling units to a place that already has 9999 units of the same commodity should give you the cash from the sale, but the display should stay 9999 if that is indeed the maximum number a location can hold. 

Going the other way, I don't think there is a cargo hold that can take 9999 of anything, so this point might be moot as this is a number that can never be reached.

stephengtuggy (approved):
LGTM

PR #1287, opened 2025-08-28
Enhancement: Documentation

BenjamenMeyer:
Add a documentation spot in the code for the changes to the Engine API, aka the Python API, so we can ensure we record changes as they happen.

Please answer the following:

Code Changes:

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Give us a space to record changes to the Engine (Python) API as the happen.
- What release is this for? 0.10.x but we should probably backport it to 0.9.x; not a huge deal if we don't though.

### DIFF
--- a/doc/Engine-API.md
+++ b/doc/Engine-API.md
@@ -10,7 +10,7 @@ if they were compatible with a given version of the engine. Starting in Vega Str
 the Engine API Version tooling was introduced.
 
 If the Engine API is not available, then the game assets can assume that the running Vega Strike Engine
-is versioni 0.7.x or earlier. No guarentees are then made to the stability of the API or anything being
+is version 0.7.x or earlier. No guarantees are then made to the stability of the API or anything being
 available.
 
 The below documents the changes to the Engine API starting with Vega Strike Engine 0.8.0, Engine API Version 1.


doc/Engine-API.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Overview
+
+The Engine API is utilized by the game assets in order to produce a working game.
+Historically this was not well documented and the game assets were unable to determine
+if they were compatible with a given version of the engine. Starting in Vega Strike Engine 0.8.0
+the Engine API Version tooling was introduced.
+
+If the Engine API is not available, then the game assets can assume that the running Vega Strike Engine
+is versioni 0.7.x or earlier. No guarentees are then made to the stability of the API or anything being
```
stephengtuggy:
```suggestion
is version 0.7.x or earlier. No guarantees are then made to the stability of the API or anything being
```

stephengtuggy (commented):
Found a couple of typos on one line, but other than that, it looks good.

doc/Engine-API.md
royfalk: I think I'm missing something. You can't run engine 0.7 with assets 0.8. The game won't run.

BenjamenMeyer: @royfalk the 0.8.0 assets should run on a 0.7.0 engine since we were avoiding changes at the Python API level during that time. I won't guarantee it but it should work.

Historically the game would run and you'd play until something didn't align and then it'd crash and you wouldn't know why. The Engine API was a method introduced to help the game assets detect that up front to prevent that scenario, and this documentation is for folks to be able to better understand how to update the game assets for newer engine versions.

PR #1296, opened 2025-08-29
Add virtual destructors to various classes

stephengtuggy:
Code Changes:

Issues Addressed:
- N/A

Known Issues Remaining:
- There are probably still be some classes in the engine that need virtual destructors.

Purpose:
- What is this pull request trying to do? Add virtual destructors to a bunch of classes that need them, as they are polymorphic

### DIFF
--- a/script/bootstrap-mac.sh
+++ b/script/bootstrap-mac.sh
@@ -54,7 +54,6 @@ echo "Detected Mac OS X ${DETECT_MAC_OS_VERSION} - ${MAC_OS_NAME}"
 # Install the stuff we know needs to get installed all the time
 brew install \
     gcc \
-    cmake \
     python3 \
     boost-python3 \
     jpeg \
@@ -62,6 +61,9 @@ brew install \
     gtk+3 \
     gtkglext \
     sdl2
+
+# Only install cmake if it isn't installed yet
+brew ls --versions cmake || brew install cmake
 
 # The following Apple-provided libraries are deprecated:
 # * OpenGL as of macOS 10.14


stephengtuggy:
Hmm. Both macOS builds are suddenly failing, That's not so good.

The error message says something about cmake already being installed. Shall we just remove cmake from the list of items to install using Homebrew in bootstrap-mac.sh?

libraries/gfx_generic/mesh.h
```diff
@@ -59,12 +59,12 @@ struct mesh_polygon {
 };
 /**
  * Mesh FX stores various lights that light up shield or hull for damage
- * They may be merged and they grow and shrink based on their TTL and TTD and delta values
+ * They may be merged, and they grow and shrink based on their TTL and TTD and delta values
```
royfalk, line 62: I'm nitpicking, I know. This PR is fine.
Why the comma?

libraries/gfx_generic/mesh.h
```diff
@@ -59,12 +59,12 @@ struct mesh_polygon {
...
  * They must be updated every frame or every physics frame if not drawn (pass in time)
  */
 class MeshFX : public GFXLight {
 public:
-///The ammt of change that such meshFX objects attenuation get
```
royfalk, line 67: Why not amount?

libraries/gfx_generic/mesh.h
stephengtuggy: That's a CLion grammar suggestion. I think CLion has it right in this case.

libraries/gfx_generic/mesh.h
stephengtuggy: Oh, good point.

stephengtuggy: Fixed in the next PR that I'm about to create.

PR #1309, opened 2025-08-31
Fix bug where repair upgrades costs 0

### LINKED ISSUE #1292: Able to repair ship upgrades for free
I'm noticing that I can repair damaged ship upgrades for 0.00. Is that intentional?

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/discussions/1285#discussioncomment-14251580_

royfalk:
Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation Only the bug fix at the moment. Will try to do it tomorrow.

Issues:
- We are in transition from upgrades as cargo and upgrades as components. Need to finish this transition to simplify code.

close #1292


engine/src/cmd/basecomputer.cpp
```diff
@@ -396,8 +397,25 @@ static float basicRepairPrice(void) {
...
+    if(component) {
+        return component->PercentOperational();
+    }
+
+    return 0.0;
+}
+
+static double RepairPrice(Unit *playerUnit, const Cargo *item) {
+    // TODO: * configuration()->general.difficulty;
+    return 0.8 * (1-GetOperational(playerUnit, item)) * item->GetPrice();
```
stephengtuggy: 0.8 is a magic number. You might want to make it a named constant. Something like:

```suggestion
    constexpr double kRepairPriceCoefficient = 0.8;
    return kRepairPriceCoefficient * (1-GetOperational(playerUnit, item)) * item->GetPrice();
```

stephengtuggy (approved):
Just one minor nitpick. Looks fine overall.

engine/src/cmd/basecomputer.cpp
royfalk: It replaced 0.5 as the magic number and I incorrectly assumed the TODO was about that. 
This was a rush job. 
I'll fix this tomorrow before I test it.

PR #1312, opened 2025-09-01
Convert AI and Unit Audio settings

### LINKED ISSUE #1180: Migrate `ai` settings to new mechanism

stephengtuggy:
Code Changes:

Issues Addressed:
- Closes #1180 

Known Issues Remaining:
- Nothing new that I know of

Purpose:
- What is this pull request trying to do? Convert two more sections' settings to the new `configuration()....` mechanism

### DIFF
--- a/engine/src/cmd/beam.cpp
+++ b/engine/src/cmd/beam.cpp
@@ -607,19 +607,20 @@ bool Beam::Collide(Unit *target, Unit *firer, Unit *superunit) {
                         if (un->cargo_hold.CanAddCargo(adder)) {
                             un->cargo_hold.AddCargo(un, adder);
                             if (un->IsPlayerShip()) {
-                                static int tractor_onboard =
-                                        AUDCreateSoundWAV(vs_config->getVariable("unitaudio", "player_tractor_cargo",
-                                                "tractor_onboard.wav"));
-                                AUDPlay(tractor_onboard, QVector(0, 0, 0), Vector(0, 0, 0), 1);
+                                static boost::optional<int> tractor_onboard;
+                                if (tractor_onboard == boost::none) {
+                                    tractor_onboard = AUDCreateSoundWAV(configuration().audio.unit_audio.player_tractor_cargo);
+                                }
+                                AUDPlay(tractor_onboard.get(), QVector(0, 0, 0), Vector(0, 0, 0), 1);
                             } else {
                                 Unit *tmp = _Universe->AccessCockpit()->GetParent();
                                 if (tmp && tmp->owner == un) {
                                     //Subunit of player (a turret)
-                                    static int tractor_onboard_fromturret =
-                                            AUDCreateSoundWAV(vs_config->getVariable("unitaudio",
-                                                    "player_tractor_cargo_fromturret",
-                                                    "tractor_onboard.wav"));
-                                    AUDPlay(tractor_onboard_fromturret, QVector(0, 0, 0), Vector(0, 0, 0), 1);
+                                    static boost::optional<int> tractor_onboard_fromturret;
+                                    if (tractor_onboard_fromturret == boost::none) {
+                                        tractor_onboard_fromturret = AUDCreateSoundWAV(configuration().audio.unit_audio.player_tractor_cargo_fromturret);
+                                    }
+                                    AUDPlay(tractor_onboard_fromturret.get(), QVector(0, 0, 0), Vector(0, 0, 0), 1);
                                 }
                             }
                             target->Kill();
--- a/libraries/cmd/audible.cpp
+++ b/libraries/cmd/audible.cpp
@@ -45,12 +45,12 @@ Audible::Audible() {
         sounds[soundType] = -1;
     }
 
-    defaultSoundNames[SoundType::engine] = vs_config->getVariable("unitaudio", "afterburner", "sfx10.wav");
-    defaultSoundNames[SoundType::shield] = vs_config->getVariable("unitaudio", "shield", "sfx09.wav");
-    defaultSoundNames[SoundType::armor] = vs_config->getVariable("unitaudio", "armor", "sfx08.wav");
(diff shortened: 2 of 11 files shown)


evertvorster (approved):
The game plays normally with this PR. 
Every function I tested is still the same, and sounds are normal too. 
Watched a swarm of ships fly around and dock with Serenity, so it all looks OK.

engine/src/cmd/beam.cpp
```diff
@@ -607,19 +607,20 @@ bool Beam::Collide(Unit *target, Unit *firer, Unit *superunit) {
                         if (un->cargo_hold.CanAddCargo(adder)) {
                             un->cargo_hold.AddCargo(un, adder);
                             if (un->IsPlayerShip()) {
-                                static int tractor_onboard =
-                                        AUDCreateSoundWAV(vs_config->getVariable("unitaudio", "player_tractor_cargo",
-                                                "tractor_onboard.wav"));
-                                AUDPlay(tractor_onboard, QVector(0, 0, 0), Vector(0, 0, 0), 1);
+                                static boost::optional<int> tractor_onboard;
+                                if (tractor_onboard == boost::none) {
+                                    tractor_onboard = AUDCreateSoundWAV(configuration().audio.unit_audio.player_tractor_cargo);
```
royfalk, line 612: I see why you did that but I wonder if we can't just create all game sounds after loading the game configuration.

engine/src/cmd/beam.cpp
```diff
@@ -607,19 +607,20 @@ bool Beam::Collide(Unit *target, Unit *firer, Unit *superunit) {
...
                                 if (tmp && tmp->owner == un) {
                                     //Subunit of player (a turret)
-                                    static int tractor_onboard_fromturret =
-                                            AUDCreateSoundWAV(vs_config->getVariable("unitaudio",
-                                                    "player_tractor_cargo_fromturret",
-                                                    "tractor_onboard.wav"));
-                                    AUDPlay(tractor_onboard_fromturret, QVector(0, 0, 0), Vector(0, 0, 0), 1);
+                                    static boost::optional<int> tractor_onboard_fromturret;
+                                    if (tractor_onboard_fromturret == boost::none) {
+                                        tractor_onboard_fromturret = AUDCreateSoundWAV(configuration().audio.unit_audio.player_tractor_cargo_fromturret);
```
royfalk, line 621: We'd need to figure out what the int values are for each, though I assume if we do them in order, we could hard code them somewhere.

libraries/cmd/audible.cpp
```diff
@@ -100,24 +100,33 @@ void Audible::playSound(SoundType type) {
 }
```
royfalk, line 101: Maybe something like:
```
static int Audible::playerShieldSound = 0;
static int Audible::playerSomethingElse = 0;

static Audible::initSounds() {
   Audible::playerShieldSound = AUDCreateSoundWAV(configuration().audio.unit_audio.player_shield_hit);
}
```

libraries/cmd/movable.cpp
```diff
@@ -777,15 +777,16 @@ void Movable::Thrust(const Vector &amt1, bool afterburn) {
...
+        static boost::optional<int> player_engine;
+        if (player_engine == boost::none) {
+            player_engine = AUDCreateSound(configuration().audio.unit_audio.player_afterburner, true);
+        }
         const float enginegain = configuration().audio.afterburner_gain_flt;
-        if (afterburn != AUDIsPlaying(playerengine)) {
+        if (afterburn != AUDIsPlaying(player_engine.get())) {
             if (afterburn) {
-                AUDPlay(playerengine, QVector(0, 0, 0), Vector(0, 0, 0), enginegain);
+                AUDPlay(player_engine.get(), QVector(0, 0, 0), Vector(0, 0, 0), enginegain);
```
royfalk, line 787: Given how spread out these are, that's not a bad idea to centralize init somewhere. It doesn't actually have to be _Audible_. Just Somewhere.
If you don't want to do this but think this isn't a bad idea, open an issue and I'll take a crack at it.
If you think it is a bad idea, tell me why.

libraries/cmd/audible.cpp
stephengtuggy: And where would we call Audible::initSounds() from?

libraries/cmd/movable.cpp
stephengtuggy: I'm still thinking about it. The main stumbling block I see is making sure to load the sounds after the config has been populated with the actual settings from the applicable files. It seems like this could be tricky to do. I think it will work the way I'm doing it in this PR, but perhaps there's another way. If so, feel free to take a crack at it.

stephengtuggy: This isn't the first time I've used `boost::optional`s to delay loading the values for static variables.

engine/src/cmd/beam.cpp
stephengtuggy: I think these are sound handles, and as such, their values may change from one run to the next.

libraries/cmd/movable.cpp
stephengtuggy: Basically, any time we're populating a complex type from game settings, like a class or struct, I've been using this `boost::optional<...>` mechanism. With the exception of strings, which are their own beast.

libraries/cmd/audible.cpp
stephengtuggy: I know! Perhaps we could check here and elsewhere for the condition where the sound filenames to load have been populated, but the sounds themselves -- and therefore their handles -- have not. When this situation is encountered, call `initSounds`.

What do you think?

stephengtuggy:
```c++
if (time_to_load_sounds()) {
    load_sounds();
}
```

royfalk (approved):
Please Open an issue to track my suggestion. 
Otherwise, looks good.

PR #1330, opened 2025-09-12
Delete more network code

royfalk:
Delete startMenuInterface, which has been deprecated in configuration. Also delete gamemenu, which isn't used.

**Code Changes:**

**Testing Done**

- Went through all the menus in the computer to look for issues. Info, news, upgrades, cargo, ship dealer, etc.
- Bought sold stuff. Bought sold upgrades.
- Played a short game. Flew to Atraxia. Picked a fight with it. Died.

**Issues:**
- Armor in manifest reports damage incorrectly. When damaged, it does not go down properly. Instead, when hull is damaged, this appears to "damage" the armor. Not related. Will open issue.


engine/objconv/basemaker/base_maker.cpp
```diff
@@ -80,252 +80,6 @@ int nb_checks = 1;
 double last_check = 1;
 double cur_check = 1;
 
-/*
```
stephengtuggy, line 83: This may have been a mistake to delete the code from this particular file. Will need to revisit.

royfalk: BaseMaker itself is something I haven't used at all. AFAICT, it's a way to run base code without running the game. 

This specific code inside was all commented out, including a second copyright notice.

If you still feel strongly about it, I suggest you either uncomment it and figure out if it's critical (probably not) or open an issue. 

There are a whole bunch of places where the original devs commented out stuff for "5 minutes". You have to be part archeologist, part seer to figure out what and why.

stephengtuggy: Sure enough, it was commented out. That's fine then. Never mind.

PR #1333, opened 2025-09-16
Separate DrawNodeSystem and NavigationSystem from drawgalaxy

royfalk:
Move these two classes to their own files.
Give them header files.

I was working on something else and was forced to clean this. Decided to split this into its own PR.

**Code Changes:**

Issues:
- Auto-pilot is not usable
- Some systems are impossible to fly. Gravity from sun or some such. Also, killed by sun when using auto-pilot.


engine/src/gfx/nav/navigation_system.cpp
```diff
@@ -0,0 +1,796 @@
...
+
+#include "navigation_system.h"
+
+#include "system_draw_node.h"
+#include "universe.h"
+#include "universe_util.h"
+
+#include "root_generic/vs_globals.h"
+
+// Can't include because drawgalaxy.cpp doesn't have a header file
```
BenjamenMeyer: we could always add a header file ;)

engine/src/gfx/nav/system_draw_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "system_draw_node.h"
+
+// Can't include header because drawgalaxy.cpp doesn't have one
```
BenjamenMeyer: especially if we need to reference it multiple times it'd probably be better to go ahead and define a header file ;)

BenjamenMeyer (approved):

Issue #492, opened 2021-07-08
Faction colors missing in navigation computer


evertvorster:
Just did a little play testing on the latest master. (2021-07-08)

There are a couple of new issues, all small. 
First is that the faction colors are missing from the Navigation Computer. 
All the other colors are still present, so it's not the color config..

evertvorster:
Yip, this one is still an issue. 
When you have a large part of the universe explored, it's helpful to have the faction colors on the lines of the navigation computer to show you where the various factions are hanging out.

evertvorster:
OK, this one is still very much an issue. With the latest 0.8.x engine, the factions are displayed in the navigation computer, and with the latest version of the 0.9.x engine, all the colors of the systems in the navigation computer are just one color.

evertvorster:
I made a short Youtube video showing how this bug affects game play:
https://youtu.be/ans6G5iGaJU

royfalk:
This not an issue with the display. 
Kills are not being recorded properly.
If you kill some ships and then dock and go to base computer | factions view, you'll not see your kills.

evertvorster:
I don't see what this has to do with kills. 
In the navigation computer, on the starmap, the factions controlling various star systems would be displayed via the star systems being colored by the faction color. 
Right now, all the star systems has the same color, so the faction colors are gone.

royfalk:
It may not be, but so far that's where my investigation is leading me. 
If it turns out that these are two separate bugs, I promise I'll create a new one for the kills and revert the name of this one.

PR #1340, opened 2025-09-18
Fix 1337 - prevent the creation of more than 3 layers

### LINKED ISSUE #1337: Units are created with six layers instead of three.
This does not impact typical gameplay but has some effect on functions relying on detecting the death of a ship.

royfalk:
Closes #1337 

**Code Changes:**

**Issues:**
- None identified

### DIFF
--- a/engine/src/cmd/planet.cpp
+++ b/engine/src/cmd/planet.cpp
@@ -417,7 +417,6 @@ void Planet::InitPlanet(QVector x,
     }
 
     curr_physical_state.position = prev_physical_state.position = cumulative_transformation.position = orbitcent + x;
-    Init();
     //static int neutralfaction=FactionUtil::GetFaction("neutral");
     //this->faction = neutralfaction;
     killed = false;
--- a/engine/src/damage/damageable_object.cpp
+++ b/engine/src/damage/damageable_object.cpp
@@ -46,6 +46,7 @@ DamageableObject::DamageableObject(std::vector<DamageableLayer*> layers):
 void DamageableObject::AddLayer(DamageableLayer* layer) {
     number_of_layers++;
     layers.push_back(layer);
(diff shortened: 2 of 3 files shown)


engine/src/cmd/planet.cpp
```diff
@@ -417,7 +417,6 @@ void Planet::InitPlanet(QVector x,
     }
 
     curr_physical_state.position = prev_physical_state.position = cumulative_transformation.position = orbitcent + x;
-    Init();
```
BenjamenMeyer, line 420: does this also need to go away from L203:
```
Planet::Planet() :
        Unit(0) {
    Init();
    SetAI(new Order());     //no behavior
}
```

I see the comment about `Init()` needing to go away in `unit_generic.h` but want to make sure this is correct as it seems necessary for a class Constructor to be calling it.

BenjamenMeyer (commented):
Question before I approve

engine/src/cmd/planet.cpp
royfalk: So there are a couple of unrelated stuff here, all mixed together.
My comment about deleting Init is from 2021. The original devs used Init as some kind of substitute for constructor delegation. I've been steadily removing stuff from the Init functions as a low hanging fruit.

By all rights, both Planet::Planet() and Unit::Unit(int) should call Init, resulting in six layers and tripping the assert. But I ran the game, both new and saved and it doesn't.

I'm loathe to touch this further here. I'm juggling other commits in parallel and this was a bug discovered while trying to figure out #492.

BenjamenMeyer: @royfalk cool - just wanted to validate. Generally I do think we should be moving to using the constructors properly instead of delegating out where we can; but yes - let's focus things where possible too.

PR #1332, opened 2025-09-13
Fix mouse cursor and move to hardware cursor

royfalk:
**Overview**

Currently, VS has a combination of software mouse and elements with non-standard coordinates. Instead of a window standard coordinates:
(0,0)-----------------(max_x,0)
|                                  |
|                                  |
|                                  |
(0,max_y)--------(max_x,max_y)

It has the following:
(-1,1)-----------------(1,1)
|                                  |
|                                  |
|                                  |
(-1,-1)--------(1,-1)

AFAICT, these are the original WC coordinates. Obviously, we can't integrate new widgets into the game until we fix this. This PR does that. I introduced new files gldrv/mouse_cursor.h/cpp to convert coordinates and otherwise manage the hardware cursor.

This PR also gets rid of the weird non-linear mouse movement.
The PR adds support for SDL2_image in order to load the PNG files.
This PR requires a change to the assets - copying the standard PNG files of the mouse cursors to be loaded by SDL.

**Code Changes:**
- [x] Have the PR Validation Tests been run? 

**What was tested:**
- Main menu
- Base computer - cargo, upgrades, dealer, news, info, ...
- Bar
- Flight
- Nav computer (in flight)

**Issues:**
- Clicking multi-player in main menu crashes the game. **Existing issue.**
- Clicking on public terminal in bar crashes the game. **Existing issue.**
- Right click in certain base menus physically place the cursor in incorrect places.
- In the upgrade computer, the on mouse is according to the old coordinates.
- Dragging of scrollbar is weird. Scroll bar is not moving with the cursor.
- Switching between two panels sharing a scrollbar (e.g. Player Info and Ship Stats) causes scroll bar to behave incorrectly. When switching from the much longer ship stats to player info, if the page is scrolled down, the scroll bar is sized incorrectly and can no longer be dragged. It can be moved by clicking on the arrows or below/above the active part of the scrollbar. (possible existing bug)

**What you need before running this?**
- Copy crosshairs.png, arrow.png and maybe mouseover.png from asset-masters to asset-production/sprites.
- Install libsdl2-image-dev

PR #1343, opened 2025-09-20
Fix issue 1342

### LINKED ISSUE #1342: Git master mouse cursor is not tracked properly
Tested this with a few resolutions in the game engine.
Where the game engine thinks the mouse is, and where it is displaying the mouse cursor is two different locations. 

The upper left pixel of the mouse cursor was the part of the mouse cursor that you had to position over something to be able to click it. 

Now it just seems way off, and it seems to change with different resolutions, so it is not even the same amount off every time.

royfalk:
- In full screen, force resolution to native monitor one.
- Also, force bases.max_height/width to native monitor resolution.

Fixes #1342

This PR has not been tested yet. It requires testing on a machine with multi-monitors, which I do not have.

**Code Changes:**
- [ ] Have the PR Validation Tests been run? No. Only ran the game up to main menu.

**Issues:**
- None yet found.

### DIFF
--- a/engine/src/gldrv/winsys.cpp
+++ b/engine/src/gldrv/winsys.cpp
@@ -251,6 +251,33 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
     width = configuration().graphics.resolution_x;
     height = configuration().graphics.resolution_y;
     const int screen_number = configuration().graphics.screen;
+
+    // Fix display in fullscreen
+    if(configuration().graphics.full_screen) {
+        // Get the native resolution of the selected screen.
+        SDL_DisplayMode currentDisplayMode;
+    
+        // Get the current display mode for display 0 (the primary display)
+        if (SDL_GetCurrentDisplayMode(screen_number, &currentDisplayMode) != 0) {
+            SDL_Quit();
+            const std::string error = (boost::format("Could not get display mode for display %1%!\nSDL_Error:\n%2%\n") % screen_number % SDL_GetError()).str();
+            VS_LOG_FLUSH_EXIT(fatal, error, -1);
+        }
+
+        VS_LOG(info, (boost::format("Running in full screen. Overriding requested screen resolution with native one: %1%x%2%\n") % currentDisplayMode.w % currentDisplayMode.h).str());
+        width = currentDisplayMode.w;
+        height = currentDisplayMode.h;
+        int* ptr_x = const_cast<int*>(&configuration().graphics.resolution_x);
+        int* ptr_y = const_cast<int*>(&configuration().graphics.resolution_y);
+        *ptr_x = width;
+        *ptr_y = height;
+        ptr_x = const_cast<int*>(&configuration().graphics.bases.max_width);
+        ptr_y = const_cast<int*>(&configuration().graphics.bases.max_height);
+        *ptr_x = width;
+        *ptr_y = height;
+    }
+
+
     SDL_Window *window = nullptr;
     if(screen_number == 0) {
         window = SDL_CreateWindow("Vega Strike",


evertvorster:
This seems to make it worse... 
The resolution that it picks is so high that I only see about 25% of the game screen, and now I can't pick a lower resolution. 
Like I said, I think the game engine is confused about the screen resolution and output resolution...

royfalk:
> This seems to make it worse...

How aggravating. Can you please upload a screen capture of it?
Also, I assume you're playing with the external monitor. Can you please disconnect it and see if this improves things?


> The resolution that it picks is so high that I only see about 25% of the game screen, and now I can't pick a lower resolution.

There should be a line in the log file that says: "Running in full screen. Overriding requested screen resolution with native one:". Can you please check what resolution it is selecting.

> Like I said, I think the game engine is confused about the screen resolution and output resolution...

It shouldn't anymore. This code uses SDL to check the native resolution of the screen specified in the config file. That's what it's using for both resolution_x/resolution_y and based.max_height/max_width.

Can you also try without scaling? I'll try and see if I have scaling enabled too.

royfalk:
> Can you also try without scaling? I'll try and see if I have scaling enabled too.

Nope, just tried scaling to 200% and it didn't change a thing. Game ran great.

royfalk:
@BenjamenMeyer can you check and see if you have the same issue in fullscreen or if it's just @evertvorster ?

BenjamenMeyer:
> @BenjamenMeyer can you check and see if you have the same issue in fullscreen or if it's just @evertvorster ?

I can give it a try but it'll be a few days until I'm on multi-monitor (laptop+external).

evertvorster (changes requested):
"Native Resolution" in fullscreen mode should be a configuration option. 


Quite a common use case exits where the user would pick a lower resolution that should be upscaled by SDL to the native resolution of the monitor. This works around font size issues and performance issues. 

By always applying the native resolution in fullscreen mode, this workaround is no longer possible.

royfalk:
Ok, but let's focus on the critical stuff first.

Fixing master so full screen shows up as full screen and mouse works.

Then we can implement low res upscaling in another feature.

evertvorster:
So, this PR takes options away that I use when testing, and I would rather not have it. 
I am quite capable of picking the native resolutions from the list of resolutions in the settings app myself. 

On my system, if I use a non-native resolution, SDL scales that to the full screen automatically. However, in this case I think the game engine is confused about the resolution of the monitor and gets the scaling wrong. 
With this PR, it still gets the scaling wrong, and now there is no way for me to pick a resolution that is scaled properly.

If I can't turn  this off, I would not want to have this feature at all. 
As you say, lets focus on other important things first, and we can circle back to this PR later.

BenjamenMeyer:
I'm going to hold off testing until @evertvorster is satisfied.

royfalk:
Evert, please work with me.

I've introduced #1332, which is a critical step forward. However, it is also a high risk PR. This PR is an attempt to fix a critical issue with that PR. If I can't figure this out, I'll have to revert it, which blocks a whole bunch of things, including in game settings.

> So, this PR takes options away that I use when testing, and I would rather not have it.

Can you live without it for a week? 

> On my system, if I use a non-native resolution, SDL scales that to the full screen automatically. However, in this case I think the game engine is confused about the resolution of the monitor and gets the scaling wrong.

Which is why I asked you a list of questions including checking the logs for the resolution SDL thinks is your native one.

> With this PR, it still gets the scaling wrong, and now there is no way for me to pick a resolution that is scaled properly.

I can't reproduce this on my machine, which is why I want and need your help. No one will merge this until you sign off on it.

> If I can't turn this off, I would not want to have this feature at all.
As you say, lets focus on other important things first, and we can circle back to this PR later.

Not an option. #1332 broke full screen. I have to figure this out or revert.

### Full screen with non-native resolution

I see two possible implementation options:
1. Use SDL to change the screen resolution. Store the old resolution and revert on exit. This would potentially leave the player with the system in a low resolution in case of crashes. This is the easy and obvious implementation.
2. Scale individual elements. This would mean changing every element or more likely the parent GUI element. Not impossible but quite challenging.

BenjamenMeyer:
> Use SDL to change the screen resolution. Store the old resolution and revert on exit. This would potentially leave the player with the system in a low resolution in case of crashes. This is the easy and obvious implementation.

FYI - we probably should install some handlers to capture crashes so we can do a few things before exiting; this could be one of those, as well as attempting to save a stack trace or other potentially useful information. It could also give a better user experience. Mac and Linux should work easily enough; Windows might have a harder time of this.

Would there still be times when our handler won't work? Certainly, but hopefully they'd be rare.

> Scale individual elements. This would mean changing every element or more likely the parent GUI element. Not impossible but quite challenging.

This is probably still a good thing to do as well.

evertvorster:
> Evert, please work with me.

Sure thing, I do work with you. What I am trying to prevent is us building on something fundamental that is broken, and will cause much more trouble later on. 

 
> I've introduced #1332, which is a critical step forward. However, it is also a high risk PR. This PR is an attempt to fix a critical issue with that PR. If I can't figure this out, I'll have to revert it, which blocks a whole bunch of things, including in game settings.
> 

Yes, that #1332 is the root cause of this. After spending a lot of time working on it, it can be disheartening when it breaks something else. 

> > So, this PR takes options away that I use when testing, and I would rather not have it.
> 
> Can you live without it for a week?

Just to be clear, this PR does not fix the issue introduced by #1332
Additionally, it causes more issues.  
It can stand on its own, and we don't have to merge it in a rush. 

> 
> > On my system, if I use a non-native resolution, SDL scales that to the full screen automatically. However, in this case I think the game engine is confused about the resolution of the monitor and gets the scaling wrong.
> 
> Which is why I asked you a list of questions including checking the logs for the resolution SDL thinks is your native one.
> 
> > With this PR, it still gets the scaling wrong, and now there is no way for me to pick a resolution that is scaled properly.
> 
> I can't reproduce this on my machine, which is why I want and need your help. No one will merge this until you sign off on it.
> 

Again, this PR does not fix the issue introduced by #1332. 
I can select the native resolution from the settings app, and still see the same issue. 

> > If I can't turn this off, I would not want to have this feature at all.
> > As you say, lets focus on other important things first, and we can circle back to this PR later.
> 
> Not an option. #1332 broke full screen. I have to figure this out or revert.
> 

It is hard for me to say this, but reverting #1332 might be the best way forward. 
Your hard work on it is not for nothing, you learned a lot about how all the pieces fit together, and at this point you know much more about it than I. 

Often starting fresh while knowing what to look out for can give you a new perspective. 

Maybe take a break from coding, and lay out the problem that #1332 was trying to solve in simple terms for us, and then we can brainstorm on the best way forward. Or at least give dumb ideas until one pops up that is not so dumb.

evertvorster:
FYI:
```
evert@Evert ~/.v/logs> cat vegastrike_2025-09-22_18_25_19.547860.log | grep -i running
[2025-09-22 18:25:19.547812]: running import sys
[2025-09-22 18:25:19.548485]: running import sys
[2025-09-22 18:25:19.548616]: running import sys
[2025-09-22 18:25:19.548626]: running import sys
[2025-09-22 18:25:19.548662]: running import VS
[2025-09-22 18:25:19.548672]: running import VS
[2025-09-22 18:25:22.387266]: running 
[2025-09-22 18:25:22.387928]: running file 'main_menunightprivateer.py'
[2025-09-22 18:25:22.555714]: running import VS
[2025-09-22 18:25:22.555852]: running import VS
[2025-09-22 18:25:22.557144]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:22.557162]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:22.651510]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:22.651582]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:25.917993]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:25.918050]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:26.001323]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:26.001409]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:26.001885]: running # <-- this disables precompiled python objects
[2025-09-22 18:25:26.001909]: running # <-- this disables precompiled python objects
```

I have tried on both my monitors and both my GPU, they return the same thing.

evertvorster:
This might help:
```
evert@Evert ~> sdl2-config --version
2.32.56
evert@Evert ~> SDL2_VIDEODRIVER= xrandr
Screen 0: minimum 16 x 16, current 5888 x 2272, maximum 32767 x 32767
eDP-1 connected primary 2560x1440+3328+832 (normal left inverted right x axis y axis) 381mm x 214mm
   2560x1440     59.96*+
   1920x1440     59.97  
   1600x1200     59.87  
   1440x1080     59.99  
   1400x1050     59.98  
   1280x1024     59.89  
   1280x960      59.94  
   1152x864      59.96  
   1024x768      59.92  
   800x600       59.86  
   640x480       59.38  
   320x240       59.29  
   1920x1200     59.88  
   1680x1050     59.95  
   1440x900      59.89  
   1280x800      59.81  
   1152x720      59.97  
   960x600       59.63  
   928x580       59.88  
   800x500       59.50  
   768x480       59.90  
   720x480       59.71  
   640x400       59.95  
   320x200       58.14  
   2048x1152     59.90  
   1920x1080     59.96  
   1600x900      59.95  
   1368x768      59.88  
   1280x720      59.86  
   1024x576      59.90  
   864x486       59.92  
   720x400       59.27  
   640x350       59.28  
HDMI-A-1 connected 3328x2080+0+0 (normal left inverted right x axis y axis) 641mm x 401mm
   3328x2080     59.92*+
   2048x1536     59.90  
   1920x1440     59.90  
   1600x1200     59.87  
   1440x1080     59.87  
   1400x1050     59.86  
   1280x1024     59.89  
   1280x960      59.94  
   1152x864      59.78  
   1024x768      59.68  
   800x600       59.86  
   640x480       59.38  
   320x240       59.29  
   2560x1600     59.94  
   1920x1200     59.88  
   1680x1050     59.95  
   1440x900      59.89  
   1280x800      59.81  
   1152x720      59.75  
   960x600       59.63  
   928x580       59.88  
   800x500       59.50  
   768x480       59.90  
   720x480       59.71  
   640x400       59.20  
   320x200       58.14  
   3200x1800     59.93  
   2880x1620     59.92  
   2560x1440     59.91  
   2048x1152     59.90  
   1920x1080     59.96  
   1600x900      59.95  
   1368x768      59.88  
   1280x720      59.86  
   1024x576      59.90  
   864x486       59.92  
   720x400       59.27  
   640x350       59.28  
evert@Evert ~> 
```

Also this:
```
[evert@Evert scripts]$ ./sdl2info
Displays: 2
  Display 0: 2560x1600 @ 60Hz
  Display 1: 2560x1440 @ 60Hz
Audio output devices: 2
  Family 17h/19h/1ah HD Audio Controller Analog Stereo
  HDA NVidia Digital Stereo (HDMI)
Joysticks: 0
```

royfalk:
> evert@Evert ~/.v/logs> cat vegastrike_2025-09-22_18_25_19.547860.log | grep -i running
> ...
> I have tried on both my monitors and both my GPU, they return the same thing.

This is great progress, sort of. 
I would expect to see something like the following:
_"Running in full screen. Overriding requested screen resolution with native one: 2560x1600"_

The fact that I don't see this tell us something important. My code potentially isn't running at all. 

Next requests:
1. The graphics section of the config.json in .vegastrike
2. If you can check your current monitor's resolution while running the game. There's usually an info button on the monitor itself.

evertvorster:
I can confirm that the code runs, as when I put the resolution to the external monitor to a low resolution and run vegastrike, it displays in a high resolution. 

Again, and I hope to make this perfectly clear this time. This is not something that I want. If I set a low resolution specifically, that is because I want the game to run in that resolution. Changing the resolution behind my back is not acceptable. 

To be perfectly clear, the game should function properly in ANY resolution. 

There should be only one source of truth for the resolution, whatever it is, and that should be used everywhere in the game where resolution matters. 

To get back to my issue with #1332:
On screen 0, the game runs acceptably, and without this patch I can display different resolutions in fullscreen.

With displaying to Screen 1, I think the game engine is taking the resolution of both monitors and adding them together and then trying to display it on my screen 1. This may be a quirk of how SDL is reporting resolutions. For screen 0, it starts at 0 for width and height, and then counts up the pixels to the screen resolution. For screen 1, it starts at the pixel after the full width of screen 0, and then counts up to the full width of screen 1. This process repeats for every monitor attached, if they are all horizontal. 

So, no need to force a specific resolution  when in fullscreen, rather look at the code used for displaying to the second monitor.

This  is a little program that chatgpt whipped up to check screen information using SDL:
[sdl2info.c](https://github.com/user-attachments/files/22483381/sdl2info.c)

You can compile it with bash on linux with:
gcc sdl2info.c -o sdl2info $(sdl2-config --cflags --libs) -lGL

On my system, it outputs the following:
```
[evert@Evert scripts]$ ./sdl2info 
SDL compile-time version: 2.32.56
SDL linked version     : 2.32.56
SDL platform           : Linux
Video driver           : wayland
Available renderers    : 5
  #0: opengl
  #1: opengles2
  #2: vulkan
  #3: gpu
  #4: software
Chosen renderer        : opengl
Displays               : 2
  Display 0 (Dell Inc. DELL U3014): 2560x1600 @ 60Hz, format 370546692
  Display 1 (BOE eDP-1-NE173QHM-NZ2): 2560x1440 @ 60Hz, format 370546692
Audio driver           : pipewire
Audio outputs          : 2
  Family 17h/19h/1ah HD Audio Controller Analog Stereo
  HDA NVidia Digital Stereo (HDMI)
Joysticks              : 0
GL_VENDOR              : AMD
GL_RENDERER            : AMD Radeon 610M (radeonsi, raphael_mendocino, LLVM 20.1.8, DRM 3.64, 
```

And to test the same with running on the other GPU:
```
6.16.5-arch1-1.1-g14)
[evert@Evert scripts]$ prime-run ./sdl2info
SDL compile-time version: 2.32.56
SDL linked version     : 2.32.56
SDL platform           : Linux
Video driver           : wayland
Available renderers    : 5
  #0: opengl
  #1: opengles2
  #2: vulkan
  #3: gpu
  #4: software
Chosen renderer        : opengl
Displays               : 2
  Display 0 (Dell Inc. DELL U3014): 2560x1600 @ 60Hz, format 370546692
  Display 1 (BOE eDP-1-NE173QHM-NZ2): 2560x1440 @ 60Hz, format 370546692
Audio driver           : pipewire
Audio outputs          : 2
  Family 17h/19h/1ah HD Audio Controller Analog Stereo
  HDA NVidia Digital Stereo (HDMI)
Joysticks              : 0
GL_VENDOR              : NVIDIA Corporation
GL_RENDERER            : NVIDIA GeForce RTX 4090 Laptop GPU/PCIe/SSE2
[evert@Evert scripts]$
```

evertvorster:
Just to put a final nail in the coffin of this specific pull request. 

Without it, on monitor 0 and with a chosen resolution of 1024 X 768, the SDL output is 2560 x 1600:
WhatsApp Image 2025-09-23 at 05 23 46

This confirms that for screen 0, SDL is scaling the resolution correctly to the native resolution of the monitor. This is not something that needs to be addressed in the game, from the game's perspective the monitor resolution is 1024 x 768, and it does not know that it is getting scaled afterwards to the native resolution of the monitor. This is wanted behavior.

stephengtuggy:
@royfalk I have to agree with @evertvorster here. Forcing a different resolution behind the user's back is not an acceptable solution.

I think we can still salvage #1332 , but it's going to take some doing. We will need to spec it out thoroughly, and list every piece of input that affects the mouse position, screen/window scaling, and so forth. The relevant factors may be somewhat different between Windows, macOS, and Linux, adding further complexity.

Roy, from what I can see, your code only chooses a single SDL mode for all monitors. That doesn't seem right. It should be able to scale different monitors independently of one another. We need to figure out how to accomplish that in SDL2.

engine/src/gldrv/winsys.cpp
```diff
@@ -251,6 +251,33 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
     height = configuration().graphics.resolution_y;
     const int screen_number = configuration().graphics.screen;
+
+    // Fix display in fullscreen
+    if(configuration().graphics.full_screen) {
+        // Get the native resolution of the selected screen.
+        SDL_DisplayMode currentDisplayMode;
+    
+        // Get the current display mode for display 0 (the primary display)
+        if (SDL_GetCurrentDisplayMode(screen_number, &currentDisplayMode) != 0) {
```
stephengtuggy: Only get the current display mode for display 0? Why? We should take into account the current display mode of each display individually, seems to me.

stephengtuggy: ... probably looping through them, or something like that.

stephengtuggy:
One of the relevant factors is the size of the window title bar (when there is one). On Windows, at least, the title bar can change size when moving between different monitors. So we will need to take that into account.

royfalk:
So, progress.

1. Game now runs in full screen in any resolution without cropped picture or black bars.
2. Game does not run well on second monitor. Documented in #1345. This is an existing bug. See more details there.

evertvorster:
> So, progress.
> 
> 1. Game now runs in full screen in any resolution without cropped picture or black bars.
> 2. Game does not run well on second monitor. Documented in [Game resolution incorrectly set for secondary monitor #1345](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1345). This is an existing bug. See more details there.

Unfortunately, for me it now does not work well on either of my monitor. Scaling of the text and background images seem good, but the location of the clickable boxes are way off.

royfalk:
> Unfortunately, for me it now does not work well on either of my monitor. Scaling of the text and background images seem good, 

Good!

> but the location of the clickable boxes are way off.

OK. What exactly is wrong and where? Is it the main menu, the base, base computer?
When you say clickable boxes are way off do you mean when you physically move your mouse to where the button is it doesn't work?

or 

When you right click, the mouse goes to the wrong place but it is clickable?

or

Something else...

evertvorster:
> > but the location of the clickable boxes are way off.
> 
> OK. What exactly is wrong and where? Is it the main menu, the base, base computer? When you say clickable boxes are way off do you mean when you physically move your mouse to where the button is it doesn't work?
> 

I never could into the base, or load a game. What I mean is that if I click on any of the buttons on the main menu, nothing happens. When right-clicking the cursor moves around a small area in the top left corner of the screen, almost like where the buttons should be is scaled to 20 - 30 % of the actual screen. 

It has been explained that the menu buttons are only "clickable boxes" that are drawn over the buttons on the background. These are way off.

engine/src/gldrv/winsys.cpp
```diff
@@ -287,11 +277,27 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
-
     if(!window) {
         VS_LOG_FLUSH_EXIT(fatal, "No window", 1);
     }
 
+    if(screen_number > 0) {
+        // Get bounds of the secondary monitor
+        SDL_Rect displayBounds;
+        if (SDL_GetDisplayBounds(screen_number, &displayBounds) != 0) {
+            SDL_Log("Failed to get display bounds: %s", SDL_GetError());
```
stephengtuggy: Can we use `VS_LOG_AND_FLUSH` here? Other than that, these latest changes look reasonable.

stephengtuggy:
It seems to me that we should have either five or six pieces of code:

1. One that runs when starting the game in full-screen mode. Save the existing screen resolution and OS-level scaling to be restored when exiting. Draw the main menu / initial screen according to these parameters.
2. One that runs when starting the game in windowed mode. Get the viewport size and OS-level scaling, and draw the main menu / initial screen accordingly.
3. One that runs in windowed mode whenever the game window is resized, moved, or the OS-level scaling changed. Get the new viewport size and OS-level scaling, and redraw the current game screen accordingly.
4. One that runs in full-screen mode whenever a screen's resolution or OS-level scaling changes? Get the new screen size and OS-level scaling, and redraw the current game screen accordingly?
5. One that runs when exiting the game from windowed mode. `winsys_exit()`, `SDL_Quit()`, `VSLogFlushProgramExiting()`, etc.
6. One that runs when exiting the game from full-screen mode. Do all of the above, plus restore the saved screen resolution and scaling.

stephengtuggy:
I suppose most of the above pieces of code should also recalculate the mouse coordinate system according to the current parameters.

royfalk:
This has been quite a journey, but I think we're nearing the end.
Current status of the PR and the area in general:

- [x] Mouse cursor in the right place but buttons in the wrong one. **fixed**
- [x] Image drawn larger or smaller than screen. **fixed**
- [x] Secondary monitors have display issues. **fixed**
- [ ] Right click iteration between buttons. **Works but mouse location is weirdly a bit offset. Probably at lower left corner of button.**
- [ ] On hover highlight in base computer incorrectly highlights wrong row. **pending**
- [ ] Joystick not tested
- [ ] Mac/Windows not tested

evertvorster:
Solid progress!

On the main menu, at various resolutions on screen 0, the buttons and the clickable areas now overlap nicely. 

Unfortunately, in the HUD, with mouse control, the controls are off. Usually, when you move your mouse to the middle of the screen, your ship does not pitch or yaw, and now it does. With high resolutions, you have to move your mouse to a different area of the screen to get no control inputs. So, the scaling is off there. 

Scaling is still wrong on the second monitor, but we can get to that one later.

royfalk:
> Unfortunately, in the HUD, with mouse control, the controls are off. Usually, when you move your mouse to the middle of the screen, your ship does not pitch or yaw, and now it does. With high resolutions, you have to move your mouse to a different area of the screen to get no control inputs. So, the scaling is off there.

How do I enable mouse support. I have a cursor but it does nothing.

> Scaling is still wrong on the second monitor, but we can get to that one later.

OK. Let's define a few things:
1. Primary monitor is whatever monitor your taskbar is on. You have a laptop and external monitor. Which is primary?
2. Do you play with your laptop lid open? What happens if you close it?
3. Which is your second monitor?
4. Please provide a screenshot of the wrong scaling.

royfalk:
> Usually, when you move your mouse to the middle of the screen, your ship does not pitch or yaw, and now it does. With high resolutions, you have to move your mouse to a different area of the screen to get no control inputs. So, the scaling is off there.

Where is the neutral position of the mouse now compared to center?

evertvorster:
> > Unfortunately, in the HUD, with mouse control, the controls are off. Usually, when you move your mouse to the middle of the screen, your ship does not pitch or yaw, and now it does. With high resolutions, you have to move your mouse to a different area of the screen to get no control inputs. So, the scaling is off there.
> 
> How do I enable mouse support. I have a cursor but it does nothing.
> 

You still need to enable it in the old vegastrike.config file, this setting has not been ported over yet. 

> > Scaling is still wrong on the second monitor, but we can get to that one later.
> 
> OK. Let's define a few things:
> 
> 1. Primary monitor is whatever monitor your taskbar is on. You have a laptop and external monitor. Which is primary?
> 2. Do you play with your laptop lid open? What happens if you close it?
> 3. Which is your second monitor?
> 4. Please provide a screenshot of the wrong scaling.

What is primary and secondary is a bit hard to define. It changes all the time, depending on where my laptop is in relation to the screen. Currently, the external screen is to the left of the laptop, and is set to be the secondary screen, with the laptop screen being primary. 

If I close the lid, and the primary screen turns off, then the external monitor switches to be primary screen. 

SDL does not really care which screen is primary, though. It always seems to number the leftmost screen screen 0

The neutral position for the mouse while flying is about 30% from the top left corner of the screen. 

The screen layout is now exactly like it was three days ago, where I provided some source code for querying SDL for monitors, and provided the output of the program provided.

evertvorster:
> > Usually, when you move your mouse to the middle of the screen, your ship does not pitch or yaw, and now it does. With high resolutions, you have to move your mouse to a different area of the screen to get no control inputs. So, the scaling is off there.
> 
> Where is the neutral position of the mouse now compared to center?

It seems to be resolution-dependent, with it being further offset when I pick lower resolutions, and closer to the middle when I pick higher resolutions. 

The settings app does not have the resolution of the screen in the list... I really think we should just go back to typing in the resolution and monitor number. 

Unfortunately, my screen capture does not show the position of the cursor in neutral position. About 30% from the top and left.

royfalk:
> You still need to enable it in the old vegastrike.config file, this setting has not been ported over yet.

This is embarrassing. I set the line:
`#set Mouse glide_mouse` in `.vegastrike/vegastrike.config`
and 
`#set Mouse warp_mouse` in `Assets-Production/vegastrike.config`

When flying and max speed, moving the mouse around does nothing. Clicking on the buttons does nothing. Clicking and moving does nothing. 

Relevant `config.json`:
```
"joystick": {
        "enabled": true,
        "throttle": false,
        "hat_enabled": false,

        "clamp_axes": true,
        "deadband": 0.05,
        "debug_digital_hatswitch": false,
        "double_mouse_position": 1280,
        "double_mouse_factor": 2.0,
        "ff_device": 0,
        "force_feedback": false,
        "force_use_of_joystick": false,
        "initial_mode": "normal",
        "joystick_exponent": 1.0,
        "mouse_blur": 0.025,
        "mouse_crosshair": "crosshairs.spr",
        "mouse_cursor": true,
        "mouse_cursor_chasecam": true,
        "mouse_cursor_pancam": false,
        "mouse_cursor_pantgt": false,
        "mouse_deadband": 0.025,
        "mouse_exponent": 3.0,
        "mouse_sensitivity": 50.0,
        "nonlinear_expamount_combat": 1.0,
        "nonlinear_expamount_nav": 1.0,
        "nonlinear_expfactor_combat": 6.0,
        "nonlinear_expfactor_nav": 6.0,
        "nonlinear_pamount_combat": 0.0,
        "nonlinear_pamount_nav": 0.0,
        "nonlinear_pfactor_combat": 2.0,
        "nonlinear_pfactor_nav": 2.0,
        "nonlinear_throttle_combat": false,
        "nonlinear_throttle_nav": true,
        "polling_rate": 0,
        "reverse_mouse_spr": false,
        "warp_mouse": true,
        "warp_mouse_zone": 100
    },

    "mouse": {
        "enabled": true,
        "FireKey": {"button": 0, "modifier": "none"},
        "inverse_x": false,
        "inverse_y": false,
        "comment": "move joystick and keyboard here. Move bindings down a level."
    },

```

Anyone?

royfalk:
Just added a commit which may solve the mouse issue. Since I can't actually setup the mouse yet, this is unfortunately more of a shotgun debug.

> SDL does not really care which screen is primary, though. It always seems to number the leftmost screen screen 0

That simplifies things enormously. If it were to treat the laptop monitor as primary, closing it mid-game would do "interesting" things to my understanding of how this works.

> The neutral position for the mouse while flying is about 30% from the top left corner of the screen.

> The screen layout is now exactly like it was three days ago, where I provided some source code for querying SDL for monitors, and provided the output of the program provided.

You misunderstood me. I'm asking about the resolution you select in config.json. I'm reasonably certain that if your config.json resolution is set at around half your native one, you'd get mouse center at 25% instead of 50%.

> Scaling is still wrong on the second monitor, but we can get to that one later.

@stephengtuggy actually pointed this out earlier for a comment but it actually got turned into code now. See if this last commit helps.

engine/src/gldrv/winsys.cpp
royfalk: This code now gets the resolution for the screen in the config. We don't need to loop because we don't care about the resolution of other screens.

royfalk:
@evertvorster , you've requested changes without actually requesting something. 
Assuming you are happy, you'll need to approve or I won't be able to merge, even if others approve.

evertvorster:
> @evertvorster , you've requested changes without actually requesting something. Assuming you are happy, you'll need to approve or I won't be able to merge, even if others approve.

I just report on what I see. 
With the latest commits in, this is what the game does. 

For simplicity's sake, we will stick to Screen 0, as that is what would be the display on a single monitor setup. 

The file below is the vegastrike.config that has the mouse enabled for in flight in the HUD. As I said before, this setting has not been ported over to the new config.
[vegastrike.config.mouse_enabled.zip](https://github.com/user-attachments/files/22570238/vegastrike.config.mouse_enabled.zip)

In the settings app, the native resolution of my monitor is not available. 
The settings app is a little restrictive in this regard. 
No matter, the game should play correctly no matter what resolution you select, as it did before we started messing with the mouse code. 

In resolutions other than native in fullscreen mode, the mouse position is still not correct in the bases and HUD. The further the resolution selected is differing from the native resolution of the monitor, the greater the offset. 

It is almost like the mouse area is not scaled to the full screen, and is limited to the resolution set in the config app, while the rest of the game is scaled properly.

 Did you look at the source code in this file?
[sdl2info.c](https://github.com/user-attachments/files/22570267/sdl2info.c)

It reads back actual screen sizes, etc. using SDL, and should be cross platform. 
It may seem a little hackish, but what would happen if you read back the native resolution that you know the game is displayed on, and scale your mouse area to that?

So, the user would select whatever resolution they choose.
If fullscreen is selected, you know the game will be scaled to the native resolution of whatever screen number is in the config file. 
Since it appears that the mouse code is not drawn onto the same size output as the game resolution, and rather scaled to the view port's actual size, it would make sense to take this into account. 

***Edit***
I know I did say that I would stick to the Monitor 0 only, but since you mentioned that code has been inserted that would touch Monitor 1, I had to try it out. 
Monitor 1 now scales correctly, but it suffers from exactly the same scaling issues with the mouse as Monitor 0. 
So, solid progress!
On a side note, the resolution for my monitor 0 is in the list of resolutions for monitor 1 in the settings app.

In the base computer, the scaling is still off as well, but this one looks to be a bit more tricky, as weirly enough, it seems to be scaled on certain areas rather than the whole screen. We can tackle that once the HUD and Base interface scales correctly.

royfalk:
> The file below is the vegastrike.config that has the mouse enabled for in flight in the HUD. 

Thanks. I now know how to enable it.

> As I said before, this setting has not been ported over to the new config.

This will need a bit more work, as we need to do the following:

- modify the settings app.
- modify the config.json.
- generate a new configuration.cpp/h files for the engine.
- modify the engine to use it.

> In the settings app, the native resolution of my monitor is not available.

File a bug against the settings app and remind me next week. 

A quick summary of what works and what doesn't, as far as I can understand it and test it.

**A quick explanation of the game UI**
To a player, everything is part of the game.
Unfortunately, the original developers implemented multiple versions of various UI functionality, including mouse behavior and screens.

- The main menu is implemented in main_menu.py. This includes the save game.
- The base (e.g. Atlantis), including the bar, launch pad and various corridors, is implemented in python as well.
- The base computer is implemented in C++ in the engine. This includes: missions, news, info (factions/ship), cargo, ship upgrades and ship dealer.

This is why the load game in main menu has an issue with overlapping lines and the one in the game doesn't. It's two different codes.

**Works**

- Image scaling is now correct on both primary and secondary monitors in full screen and window mode.
- Main menu works, including right click.

**Doesn't work**

- Base - right click is broken.
- Base - on hover (lights up the target location) is in the wrong place. 
- Flight - mouse (full screen) is limited to the config.json resolution and not the full screen one. This is why the mouse center moves up and left.
- Flight - mouse (window) is captured by the window. The game calls SDL_WarpMouseInWindow. This prevents the mouse from leaving the window. Also, in 50% resolution, there seem to be 4 mouse centers at (0.25,0.25), (0.25, 0.75), (0.75, 0.25) and (0.75, 0.75).

Did I get this right? Did I miss something?

royfalk:
Fixed issues with mouse in flight. Note that this isn't mouse support per se. It's actually joystick support.
This is why, if you have a joystick connected, the mouse cursor will not do anything.

royfalk:
Last commit fixes right click in base - It's now in the correct position.

evertvorster:
> Last commit fixes right click in base - It's now in the correct position.

Solid Progress!

Bases and HUD now have the mouse co-ordinates perfect, no matter the resolution, on both primary and secondary monitors. 
This is almost bug-free, apart from one little thing that is still outstanding:

On the bases and in the base computer, where the game engine thinks the cursor is in terms of highlighting is still being scaled wrongly. 
In the base interface, if you right-click to get to an option, usually that option is highlighted, now, you have to move your mouse towards the center of the screen to see the highlight. 

With the base computer it seems to be the same thing. 
It is as if the highlight tracking is scaled from the middle of the monitor. 

Please note that this is visual only, where you actually have to click for the option is perfectly scaled, just the highlighting visual feedback is off. 
Also, this does not seem resolution dependent, ie: if I match the desired resolution of the game to the native resolution of the output, the highlighting is still scaled by the same amount.

royfalk:
> This is almost bug-free, apart from one little thing that is still outstanding...

I was aware of this and even documented it in #1332. It's now fixed.

royfalk:
@BenjamenMeyer , @stephengtuggy any comments or requests before I merge?
I suggest you re-review my code changes.

engine/src/gfx/nav/navscreen.cpp
```diff
@@ -109,8 +110,23 @@ void NavigationSystem::mouseMotion(int x, int y) {
...
+        double resolution_y_dbl = configuration().graphics.resolution_y;
+        double x_factor = resolution_x_dbl / native_resolution_x_dbl;
+        double y_factor = resolution_y_dbl / native_resolution_y_dbl;
+        x *= x_factor;
+        y *= y_factor;
+    }
+
     mousex = x;
     mousey = y;
+    std::cout << "2 x: " << x << " y: " << y << std::endl;
```
BenjamenMeyer: do we still want to be sending this to `std::cout`?

BenjamenMeyer (changes requested):
one minor change

BenjamenMeyer (approved):

stephengtuggy:
I am really happy to see these issues all fixed, and the game using the hardware mouse pointer now. Great work, @royfalk .

PR #1348, opened 2025-09-30
Introduce Dear ImGui

royfalk:
This is more of a technology demonstrator and draft PR than something intended to be merged as is.
Dear ImGui is a modern C++ library for GUI and widgets.

Right now, the ImGui code is hard coded into a library at /library/imgui. Unless someone says otherwise, step 1 would be a commit to add it as sub-module.

Other things I plan to work on:

- Start replacing the old text functions with ImGui.
- Create some wrappers around it and expose to python.
- Develop a JSON parser as a way for non-programmers to write base GUI.
- Plug ImGui code into the existing widgets.

I've been working on main menu (hence the branch name) in order to gain experience with this library.
To test my work, go to /build/.../libraries/gui and run test_gui.

In theory, python could use pyimgui. However, this requires venv and also some way to share the window and renderer with python anyway.


evertvorster:
Hmmm, it is far too early for me to have an opinion on this. 
I get a compile error:
```
 44%] Building CXX object libraries/gui/CMakeFiles/vegastrike_gui.dir/widgets/clickable_text.cpp.o
[ 44%] Building CXX object libraries/gui/CMakeFiles/vegastrike_gui.dir/widgets/toggleable_text.cpp.o
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp: In function ‘Credits ParseJSON(const std::string&)’:
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp:75:18: error: ‘error_code’ is not a member of ‘boost::json’; did you mean ‘make_error_code’?
   75 |     boost::json::error_code ec;
      |                  ^~~~~~~~~~
      |                  make_error_code
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp:76:62: error: ‘ec’ was not declared in this scope
   76 |     boost::json::value jv = boost::json::parse(json_content, ec);
      |                                                              ^~
make[2]: *** [libraries/gui/CMakeFiles/vegastrike_gui.dir/build.make:79: libraries/gui/CMakeFiles/vegastrike_gui.dir/credits.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:616: libraries/gui/CMakeFiles/vegastrike_gui.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 44%] Linking CXX static library libvegastrike_cmd.a
[ 53%] Built target vegastrike_cmd
make: *** [Makefile:156: all] Error 2
==> ERROR: A failure occurred in build().
    Aborting...
```

Python venv does add a lot of complexity to packaging, and is something that is actively discouraged in Arch linux. I may be naive here, but let the game engine handle graphics, and Python just do calls to the game engine where needed. 

This of course means that the configuration eventually have to be C++, SDL2 and imGui, but the trade-off here is the initial pain of implementation versus subtle incompatibilities that are hard to troubleshoot further down the line.

From my very limited experience, Python initially makes prototyping something easy, but once it gets complicated it becomes a nightmare. For some reason Python dependencies are a pain to package, and seem to be abandoned not long after their creation, and good luck changing from one dependency to another. C++, on the other hand, is a pain to get right, but it scales much better and it's libraries are much more stable.

If imGui absolutely has to be in the source, making the imgui a submodule is the way to go, as you won't be having to update our copy all the time, and our code base will be smaller. So, much less day to day maintenance.   

The drawback of this approach of course is if a breaking change is made in imGui code, then we will have to fix our code to be compatible again. Seeing as imgui is fairly mature, the likelihood of this happening is fairly small.

BenjamenMeyer (commented):
I took a quick cursory look; I'll have to take a deeper look another time. I'm mostly writing this regarding how we integrate the build with ImGui...

So I'm not a fan of using Git Submodules because Git Submodules add complications. Yes, I've read https://github.com/ocornut/imgui/wiki/Getting-Started and while I generally like ImGui I'm not a fan of how they're packaging it; it shouldn't be hard to make a Static Library that can be utilized or even a series of static libraries to offer the different backends.

I've also read a number of the ImGui threads about CMake, and they seem to be anathema to doing anything more modern in their build system - I like https://github.com/adembudak/CMakeForImGui did for them, but something tells me it won't get added in.

Honestly, I'd like to see something like that or https://github.com/ocornut/imgui/discussions/5970#discussioncomment-9900483 so we could do a simple `FIND_PACKAGE(imgui)` kind of thing in our CMake and just let it be another dependency we pull in, and that seems doable on at least some platforms (see below); MacOS may be harder since you can't `brew install imgui` .  All that said...

We need to make the experience seamless - fully integrated into CMake somehow so that the general build process doesn't change:
```
$ mkdir build
$ cd build
$ cmake ..
$ make
```
Then I have no issue. Keep in mind, though, that we will need to be able to include it in our source packaging (zip, etc) too. I don't know how GitHub will manage it with their provided Zip files of the source when we do releases, I doubt it accounts for git submodules.

I'm not particularly fond of having the code in our codebase either as that is more for us to maintain and figure out how to update when it does change. So I'd like to see a dependency in some form if at all possible.

Certainly open to having the discussion.

Platforms that have it available:
- https://vcpkg.io/en/package/imgui
- Debian: `apt-get install imgui-dev`

evertvorster:
Arch also has imgui available as a standalone:
```
aur/imgui-git 1.92.2b+41.r9271.20250904.02af06ea5-1 (+1 0.10) 
    Dear ImGui: Bloat-free Graphical User interface for C++ with minimal dependencies.
aur/imgui-full 1.92.1-1 (+0 0.00) 
    Bloat-free Graphical User interface for C++
aur/imgui 1.92.3-1 (+6 0.00) 
    Bloat-free Graphical User interface for C++
```

evertvorster:
Hmm, I see that imgui is not really packaged for mac, but seems to be available for the rest of the supported OS. What's stopping us from packaging it for mac?

royfalk:
> I get a compile error:

Try `sudo apt install libboost-system-dev` or equivalent.

royfalk:
> let the game engine handle graphics, and Python just do calls to the game engine where needed.

So pyimgui isn't an implementation of imgui in python. It simply exposes C++ imgui to python. In any case, whatever we do, I want people who don't know C++ to be able to have full or near full control over the interface, just like https://github.com/vegastrike/Assets-Production/pull/19.

> This of course means that the configuration eventually have to be C++, SDL2 and ImGui, 

That was always going to be the case, except maybe with wxwidgets or qt instead of ImGui.

royfalk:
> So I'm not a fan of using Git Submodules because Git Submodules add complications. 

I'll defer to your expertise in this.

> it shouldn't be hard to make a Static Library that can be utilized or even a series of static libraries to offer the different backends.

No need for multiple. I've settled on SDL2 backend for now.
I'm more focused on getting this thing to work and moving on, rather than solving ImGui issues.

**My suggestion:**
Create a static library in VS with a subset of the ImGui code. Every 6-12 months, update those files manually.

It ain't pretty, but this will do the job.

evertvorster:
> **My suggestion:** Create a static library in VS with a subset of the ImGui code. Every 6-12 months, update those files manually.
> 
> It ain't pretty, but this will do the job.

I'm with @BenjamenMeyer on this one... I would rather have an imgui dependency, and let other people maintain that package as they already do now. 
This of course leaves macos out in the cold, as a package for imgui does not exist for macos at the moment. We could package imgui ourselves, just for macos, and distribute it with vegastrike on that platform. 

With imgui being a dependency, you can treat it just like any other dependency in the code, and not worry about it further.... Unless I am missing something fundamental?

evertvorster:
> > I get a compile error:
> 
> Try `sudo apt install libboost-system-dev` or equivalent.

On arch, there is no such package. It appears to me that boost-libs does install a "system" library:
```
evert@Evert ~> yay -Ql boost-libs
boost-libs /usr/
boost-libs /usr/lib/
boost-libs /usr/lib/libboost_atomic.so.1.88.0
boost-libs /usr/lib/libboost_charconv.so.1.88.0
boost-libs /usr/lib/libboost_chrono.so.1.88.0
boost-libs /usr/lib/libboost_cobalt.so.1.88.0
boost-libs /usr/lib/libboost_container.so.1.88.0
boost-libs /usr/lib/libboost_context.so.1.88.0
boost-libs /usr/lib/libboost_contract.so.1.88.0
boost-libs /usr/lib/libboost_coroutine.so.1.88.0
boost-libs /usr/lib/libboost_date_time.so.1.88.0
boost-libs /usr/lib/libboost_fiber.so.1.88.0
boost-libs /usr/lib/libboost_filesystem.so.1.88.0
boost-libs /usr/lib/libboost_graph.so.1.88.0
boost-libs /usr/lib/libboost_graph_parallel.so.1.88.0
boost-libs /usr/lib/libboost_iostreams.so.1.88.0
boost-libs /usr/lib/libboost_json.so.1.88.0
boost-libs /usr/lib/libboost_locale.so.1.88.0
boost-libs /usr/lib/libboost_log.so.1.88.0
boost-libs /usr/lib/libboost_log_setup.so.1.88.0
boost-libs /usr/lib/libboost_math_c99.so.1.88.0
boost-libs /usr/lib/libboost_math_c99f.so.1.88.0
boost-libs /usr/lib/libboost_math_c99l.so.1.88.0
boost-libs /usr/lib/libboost_math_tr1.so.1.88.0
boost-libs /usr/lib/libboost_math_tr1f.so.1.88.0
boost-libs /usr/lib/libboost_math_tr1l.so.1.88.0
boost-libs /usr/lib/libboost_mpi.so.1.88.0
boost-libs /usr/lib/libboost_mpi_python313.so.1.88.0
boost-libs /usr/lib/libboost_nowide.so.1.88.0
boost-libs /usr/lib/libboost_numpy313.so.1.88.0
boost-libs /usr/lib/libboost_prg_exec_monitor.so.1.88.0
boost-libs /usr/lib/libboost_process.so.1.88.0
boost-libs /usr/lib/libboost_program_options.so.1.88.0
boost-libs /usr/lib/libboost_python313.so.1.88.0
boost-libs /usr/lib/libboost_random.so.1.88.0
boost-libs /usr/lib/libboost_regex.so.1.88.0
boost-libs /usr/lib/libboost_serialization.so.1.88.0
boost-libs /usr/lib/libboost_stacktrace_addr2line.so.1.88.0
boost-libs /usr/lib/libboost_stacktrace_basic.so.1.88.0
boost-libs /usr/lib/libboost_stacktrace_from_exception.so.1.88.0
boost-libs /usr/lib/libboost_stacktrace_noop.so.1.88.0
boost-libs /usr/lib/libboost_system.so.1.88.0
boost-libs /usr/lib/libboost_thread.so.1.88.0
boost-libs /usr/lib/libboost_timer.so.1.88.0
boost-libs /usr/lib/libboost_type_erasure.so.1.88.0
boost-libs /usr/lib/libboost_unit_test_framework.so.1.88.0
boost-libs /usr/lib/libboost_url.so.1.88.0
boost-libs /usr/lib/libboost_wave.so.1.88.0
boost-libs /usr/lib/libboost_wserialization.so.1.88.0
boost-libs /usr/lib/python3.13/
boost-libs /usr/lib/python3.13/site-packages/
boost-libs /usr/lib/python3.13/site-packages/boost/
boost-libs /usr/lib/python3.13/site-packages/boost/__init__.py
boost-libs /usr/lib/python3.13/site-packages/boost/__pycache__/
boost-libs /usr/lib/python3.13/site-packages/boost/__pycache__/__init__.cpython-313.opt-1.pyc
boost-libs /usr/lib/python3.13/site-packages/boost/__pycache__/__init__.cpython-313.opt-2.pyc
boost-libs /usr/lib/python3.13/site-packages/boost/__pycache__/__init__.cpython-313.pyc
boost-libs /usr/lib/python3.13/site-packages/boost/mpi.so
```

evertvorster:
> > I get a compile error:
> 
> Try `sudo apt install libboost-system-dev` or equivalent.

The only place on the web that this error popped up before was in 2021:
https://github.com/boostorg/json/issues/585

royfalk:
Did running apt or equivalent fix it for you?

stephengtuggy:
> > **My suggestion:** Create a static library in VS with a subset of the ImGui code. Every 6-12 months, update those files manually.
> > It ain't pretty, but this will do the job.
> 
> I'm with @BenjamenMeyer on this one... I would rather have an imgui dependency, and let other people maintain that package as they already do now. This of course leaves macos out in the cold, as a package for imgui does not exist for macos at the moment. We could package imgui ourselves, just for macos, and distribute it with vegastrike on that platform.
> 
> With imgui being a dependency, you can treat it just like any other dependency in the code, and not worry about it further.... Unless I am missing something fundamental?

This might be a good time to start using vcpkg on macOS as well as on Windows. `Dear ImGui` should be available on macOS via vcpkg, just like with other platforms that vcpkg supports, AFAIK.

Then two of our three platforms will use the same package manager. There could be a lot of advantages to that. What do you guys think?

royfalk:
> This might be a good time to start using vcpkg on macOS as well as on Windows. 

I have no opinion about this. I try and leave this whole area of CICD and packaging in your capable hands. Whatever you and @BenjamenMeyer think is fine with me.

> Dear ImGui should be available on macOS via vcpkg...

A quick search shows that it is available. However, if we go this route it means that if a platform does not have imgui, it will not be supported.

engine/src/gldrv/image_utils.cpp
```diff
@@ -0,0 +1,112 @@
...
+        index = 0;
+    }
+ 
+    float x = ImGui::GetCursorPosX();
+    x = margin;
+    ImGui::SetCursorPosX(x);
+
+    float y = ImGui::GetCursorPosY();
+    y = margin;
+    ImGui::SetCursorPosY(y);
```
stephengtuggy, line 97: What exactly are we doing here? It looks like this code will have the effect of repositioning the ImGui cursor to `(margin, margin)`. Is that what we want? What does `margin` refer to here, exactly?

libraries/gui/widgets/splash_screen.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+void RenderSplashScreen(const std::string message, int width, int height) {
+    if(!gui_initialized) {
+        return;
+    }
+
+    // This is used to generate a random name for the ImGui window,
+    // in order to prevent collisions.
+    const std::string* message_ptr = &message;
+    const std::string window_name = std::to_string((unsigned long long)message_ptr);
```
stephengtuggy:
```suggestion
    const std::string window_name = std::to_string(static_cast<uintmax_t>(message_ptr));
```

?

libraries/gui/widgets/widget.h
```diff
@@ -0,0 +1,41 @@
...
+#ifndef VEGA_STRIKE_LIBRARIES_GUI_WIDGETS_WIDGET_H
+#define VEGA_STRIKE_LIBRARIES_GUI_WIDGETS_WIDGET_H
+
+#include <vector>
+#include <string>
+
+class Widget {
+public:
+    virtual void Draw() = 0;
+    virtual ~Widget() {}
```
stephengtuggy:
```suggestion
    virtual ~Widget() = default;
```

?

engine/src/gldrv/image_utils.cpp
royfalk: This code is a bit misleading as it is displaying the splash screen image.
As such, we are applying a margin to the left, right and top.
However, we're leaving some space at the bottom to display text.

libraries/gui/widgets/splash_screen.cpp
royfalk: This gives `invalid ‘static_cast’ from type ‘const std::string*’ {aka ‘const std::__cxx11::basic_string<char>*’} to type ‘uintmax_t’ {aka ‘long unsigned int’}`

How about:
```
const std::string* message_pointer = &message;
uintptr_t pointer_address = reinterpret_cast<uintptr_t>(message_pointer);  
const std::string window_name = std::to_string(pointer_address);
```

BenjamenMeyer: could we make a simpler hash function that just takes the message and generates a hash? Perhaps even a UUIDv7? (Boost has the necessarily functionality for us - https://www.boost.org/doc/libs/1_89_0/libs/uuid/doc/html/uuid.html#time_generator_v7).

BenjamenMeyer:
> > > **My suggestion:** Create a static library in VS with a subset of the ImGui code. Every 6-12 months, update those files manually.
> > > It ain't pretty, but this will do the job.
> > 
> > 
> > I'm with @BenjamenMeyer on this one... I would rather have an imgui dependency, and let other people maintain that package as they already do now. This of course leaves macos out in the cold, as a package for imgui does not exist for macos at the moment. We could package imgui ourselves, just for macos, and distribute it with vegastrike on that platform.
> > With imgui being a dependency, you can treat it just like any other dependency in the code, and not worry about it further.... Unless I am missing something fundamental?
> 
> This might be a good time to start using vcpkg on macOS as well as on Windows. `Dear ImGui` should be available on macOS via vcpkg, just like with other platforms that vcpkg supports, AFAIK.

I'm not sure. Reading through [Homebrew Support](https://docs.brew.sh/Support-Tiers) and [vcpkg Support](https://learn.microsoft.com/en-us/vcpkg/concepts/supported-hosts) Microsoft isn't (a) isn't as clear about some thing related to support in some ways (namely how they talk about Linux, but we're not looking at using it there, so kind of moot) and (b) cuts off support much sooner than Homebrew does for MacOS, which has a greater likelihood of effecting our users.

> Then two of our three platforms will use the same package manager. There could be a lot of advantages to that. What do you guys think?

I generally lean towards using the most native method for a given platform. For Mac that tends to be macPorts and Homebrew, so one of those makes good sense - and even then Homebrew has a lot more mindshare. For Windows vcpkg makes sense. While Microsoft may be supporting Linux and Mac for vcpkg, it's likely to help their own limited cross-platform builds; until you mentioned it I wasn't even aware that vcpkg would support Mac and I've been on Mac (for work) for most of the last 6 years - I'm aware of macPorts but I'm not sure I even know anyone that actually uses it - Homebrew likely won the "war".

Another option is for us to do a [Homebrew Tap](https://docs.brew.sh/Taps) within our org and then host our own method of install imgui for Mac. If there's enough interest it could get promoted into Homebrew itself, but this gives us a faster ability to onboard it and control it.

Thoughts?

stephengtuggy:
Do we really want to build and maintain our own Homebrew Tap? Seems like that could be a lot of work. I'd rather use what someone else has already built, if possible.

BenjamenMeyer:
> Do we really want to build and maintain our own Homebrew Tap? Seems like that could be a lot of work. I'd rather use what someone else has already built, if possible.

I'd rather not do it ourselves if we can help it; but I wouldn't rule it out if there wasn't another choice.
If we did, I'd be make it clear about our intentions, need, and that if there is enough interest we would want/be willing to to turn it over to the Homebrew developers for the long term and get out of the way so the community could own it.

stephengtuggy:
> > Do we really want to build and maintain our own Homebrew Tap? Seems like that could be a lot of work. I'd rather use what someone else has already built, if possible.
> 
> I'd rather not do it ourselves if we can help it; but I wouldn't rule it out if there wasn't another choice. If we did, I'd be make it clear about our intentions, need, and that if there is enough interest we would want/be willing to to turn it over to the Homebrew developers for the long term and get out of the way so the community could own it.

I suppose this isn't a terrible idea.

royfalk:
> Do we really want to build and maintain our own Homebrew Tap? Seems like that could be a lot of work. I'd rather use what someone else has already built, if possible.

> I'd rather not do it ourselves if we can help it; but I wouldn't rule it out if there wasn't another choice. If we did, I'd be make it clear about our intentions, need, and that if there is enough interest we would want/be willing to to turn it over to the Homebrew developers for the long term and get out of the way so the community could own it.

> I suppose this isn't a terrible idea.

At this point I'd like to reach some decision and move on. 
**Options:**
1. Plug the code in the game for now and wait for support for imgui to arrive either from our team or elsewhere.
2. Wait for homebrew implementation from one of you guys and then proceed.
3. Drop this and move on to other parts of the game.
4. Something else?

BenjamenMeyer:
> 2\. Wait for homebrew implementation from one of you guys and then proceed.

I'd like to take a shot at the homebrew side; it might not until next week though.
In the mean time, we can (temporarily) put it into the libraries folder so we have a backup if the homebrew thing is too hard.

NOTE: Homebrew can be tested under Linux too; I haven't tried it before but it's technically supported.

evertvorster:
Just to note, this still does not compile for me:
```
[ 16%] Building CXX object libraries/cmd/CMakeFiles/vegastrike_cmd.dir/jump_capable.cpp.o
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp: In function ‘Credits ParseJSON(const std::string&)’:
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp:75:18: error: ‘error_code’ is not a member of ‘boost::json’; did you mean ‘make_error_code’?
   75 |     boost::json::error_code ec;
      |                  ^~~~~~~~~~
      |                  make_error_code
[ 16%] Building CXX object engine/objconv/CMakeFiles/vega-meshtool.dir/mesher/Modules/BFXM_to_Wavefront.cpp.o
/home/evert/Aur/vegastrike-engine-git/src/Vega-Strike-Engine-Source/libraries/gui/credits.cpp:76:62: error: ‘ec’ was not declared in this scope
   76 |     boost::json::value jv = boost::json::parse(json_content, ec);
      |                                                              ^~
make[2]: *** [libraries/gui/CMakeFiles/vegastrike_gui.dir/build.make:93: libraries/gui/CMakeFiles/vegastrike_gui.dir/credits.cpp.o] Error 1
```

I do have the boost development headers installed:
```
evert@Evert ~/A/vegastrike-engine-git (master) [4]> pacman -Ss boost
cachyos-extra-znver4/boost 1.89.0-2.1 [installed]
    Free peer-reviewed portable C++ source libraries (development headers)
cachyos-extra-znver4/boost-libs 1.89.0-2.1 [installed]
    Free peer-reviewed portable C++ source libraries (runtime libraries)
```

royfalk:
I'm closing this PR in favour of #1417 and following smaller PRs.
I'll return to this to PR for its review comments and incorporate them.

PR #1350, opened 2025-10-04
Fix bug keeping CMake from recognizing Debian Trixie as supported

### LINKED ISSUE #1282: Support Debian Trixie

stephengtuggy:
Code Changes:
- [x] This is a CI / build system change only

Issues:
- Closes #1282 

Purpose:
- What is this pull request trying to do? Fix a one-line bug in engine/CMakeLists.txt that causes CMake to spit out a spurious warning that Debian Trixie is not supported (when in fact it is).


stephengtuggy:
> Not sure why this fixes but looks harmless enough.

That whole section of CPack logic is supposed to rely on `LSB_LINUX_DISTRIBUTION_CODENAME` to determine the distro/version. (This was a copy/paste error.)

engine/CMakeLists.txt
```diff
@@ -992,7 +992,7 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
                         SET(BOOST_VER "${Boost_DERIVED_VERSION}")
                     ENDIF (Boost_DERIVED_VERSION STREQUAL "")
                     SET(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libjpeg-turbo8, libpng16-16, freeglut3, libgtk-3-0, libvorbis0a, libopenal1, libsdl2-gfx-1.0-0, libsdl2-image-2.0-0, xdg-utils, libgl1, libopengl0, ${VEGA_STRIKE_GLU_DEPENDENCY}, libboost-atomic${BOOST_VER}, libboost-chrono${BOOST_VER}, libboost-date-time${BOOST_VER}, libboost-filesystem${BOOST_VER}, libboost-log${BOOST_VER}, libboost-python${BOOST_VER}, libboost-regex${BOOST_VER}, libboost-thread${BOOST_VER}, libboost-program-options${BOOST_VER}")
-                ELSEIF (DEBIAN_RELEASE_VERSION STREQUAL "trixie")
```
BenjamenMeyer, line 995: This is interesting...and perhaps we can think more about this too - `DEBIAN_RELEASE_VERSION` is set on L910 and should be capturing the first portion of the contents of the file `/etc/debian_version`, which for trixie is `trixie/sid` so `DEBIAN_RELEASE_VERSION` should be `trixie` already. This is the method of determining the Debian distribution information since long before cross-distro tooling allowed such detection.

LSB_LINUX_DISTRIBUTION_CODENAME, however, comes from the LSB tooling we do in `cmake/FindLinuxDistro.cmake` and relies on the `/etc/os-release` data being correct. It's a standard for LSB Compliant Distributions, which Debian is, and provides a cross-distro method to detect the distro information.

For our official support we should be able to rely on the LSB tooling; though some rebranded distros may not support it or even remove it...but I'm not sure that should necessarily be _our_ problem.

All that is to say... we could probably simplify the CMake file to rely more on the LSB_LINUX_DISTRIBUTION_CODENAME and some of the other constants we generate out of `cmake/FindLinuxDistro.cmake`; if need be we could move some of this logic there but do we really need it?

stephengtuggy: We can consider those items for the future, I suppose. For now, this fixes the immediate issue, so merging it.

BenjamenMeyer: @stephengtuggy created https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1362

PR #1353, opened 2025-10-04
Fix null-unit segfaults

### LINKED ISSUE #1231: Segfault trying to tractor in some cargo

### LINKED ISSUE #1352: Segfault when player's ship's `velocity_reference` is null

stephengtuggy:
Code Changes:

Issues Addressed:
- Closes #1231 
- Closes #1352 

Purpose:
- What is this pull request trying to do? Fix several segfaults that occur when trying to dereference null units from the player's ship

### DIFF
--- a/engine/src/gfx/cockpit.cpp
+++ b/engine/src/gfx/cockpit.cpp
@@ -51,9 +51,7 @@
 #include "src/star_system.h"
 #include "cmd/unit_generic.h"
 #include "cmd/movable.h"
-#include "cmd/collection.h"
 #include "cmd/unit_util.h"
-#include "cmd/unit_find.h" //for radar iteration.
 #include "cmd/base_util.h"
 #include "gfx/hud.h"
 #include "gfx/vdu.h"
@@ -75,12 +73,9 @@
 #include "gfx_generic/mesh.h"
 #include "src/universe_util.h"
 #include "src/in_mouse.h"
-#include "gui/glut_support.h"
 #include "src/audiolib.h"
-#include "src/save_util.h"
 #include "cmd/base.h"
 #include "src/in_kb_data.h"
-#include "src/main_loop.h"
 #include <set>
 #include <string>
 #include "cmd/unit_const_cache.h"
@@ -93,7 +88,6 @@
 #include "gfx/cockpit_gfx.h"
 #include "cmd/dock_utils.h"
 #include "vega_cast_utils.h"
-#include "resource/random_utils.h"
 #include "resource/random_utils.h"
 
 #include <cstddef>
@@ -215,7 +209,7 @@ void GameCockpit::DoAutoLanding(Unit *un, Unit *target) {
     diffvec.Normalize();
 
     static bool haswarned = false;
-    static void *lastwarned = NULL;
+    static void *lastwarned = nullptr;
     static float docktime = -FLT_MAX;
     if (dist < lessthan && haswarned && lastwarned == target) {
         //CrashForceDock(target,un,true);
@@ -230,7 +224,7 @@ void GameCockpit::DoAutoLanding(Unit *un, Unit *target) {
             un->SetPosAndCumPos(UniverseUtil::SafeEntrancePoint(
                     target->Position() + diffvec * (rsize + moveout + un->rSize()),
                     un->rSize() * 1.1));
-            lastwarned = NULL;
+            lastwarned = nullptr;
         }
     } else if (dist < warnless) {
         if (lastwarned != target || !haswarned) {
@@ -275,7 +269,7 @@ void GameCockpit::DoAutoLanding(Unit *un, Unit *target) {
         }
     } else if (lastwarned == target) {
         haswarned = false;
-        lastwarned = NULL;
+        lastwarned = nullptr;
     }
 }
 
@@ -283,7 +277,7 @@ void GameCockpit::AutoLanding() {
     const bool autolanding_enable = configuration().physics.auto_landing_enable;
     if (autolanding_enable) {
         Unit *player = GetParent();
-        if (player == NULL) {
+        if (player == nullptr) {
             return;
         }
 
@@ -294,7 +288,7 @@ void GameCockpit::AutoLanding() {
             }
 
             Unit *target = it->ref.unit;
-            if (target == NULL) {
+            if (target == nullptr) {
                 continue;
             }
 
@@ -314,7 +308,7 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
     const bool display_in_meters = configuration().physics.display_in_meters;
     const bool lie = configuration().physics.game_speed_lying;
     static float fpsval = 0;
-    const float fpsmax = 1;
+    constexpr float fpsmax = 1;
     static float numtimes = fpsmax;
     Unit *tmpunit;
 
@@ -379,14 +373,14 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
                 if (go == 0) {
                     static soundContainer ejectstopsound;
                     if (ejectstopsound.sound < 0) {
-                        const std::string str = configuration().cockpit_audio.overload_stopped;
+                        const std::string& str = configuration().cockpit_audio.overload_stopped;
                         ejectstopsound.loadsound(str);
                     }
                     ejectstopsound.playsound();
                 } else {
                     static soundContainer ejectsound;
                     if (ejectsound.sound < 0) {
-                        const std::string str = configuration().cockpit_audio.overload;
+                        const std::string& str = configuration().cockpit_audio.overload;
                         ejectsound.loadsound(str);
                     }
                     ejectsound.playsound();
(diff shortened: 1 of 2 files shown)


engine/src/gfx/cockpit.cpp
```diff
@@ -396,7 +393,7 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
             float distance;
             const float locklight_time = configuration().graphics.locklight_time_flt;
             bool res = false;
-            if (tmpunit = target->Threat()) {
+            if ((tmpunit = target->Threat())) {
```
royfalk: As nice and elegant as this code is, I'd appreciate something along of:

```
tmpunit = target->Threat();
if(tmpunit) {
...
```

engine/src/gfx/cockpit.cpp
```diff
@@ -1781,9 +1784,6 @@ void GameCockpit::Draw() {
             die = false;
         }
         if (un->Threat() != nullptr) {
-            if (0 && getTimeCompression() > 1) {
-                reset_time_compression(std::string(), PRESS);
```
royfalk, line 1785: Did you also delete reset_time_compression?

engine/src/gfx/cockpit.cpp
```diff
@@ -1944,20 +1944,20 @@ void GameCockpit::Draw() {
 }
 
 int GameCockpit::getScrollOffset(unsigned int whichtype) {
-    for (unsigned int i = 0; i < vdu.size(); i++) {
-        if (vdu[i]->getMode() & whichtype) {
-            return vdu[i]->scrolloffset;
+    for (std::size_t i = 0; i < vdu.size(); ++i) {
```
royfalk: Why not go full modern with 

```
for(auto& vdu_unit : vdu) {
...

Note that `vdu` really should be `vdus`.
```

engine/src/gfx/cockpit.cpp
```diff
@@ -2140,11 +2139,14 @@ void GameCockpit::ScrollAllVDU(int howmuch) {
 }
 
 void GameCockpit::SetStaticAnimation() {
-    const string comm_static = configuration().graphics.comm_static;
-    static Animation Statuc(comm_static.c_str());
+    const std::string& comm_static = configuration().graphics.comm_static;
+    static boost::optional<Animation> Statuc{};
```
royfalk: Surely this is a typo and they meant status?!

royfalk (commented):
I'd appreciate it if you split the obvious stuff out of this PR.
I know it is a bit more work but it's an auto-approve from me and those few lines of code that matter can the review they need.

Having said that, it looks fine to me except the following:
1. You are right that the two PRs are related in the sense that both try to solve the problem of what to do if a ship is killed while other units hold their pointers.
2. I don't understand the code in unit_generic. I'd need to review the deletion code more before I can review this.

But, overall it looks like a great step forward.

engine/src/gfx/cockpit.cpp
stephengtuggy: Fair enough. CLion recommended that I put the assignment inside an extra pair of parentheses. But splitting it out onto its own line makes sense too.

engine/src/gfx/cockpit.cpp
stephengtuggy: I could probably do that as well. I'm a bit lazy a lot of the time, and only go "full modern," as you put it, when CLion gives me an easy shortcut to click to accomplish that.

engine/src/gfx/cockpit.cpp
stephengtuggy: No, they meant `static`. Of course, that's a keyword, so they intentionally misspelled it. At least as far as I can tell.

engine/src/gfx/cockpit.cpp
stephengtuggy: No, I don't think I did. Let me see if it can be deleted as well.

stephengtuggy: There is still one other place where `reset_time_compression` is used: https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/config_xml.cpp#L524

engine/src/gfx/cockpit.cpp
stephengtuggy: Done.

libraries/cmd/unit_generic.cpp
```diff
@@ -1397,11 +1397,21 @@ void Unit::ProcessDeleteQueue() {
 #ifdef DESTRUCTDEBUG
         if ( unit_delete_queue.back()->isSubUnit() ) {
             VS_LOG(debug, "Subunit Deleting (related to double dipping)");
+            unit_delete_queue.pop_back();
+            continue;
```
stephengtuggy, line 1429: This code only comes into effect when `DESTRUCTDEBUG` is defined. So not very often.

libraries/cmd/unit_generic.cpp
```diff
@@ -1397,11 +1397,21 @@ void Unit::ProcessDeleteQueue() {
...
         }
 #endif
         Unit *mydeleter = unit_delete_queue.back();
         unit_delete_queue.pop_back();
+
+        // Avoid segfault when the unit getting destroyed is the player's current target
+        Unit* parent = _Universe->AccessCockpit()->GetParent();
+        if (parent && parent->Target() == mydeleter) {
+            parent->SetTarget(nullptr);
+        }
```
stephengtuggy, line 1439: Here, to avoid segfaults, we simply traverse the object hierarchy starting with `_Universe->AccessCockpit()->GetParent()`, which returns the player's ship. This way, we don't have to add new parameters all over the place.

libraries/cmd/unit_generic.cpp
```diff
@@ -3196,7 +3206,7 @@ enum Unit::tractorHow Unit::getTractorability() const {
     static bool tractorability_mask_init = false;
     static unsigned char tractorability_mask;
     if (!tractorability_mask_init) {
-        std::string stractorability_mask = configuration().physics.player_tractorability_mask;
+        const std::string& stractorability_mask = configuration().physics.player_tractorability_mask;
```
stephengtuggy, line 3237: minor optimization

stephengtuggy, line 3237: minor optimization

libraries/cmd/unit_generic.cpp
```diff
@@ -3461,9 +3471,9 @@ void Unit::UpdatePhysics3(const Transformation &trans,
         Velocity = Velocity * (1 - SPACE_DRAG);
     }
 
-    std::string locking_sound_name = configuration().audio.unit_audio.locking;
+    const std::string& locking_sound_name = configuration().audio.unit_audio.locking;
     //enables spiffy wc2 torpedo music, default to normal though
-    std::string locking_sound_torp_name = configuration().audio.unit_audio.locking_torp;
+    const std::string& locking_sound_torp_name = configuration().audio.unit_audio.locking_torp;
```
stephengtuggy, line 3504: minor optimizations

libraries/cmd/unit_generic.cpp
```diff
@@ -1397,11 +1397,21 @@ void Unit::ProcessDeleteQueue() {
...
         unit_delete_queue.pop_back();
+
+        // Avoid segfault when the unit getting destroyed is the player's current target
+        Unit* parent = _Universe->AccessCockpit()->GetParent();
+        if (parent && parent->Target() == mydeleter) {
+            parent->SetTarget(nullptr);
+        }
+
         delete mydeleter;                        ///might modify unitdeletequeue
+        mydeleter = nullptr;
```
stephengtuggy, line 1442: make sure to set this to nullptr so that we know not to dereference it later

engine/src/gfx/cockpit.cpp
```diff
@@ -1207,10 +1210,10 @@ void GameCockpit::Shake(float amt, int dtype) {
...
 
 static void DrawDamageFlash(int dtype) {
-    const int numtypes = 3;
-    const string shieldflash = configuration().graphics.shield_flash_animation;
-    const string armorflash = configuration().graphics.armor_flash_animation;
-    const string hullflash = configuration().graphics.hull_flash_animation;
+    constexpr int numtypes = 3;
+    const std::string& shieldflash = configuration().graphics.shield_flash_animation;
+    const std::string& armorflash = configuration().graphics.armor_flash_animation;
+    const std::string& hullflash = configuration().graphics.hull_flash_animation;
```
stephengtuggy, line 1216: more minor optimizations

engine/src/gfx/cockpit.cpp
stephengtuggy: Done

engine/src/gfx/cockpit.cpp
royfalk: <sigh>
<heavy sigh>
<SCREEEAAM>
`
static boost::optional<Animation> Static_animation{};`

stephengtuggy: Good suggestion. Done

evertvorster:
I have played this for an hour or so, and it crashed, but the log does not really state what went wrong. 
How would I go about figuring out  why the game is crashing?

My crashes happen when I jump, just FYI

engine/src/gfx/cockpit.cpp
BenjamenMeyer: I prefer splitting it out as its easier to capture for logs or debugging; it also avoids potential issues with testing the wrong thing in the IF-conditional.

stephengtuggy: @BenjamenMeyer I believe this change has been made. Noted for the future.

stephengtuggy:
> I have played this for an hour or so, and it crashed, but the log does not really state what went wrong. How would I go about figuring out why the game is crashing?
> 
> My crashes happen when I jump, just FYI

@evertvorster you could try running the game in a debugger. Either a command-line debugger such as `gdb` or `lldb`, or a full-fledged graphical debugger inside an IDE such as CLion. However, you're not guaranteed to be able to see the actual cause of the crash. It's just more likely that you'll be able to see it if you run inside a debugger.

TBH, to me, being able to play for an hour or so without crashes sounds like significant progress. In the past, the game has often crashed on me much sooner than that.

BenjamenMeyer (approved):

PR #1364, opened 2025-10-29
SDL3 Port for the 0.10 branch 0.10.x sdl3

pmxy:

Please answer the following:

Code Changes:

Issues:

Purpose:
- What release is this for?

### DIFF
--- a/engine/src/components/ecm.cpp
+++ b/engine/src/components/ecm.cpp
@@ -162,6 +162,9 @@ void ECM::_upgrade(const std::string key) {
     boost::split(result, ecm_string, boost::is_any_of("/"));
     int max_ecm = 0;
     int current_ecm = 0;
+    
+    // pmx-20251103 str::stod() fails on the decimal point on french locale (decimal separator is ',')
+    const char* loc = std::setlocale(LC_NUMERIC, "en_US.UTF-8");
 
     try {
         switch(result.size()) {
@@ -182,4 +185,7 @@ void ECM::_upgrade(const std::string key) {
 
     ecm = Resource<int>(current_ecm, 0, max_ecm);
     installed = true;
+
+    // Restore locale
+    std::setlocale(LC_NUMERIC, loc);
 }
--- a/engine/src/components/shield.cpp
+++ b/engine/src/components/shield.cpp
@@ -161,8 +161,11 @@ void Shield::Load(std::string unit_key) {
             if(string_value.empty()) {
                 break;
             }
-
+// pmx-20251103 str::stod() fails on the decimal point on french locale (decimal separator is ',')
+            const char* loc = std::setlocale(LC_NUMERIC, "en_US.UTF-8");
             double value = std::stod(string_value);
+// Restore locale
+            std::setlocale(LC_NUMERIC, loc);    
             shield_values.push_back(Resource<double>(value, 0.0, value));
             ++shield_count;
         }


pmxy:
This update corrects the exit code problem found on the branch 0.90 and is also present in 0.10

engine/src/gfx/cockpit.cpp
```diff
@@ -538,17 +542,29 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
...
-            }
-            if (numtimes <= 0) {
-                numtimes = fpsmax;
-                fpsval = GetElapsedTime();
-            }
-            if (fpsval) {
-                return 1.0F / fpsval;
+        case UnitImages<void>::COCKPIT_FPS:  {
+            float dt = GetElapsedTime(); 
+#if 0 // pmx-20251025  Basic first order smoothing filter.
```
BenjamenMeyer, line 547: instead of using `#if 0` it would be good if you used a Processor Definition - `#if <name>` that could be defined as a build option via CMake to make it an easy choice by simply passing `-D<name>` as a command-line parameter to `cmake`.

engine/src/gfx/technique.cpp
```diff
@@ -370,8 +370,10 @@ Technique::Technique(const string &name) :
     } else {
         filename = sub_technique_filename;
     }
-
-    pt::ptree tree;
+    
+    // if (!boost::filesystem::exists(filename)) {return;}
+    
+        pt::ptree tree;
```
BenjamenMeyer, line 376: there's essentially no change here; please clean this up

engine/src/gldrv/winsys.cpp
```diff
@@ -493,7 +542,10 @@ void winsys_shutdown() {
 void winsys_show_cursor(bool visible) {
     static bool vis = true;
     if (visible != vis) {
-        SDL_ShowCursor(visible);
+        if (visible)
+            SDL_ShowCursor();
+        else
+            SDL_HideCursor();
```
BenjamenMeyer, line 522: NOTE: We require the braces even for one liners like this to prevent potential bugs. Please update:
```c++
if (visible) {
   SDL_ShowCursor();
} else {
   SDL_HideCursor();
}
```

engine/src/gldrv/winsys.h
```diff
@@ -156,22 +159,23 @@ typedef enum {
...
+    WSK_MOD_RSHIFT = 0x0002, // SDL_KMOD_RSHIFT,
+    WSK_MOD_LCTRL = 0x0040u, // SDL_KMOD_LCTRL,
+    WSK_MOD_RCTRL = 0x0080u, // SDL_KMOD_RCTRL,
+    WSK_MOD_LALT = 0x0100, // SDL_KMOD_LALT,
+    WSK_MOD_RALT = 0x0200u, // SDL_KMOD_RALT,
+    WSK_MOD_LMETA = 0x0400, // SDL_KMOD_LGUI, // This is an issue for these two entries. We are ignoring the modifier.
+    WSK_MOD_RMETA = 0x0800, // SDL_KMOD_RGUI, // Need to figure out how to do modifier in sdl2.
+    WSK_MOD_NUM = 0x1000u, // SDL_KMOD_NUM,
+    WSK_MOD_CAPS = 0x2000u, // SDL_KMOD_CAPS,
+    WSK_MOD_MODE = 0x4000 // SDL_KMOD_MODE // AltGr
```
BenjamenMeyer, line 178: it would be better to be mapping the names instead of copying the values; it's okay that the name changes, but it's be a lot harder to track the value changing.

engine/src/gldrv/winsys.h
```diff
@@ -190,8 +194,8 @@ typedef enum {
 } winsys_mouse_button_t;
 
 typedef enum {
-    WS_MOUSE_DOWN = SDL_PRESSED,
-    WS_MOUSE_UP = SDL_RELEASED
+    WS_MOUSE_DOWN = true,
+    WS_MOUSE_UP = false
```
BenjamenMeyer, line 198: same here

engine/src/in_sdl.cpp
```diff
@@ -187,18 +187,24 @@ void ProcessJoystick(int whichplayer) {
...
             for (int j = 0; j < NUMJBUTTONS; j++) {
                 KBSTATE *state = &JoystickState[JOYSTICK_SWITCH][i][j];
                 JSHandlerCall *handler = &JoystickBindings[JOYSTICK_SWITCH][i][j];
-                if ((buttons & (1 << j))) {
+                if ((buttons & ((long long)(1) << j))) {
                     if (*state == UP) {
                         (*handler->function)
                                 (handler->data, PRESS);
                         *state = DOWN;
+                        std::cerr << "Player ["<< std::to_string(i)
```
BenjamenMeyer, line 199: we use a Boost Logger for everything once it has been configured. Please update this to use the boost logger or a VS log function. @stephengtuggy could advice if needed

BenjamenMeyer (changes requested):
1. Please adjust formatting to match existing styles. There's a lot of changes listed that are simply adjusting the space around a `*`.
2. Please ensure all blocks use braces, even one liners.
3. Please ensure everything is sent to the Logger; std::cout/std::cerr are only used during early phase initialization before the logger is initialized.
4. There is an interesting option of some smoothing that would be great to turn into a build option so people can test both modes. This would just require updating to use a C++ Preprocessor definition and appropriate CMake options to support.
5. Let's avoid magic numbers where possible; mapping from one set of constants in a 3rd party library to our own internal constants by using their names is better than using their values.
6. before we can let the build run we need to add SDL3 to our base images so its available our CI system

Generally I do like the changes; though I'm going to have to defer final approval to @royfalk since he's the most familiar with the SDL2 code. But we do need a few adjustments first.

evertvorster:
This now depends on sdl3_image, which is an AUR package. 
Once that is installed, it compiles fine. 

When you change views, you can't change back to cockpit view.
You change views by hitting the number keys, 1 being in-cockpit, 2 & 3 are left and right, and 4 is behind. 5 & 6 are exterior views. 
Play around with these a bit when the ship is launched, and you will see what I mean. 

I have not done extensive testing, this is the first major issue that I picked up.

engine/src/in_sdl.cpp
stephengtuggy: Use of `VS_LOG`, `VS_LOG_AND_FLUSH`, and `VS_LOG_FLUSH_EXIT` is preferred. This logging call would become something like:

```
                VS_LOG(debug, (boost::format("Player [%1%] JS Button DOWN %2%") % i % j));
```

Notice the use of the boost::format interpolated string. Also notice that I chose a relatively verbose log level, `debug`, since this message might get logged often. You could choose another logging level if you think it's more appropriate here.

stephengtuggy: Thanks for the contribution. I look forward to getting this cleaned up and merged soon. Moving to SDL3 is certainly a worthy goal.

engine/src/in_sdl.cpp
```diff
@@ -187,18 +187,24 @@ void ProcessJoystick(int whichplayer) {
                     (*handler->function)(handler->data, *state);
                 }
             }             //digital_hatswitch
+
             for (int j = 0; j < NUMJBUTTONS; j++) {
                 KBSTATE *state = &JoystickState[JOYSTICK_SWITCH][i][j];
                 JSHandlerCall *handler = &JoystickBindings[JOYSTICK_SWITCH][i][j];
-                if ((buttons & (1 << j))) {
+                if ((buttons & ((long long)(1) << j))) {
```
stephengtuggy, line 194: Could this simply be written as:


```suggestion
                if ((buttons & (1LL << j))) {
```

?

engine/src/gldrv/winsys.cpp
```diff
@@ -202,43 +202,69 @@ int native_resolution_x;
 int native_resolution_y;
 
 
+/*---------------------------------------------------------------------------*/
+// pmx-20251026
+bool get_sdl_display_name_by_nr(int screen_number, std::string& screen_name, SDL_DisplayID& id) {
+    std::ostringstream display_names;
+    int num_displays = 0;
+    bool found = false;
+    std::cerr << "===============================================================" << std::endl;
```
stephengtuggy: Please use `VS_LOG` or one of the related macros here as well

engine/src/gldrv/winsys.cpp
```diff
@@ -202,43 +202,69 @@ int native_resolution_x;
...
+        for (i = 0; i < num_displays; ++i) {
+            SDL_DisplayID instance_id = displays[i];
+            const char* name = SDL_GetDisplayName(instance_id);
+            display_names << "Display " << i << ": " << (name ? name : "Unknown") << "\n";
+            if (i == screen_number) {
+                screen_name = name ? name : "Unknown";
+                id = instance_id;
+                found = true;
+            }
+            std::cerr << "Display " << i << ": " << name << std::endl;
```
stephengtuggy: And here

engine/src/gldrv/winsys.cpp
```diff
@@ -202,43 +202,69 @@ int native_resolution_x;
...
+        }
+        SDL_free(displays);
+    }
+    std::cerr << "===============================================================" << std::endl;
```
stephengtuggy: Etc.

engine/CMakeLists.txt
```diff
@@ -85,25 +85,25 @@ IF (NOT DISABLE_CLIENT) ##########
...
     #Find SDL
-    FIND_PACKAGE(SDL2 REQUIRED)
-    IF (SDL2_FOUND)
-        LIST(APPEND VSE_TST_INCLUDES ${SDL2_INCLUDE_DIRS})
-        LIST(APPEND VSE_TST_LIBS ${SDL2_LIBRARIES})
-        MESSAGE(STATUS "SDL2 Found")
+    find_package(SDL3 REQUIRED CONFIG REQUIRED COMPONENTS SDL3)
+    IF (SDL3_FOUND)
+        LIST(APPEND VSE_TST_INCLUDES SDL3::Headers)
+        LIST(APPEND VSE_TST_LIBS SDL3::SDL3-shared)
```
stephengtuggy, line 91: Does this work? I've tried something similar with other dependencies in the past, but it didn't work. I wasn't able to add library targets containing `::` like this to a list. I could only specify them directly on the target library or executable. I will be interested to see if this works.

script/bootstrap_mint_sdl3
```diff
@@ -0,0 +1,117 @@
...
+    fi 
+    cd SDL3-source
+    git checkout
+    mkdir -p  build-${CMAKE_BUILD_TYPE}
+    cmake -G Ninja -S ./ -B build-${CMAKE_BUILD_TYPE} -DCMAKE_INSTALL_PREFIX:PATH=${INSTALL_DIR} -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DSDL_EXAMPLES=OFF -DSDL_STATIC=OFF -DSDL_SHARED=ON
+    ninja -C build-${CMAKE_BUILD_TYPE}
+     
+    echo "INSTALL INTO ${INSTALL_DIR}"
+    sudo ninja -C build-${CMAKE_BUILD_TYPE} install
+    sudo sudo ldconfig
```
stephengtuggy, line 88: `sudo sudo` ?

stephengtuggy:
I learned at least two new things from this PR: `std::rint` and `std::calloc`. Good stuff!

evertvorster:
With the latest commit, switching views now works as before. I'll do some extra play testing, but for now I am not finding any flaws from a playability point of vieew

engine/src/gldrv/winsys.cpp
```diff
@@ -208,24 +208,20 @@ bool get_sdl_display_name_by_nr(int screen_number, std::string& screen_name, SDL
...
     int num_displays = 0;
     bool found = false;
-    std::cerr << "===============================================================" << std::endl;
     SDL_DisplayID* displays = SDL_GetDisplays(&num_displays);
     if (displays) {
         int i;
         for (i = 0; i < num_displays; ++i) {
             SDL_DisplayID instance_id = displays[i];
             const char* name = SDL_GetDisplayName(instance_id);
-            display_names << "Display " << i << ": " << (name ? name : "Unknown") << "\n";
```
BenjamenMeyer: this is all good debug information. I'd rather see it moved to VS_LOG than removed

evertvorster:
Keys for flight control now work as before. 
Great work!

engine/src/in_kb.cpp
```diff
@@ -193,20 +194,27 @@ void glut_keyboard_cb(unsigned int ch, unsigned int mod, bool release, int x, in
...
-        for (int a = 0; a < KEYMAP_SIZE; a++) {
-            const std::string map_key = std::to_string(i) + "-" + std::to_string(a);
-
-            if (keyBindings[map_key].state == DOWN) {
-                keyBindings[map_key].function(keyBindings[map_key].data, RELEASE);
-                keyBindings[map_key].state = UP;
-            }
+    // pmx-20251030 This should work with any flavour of C++, even the oldest.
+    std::map<std::string, HandlerCall>::iterator it;
+    for (it=keyBindings.begin(); it != keyBindings.end(); it++) {
```
BenjamenMeyer, line 199: it's okay to use `auto` for the type in this case

BenjamenMeyer (changes requested):

engine/src/in_kb.cpp
stephengtuggy: I agree with @BenjamenMeyer here. Preserving compatibility with very old versions of C++ is not one of our goals. Instead, we try to strike a good balance between targeting a version of C++ that is fairly up to date, but not so up to date as to lack support on the major OS distro versions. Right now, we are targeting C++14.

script/bootstrap_mint_sdl3
pmxy: Yes, sudo, because I cheated and installed the SDL3 libs, includes and cmake macros in /usr/local/, so CMake can find it without touching the guts of the CMakeLists.txt files. 
The other benefit was that I  didn't clutter the main installation, most officially packaged Linux Mint libs are installed in /usr.

Please, remember that I had to make CMake happy first, to simply be able to evaluate the feasibility of the SDL3 port.

This script was merely a transitional helper for myself and other tinkerers.
In fact, I expected people to first try the early code outside of the github CI system, on their own machines. The hacker's way :)

Anyway, all this is currently rewritten and will only install locally (WIP).

stephengtuggy: On this particular line, you put `sudo` twice. That was my point.

stephengtuggy:
Obsoleted by #1474

PR #1366, opened 2025-10-29
SDL3 port to branch 0.9 0.9.x sdl3

pmxy:
Please discard previous 0.9 PR... (wrong base branch)

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -360,44 +360,202 @@ function bootstrapOnLinuxMint ()
     fi
 
     case "$LINUX_CODENAME" in
-        "wilma")
+    "wilma"|"zara")
             apt-get -qy install \
-                            git \
-                            cmake \
-                            python3-dev \
-                            build-essential \
-                            automake \
-                            autoconf \
-                            libarchive-dev \
-                            libpng16-16 \
-                            libpng-dev \
-                            libpng-tools \
-                            libjpeg-turbo8-dev \
-                            libexpat1-dev \
-                            libgdk-pixbuf-2.0-dev \
-                            libgtk-3-dev \
-                            libopenal-dev \
-                            libogg-dev \
-                            libvorbis-dev \
-                            libglvnd-dev \
-                            libgl1-mesa-dev \
-                            libsdl3-dev \
-                            libsdl3-image-dev \
-                            libopengl0 \
-                            libpostproc-dev \
-                            freeglut3-dev \
-                            libboost-python-dev \
-                            libboost-log-dev \
-                            libboost-regex-dev \
-                            libboost-json-dev \
-                            libboost-program-options-dev \
-                            libxmu-dev \
-                            clang \
-                            lsb-release
+                      git \
+                      cmake \
+                      python3-dev \
+                      build-essential \
+                      automake \
+                      autoconf \
+                      libarchive-dev \
+                      libpng16-16 \
+                      libpng-dev \
+                      libpng-tools \
+                      libjpeg-turbo8-dev \
+                      libexpat1-dev \
+                      libgdk-pixbuf-2.0-dev \
+                      libgtk-3-dev \
+                      libopenal-dev \
+                      libogg-dev \
+                      libvorbis-dev \
+                      libglvnd-dev \
+                      libgl1-mesa-dev \
+                      libsdl3 \
+                      libsdl3_image \
+                      libopengl0 \
+                      libpostproc-dev \
+                      freeglut3-dev \
+                      libboost-python-dev \
+                      libboost-log-dev \
+                      libboost-regex-dev \
+                      libboost-program-options-dev \
+                      libxmu-dev \
+                      gcc \
+                      lsb-release
+      ;;
+    "virginia")
+      apt-get -qy install \
+                      git \
+                      cmake \
+                      python3-dev \
+                      build-essential \
+                      automake \
+                      autoconf \
+                      libpng16-16 \
+                      libpng-dev \
+                      libpng-tools \
+                      libjpeg-turbo8-dev \
+                      libexpat1-dev \
+                      libgdk-pixbuf-2.0-dev \
+                      libgtk-3-dev \
+                      libopenal-dev \
+                      libogg-dev \
+                      libvorbis-dev \
+                      libglvnd-dev \
+                      libgl1-mesa-dev \
+                      libsdl1.2-compat-dev \
+                      libsdl2-dev \
+                      libopengl0 \
+                      libpostproc-dev \
+                      freeglut3-dev \
+                      libboost-python-dev \
+                      libboost-log-dev \
+                      libboost-regex-dev \
+                      libboost-program-options-dev \
+                      libxmu-dev \
+                      clang \
+                      lsb-release
             ;;
-        "virginia"|"victoria"|"vera"|"vanessa"|"ulyana")
-            echo "Sorry, Linux Mint ${LINUX_CODENAME} is no longer supported"
-            exit 2
+    "victoria")
+      apt-get -qy install \
+                      git \
+                      cmake \
+                      python3-dev \
+                      build-essential \
+                      automake \
+                      autoconf \
+                      libpng16-16 \
+                      libpng-dev \
+                      libpng-tools \
+                      libjpeg-turbo8-dev \
+                      libexpat1-dev \
+                      libgdk-pixbuf-2.0-dev \
+                      libgtk-3-dev \
+                      libopenal-dev \
+                      libogg-dev \
+                      libvorbis-dev \
+                      libglvnd-dev \
+                      libgl1-mesa-dev \
+                      libsdl1.2-compat-dev \
+                      libsdl2-dev \
+                      libopengl0 \
+                      libpostproc-dev \
+                      freeglut3-dev \
+                      libboost-python-dev \
+                      libboost-log-dev \
+                      libboost-regex-dev \
+                      libboost-program-options-dev \
+                      libxmu-dev \
+                      clang \
+                      lsb-release
+      ;;
+    "vera")
+      apt-get -qy install \
+                      git \
+                      cmake \
+                      python3-dev \
+                      build-essential \
+                      automake \
+                      autoconf \
+                      libpng16-16 \
+                      libpng-dev \
+                      libpng-tools \
+                      libjpeg-turbo8-dev \
+                      libexpat1-dev \
+                      libgdk-pixbuf-2.0-dev \
+                      libgtk-3-dev \
+                      libopenal-dev \
+                      libogg-dev \
+                      libvorbis-dev \
+                      libglvnd-dev \
+                      libgl1-mesa-dev \
+                      libsdl1.2-compat-dev \
+                      libsdl2-dev \
+                      libopengl0 \
+                      libpostproc-dev \
+                      freeglut3-dev \
+                      libboost-python-dev \
+                      libboost-log-dev \
+                      libboost-regex-dev \
+                      libboost-program-options-dev \
+                      libxmu-dev \
+                      clang \
+                      lsb-release
+      ;;
+    "vanessa")
+      apt-get -qy install \
+                      git \
+                      cmake \
+                      python3-dev \
+                      build-essential \
+                      automake \
+                      autoconf \
+                      libpng16-16 \
+                      libpng-dev \
+                      libpng-tools \
+                      libjpeg-turbo8-dev \
+                      libexpat1-dev \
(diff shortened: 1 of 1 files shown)


engine/CMakeLists.txt
```diff
@@ -1440,23 +1448,7 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
-                ELSEIF (DEBIAN_RELEASE_VERSION STREQUAL "stretch")
-                    SET(CPACK_DEBIAN_PACKAGE_DEPENDS "libpython3.5")
-                    IF (Boost_DERIVED_VERSION STREQUAL "")
-                        SET(BOOST_VER "1.62.0")
-                    ELSE (Boost_DERIVED_VERSION STREQUAL "")
-                        SET(BOOST_VER "${Boost_DERIVED_VERSION}")
-                    ENDIF (Boost_DERIVED_VERSION STREQUAL "")
-                    SET(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libjpeg62-turbo, libpng16-16, freeglut3, libgtk-3-0, libvorbis0a, libopenal1, libsdl2-gfx2.0.14, xdg-utils, libgl1, libopengl0, libboost-atomic${BOOST_VER}, libboost-chrono${BOOST_VER}, libboost-date-time${BOOST_VER}, libboost-filesystem${BOOST_VER}, libboost-log${BOOST_VER}, libboost-python${BOOST_VER}, libboost-regex${BOOST_VER}, libboost-system${BOOST_VER}, libboost-thread${BOOST_VER}, libboost-program-options${BOOST_VER}")
```
BenjamenMeyer, line 1459: Please don't drop the other Debian distros. 0.9.x is locked on which distros it supports. We will add/remove only when absolutely necessary for our release branch; most of that is done via our master branch for our future releases.

engine/src/cg_global.cpp
```diff
@@ -33,9 +33,7 @@
 #include "vs_globals.h"
 
 #include <glh/glh_extensions.h>
-// #include <iostream>
-#include <string>
-#include <shared/data_path.h>
+g#include <shared/data_path.h>
```
BenjamenMeyer, line 36: typo?

engine/src/cg_global.cpp
```diff
@@ -33,9 +33,7 @@
 #include "vs_globals.h"
 
 #include <glh/glh_extensions.h>
-// #include <iostream>
-#include <string>
```
BenjamenMeyer, line 37: if `std::string` is used in the file then the include should be done here too.

engine/src/cmd/unit_csv_factory.h
```diff
@@ -209,7 +209,8 @@ inline int UnitCSVFactory::GetVariable(std::string unit_key, std::string const &
     }
     try {
         return std::stoi(result);
-    } catch (std::invalid_argument &) {
+    // } catch (std::invalid_argument &) {
+    } catch (...) {
```
BenjamenMeyer, line 213: Please clean up

engine/src/gldrv/winsys.h
```diff
@@ -133,22 +136,23 @@ typedef enum {
...
+    WSK_MOD_RSHIFT = 0x0002, // SDL_KMOD_RSHIFT,
+    WSK_MOD_LCTRL = 0x0040u, // SDL_KMOD_LCTRL,
+    WSK_MOD_RCTRL = 0x0080u, // SDL_KMOD_RCTRL,
+    WSK_MOD_LALT = 0x0100, // SDL_KMOD_LALT,
+    WSK_MOD_RALT = 0x0200u, // SDL_KMOD_RALT,
+    WSK_MOD_LMETA = 0x0400, // SDL_KMOD_LGUI, // This is an issue for these two entries. We are ignoring the modifier.
+    WSK_MOD_RMETA = 0x0800, // SDL_KMOD_RGUI, // Need to figure out how to do modifier in sdl2.
+    WSK_MOD_NUM = 0x1000u, // SDL_KMOD_NUM,
+    WSK_MOD_CAPS = 0x2000u, // SDL_KMOD_CAPS,
+    WSK_MOD_MODE = 0x4000 // SDL_KMOD_MODE // AltGr
```
BenjamenMeyer, line 155: Please just update the names

engine/src/gldrv/winsys.h
```diff
@@ -167,8 +171,8 @@ typedef enum {
 } winsys_mouse_button_t;
 
 typedef enum {
-    WS_MOUSE_DOWN = SDL_PRESSED,
-    WS_MOUSE_UP = SDL_RELEASED
+    WS_MOUSE_DOWN = true,
+    WS_MOUSE_UP = false
```
BenjamenMeyer, line 175: Please keep the names

run-vs.sh
BenjamenMeyer, line 1: this file probably shouldn't be part of the repository. We don't make any guarantee on where the assets are going to be

engine/src/gfx/aligned.h
```diff
@@ -29,6 +29,7 @@
 
 #include <memory>
 #include <cstddef>
+#include <cstdint>
```
royfalk, line 32: Why add a header without any other changes?
If it's used in a cpp file, why not add the header there instead?

engine/src/gfx/cockpit.cpp
```diff
@@ -636,17 +644,29 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
...
-                numtimes -= .1 + fpsval;
-            }
-            if (numtimes <= 0) {
-                numtimes = fpsmax;
-                fpsval = GetElapsedTime();
-            }
-            if (fpsval) {
-                return 1. / fpsval;
+        case UnitImages<void>::COCKPIT_FPS:  {
+            float dt = GetElapsedTime();
```
royfalk, line 648: Please extract this to a separate PR and function.
Also, this belongs in master. Adding fixes like this in obsolete branches means someone needs to port them to master or they'll be lost.

engine/src/gfx/technique.cpp
```diff
@@ -371,9 +371,15 @@ Technique::Technique(const string &name) :
     } else {
         filename = sub_technique_filename;
     }
-
+
```
royfalk, line 374: Nice cleanup but again belongs in master and a separate PR than SDL. I myself am fine with several small fixes in one PR, but switch to SDL 3 ain't one.

Also, this changes the behavior from crash the game to keep going. I don't know enough about this code but specifically here I prefer to crash early, rather than chase some graphic artifact because a technique didn't compile. 
Please change VS_LOG to VS_LOG_FLUSH_EXIT which writes the log and exits.

royfalk (commented):
I've started reviewing this and then stopped.

I'd like to see a minimal PR on master that only deals with a port to SDL3. Before touching the joystick code and certainly without the very welcome bug fixes I saw sprinkled in the code.

I prefer not to touch 0.9 at all if possible. I realise auto pilot isn't working as well but we're not going back. The way forward is to figure out the regression and fix it.
I actually plan to look at this and maybe rewrite it using AI of some sort.

If you have less free time to jump through these hoops, I'll gladly do this work and attribute it to you.

stephengtuggy:
> 1. 0.9.x cannot drop support for platforms unless those platforms via this PR
> 
>     2. similar comments to [#1364](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1364#pullrequestreview-3393420421)
> 
>     3. We need to add SDL3 to our base images before this can be built in CI
> 
> 
> Since we don't normally accept major changes of this nature in our release branches, I'm mixed on whether to accept this or whether to make a separate branch/release (0.9.x-SDL3). @stephengtuggy @royfalk thoughts?

If we're going to accept this change, it should definitely be on a separate branch and in a separate release series (i.e. 0.9.x-SDL3, as you mentioned, @BenjamenMeyer ). I'm leaning away from merging the SDL3 changes into 0.9.x at all though. I think we should move to SDL3 on master/0.10.x and later only. This will be a lot less work, and more in keeping with where we have been doing active development over the last few months.

engine/src/gfx/technique.cpp
stephengtuggy: @royfalk Good point that this changes the game's behavior. I do want to mention, however, that it's fairly common for techniques to fail to compile. We can discuss in a separate issue and PR whether to incorporate this change.

engine/src/gldrv/winsys.cpp
```diff
@@ -474,80 +597,81 @@ void winsys_process_events() {
...
+                cleanexit = true;
+                keepRunning = false;
+                break;
+            default:
+                break;
             }
-            SDL_LockAudio();
-            SDL_UnlockAudio();
+            // SDL_LockAudio();
+            // SDL_UnlockAudio();
```
stephengtuggy, line 674: Is it really a good idea to comment these out? I'm pretty sure they were there for a reason.

engine/src/gldrv/winsys.cpp
```diff
@@ -474,80 +597,81 @@ void winsys_process_events() {
         memset(keysym_to_unicode, 0, sizeof(keysym_to_unicode));
     }
     while (keepRunning) {
-        SDL_LockAudio();
-        SDL_UnlockAudio();
+        // SDL_LockAudio();
+        // SDL_UnlockAudio();
```
stephengtuggy, line 601: Same here -- should these really be commented out?

engine/src/gldrv/winsys.cpp
pmxy: Phased out in SDL3, which doesn't requires this kind of lock anymore, as per the SDL2->SDL3 porting documents.
https://wiki.libsdl.org/SDL3/README-migration

engine/src/gfx/technique.cpp
pmxy: This change is only on empty line, probably because of a stray character in one of the myriad of  files automatically opened when tracing code execution with gdb. Nothing was touched in the guts of the graphics code.

engine/src/gfx/cockpit.cpp
pmxy: All this exist in 0.10 too.

engine/src/gfx/aligned.h
pmxy: Answer : I probably needed it at one time, during the evolution of the code.

pmxy: OK, I remember !!
There  is a bug in the "official" 0.9. x branch that prevent to compile (at least with the compiler I use, the latest gcc) the uint8, uint16 etc... types were not recognized, hence the <cstdint> include.

engine/src/gldrv/winsys.h
pmxy: SDL_PRESSED and SDL_RELEASED don't exist anymore in SDL3, and must be replaced by true and false.
https://wiki.libsdl.org/SDL3/README-migration

engine/src/gldrv/winsys.h
pmxy: I explained on gitter. BTW, the symbolic constants names are still there. Feel free to give a fix, as I didn't find any other way to get this part of the code to compile, strangely.

engine/CMakeLists.txt
pmxy: FIxed. Mishap during the transfer from 0.10.

engine/src/cg_global.cpp
pmxy: ? probably corrected in more recent PR, as, indeed, it wouldn't compile (although it is into a #if defined(CG_SUPPORT) clause (no idea what it is).
I just checked , the latest code is clean.

engine/src/cg_global.cpp
pmxy: It is in the latest PR.

engine/src/gldrv/winsys.cpp
pmxy: Same cause, same effect :)

BenjamenMeyer:
moved this to a fork of 0.9.x for SDL3.

Note: we still need to add SDL3 to the base images.

engine/src/gfx/aligned.h
stephengtuggy: Yes, I think I ran into this during my work on my SDL3 PR for 0.10.x, with Rocky Linux 10.0. `<cstdint>` does need to be included, evidently.

stephengtuggy:
We are not planning to implement this change on 0.9.x.

stephengtuggy:
See discussion on #1407

PR #1335, opened 2025-09-17
Bug Fix: Add SDL2-image to the bootstrap

BenjamenMeyer:
- Fix CodeQL and local builders using `script/bootstrap` to configure the system

Thank you for submitting a pull request and becoming a contributor to the Vega Strike Core Engine.

Please answer the following:

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation
- [x] This is a documentation change only

Issues:
- Please list any related issues

Purpose:
- What is this pull request trying to do?
- What release is this for?
- Is there a project or milestone we should apply this to?

PR #1367, opened 2025-10-30
Add a derelict function

royfalk:
And now for something a little lighter. I've started working on the assets side of the game a little. First on the agenda is a derelict ship. It would create a nearly destroyed unit to float in space.

In the future, this could be used in a search for a missing ship. It could either be derelict, destroyed (just some debris floating) or be found OK. 

**Code Changes:**

**Tests made**
This is pretty localised. I've tested the code itself and made sure the derelict appears and functions as expected.

**Pending**
[Upload the assets side code.](https://github.com/vegastrike/Assets-Production/pull/222).

Issues:
- Couldn't change the textures. This would require a better understanding of the mesh/texture code. Right now a ship texture is tied to faction.


BenjamenMeyer:
NOTE: all the build issues seem to stem from SDL2_Image not being found by CMake.
Not sure why this is an issue for the CodeQL builds as they should share the same base as our Linux builds; but the Mac and Windows builds seem to be missing something.

royfalk:
> NOTE: all the build issues seem to stem from SDL2_Image not being found by CMake. Not sure why this is an issue for the CodeQL builds as they should share the same base as our Linux builds; but the Mac and Windows builds seem to be missing something.

This is clearly a regression of some sort. This code does not touch SDL2_Image.

royfalk:
Looking at #1332, it seems that's where we added sdl2-image and that's also where we fixed this.
I've created https://github.com/vegastrike/build-system-docker-images/pull/64 and you added #1335.
I doubt we would have merged this if the checks were unsuccessful.

@BenjamenMeyer Do you understand what happened here?

BenjamenMeyer:
> Looking at #1332, it seems that's where we added sdl2-image and that's also where we fixed this. I've created [vegastrike/build-system-docker-images#64](https://github.com/vegastrike/build-system-docker-images/pull/64) and you added #1335. I doubt we would have merged this if the checks were unsuccessful.
> 
> @BenjamenMeyer Do you understand what happened here?

Looking at https://github.com/vegastrike/Vega-Strike-Engine-Source/network this branch doesn't seem to have been merged with master recently enough to pull in a number of changes, such as the sdl2_image change.

So I'm guessing the way the merge goes it ends up leaving something out... an odd git situation I guess.

Looking at:
```bash
$ git diff master task_derelic_2
```
There's a lot of changes; however, doing a rebase in this case may make sense and reduces the changes to a very small set:
```bash
$ git checkout task_derelic_2
$ git rebase master
```
While I generally don't like rebasing, in this case I think its probably the right thing to do and what you wanted. You'll just have to do a force push after doing so.

BenjamenMeyer (approved):

PR #1389, opened 2025-11-13
blender_xmesh_import.py: Fix syntax errors

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Fix a CodeQL warning by correcting the syntax in the Python file it's complaining about

### DIFF
--- a/engine/objconv/blender_xmesh_import.py
+++ b/engine/objconv/blender_xmesh_import.py
@@ -7,8 +7,6 @@ Group: 'Import'
 Tooltip: 'Import VegaStrike Models (.xmesh)'
 """
 
-from __future__ import print_function
-
 __author__	= "Alex 'CubOfJudahsLion' Feterman"
 __url__		= ("blender", "http://www.blender.org", "Author's homepage, http://geocities.com/cubofjudahslion")
 __version__	= "0.1.2"


engine/objconv/blender_xmesh_import.py
```diff
@@ -7,6 +7,8 @@
 Tooltip: 'Import VegaStrike Models (.xmesh)'
 """
 
+from __future__ import print_function
```
BenjamenMeyer: we should be entirely Python3 so this shouldn't necessary

stephengtuggy: Good point.

stephengtuggy: Fixed.

engine/objconv/blender_xmesh_import.py
```diff
@@ -37,10 +37,9 @@
 	Generates UV mappings, but for the texture to be activated, go to
 	the texture buttons and 
 """
-from __future__ import print_function
 try:
 	range = xrange
```
BenjamenMeyer, line 41: FYI - `xrange` is Py2's version of Py3's `range` - this won't break anything but could be cleaned up.
Not going to block this PR on it, but we probably should do a review of our Python code to cleanup things like this.

BenjamenMeyer (approved):

PR #1391, opened 2025-11-16
SDL3 via FetchContent_MakeAvailable

### LINKED ISSUE #1351: Add Rocky Linux 10.0 support to the Vega Strike Game Engine

### LINKED ISSUE #1369: Upgrade to SDL3
Currently @pmxy is working on porting the game to SDL3. He is working on branches 9 and 10 and not master.

There are significant issues with getting SDL3 to play nice with package managers and CMake.

### LINKED ISSUE #1395: `std::stod`, `std::stoi`, etc. do not convert strings to numbers properly with French locale
As reported (and fixed) by @pmxy

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Closes #1369 
- Closes #1351 
- Fixes #1395 

Purpose:
- What is this pull request trying to do? Upgrade to SDL3 using an alternate approach for the dependencies. Also add Rocky Linux 10.0 support/builds.


stephengtuggy:
@evertvorster In this PR, I believe I improved some things related to the `INSTALL_GTEST` and `USE_GTEST` settings. You should now be able to set them using environment variables of the same names. `"ON"` and `"1"` are truthy; everything else is falsy.

And you should now be able to enable `USE_GTEST` and have it work with `INSTALL_GTEST` disabled. I doubt it worked before, looking at the code.

I will be very curious to hear what results you get.

evertvorster:
Hmm, during the configure phase, there is a fairly large download, just after tee 
```
** Build Type: RelWithDebInfo
``` 
message. 
Then again just after the 
```
-- SDL3 found via FetchContent_MakeAvailable
```
 message. 

What is it downloading? From a security point of view, downloading things during compilation is frowned upon, and doubly so when it is not made clear what is being downloaded. 

With a bit of research, I see that there are now a few git submodules for packages that are already on Arch. For the Arch build process, it always starts with a clean build, which means that these sources will be downloaded every single time the package is compiled. 
You can probably already see where I am going with this, but let me spell it out clearly:
On Arch, these downloads are unwanted.

When I am on a very slow and restrictive network, I will be unable to test this branch because this download will just fail, and I will get in trouble for downloading large amounts of data. 
How do I turn this off?

I have used the -DINSTALL_GTEST=OFF compiler flag, which stopped the download of gtest in the past. 
I'm hoping for a similar flag here, or maybe the build process looking for the needed packages in the system and not downloading them if found.

They should be easily detectable with pkg-config:
```
evert@Evert ~/A/vegastrike-engine-git (master)> yay -Ql sdl3 | grep ".pc"
sdl3 /usr/lib/pkgconfig/sdl3.pc
evert@Evert ~/A/vegastrike-engine-git (master)> yay -Ql sdl3_image | grep ".pc"
sdl3_image /usr/lib/pkgconfig/sdl3-image.pc
```

engine/src/cmd/basecomputer.cpp
```diff
@@ -540,25 +540,23 @@ void BaseComputer::constructControls(void) {
         return;
     }
 
-    if (m_displayModes.size() != 1) {
```
royfalk, line 543: Why delete the if?
I looked at the code and I think the original devs wanted to reuse the display for a single mode outside of base computer. Which is why you wouldn't get the base, player name, etc.
Once scenario would be the load new game at the start (which now uses hand written python code instead). 
If we ever want to go back to reusing this code in that scenario, this would need to go back in.

engine/src/components/shield.cpp
```diff
@@ -83,15 +84,7 @@ void Shield::Load(std::string unit_key) {
...
-        try {
-            number_of_facets = std::stoi(shield_facets_string);
-        } catch (std::invalid_argument const& ex) {
-            VS_LOG(error, (boost::format("%1%: %2% trying to convert shield_facets_string '%3%' to int") % __FUNCTION__ % ex.what() % shield_facets_string));
-            number_of_facets = 1;
-        } catch (std::out_of_range const& ex) {
-            VS_LOG(error, (boost::format("%1%: %2% trying to convert shield_facets_string '%3%' to int") % __FUNCTION__ % ex.what() % shield_facets_string));
-            number_of_facets = 1;
-        }
+        number_of_facets = locale_aware_stoi(shield_facets_string, 1);
```
royfalk, line 87: Nice.

engine/src/components/tests/balancing_tests.cpp
```diff
@@ -28,12 +28,11 @@
 
 #include <gtest/gtest.h>
 
+#include "vegastrike.h"
 #include "components/energy_container.h"
 #include "components/reactor.h"
 #include "configuration/game_config.h"
 
-double simulation_atom_var = 0.1;
```
royfalk, line 35: I assume you included vegastrike.h to remove this line.
But, is this wise? 

The whole point of this test is that it tests balancing under specific parameters. Taking the most basic parameter and making it variable, could lead to failures. Looking at the tests, it looks like they're hard coded to 0.1. See lines 162-163.

engine/src/gfx/cockpit.cpp
```diff
@@ -307,13 +301,15 @@ void GameCockpit::AutoLanding() {
 
 
 float GameCockpit::LookupUnitStat(int stat, Unit *target) {
+    if (!target) {
```
royfalk, line 304: It'd be worth opening an issue to move this upward. 
A lot of the calls to this are in a for loop for all stats.
Also, I wonder if it can even happen. You are supposed to go through cockpit. Can you have a cockpit without a unit?
I'm not saying we shouldn't check but that we should move the unit checks as high as they can go.

BenjamenMeyer:
@stephengtuggy thanks - this really cleans up the other PR and removes a lot of things that I was blocking it on.
Overall LGTM but let's address @evertvorster 's download issue and @royfalk's questions/issues.

engine/src/cmd/basecomputer.cpp
stephengtuggy: Hmm. I'm not following here. `the original devs wanted to reuse the display for a single mode outside of base computer` ? Can you elaborate further?

@pmxy made this change in their PR, and it looked to me like it fixed an issue we hadn't noticed previously, where some variables were left uninitialized when there was only one display (mode?).

stephengtuggy:
@evertvorster Good point. We should tell users what we're downloading. I will add that to the CMake logic.

Also, I didn't realize that `-DINSTALL_GTEST=OFF` would no longer work. I will see if I can get that working again. For now, though, you should simply be able to set the `INSTALL_GTEST` environment variable to `"OFF"` for the same effect. And similarly, setting `INSTALL_SDL3` and `INSTALL_SDL3_IMAGE` environment variables to `"OFF"` should disable the other two downloads.

engine/src/gfx/cockpit.cpp
stephengtuggy: I believe this change was already made to the code on the `master` branch. For some reason, it wasn't included in the `master-sdl3` branch, so it's showing up in this PR. But it's not really a new change.

Also, there are several different conditions being checked here, and none of them are for the current unit, AFAICT. We are checking: 1) do we have a non-null `target`? 2) Do we have a non-null `target->Threat()`? 3) Do we have a non-null `target->VelocityReference()`?

engine/src/gfx/cockpit.cpp
```diff
@@ -434,6 +431,12 @@ float GameCockpit::LookupUnitStat(int stat, Unit *target) {
         case UnitImages<void>::MAXCOMBATKPS:
         case UnitImages<void>::MAXCOMBATABKPS: {
             const bool use_relative_velocity = configuration().graphics.hud.display_relative_velocity;
+            if (!target) {
+                return 0.0f;
+            }
```
stephengtuggy: This is the only `if` check I see in this file that could probably be eliminated, since we already tested this condition at the top of the function.

engine/src/components/tests/balancing_tests.cpp
stephengtuggy: I think it's more like I included vegastrike.h because that's where I originally placed the `locale_aware_sto...` functions. Then I moved them to a different header, and forgot to remove the vegastrike.h include from this file. Then I got an error about `simulation_atom_var` being defined twice, and solved it with this.

Maybe I should just take out vegastrike.h again, and see if that fixes it.

engine/src/gfx/cockpit.cpp
stephengtuggy: Removed.

engine/src/components/tests/balancing_tests.cpp
stephengtuggy: Removing vegastrike.h does not fix it. `simulation_atom_var` still needs to be declared -- as `extern`, in this file -- and as such, it cannot be initialized on the same line. So I am now initializing it at the beginning of each relevant test body.

I'm still not sure exactly what happened here.

engine/src/components/shield.cpp
stephengtuggy: Glad you like it

stephengtuggy:
> @stephengtuggy thanks - this really cleans up the other PR and removes a lot of things that I was blocking it on. Overall LGTM but let's address @evertvorster 's download issue and @royfalk's questions/issues.

@BenjamenMeyer cool. I believe I have addressed @evertvorster 's issues, along with some of @royfalk 's. Discussion continuing on a couple of others.

engine/src/gfx/cockpit.cpp
stephengtuggy: Never mind. I found several extra `if (!target)` checks and removed them. Not sure why I didn't see them before.

stephengtuggy:
If we get #1394 reviewed and merged first, this PR should look a lot cleaner. It will then include only the truly new commits.

stephengtuggy:
> If we get #1394 reviewed and merged first, this PR should look a lot cleaner. It will then include only the truly new commits.

This is now complete.

evertvorster:
> @evertvorster Good point. We should tell users what we're downloading. I will add that to the CMake logic.
> 
> Also, I didn't realize that `-DINSTALL_GTEST=OFF` would no longer work. I will see if I can get that working again. For now, though, you should simply be able to set the `INSTALL_GTEST` environment variable to `"OFF"` for the same effect. And similarly, setting `INSTALL_SDL3` and `INSTALL_SDL3_IMAGE` environment variables to `"OFF"` should disable the other two downloads.

While this works, it appears quite hackish to have to set environment variables AND -D variables. 
For instance, you have to set both INSTALL_GTEST=OFF and -DUSE_GTEST=OFF, else configuration will fail. 

I was under the impression that the "-D" cmake options are somehow better than setting environment variables. 

However, this is not a blocker for me at this point, because you can actually turn off the download of gtest, sdl3 and sdl3_image. 

Let me give this a play test and see if there is anything untoward. The last time I played this PR, it worked quite well.

stephengtuggy:
@evertvorster you should be able to use only environment variables to set all 4 settings, if I'm not mistaken. Let me know if that's not the case.

BenjamenMeyer:
> > @evertvorster Good point. We should tell users what we're downloading. I will add that to the CMake logic.
> > Also, I didn't realize that `-DINSTALL_GTEST=OFF` would no longer work. I will see if I can get that working again. For now, though, you should simply be able to set the `INSTALL_GTEST` environment variable to `"OFF"` for the same effect. And similarly, setting `INSTALL_SDL3` and `INSTALL_SDL3_IMAGE` environment variables to `"OFF"` should disable the other two downloads.
> 
> While this works, it appears quite hackish to have to set environment variables AND -D variables. For instance, you have to set both INSTALL_GTEST=OFF and -DUSE_GTEST=OFF, else configuration will fail.
> 
> I was under the impression that the "-D" cmake options are somehow better than setting environment variables.

Yes per https://gitlab.kitware.com/cmake/community/-/wikis/faq#how-can-i-get-or-set-environment-variables `-D` is preferred since it caches the value so its can be easily re-used.

stephengtuggy:
> > > @evertvorster Good point. We should tell users what we're downloading. I will add that to the CMake logic.
> > > Also, I didn't realize that `-DINSTALL_GTEST=OFF` would no longer work. I will see if I can get that working again. For now, though, you should simply be able to set the `INSTALL_GTEST` environment variable to `"OFF"` for the same effect. And similarly, setting `INSTALL_SDL3` and `INSTALL_SDL3_IMAGE` environment variables to `"OFF"` should disable the other two downloads.
> > 
> > 
> > While this works, it appears quite hackish to have to set environment variables AND -D variables. For instance, you have to set both INSTALL_GTEST=OFF and -DUSE_GTEST=OFF, else configuration will fail.
> > I was under the impression that the "-D" cmake options are somehow better than setting environment variables.
> 
> Yes per https://gitlab.kitware.com/cmake/community/-/wikis/faq#how-can-i-get-or-set-environment-variables `-D` is preferred since it caches the value so its can be easily re-used.

You should also be able to use `-D` to set all 4, if you prefer. 😄

engine/src/cmd/basecomputer.cpp
royfalk: I think the devs originally meant this code to also be used for load game at the start of the game. That means you are not technically in a base and so you don't have base name, money, date, etc.
They later "solved" this by implementing this in python. 
If my guess is correct, single mode means "load save game only" without news, trade, etc.

engine/src/components/tests/balancing_tests.cpp
royfalk: I think the test files were part of a much smaller compilation unit. Once vegastrike.h got pulled into it somewhere, the global is there, and my sim_atom_var breaks the **One Definition Rule**.

engine/src/cmd/basecomputer.cpp
stephengtuggy: I see what you're saying. Looks like you're right. I will make a note to reintroduce this `if` block.

PR #1413, opened 2025-11-19
Change TextAreaItem to use a deque of smart pointers to hold its child items

### LINKED ISSUE #1414: Segfault on Main Menu "Multiplayer" option
When I click on "Multiplayer" on the Main Menu, I get an immediate segfault currently. At least if I do so after first clicking on Credits and Introduction.

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Fixes #1414 on `master`

Purpose:
- What is this pull request trying to do? Fix a CodeQL security warning or two by replacing `TextAreaItem`'s `child` array with a `deque` of smart pointers. Also optimize the relevant files, clean up some additional warnings, replace C-style casts and `static_cast`s of UI controls with `vega_dynamic_cast_ptr`, and fix #1414 .

### DIFF
--- a/engine/src/cmd/base_util.cpp
+++ b/engine/src/cmd/base_util.cpp
@@ -670,6 +670,7 @@ void CompPython(int room,
             EnumMap::Pair("News", BaseComputer::NEWS),
             EnumMap::Pair("Info", BaseComputer::INFO),
             EnumMap::Pair("LoadSave", BaseComputer::LOADSAVE),
+            EnumMap::Pair("Network", BaseComputer::NETWORK),
             EnumMap::Pair("UNKNOWN", BaseComputer::LOADSAVE),
     };
     static const EnumMap modemap(modelist, sizeof(modelist) / sizeof(*modelist));
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -284,7 +284,8 @@ static const ModeInfo modeInfo[] = {
         ModeInfo("Missions BBS  ", "Missions", "MissionsMode", "MissionsGroup"),
         ModeInfo("GNN News  ", "News", "NewsMode", "NewsGroup"),
         ModeInfo("Info/Stats  ", "Info", "InfoMode", "InfoGroup"),
-        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup")
+        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup"),
+        ModeInfo("Network ", "Network", "NetworkMode", "NetworkGroup")
 };
 
 bool BaseComputer::actionDone(const EventCommandId &command, Control *control) {
@@ -318,6 +319,7 @@ const BaseComputer::WctlTableEntry WctlBase<BaseComputer>::WctlCommandTable[] = 
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::MISSIONS].command, "", &BaseComputer::changeToMissionsMode),
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::INFO].command, "", &BaseComputer::changeToInfoMode),
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::LOADSAVE].command, "", &BaseComputer::changeToLoadSaveMode),
+        BaseComputer::WctlTableEntry(modeInfo[BaseComputer::NETWORK].command, "", &BaseComputer::changeToNetworkMode),
         BaseComputer::WctlTableEntry("BuyCargo", "", &BaseComputer::buyCargo),
         BaseComputer::WctlTableEntry("Buy10Cargo", "", &BaseComputer::buy10Cargo),
         BaseComputer::WctlTableEntry("BuyAllCargo", "", &BaseComputer::buyAllCargo),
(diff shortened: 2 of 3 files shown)


engine/src/cmd/basecomputer.cpp
```diff
@@ -1773,7 +1773,7 @@ void BaseComputer::updateTransactionControlsForSelection(TransactionList *tlist)
                 if (item.GetCategory().find("My_Fleet") != string::npos) {
                     //This ship is in my fleet -- the price is just the transport cost to get it to
                     //the current base.  "Buying" this ship makes it my current ship.
-                    tempString = (boost::format("#b#Transport cost: %$1.2f#-b#n1.5#") % item.GetPrice()).str();
+                    tempString = (boost::format("#b#Transport cost: %1$.2f#-b#n1.5#") % item.GetPrice()).str();
```
BenjamenMeyer, line 1845: is this correct? seems like the original would have been right since `1.2f` is usually used to denote 1 whole number + 2 decimals (`1.23`, `0.23`)... not sure what the `$` is for though since IIRC the `%` denotes the variable placement with boost.

If it is correct then place add a comment about it so someone doesn't change it in the future; but I highly suspect the original was correct.

engine/src/gui/text_area.cpp
```diff
@@ -181,21 +197,18 @@ void TextArea::Refresh(void) {
     }
 }
 
-void TextArea::RenderText(void) {
+void TextArea::RenderText() {
```
BenjamenMeyer, line 183: technically the `void` parameter is implied so the original was just being more explicit. If you go back to older compilers (early 1980's) then that would have been common to see.

stephengtuggy: Agreed. CLion (or more specifically, clang-tidy) suggested I remove these, I guess because they are no longer necessary with newer compilers. Would you like me to add them back?

engine/src/cmd/basecomputer.cpp
stephengtuggy: This is a positional format string, as per the Unix98 [Open-group printf](http://www.opengroup.org/onlinepubs/7908799/xsh/fprintf.html) precise syntax. This is supported by Boost Format; see https://www.boost.org/doc/libs/latest/libs/format/doc/format.html#printf_directives .

The value I changed it to is correct, and matches one of the other format strings in basecomputer.cpp. The old value was incorrect.

stephengtuggy: Explanatory comment added.

engine/src/cmd/base_util.cpp
```diff
@@ -670,6 +670,7 @@ void CompPython(int room,
             EnumMap::Pair("News", BaseComputer::NEWS),
             EnumMap::Pair("Info", BaseComputer::INFO),
             EnumMap::Pair("LoadSave", BaseComputer::LOADSAVE),
+            EnumMap::Pair("Network", BaseComputer::NETWORK),
```
stephengtuggy, line 673: It appears that we have to keep this until we do away with the corresponding Main Menu item.

engine/src/cmd/basecomputer.cpp
```diff
@@ -284,7 +284,8 @@ static const ModeInfo modeInfo[] = {
         ModeInfo("Missions BBS  ", "Missions", "MissionsMode", "MissionsGroup"),
         ModeInfo("GNN News  ", "News", "NewsMode", "NewsGroup"),
         ModeInfo("Info/Stats  ", "Info", "InfoMode", "InfoGroup"),
-        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup")
+        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup"),
+        ModeInfo("Network ", "Network", "NetworkMode", "NetworkGroup")
```
stephengtuggy, line 288: Same with this

engine/src/cmd/basecomputer.cpp
```diff
@@ -318,6 +319,7 @@ const BaseComputer::WctlTableEntry WctlBase<BaseComputer>::WctlCommandTable[] =
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::MISSIONS].command, "", &BaseComputer::changeToMissionsMode),
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::INFO].command, "", &BaseComputer::changeToInfoMode),
         BaseComputer::WctlTableEntry(modeInfo[BaseComputer::LOADSAVE].command, "", &BaseComputer::changeToLoadSaveMode),
+        BaseComputer::WctlTableEntry(modeInfo[BaseComputer::NETWORK].command, "", &BaseComputer::changeToNetworkMode),
```
stephengtuggy, line 322: Etc.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: Thanks

engine/src/gui/text_area.cpp
BenjamenMeyer: I don't know if they were ever truly required - if so, they haven't been in so long that most folks don't even know it could even be added that way. It's fine to clean that stuff up - it'll be more normal for how most devs expect to read it.

engine/src/cmd/base_util.cpp
BenjamenMeyer: let's just do away with the main menu item.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: same. the menu item should go away

PR #1416, opened 2025-11-21
Change TextAreaItem to use a deque of smart pointers to hold its child items -- SDL3 branch

### LINKED ISSUE #1414: Segfault on Main Menu "Multiplayer" option

### LINKED ISSUE #1419: On Windows 11, 0.10.x SDL3 variant exits with error about unsupported image format

### LINKED ISSUE #1421: On Windows 11, 0.10.x SDL3 variant crashes with access violation trying to iterate through non-existent joysticks

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Fixes #1414 on `master-sdl3`
- Fixes #1419 
- Fixes #1421 

Purpose:
- What is this pull request trying to do? Port #1413 to the `master-sdl3` branch
- What release is this for? 0.10.x (SDL3 variant)
- Is there a project or milestone we should apply this to? 0.10.x, presumably

### DIFF
--- a/engine/src/in_joystick.cpp
+++ b/engine/src/in_joystick.cpp
@@ -64,7 +64,7 @@ static int miny = -1;
 static int maxz = 1;
 static int minz = -1;
 
-JoyStick *joystick[MAX_JOYSTICKS]; //until I know where I place it
+JoyStick *joystick[MAX_JOYSTICKS]{}; //until I know where I place it
 int num_joysticks = 0;
 
 void modifyDeadZone(JoyStick *j) {
@@ -187,13 +187,17 @@ void InitJoystick() {
     if (num_joysticks > MAX_JOYSTICKS) {
         num_joysticks = MAX_JOYSTICKS;
     }
-    VS_LOG(info, (boost::format("%1% joysticks were found.\n\n") % num_joysticks));
-    VS_LOG(info, "The names of the joysticks are:\n");
+    VS_LOG(important_info, (boost::format("%1% joysticks were found.\n\n") % num_joysticks));
+    VS_LOG(important_info, "The names of the joysticks are:\n");
     if (joysticks) {
-        for (int i1 = 0; i1 < num_joysticks; ++i1) {
-            const SDL_JoystickID instance_id = joysticks[i1];
-            VS_LOG(info, (boost::format("    %1%\n") % SDL_GetJoystickNameForID(instance_id)));
-            joystick[i1] = new JoyStick(i1, instance_id);
+        for (int i1 = 0; i1 < MAX_JOYSTICKS; ++i1) {
+            if (i1 < num_joysticks) {
+                const SDL_JoystickID instance_id = joysticks[i1];
+                VS_LOG(important_info, (boost::format("    %1%\n") % SDL_GetJoystickNameForID(instance_id)));
+                joystick[i1] = new JoyStick(i1, instance_id);
+            } else {
+                joystick[i1] = new JoyStick(i1, 0);
+            }
         }
     }
     SDL_free(joysticks);
@@ -207,24 +211,27 @@ void InitJoystick() {
         VS_LOG(info, (boost::format("Glut detects %1% joystick(s)") % (i+1)));
         joystick[0] = new JoyStick(0, 0);
     }
+    for (int i2 = 0; i2 < MAX_JOYSTICKS; ++i2) {
+        joystick[i2] = new JoyStick(i2, 0);
+    }
 #endif
 #endif
 }
 
 void DeInitJoystick() {
-    for (int i = 0; i < MAX_JOYSTICKS; i++) {
-        delete joystick[i];
+    for (auto & i : joystick) {
+        delete i;
     }
     num_joysticks = 0;
 }
 
-JoyStick::JoyStick(int which, SDL_JoystickID instanceID) : mouse(which == MOUSE_JOYSTICK) {
+JoyStick::JoyStick(const int which, const SDL_JoystickID instance_id) : mouse(which == MOUSE_JOYSTICK) {
     for (int j = 0; j < MAX_AXES; ++j) {
         axis_axis[j] = -1;
         axis_inverse[j] = false;
         joy_axis[j] = 0;
     }
-    this->instanceID = instanceID;
+    instanceID = instance_id;
 
     joy_buttons = 0;
     player = 0; //which;     //by default bind players to whichever joystick it is
@@ -244,14 +251,13 @@ JoyStick::JoyStick(int which, SDL_JoystickID instanceID) : mouse(which == MOUSE_
 
 #else
 #ifdef HAVE_SDL
-    // num_joysticks = SDL_NumJoysticks(); // pmx-20251020 Commented : already done in the init stage, and num_joysticks is a global.
     if (which >= num_joysticks) {
         if (which != MOUSE_JOYSTICK) {
             joy_available = false;
         }
         return;
     }
-    joy = SDL_OpenJoystick(instanceID);     //joystick nr should be configurable
+    joy = SDL_OpenJoystick(instance_id);     //joystick nr should be configurable
     if (joy == nullptr) {
         VS_LOG(warning, (boost::format("warning: no joystick nr %1%\n") % which));
         joy_available = false;
--- a/engine/src/in_joystick.h
+++ b/engine/src/in_joystick.h
@@ -77,7 +77,7 @@ class JoyStick {
     void GetMouse(float &x, float &y, float &z, int &buttons);
 public:
 //initializes the joystick
-    JoyStick(int which, SDL_JoystickID instanceID);
+    JoyStick(int which, SDL_JoystickID instance_id);
 //engine calls GetJoyStick to get coordinates and buttons
     void GetJoyStick(float &x, float &y, float &z, long long& buttons);
     bool isAvailable(void);


engine/src/audio/Types.h
```diff
@@ -40,6 +40,7 @@
 #define WeakPtr boost::weak_ptr
 #define AutoPtr boost::scoped_ptr
 #define SharedFromThis boost::enable_shared_from_this
+#define MakeShared boost::make_shared
```
royfalk, line 43: I hate that C++ has ugly `boost::make_shared` and `std::string` but every time we try to improve with `#define` or `using`, we make it a bit harder for new people to join. 
Embrace the ugly! ;)

engine/src/cmd/base_util.cpp
```diff
@@ -670,6 +670,7 @@ void CompPython(int room,
             EnumMap::Pair("News", BaseComputer::NEWS),
             EnumMap::Pair("Info", BaseComputer::INFO),
             EnumMap::Pair("LoadSave", BaseComputer::LOADSAVE),
+            EnumMap::Pair("Network", BaseComputer::NETWORK),
```
royfalk, line 673: I assume this fixes the crash?

engine/src/cmd/basecomputer.cpp
```diff
@@ -284,7 +284,8 @@ static const ModeInfo modeInfo[] = {
         ModeInfo("Missions BBS  ", "Missions", "MissionsMode", "MissionsGroup"),
         ModeInfo("GNN News  ", "News", "NewsMode", "NewsGroup"),
         ModeInfo("Info/Stats  ", "Info", "InfoMode", "InfoGroup"),
-        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup")
+        ModeInfo("Load / Save ", "LoadSave", "LoadSaveMode", "LoadSaveGroup"),
```
royfalk: I'm going to assume you simply reverted my deletions here.
I'm not a big fan of this. 
It shouldn't be too hard to figure out how to prevent a crash without all this jazz.

Why don't I take a stab at it tomorrow and you separate it out of the PR?

engine/src/components/radar.cpp
```diff
@@ -99,15 +99,7 @@ void CRadar::Load(std::string unit_key) {
...
-        std::cout << "Try stoi " << unit_key << std::endl;
-        unsigned int value = stoi(iffval, 0);
-        std::cout << "Success stoi " << unit_key << std::endl;
-        if (value == 0) {
-            // Unknown value
-            capabilities = RadarCapabilities::NONE;
-        } else {
-            capabilities = value;
-        }
+        capabilities = locale_aware_stoi(iffval, RadarCapabilities::NONE);
```
royfalk, line 102: Nice.

engine/src/gui/text_area.cpp
```diff
@@ -105,8 +106,8 @@ TextArea::TextArea(float x, float y, float wid, float hei, int scrollbar) {
     vertical_left_of_text = 0.02;
     text_spacing = (font_size_float / 100) + (horizontal_spacer * 2);
 
-    //The parent TextAreaItem. This is the only link where parent == NULL. It is not displayed and handled only internally
-    ItemList = new TextAreaItem("", "", NULL);
+    //The parent TextAreaItem. This is the only link where parent == nullptr. It is not displayed and handled only internally
+    item_tree_top = MakeShared<TextAreaItem>("", "", nullptr);
```
royfalk, line 110: Is this the real meat of the PR?

engine/src/gui/text_area.cpp
```diff
@@ -181,21 +180,18 @@ void TextArea::Refresh(void) {
     }
 }
 
-void TextArea::RenderText(void) {
```
royfalk, line 184: This whole code feels like it was taken from some C program or written by some C programmer. `RenderText(void)` and init all variables at the start of the function... weird.

engine/src/gui/text_area.cpp
```diff
@@ -208,85 +204,80 @@ void TextArea::RenderTextItem(TextAreaItem *current, int level) {
         if (count < top_item_number) {
             count++;
         } else {
-            new_y = ycoord[5] - (text_spacing * (count + 1 - top_item_number)) + horizontal_spacer;
-            new_x = xcoord[5] + (horizontal_per_level * (level - 1));
+            float new_x = 0;
```
royfalk: Why separate to two lines?

engine/src/gui/text_area.cpp
```diff
@@ -208,85 +204,80 @@ void TextArea::RenderTextItem(TextAreaItem *current, int level) {
...
+    if (!search) {
         return;
     }
-    if (search->description != 0) {
+    if (search->description != nullptr) {
         free(search->description);
+        search->description = nullptr;
     }
-    search->description = strdup(description);
+    search->description = vega_str_dup(description);
```
royfalk, line 241: Definitely C. Why would anyone write char* when std::string is available. I wonder how hard it would be to replace` char*` with `std::string`?

engine/src/gui/text_area.cpp
```diff
@@ -626,90 +581,61 @@ TextAreaItem::~TextAreaItem(void) {
...
-                temp = child[cur2];
-                child[cur2] = child[cur2 + 1];
-                child[cur2 + 1] = temp;
-            }
-        }
+void TextAreaItem::Sort() {
+    for (const auto& cur : children) {
+        cur->Sort();
     }
+    std::sort(children.begin(), children.end(),
```
royfalk, line 606: Nice!

engine/src/cmd/base_util.cpp
stephengtuggy: Yes. Together with the other network stuff I put back in

engine/src/gui/text_area.cpp
stephengtuggy: It's certainly part of it. Along with all the lines where I replaced raw pointers with `WeakPtr` or `SharedPtr`. And the call to `shared_from_this()`. And the `deque` of `SharedPtr` for child items, in place of the array of raw pointers.

engine/src/gui/text_area.cpp
stephengtuggy: Definitely

engine/src/gui/text_area.cpp
stephengtuggy: Glad you like it 😄 . I think it's pretty elegant, myself

engine/src/gui/text_area.cpp
stephengtuggy: Probably not hard. That can be the next refactor, I suppose.

engine/src/gui/text_area.cpp
stephengtuggy: Oh. Duh. I could declare `new_x` and `new_y` and initialize them to their final values on one line each, couldn't I? Good catch

engine/src/vega_cast_utils.h
```diff
@@ -91,20 +91,17 @@ inline long locale_aware_stol(const std::string& s, const long default_value = 0
...
+inline TargetType* vega_dynamic_cast_ptr(SourceType* from, const bool allow_null = false) {
     TargetType* ret_val = nullptr;
     try {
         ret_val = dynamic_cast<TargetType*>(from);
     } catch (std::bad_cast& e) {
         VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Error '%1%' casting type %2%* to %3%*") % e.what() % typeid(SourceType).name() % typeid(TargetType).name()));
         VSExit(-422);
-    } catch (std::exception& e) {
-        VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Error '%1%' casting type %2%* to %3%*") % e.what() % typeid(SourceType).name() % typeid(TargetType).name()));
-        VSExit(-423);
```
stephengtuggy, line 103: CLion informs me that this block is unreachable code. So I guess `std::bad_cast` is the only type of exception that can be thrown here?

It certainly cleans the code up nicely to take this out.

engine/src/gui/text_area.h
```diff
@@ -211,56 +211,54 @@ class TextArea {
...
-    char *name;
-    char *description;
+    char *   name;
+    char *   description;
 
-//The size of the array is 10 * child_count_multiplier. Allows for an expanding array
-    int child_count_multiplier;
-    int child_count;
-    TextAreaItem **child;
+    std::deque<SharedPtr<TextAreaItem>> children;
```
stephengtuggy, line 253: This is one of the key changes, right here

engine/src/gui/text_area.cpp
```diff
@@ -626,90 +581,61 @@ TextAreaItem::~TextAreaItem(void) {
...
-    TextAreaItem *match;
-    if (search_name == NULL) {
-        return this;
+SharedPtr<TextAreaItem> TextAreaItem::FindChild(const char* search_name) {
+    if (search_name == nullptr) {
+        return shared_from_this();
     }
     if (strcmp(name, search_name) == 0) {
-        return this;
+        return shared_from_this();
```
stephengtuggy, line 591: This is another key change. In order to use shared pointers with this class, we have to be able to return `shared_from_this()` where applicable. And in order for that to happen, this class has to inherit publicly from `boost::enable_shared_from_this<TextAreaItem>`.

stephengtuggy: Which it does, via an alias, `SharedFromThis`.

engine/src/gui/text_area.cpp
stephengtuggy: There. Filed as https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1418

engine/src/vega_cast_utils.h
```diff
@@ -91,20 +91,17 @@ inline long locale_aware_stol(const std::string& s, const long default_value = 0
 }
 
 template<class TargetType, class SourceType>
-inline TargetType* vega_dynamic_cast_ptr(SourceType* from) {
+inline TargetType* vega_dynamic_cast_ptr(SourceType* from, const bool allow_null = false) {
```
stephengtuggy, line 94: @royfalk This parameter, and [the commit](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1416/commits/392589c20aeb3ffb770488522bcf0ffc145db65f) in which it was introduced, are new in this PR as opposed to the non-SDL3 version. (I plan to port this over there too.)

engine/src/gui/text_area.cpp
stephengtuggy: Done

engine/src/gui/text_area.cpp
BenjamenMeyer: > Definitely C. Why would anyone write char* when std::string is available. I wonder how hard it would be to replace` char*` with `std::string`?

For the most part it's easy unless you're doing buffer allocation - then it can be easier to allocate it using `char*` and then convert after.

evertvorster:
FWIW, there is no segfault when selecting multiplayer with this PR on linux.

stephengtuggy:
> FWIW, there is no segfault when selecting multiplayer with this PR on linux.

@evertvorster Yeah, I wasn't seeing segfaults on Linux either. But besides that, the segfaults on Windows are now fixed too, I believe.

Thanks for testing this.

stephengtuggy:
This PR is definitely starting to suffer from scope creep. I think I'll leave it as it is unless a showstopper pops up, and work on the remaining Windows-specific issues in another branch.

engine/src/gui/text_area.cpp
BenjamenMeyer: > Why would anyone write char* when std::string is available.

Keep in mind, when VS started around 2000 `std::string` barely existed and wasn't very well known. Most devs just wrote their own version or used the C tooling at that point. That quickly changed by 2003/2005, but legacy code at that point wasn't really updated to use it.

https://en.wikipedia.org/wiki/C%2B%2B_string_handling

stephengtuggy:
There was still a critical, crash-inducing bug in the joystick and mouse logic. That is fixed now, so I think this PR can finally be merged.

engine/src/vega_cast_utils.h
stephengtuggy: Filed as #1462

stephengtuggy: Done porting over

PR #1417, opened 2025-11-23
Initial commit of imgui as a Vegastrike library

royfalk:
This is a restart of #1348, now as a series of stand-alone, small PRs.
Initial commit only has the library and a small test program.

**Code Changes:**
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation **Not relevant yet. This commit does not use imGui in the main game.**

**Issues:**
- This commit introduces imGui as a VS library but code is hard coded. It is not pulling the latest from the imGui git repo or some package manager.


stephengtuggy:
@royfalk Last I knew, I thought @BenjamenMeyer was going to try creating a Homebrew tap and port. I thought that was the route we were going to take.

BenjamenMeyer:
> @royfalk Last I knew, I thought @BenjamenMeyer was going to try creating a Homebrew tap and port. I thought that was the route we were going to take.

Yes. Sorry - it's been chaotic for me lately. I'll try to set some time aside this next week or so to get that setup.
Created https://github.com/vegastrike/homebrew-vegastrike as our own Homebrew Tap where we can add our custom formula.

https://github.com/vegastrike/homebrew-vegastrike/issues/1

BenjamenMeyer (changes requested):
I'm good with this while we figure out how to get the Homebrew Tap solution figured out. Once that's working we can remove this and convert over.

Windows already has https://vcpkg.io/en/package/imgui
And imgui is already available on linux distros via the package managers.
@royfalk could you change this to only be used if CMake doesn't find a system installed one? Or make it a CMake Option where its disabled by default on platforms that are known to have imgui available already?

royfalk:
> @royfalk could you change this to only be used if CMake doesn't find a system installed one? Or make it a CMake Option where its disabled by default on platforms that are known to have imgui available already?

Can we choose **lazy**? This works and works well for now and imGui was always designed to be this self contained chunk of code to be dropped in some other project.

Why don't we just leave this for 6-12 months and see if **somebody else** didn't do it in the meantime?

Can I do it? Probably. 
But I'd prefer to change the code of the game to make it better instead. 

One thing I'm really concerned about it what have we achieved in these several years? The engine has been refactored and we added support for modern OS. But the game itself hasn't really progressed. The campaign isn't there. Graphics aren't better. I think we need to prioritize these.

If imGui is deployed, we can get rid of the old fonts and replace them with something modern. This would also solve issues of spacing that plague some screens.

evertvorster:
There are no compile errors with this PR. 
However, there are also no new files installed with the make install command. 

As with the SDL3 stuff, can't we have a -DINSTALL_IMGUI=OFF flag?
Arch has imgui as a package, and it would be dead easy for me to have it installed as a dependency for vegastrike.

stephengtuggy:
> There are no compile errors with this PR. However, there are also no new files installed with the make install command.
> 
> As with the SDL3 stuff, can't we have a -DINSTALL_IMGUI=OFF flag? Arch has imgui as a package, and it would be dead easy for me to have it installed as a dependency for vegastrike.

I would also like to see a `-DINSTALL_IMGUI=OFF` flag. I can perhaps work on implementing that if you'd rather not, @royfalk .

stephengtuggy:
> > There are no compile errors with this PR. However, there are also no new files installed with the make install command.
> > As with the SDL3 stuff, can't we have a -DINSTALL_IMGUI=OFF flag? Arch has imgui as a package, and it would be dead easy for me to have it installed as a dependency for vegastrike.
> 
> I would also like to see a `-DINSTALL_IMGUI=OFF` flag. I can perhaps work on implementing that if you'd rather not, @royfalk .

Filed as #1424

BenjamenMeyer (approved):
Approving given follow-up of https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1424

PR #1453, opened 2025-11-28
Detect whether on Linux or macOS using `uname` -- `master` branch

### LINKED ISSUE #1443: We do not currently have real OS detection logic for Linux vs. macOS
**Describe the bug**
For bootstrapping purposes, among other things, we need to follow different code paths for Linux vs. macOS vs. Windows. We should auto-detect which one we are running on, but we have no such functionality today. Instead, we just read an environment variable, `MY_OS_NAME`.

**To Reproduce**
You can probably fool the system into thinking it's on macOS simply by setting `MY_OS_NAME` to "darwin" or something.

**Expected behavior**
The bootstrap script should call `uname` or similar to auto-detect which operating system we are truly running on.

**Please complete the following information):**
 - OS: [e.g. iOS] All
 - Software Versions [e.g. 22] N/A

**Additional context**
This same issue applied to the Build System Docker Images, as well; see https://github.com/vegastrike/build-system-docker-images/issues/89

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Addresses #1443 on `master`

Purpose:
- What is this pull request trying to do? Detect properly which operating system we are on; fix #1443 on the `master` branch

### DIFF
--- a/script/bootstrap-on-linux.sh
+++ b/script/bootstrap-on-linux.sh
@@ -30,7 +30,7 @@
 set -e
 
 echo "------------------------------------------"
-echo "--- bootstrap-on-linux.sh | 2025-11-27 ---"
+echo "--- bootstrap-on-linux.sh | 2025-12-01 ---"
 echo "------------------------------------------"
 
 UPDATE_ALL_SYSTEM_PACKAGES="$1"
@@ -101,14 +101,9 @@ function bootstrapOnDebian()
                             make \
                             pkg-config \
                             ninja-build \
-                            gnome-desktop-testing \
-                            libasound2-dev \
-                            libpulse-dev \
-                            libaudio-dev libfribidi-dev libjack-dev libsndio-dev libx11-dev libxext-dev \
-                            libxrandr-dev libxcursor-dev libxfixes-dev libxi-dev libxss-dev libxtst-dev \
-                            libxkbcommon-dev libdrm-dev libgbm-dev libgl1-mesa-dev libgles2-mesa-dev \
-                            libegl1-mesa-dev libdbus-1-dev libibus-1.0-dev libudev-dev \
-                            libpipewire-0.3-dev libwayland-dev libdecor-0-dev liburing-dev
+                            libaudio-dev \
+                            libfribidi-dev \
+                            libwayland-dev
             ;;
         "bookworm")
             echo "Bookworm does NOT support SDL3"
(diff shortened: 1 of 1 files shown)


script/bootstrap
```diff
@@ -30,1133 +30,40 @@
...
 
 echo "------------------------------"
-echo "--- bootstrap | 2025-11-16 ---"
+echo "--- bootstrap | 2025-11-27 ---"
 echo "------------------------------"
 
-UPDATE_ALL_SYSTEM_PACKAGES="$1"
+platform="$(uname | tr '[:upper:]' '[:lower:]')"
 
-if [ -f /etc/os-release ]
```
stephengtuggy, line 38: Linux-specific logic moved to bootstrap-on-linux.sh

script/bootstrap-on-linux.sh
```diff
@@ -0,0 +1,1162 @@
...
+else
+    echo "os-release file not found; unable to continue"
+    exit 1
+fi
+LINUX_ID=$(grep ^ID= $OS_RELEASE_LOCATION | sed 's/^ID=//' | tr -d '"\n')
+echo "LINUX_ID = ${LINUX_ID}"
+LINUX_CODENAME=$(grep ^VERSION_CODENAME= $OS_RELEASE_LOCATION | sed 's/^VERSION_CODENAME=//' | tr -d '"\n')
+echo "LINUX_CODENAME = ${LINUX_CODENAME}"
+LINUX_VERSION_ID=$(grep ^VERSION_ID= $OS_RELEASE_LOCATION | sed 's/^VERSION_ID=//' | tr -d '"\n')
+echo "LINUX_VERSION_ID = ${LINUX_VERSION_ID}"
```
BenjamenMeyer, line 53: FYI- there's a tool `lsb_release` that technically is supposed to report all the info in `/etc/lsb-release`/`/etc/os-release` which should provide all this information. I didn't use it before because I've had little success with figuring it out myself, but it's probably a better mechanism than parsing the files ourselves. $0.02 in case you wanted to pursue it, but not blocking...

Then again, reading through https://www.freedesktop.org/software/systemd/man/latest/os-release.html and https://0pointer.de/blog/projects/os-release this seems to be replacing the LSB utilities, and the examples on freedesktop.org show it being just sourced in:
```
Example 3. Reading os-release in [sh(1)](https://man7.org/linux/man-pages/man1/sh.1.html)

#!/bin/sh -eu
# SPDX-License-Identifier: MIT-0

test -e /etc/os-release && os_release='/etc/os-release' || os_release='/usr/lib/os-release'
. "${os_release}"

echo "Running on ${PRETTY_NAME:-Linux}"

if [ "${ID:-linux}" = "debian" ] || [ "${ID_LIKE#*debian*}" != "${ID_LIKE}" ]; then
    echo "Looks like Debian!"
fi
```
so we might just do that instead of trying to parse them out of the file.

BenjamenMeyer (approved):

PR #1458, opened 2025-11-30
Detect whether on Linux or macOS using `uname` -- 0.9.x, SDL3 variant

### LINKED ISSUE #1443: We do not currently have real OS detection logic for Linux vs. macOS

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Addresses #1443 on 0.9.x_sdl3

Purpose:
- What is this pull request trying to do? Fix #1443 on 0.9.x_sdl3
- What release is this for? 0.9.x, SDL3 variant

### DIFF
--- a/script/bootstrap-on-linux.sh
+++ b/script/bootstrap-on-linux.sh
@@ -30,7 +30,7 @@
 set -e
 
 echo "------------------------------------------"
-echo "--- bootstrap-on-linux.sh | 2025-11-29 ---"
+echo "--- bootstrap-on-linux.sh | 2025-12-01 ---"
 echo "------------------------------------------"
 
 UPDATE_ALL_SYSTEM_PACKAGES="$1"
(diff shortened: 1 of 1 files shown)


script/bootstrap
```diff
@@ -20,2066 +21,49 @@
...
-                      libboost-log-dev \
-                      libboost-regex-dev \
-                      libboost-program-options-dev \
-                      libxmu-dev \
-                      clang \
-                      lsb-release
-      ;;
+platform="$(uname | tr '[:upper:]' '[:lower:]')"
+
+case "$platform" in
```
royfalk, line 38: I like this. Now we're just missing bootstrap-on-windows.sh (bat?!).

script/bootstrap-on-linux.sh
```diff
@@ -0,0 +1,2469 @@
...
+                          rpm-build \
+                          make \
+                          clang
+      ;;
+    "9.0")
+      dnf -y install dnf-plugins-core
+      dnf config-manager --set-enabled devel
+      dnf -y update
+      dnf -y install \
+                          git \
```
royfalk: This is a 2500 lines file. Most of what's here is packages.
I wonder if we can somehow shorten it or make it more maintainable. 
I can easily see us missing something on the next version of Linux mint.

royfalk (commented):
My two cents:

script/bootstrap
BenjamenMeyer: Windows uses a very different mechanism with the vcpkg.json tooling.

script/bootstrap-on-linux.sh
BenjamenMeyer: I tried that at the beginning of things, and we've starting shortening it a little (see the 10/10.1 version).
I do agree it'd be nice if we could do a simple list for each type of installer, but different distros may end up with some slightly different requirements. Theoretically, Debian/Ubuntu/Mint should all have the same dependencies since Mint and Ubuntu more or less repackage Debian; but practically there are some variances, and often this happens between the releases too.

Perhaps we could compare and only split out when there is an actual differences between versions at a distro level; but that probably won't save much - RPM-based distros have the most differences IIRC.

script/bootstrap
stephengtuggy: I think you're talking about bootstrap.ps1. 😄

BenjamenMeyer: @stephengtuggy yes, which only works in PowerShell. Sadly this is one area where Windows does things so differently that its not very compatible, and doesn't have a standard Unix shell utility by default.

Issue #1248, opened 2025-08-03
Use C++ OpenTelemetry library?


stephengtuggy:
Seems like that would introduce a lot of unnecessary boilerplate code. And the tracking class would know too much about the innards of the class it was tracking.

Another possibility I have considered is to use the [C++ OpenTelemetry library](https://opentelemetry.io/docs/languages/cpp/) to measure the performance of the code. It would replace all these `#ifdef`s, and would do so in an industry-standard way that is suited for just this type of job. What do you think of that option?

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1232#discussion_r2249502189_

BenjamenMeyer:
Question: can we make this work without requiring additional infrastructure like Grafana? If not, then I don't think it'd work as that would make for a lot more dependencies. Certainly open to exploring it.

stephengtuggy:
> Question: can we make this work without requiring additional infrastructure like Grafana? If not, then I don't think it'd work as that would make for a lot more dependencies. Certainly open to exploring it.

I think we can. I'm thinking about implementing a [custom Exporter](https://opentelemetry-cpp.readthedocs.io/en/latest/otel_docs/classopentelemetry_1_1sdk_1_1trace_1_1SpanExporter.html) that records metrics to a local SQLite database. Either that, or use the standard [Exporter to export in OTLP native format](https://opentelemetry.io/docs/languages/cpp/exporters/#otlp), and then we can put any one of a number of receivers on the other end of that as we like.

stephengtuggy:
It occurs to me, though, that some users might be skittish about the possibility of telemetry being added to the game. Perhaps it's best to keep the telemetry local-only. If we add it at all, that is.

BenjamenMeyer:
2 things on this

1. yes it must stay local and entirely within the user's control
2. we should have some tooling on top of it that users can run themselves, see the information, and generate reports they could file to us *if* they chose to.
3. it should be turned off by default and the user must actively turn it on to enable

OpenTelemetry has a lot of neat things; and our historic user base has been very sensitive to privacy, etc. I know it's a fine line to walk, though I'm sure we can do it properly if we respect the above 3 points.

BenjamenMeyer:
> > Question: can we make this work without requiring additional infrastructure like Grafana? If not, then I don't think it'd work as that would make for a lot more dependencies. Certainly open to exploring it.
> 
[...]

PR #1459, opened 2025-12-01
More on OpenTelemetry-Cpp

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- #1248 

Purpose:
- What is this pull request trying to do? Give OpenTelemetry-Cpp another try
- What release is this for? 0.10.x, SDL3 variant
- Is there a project or milestone we should apply this to? 0.10.x, I suppose

### DIFF
--- a/engine/src/configuration/tests/configuration_tests.cpp
+++ b/engine/src/configuration/tests/configuration_tests.cpp
@@ -47,7 +47,7 @@ constexpr int kIterations = 100000;
 
 TEST(QVector, Cast_Performance) {
     VS_LOG_AND_FLUSH(important_info, "Starting QVector Cast() performance test");
-    const auto start_time = std::chrono::steady_clock::now();
+    const double start_time = realTime();
     for (int i = 0; i < kIterations; ++i) {
         const QVector q_vector(randomDouble(), randomDouble(), randomDouble());
         // ReSharper disable once CppDFAUnreadVariable
@@ -55,15 +55,14 @@ TEST(QVector, Cast_Performance) {
         // ReSharper disable once CppDFAUnusedValue
         const Vector vector = q_vector.Cast();
     }
-    const auto end_time = std::chrono::steady_clock::now();
-    const auto duration = std::chrono::duration<double, std::micro>(end_time - start_time);
-    const auto duration_in_seconds = std::chrono::duration_cast<std::chrono::seconds>(end_time - start_time);
-    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% microseconds (%2% whole seconds) for %3% iterations") % duration.count() % duration_in_seconds.count() % kIterations));
+    const double end_time = realTime();
+    const double duration = end_time - start_time;
+    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% second(s) for %2% iterations") % duration % kIterations));
 }
 
 TEST(Vector, Cast_Performance) {
     VS_LOG_AND_FLUSH(important_info, "Starting Vector Cast() performance test");
-    const auto start_time = std::chrono::steady_clock::now();
+    const double start_time = realTime();
     for (int i = 0; i < kIterations; ++i) {
         const Vector vector(randomDouble(), randomDouble(), randomDouble());
         // ReSharper disable once CppDeclaratorNeverUsed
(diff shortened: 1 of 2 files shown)


engine/src/vs_logging.cpp
```diff
@@ -46,6 +45,103 @@
...
+    return provider->GetLogger(__FILE__);
+}
+
+}  // namespace
+
+#endif
+
 namespace VegaStrikeLogging {
 
 void VegaStrikeLogger::InitLoggingPart2(const uint8_t debug_level,
```
royfalk, line 147: Not a big fan of init2. Is there anything we can do here?
Also, I don't see where this is called from. This may be a limitation of the GitHub web interface.

Yes, I know I've done it as well.
I can only plead that I was a very poor C++ programmer back then.

stephengtuggy: This "init2" method has been around for ages -- several years at this point. It gets called from `main` once `main` has collected enough information to know what file to log to and with what level of detail. Mostly the level of detail.

It could perhaps be named differently. I'm open to suggestions.

stephengtuggy:
> A couple of comments:
> 
>     1. It would help to know why we're doing this. A brief text and a link to the discussion would remind anyone not up to date.

Point taken. I wanted to first see if the code would even work though. So far, it hasn't.

It hasn't been decided yet what OpenTelemetry will replace, if anything. It does have logging functionality, so it might replace -- or provide an alternate implementation for -- our logging. It might also replace the code that measures how long certain things take and then logs that. That was its original intended purpose.

>     2. How is this better than boost logs? Both write to files in our configuration.

It looks like, if we use OpenTelemetry logging, it can automatically rotate files; keep a link to the latest one; and delete old files to avoid filling up drives. Boost Logging doesn't have most of those features.
 
>     3. I don't have a lot to contribute in the non-code part of the PR. I don't really understand any of it.
> 
>     4. I don't have a lot to contribute in the code part of the PR but that's because it looks good.
> 
>     5. I see this is a draft, so I may have reviewed prematurely. However, I try to go over notifications, so newer gets reviewed first. This has some drawbacks, so I apologise if I hit another PR in a few minutes that answers a lot of these.

stephengtuggy:
This PR has gotten absolutely gigantic. And that's not just in the number of lines changed. Introducing OpenTelemetry-Cpp into our codebase would require several large, sweeping changes. Among other things, it uses vcpkg on all supported platforms, pretty much forcing us to use vcpkg everywhere too. (If you don't provide your own copy of vcpkg, it uses its own, internal, outdated one instead, apparently.)

Some of the changes are changes we probably need to make anyway. Like switching to using imported targets wherever possible in CMake. But making those changes all at once like this is not likely to work very well.

engine/src/vs_logging.cpp
stephengtuggy: Perhaps `configure_logging_verbosity` or something?

stephengtuggy:
So many CMake presets! The number of presets for the project is bonkers. Every time we add a new option (or at least, most of the time) I feel the need to add presets with the different settings for it. That ends up multiplying the total number of presets by 2 or 3. Sometimes more. We are up to over 300 now!

Surely there's a better way to do this, but at the moment, I'm not sure what it is

I did, at least, find and download a Python script that helps me generate and manage all the different permutations of options.

BenjamenMeyer:
> 1. It would help to know why we're doing this. A brief text and a link to the discussion would remind anyone not up to date.

The primary thing OpenTelemetry is for, IMHO, is for metrics and insights into the code from performance perspectives.
It gets driven by a lot of logging in that it records its information to logs that other tooling like Graphana and Prometheus utilize. It's pretty standardized and there's lots of tools around it so _if_ we can use it it may make sense since the alternative would be to build a lot of that stuff ourselves even if we have to piecemeal different libraries together to collect the information or generate the reports.

stephengtuggy:
> > 1. It would help to know why we're doing this. A brief text and a link to the discussion would remind anyone not up to date.
> 
> The primary thing OpenTelemetry is for, IMHO, is for metrics and insights into the code from performance perspectives. It gets driven by a lot of logging in that it records its information to logs that other tooling like Graphana and Prometheus utilize. It's pretty standardized and there's lots of tools around it so _if_ we can use it it may make sense since the alternative would be to build a lot of that stuff ourselves even if we have to piecemeal different libraries together to collect the information or generate the reports.

@BenjamenMeyer Agreed.

stephengtuggy:
Closing since the target branch (`master-sdl3`) is obsolete, and this PR would need to be reworked to target `master`.

PR #1464, opened 2025-12-04
lin_time fixes

stephengtuggy:

Please answer the following:

Code Changes:
- [x] (Largely) Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Make some changes to the `lin_time.cpp` function implementations to make them more technically correct. Like adding runtime checks that, according to the documentation, they should have.
- What release is this for? 0.10.x, SDL3 variant

### DIFF
--- a/engine/src/configuration/tests/configuration_tests.cpp
+++ b/engine/src/configuration/tests/configuration_tests.cpp
@@ -57,7 +57,7 @@ TEST(QVector, Cast_Performance) {
     }
     const double end_time = realTime();
     const double duration = end_time - start_time;
-    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% for %2% iterations") % duration % kIterations));
+    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% second(s) for %2% iterations") % duration % kIterations));
 }
 
 TEST(Vector, Cast_Performance) {
@@ -72,7 +72,7 @@ TEST(Vector, Cast_Performance) {
     }
     const double end_time = realTime();
     const double duration = end_time - start_time;
-    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished Vector Cast() performance test. Took %1% for %2% iterations") % duration % kIterations));
+    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished Vector Cast() performance test. Took %1% second(s) for %2% iterations") % duration % kIterations));
 }
 
 class B {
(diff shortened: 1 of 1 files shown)


stephengtuggy:
Note that I am using the updated `realTime()` to time several performance tests, and it is returning far more accurate results than what I was getting yesterday.

engine/src/configuration/tests/configuration_tests.cpp
```diff
@@ -47,34 +47,32 @@ constexpr int kIterations = 100000;
...
         // ReSharper disable once CppDFAUnusedValue
         const Vector vector = q_vector.Cast();
     }
-    const auto end_time = std::chrono::steady_clock::now();
-    const auto duration = std::chrono::duration<double, std::micro>(end_time - start_time);
-    const auto duration_in_seconds = std::chrono::duration_cast<std::chrono::seconds>(end_time - start_time);
-    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% microseconds (%2% whole seconds) for %3% iterations") % duration.count() % duration_in_seconds.count() % kIterations));
+    const double end_time = realTime();
+    const double duration = end_time - start_time;
+    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% second(s) for %2% iterations") % duration % kIterations));
```
BenjamenMeyer, line 60: Might this be better to have as a type:
```
class VegaStrikeTimer {
public:
    void start(); // to re-use, just call again
    void end();
    void log(const std::string& _prefix);
    double duration() const;
}
....

// most code would look like:
VegaStrikeTimer timer;
timer.start();
...
timer.end();
timer.log("Finished QVector Cast() performance test");
```
You could even add a parameter for the constructor to auto-start the timer and save another line in the usage code.

If we wanted to get really fancy then you could add a stack to it so multiple calls to `start` would add another layer on the stack that would have to be equally `end`'d to fully stop the timer; then you'd need a `reset()` to fully reset it.

Just a thought... not blocking right now on it.... though..IIRC this might have been one of the questions that led to looking at OpenTelemetry too.

stephengtuggy: Yes, this is largely the sort of thing that OpenTelemetry is supposed to take care of... if we end up going that route.

stephengtuggy:
@BenjamenMeyer are you going to approve? 😃

stephengtuggy:
Whoops, closed this by mistake.

PR #1465, opened 2025-12-05
Add `allow_null` parameter to `vega_dynamic_cast_ptr`

### LINKED ISSUE #1462: Add optional bool paramter `allow_null` to `vega_dynamic_cast_ptr`
@royfalk This parameter, and [the commit](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1416/commits/392589c20aeb3ffb770488522bcf0ffc145db65f) in which it was introduced, are new in this PR as opposed to the non-SDL3 version. (I plan to port this over there too.)

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1416#discussion_r2554328479_

stephengtuggy:
…and start using it in several places in basecomputer.cpp

Please answer the following:

Code Changes:

Issues:
- Closes #1462 

Purpose:
- What is this pull request trying to do? Port this set of changes from `master-sdl3` to `master`, as promised
- What release is this for? 0.10.x, non-SDL3 variant

### DIFF
--- a/engine/src/vega_cast_utils.h
+++ b/engine/src/vega_cast_utils.h
@@ -89,20 +89,17 @@ inline long locale_aware_stol(const std::string& s, const long default_value = 0
 }
 
 template<class TargetType, class SourceType>
-inline TargetType* vega_dynamic_cast_ptr(SourceType* from) {
+inline TargetType* vega_dynamic_cast_ptr(SourceType* from, const bool allow_null = false) {
     TargetType* ret_val = nullptr;
     try {
         ret_val = dynamic_cast<TargetType*>(from);
     } catch (std::bad_cast& e) {
         VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Error '%1%' casting type %2%* to %3%*") % e.what() % typeid(SourceType).name() % typeid(TargetType).name()));
         VSExit(-422);
-    } catch (std::exception& e) {
-        VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Error '%1%' casting type %2%* to %3%*") % e.what() % typeid(SourceType).name() % typeid(TargetType).name()));
-        VSExit(-423);
     }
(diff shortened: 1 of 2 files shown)


engine/src/vega_cast_utils.h
```diff
@@ -89,20 +89,17 @@ inline long locale_aware_stol(const std::string& s, const long default_value = 0
...
         VSExit(-422);
-    } catch (std::exception& e) {
-        VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Error '%1%' casting type %2%* to %3%*") % e.what() % typeid(SourceType).name() % typeid(TargetType).name()));
-        VSExit(-423);
     }
-    if (ret_val == nullptr) {
+    if (ret_val == nullptr && !allow_null) {
         VS_LOG_AND_FLUSH(fatal, (boost::format("Fatal Failure to Cast type %1%* to %2%* -- nullptr encountered") % typeid(SourceType).name() % typeid(TargetType).name()));
-        VSExit(-422);
+        VSExit(-424);
```
BenjamenMeyer, line 102: not blocking on this one, but we should probably come up with a mechanism for either:
- generating the exit codes (bit mapping to reflect multiple things)
- use an locked enum for the exit codes so we can easily maintain the list of exit codes and know which ones are used where by name instead of by number, making this kind of change easier since it'd be in one place.

stephengtuggy: In point of fact, we should probably use the SDL3 return codes, which have only two possibilities: success, or failure (error). That would do away with all these custom exit codes entirely.

BenjamenMeyer: we have many other reason than SDL issues for why we terminated. Our exit codes should be our own.

stephengtuggy: Oh, maybe so

PR #1466, opened 2025-12-06
lin_time fixes -- `master`

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- N/A

Purpose:
- What is this pull request trying to do? Port #1464 to `master`
- What release is this for? 0.10.x, non-SDL3 variant

### DIFF
--- a/engine/src/configuration/tests/configuration_tests.cpp
+++ b/engine/src/configuration/tests/configuration_tests.cpp
@@ -57,7 +57,7 @@ TEST(QVector, Cast_Performance) {
     }
     const double end_time = realTime();
     const double duration = end_time - start_time;
-    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% for %2% iterations") % duration % kIterations));
+    VS_LOG_AND_FLUSH(important_info, (boost::format("Finished QVector Cast() performance test. Took %1% second(s) for %2% iterations") % duration % kIterations));
 }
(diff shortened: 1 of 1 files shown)


stephengtuggy:
Based on the [online documentation I've been reading](https://stackoverflow.com/a/50245503/5067822), different platforms have different levels of support for `CLOCK_MONOTONIC`, `CLOCK_REALTIME`, etc. Sometimes you even have to perform a runtime check to switch between them.

We could perhaps just use `SDL_GetTicks` everywhere. If we can, that would be great. Perhaps I'll try that. Otherwise, we'll probably need to keep this admittedly ugly code for the foreseeable future.

stephengtuggy:
I could see getting rid of time compression as well. I will take a look at that.

stephengtuggy:
`realTime()` has been around for quite a while. It is certainly ugly, but it gives much more accurate results when timing how long a particular section of code takes. That's why I switched back to using that instead of calling `std::chrono` directly.

In the near future, we might replace all of this with OpenTelemetry stuff. At least when users opt into it. We shall see.

stephengtuggy:
Note that at present, these changes are included in #1474 , as they were on the `master-sdl3` branch. Is that okay, or should I take them out?

stephengtuggy:
Closing. Will want to revisit this for both 0.10.x and 0.11.x, hopefully simplifying the implementation down to one version/branch for each function.

Issue #1467, opened 2025-12-08
Game crash on base computer (master)


royfalk:
```
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
 File is a directory ! 
 File is a directory ! 
 File is a directory ! 
 File is a directory ! 
Fatal Failure to Cast type 7Control* to 13StaticDisplay* -- nullptr encountered
```

royfalk:
`Fatal Failure to Cast type 7Control* to 13StaticDisplay* -- nullptr encountered`

This means no core dump! We need to either print the offending caller filename and line number or crash here.

royfalk:
Crash is here:
```
//Set the string in the base title control.
StaticDisplay *baseTitleDisplay = vega_dynamic_cast_ptr<StaticDisplay>(window()->findControlById("baseTitle"));
```
findControlById returns null.

stephengtuggy:
> `Fatal Failure to Cast type 7Control* to 13StaticDisplay* -- nullptr encountered`
> 
> This means no core dump! We need to either print the offending caller filename and line number or crash here.

Good point. I still think it's best if we catch this type of error specifically. But we probably do need to output the offending caller in some format or other. Or a complete call stack, perhaps.

PR #1468, opened 2025-12-08
Fix bug #1467

royfalk:
Fixes a crash in base computer.

Fixes: #1467 

Code Changes:
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. No. Only check this fixes the error in question.

Issues:
- The current checks in vega_dynamic_cast_ptr don't generate core dumps.
- Base computer is filled with hard coded values that make it easy to generate such a bug.


stephengtuggy:
With this fix in place, does the "Multiplayer" main menu item crash, or not?

PR #1470, opened 2025-12-08
Task fix 1363

### LINKED ISSUE #1363: Hide mouse cursor in flight
With the changes made to mouse code, the mouse cursor in flight (green crosshairs) is now shown always. This is not desirable. 

Relevant code is in gldrv/mousecursor.h/cpp.
When docking and undocking, the game changes the cursor by calling changeCursor.

Required:

- Add placeholder bool show_mouse_cursor_in_flight or some such in configuration/configuration.h
- In the changeCursor function, add show and hide based on configuration.
- Optionally, check if cursor is visible before calling show/hide.
- Optionally, look at code derived from searching for crosshairs.spr. It's not supposed to be called ans should be deleted.

In assets, we will need to modify the config.json and then rebuild the configuration files. I'll loop in @BenjamenMeyer and @stephengtuggy to discuss how to best do this.

This is not meant to be an exhaustive list but rather pointers to what needs to be done.

royfalk:
Fix #1363 

**Code Changes:**
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. _No. This is a first stab at a fix._

Issues:
- config.json can disable mouseShow but actual mouse in flight is controlled by vegastrike.config.

### DIFF
--- a/libraries/cmd/unit_generic.cpp
+++ b/libraries/cmd/unit_generic.cpp
@@ -2006,7 +2006,11 @@ int Unit::ForceDock(Unit *utdw, unsigned int whichdockport) {
 
     // Change mouse pointer to arrow
     if(IsPlayerShip()) {
-        changeCursor(CursorType::arrow);
+        if(configuration().mouse.enabled) {
+            changeCursor(CursorType::arrow);
+        } else {
+            showCursor();
+        }
     }
 
     return whichdockport + 1;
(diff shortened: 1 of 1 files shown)


royfalk:
Great. Given that actual mouse is still defined by the old config and this only takes the visibility from mouse.enabled, I'd call that a success.

kheckwrecker:
In that case, looks good to me. Thanks for this PR.

Issue #1347, opened 2025-09-30
Base text highlighting not quite right


evertvorster:
The highlighting of clickable areas in the base is still not quite right. 
This may be the last remaining scaling issue. 

This is on my secondary monitor, but this issue might be present on the main monitor too. 

Best description is probably in a video:
https://youtu.be/b_wvwLFE3gE

kheckwrecker:
I can confirm it is present on the main monitor too.

stephengtuggy:
As part of #1474 , I fixed an apparent bug in the text-display code, where text snippets that were marked as align-left or align-right were displayed centered instead.

I'm thinking that broke some display items that were wrongly marked as align-left or align-right, but that never mattered before, but now it does. Maybe I should revert that code change for now.

royfalk:
@stephengtuggy I went over #1474 quickly and I wasn't able to find your fix. Can you point me to the correct commit or file/line?
Also, this is also in master. Someone would need to go back and check if this wasn't always the case.

stephengtuggy:
Sorry, correction. I "fixed" the text alignment in commit fae7ad032dafc79c711298459c6b70997cdea057 on 2025-11-17, which made it into master in a previous PR. I haven't reverted it yet. That will need to be done shortly.

stephengtuggy:
I think there may actually be two separate issues here. One is the **highlighting** -- the original issue that @evertvorster reported in September. The other is the text **alignment** -- the newer issue that was introduced in November by my attempted fix. The two issues have a similar effect, of making the base text difficult to read, but they have different causes.

PR #1474, opened 2025-12-13
Mainstream SDL3

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- #1369 

Purpose:
- What is this pull request trying to do? Merge the SDL3 changes into master

### DIFF
--- a/libraries/gui/pause_screen.cpp
+++ b/libraries/gui/pause_screen.cpp
@@ -53,7 +53,7 @@ void pauseGame() {
         while (SDL_PollEvent(&event)) {
             ImGui_ImplSDL3_ProcessEvent(&event);
             
-            if (event.type == SDL_KEYDOWN && event.key.repeat == 0) {
+            if (event.type == SDL_EVENT_KEY_DOWN && event.key.repeat == 0) {
                 done = true;
             }
         }


libraries/imgui/backends/imgui_impl_sdlgpu3.cpp
stephengtuggy, line 1: This file accounts for ~20% of the lines added in this PR.

libraries/root_generic/lin_time.cpp
```diff
@@ -172,17 +172,20 @@ void InitTime() {
     }
     lasttime = dblnewtime - .0001;
 
-#elif defined (_POSIX_MONOTONIC_CLOCK)
-    struct timespec ts;
-    clock_gettime(CLOCK_MONOTONIC, &ts);
+    #elif defined (_POSIX_MONOTONIC_CLOCK)
```
stephengtuggy, line 175: The lin_time.cpp changes snuck into `master_sdl3` while remaining under review on the `master` branch. I would like to bring them in for now, and then make additional changes to these files in the near future.

engine/src/gldrv/winsys.cpp
```diff
@@ -201,35 +202,205 @@ void winsys_warp_pointer(int x, int y) {
...
+SDL_Rect total_display_bounds{};
+SDL_Rect usable_display_bounds{};
+SDL_Rect total_window_size_in_pixels{};
+SDL_Rect usable_window_size_in_pixels{};
+SDL_Rect total_logical_window_size{};
+SDL_Rect usable_logical_window_size{};
+
+/*---------------------------------------------------------------------------*/
+// pmx-20251026
+bool get_sdl_display_name_by_nr(int screen_number, std::string& screen_name, SDL_DisplayID& id) {
```
royfalk: Better make it get_sdl_display_name_by_number. It's long either way and best practice to no longer shorten names.

engine/src/gldrv/winsys.cpp
```diff
@@ -201,35 +202,205 @@ void winsys_warp_pointer(int x, int y) {
...
+        SDL_LogInfo(SDL_LOG_CATEGORY_APPLICATION, "%s", log_msg.c_str());
+    } else {
+        const std::string log_msg = (boost::format("%1%: SDL_GetDisplayBounds(...) encountered an error: %2%") % __FUNCTION__ % SDL_GetError()).str();
+        VS_LOG_AND_FLUSH(error, log_msg);
+        SDL_LogError(SDL_LOG_CATEGORY_APPLICATION, "%s", log_msg.c_str());
+        SDL_ClearError();
+        return false;
+    }
+
+    if (SDL_GetDisplayUsableBounds(instance_ID, &usable_display_bounds)) {
```
royfalk, line 335:
```
if (SDL_GetDisplayUsableBounds(instance_ID, &usable_display_bounds)) {
   const std::string log_msg = (boost::format("%1%: usable display bounds: x=%2%, y=%3%, w=%4%, h=%5%") % __FUNCTION__ % usable_display_bounds.x % usable_display_bounds.y % usable_display_bounds.w % usable_display_bounds.h).str();
   print_good(log_msg);
} else {
   print_bad(const std::string log_msg = (boost::format("%1%: SDL_GetWindowSizeInPixels(...) encountered an error: %2%") % __FUNCTION__ % SDL_GetError()).str();
}
```

Also, you don't need to recreate both good and bad:
```
const std::string suffix =  (boost::format("x=%2%, y=%3%, w=%4%, h=%5%")  % usable_display_bounds.x % usable_display_bounds.y % usable_display_bounds.w % usable_display_bounds.h).str();
```
The __FUNCTION__ is identical for all and the rest of the prefix is identical for good and bad.

On a more serious note, boost::format follows the C pattern of variadic functions. These are not checked at compile time (no type safety) and are error-prone. This is according to Stroustrup. He kept them for backward compatibility so as not to rock the boat too much. 

If you got one of the error boost::format wrong, this will not issue an error but will crash the game.

Better to:
```
boost::format("%1% Attempting to run SDL_GetDisplayUsableBounds(instance_ID, (%2%, %3%, %4%, %5%) ...
if(SDL_GetDisplayUsableBounds(instance_ID, &usable_display_bounds) {
   good
} else {
   bad
}
```
```

File: engine/src/gldrv/winsys.cpp [hunk 4]
```diff
@@ -201,35 +202,205 @@ void winsys_warp_pointer(int x, int y) {
...
-        } else {
-            native_resolution_x = currentDisplayMode.w;
-            native_resolution_y = currentDisplayMode.h;
-        }
-    } else {
-        video_flags |= SDL_WINDOW_RESIZABLE;
-
-        native_resolution_x = configuration().graphics.resolution_x;
-        native_resolution_y = configuration().graphics.resolution_y;
+static bool setup_sdl_video_mode(int* argc, char** argv) {
```
royfalk: Unless one of us developers is using this, the whole argc/argv here can go.

engine/src/gldrv/winsys.cpp
```diff
@@ -201,35 +202,205 @@ void winsys_warp_pointer(int x, int y) {
...
+    int screen_number = 0;
+    Uint32 video_flags = 0;
+    int bpp = 0;
+    SDL_Rect display_bounds;
+    std:string screen_name = "";
+    SDL_DisplayID instance_ID;
+    SDL_DisplayMode* mode_for_ID = static_cast<SDL_DisplayMode*>(std::calloc(sizeof(SDL_DisplayMode), 1));
+    if (mode_for_ID == nullptr) {
+        VS_LOG_FLUSH_EXIT(fatal, "Memory allocation error", 1);
```
royfalk: Either give me the file and line number or just crash the game and give me a stack. Right now we have a single "Memory allocation error" but I can totally see us or some other poor bastard trying to pick between several.

engine/src/gldrv/winsys.cpp
```diff
@@ -265,136 +436,136 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
+                SDL_GetDisplayBounds(instance_ID, &display_bounds);
+                if ((mode->w == width) && (mode->h == height)) {
+                    found = true;
+                    std::memcpy(mode_for_ID, mode, sizeof(SDL_DisplayMode)); // pmx-20251026 I'm not sure of the life length of the data pointed
+                    // by 'mode', for the time being, I prefer to copy. May recheck later.
+                    break;
+                }
+            }
+        }
+        if (found == false) {
```
royfalk, line 466: I don't agree with this "rabbit hole". Consider the following scenario:
1. I have a monitor with 1400x1050 (non-standard) and not in the drop-down menu.
2. I manually edit and get it wrong - 1040x1500.
3. The function picks SDL_GetClosestFullscreenDisplayMode 1200x800 (or whatever).
4. I don't know I got it wrong but the display looks a bit off. 
5. I open a bug.

engine/src/gldrv/winsys.cpp
```diff
@@ -265,136 +436,136 @@ static bool setup_sdl_video_mode(int *argc, char **argv) {
...
+        VS_LOG_AND_FLUSH(error, (boost::format("%1%: SDL_SetWindowSize(...) encountered an error: %2%") % __FUNCTION__ % SDL_GetError()));
+        SDL_ClearError();
     }
 
-    if (SDL_SetHint(SDL_HINT_RENDER_DRIVER, "opengl")) {
-        VS_LOG_AND_FLUSH(important_info, "SDL_SetHint(SDL_HINT_RENDER_DRIVER, ...) succeeded");
-    } else {
-        VS_LOG_AND_FLUSH(error, (boost::format("SDL_SetHint(SDL_HINT_RENDER_DRIVER, ...) failed. Error: %1%") % SDL_GetError()));
+    if (!SDL_GetCurrentRenderOutputSize(renderer, &native_resolution_x, &native_resolution_y)) {
+        VS_LOG_AND_FLUSH(error, (boost::format("%1%: SDL_GetCurrentRenderOutputSize(...) encountered an error: %2%") % __FUNCTION__ % SDL_GetError()));
```
royfalk: At a minimum, we should wrap these VS_LOG_AND_FLUSH(SDL_GetError) and SDL_ClearError in a function.

engine/src/gldrv/winsys.cpp
```diff
@@ -489,11 +654,16 @@ void winsys_shutdown() {
  *  \author  jfpatry
  *  \date    Created:  2000-10-19
  *  \date    Modified: 2000-10-19
+ *  \date    Modified: 2025-11-14 - stephengtuggy in conjunction with pmxy
```
royfalk: The original author's name is listed at the top of the file. It's considered bad practice to keep a change log as part of the file in git.

engine/src/gldrv/winsys.cpp
```diff
@@ -633,16 +789,17 @@ void winsys_atexit(winsys_atexit_func_t func) {
  *  \author  jfpatry
  *  \date    Created:  2000-10-20
  *  \date    Modified: 2000-10-20
+ *  \date    Modified: 2025-11-14 - stephengtuggy in conjunction with pmxy
  */
 void winsys_exit(int code) {
-    // Reverting resolution by exiting fullscreen
-    SDL_SetWindowFullscreen(window, 0);
+    // // Reverting resolution by exiting fullscreen
+    // SDL_SetWindowFullscreen(window, false);
```
royfalk: So you commented it out and changed 0 to false? 
Why remove the functionality?

engine/src/gldrv/winsys.h
```diff
@@ -156,22 +156,23 @@ typedef enum {
     WSK_LMETA = SDLK_LGUI,
     WSK_BREAK = SDLK_PAUSE,
     WSK_PAUSE = SDLK_PAUSE,
-    WSK_LAST = SDL_NUM_SCANCODES // Could be an issue. Needs investigating. See https://wiki.libsdl.org/SDL2/MigrationGuide.
+    WSK_LAST = SDL_SCANCODE_COUNT // Could be an issue. Needs investigating. See https://wiki.libsdl.org/SDL2/MigrationGuide.
```
royfalk, line 159: We can delete this comment. If we have a user with a non-QWERTY keyboard and he had an issue, we would have heard about it. In any case, such a user would have probably remapped the keys anyway (or maybe not).

engine/src/in_joystick.cpp
```diff
@@ -175,58 +179,69 @@ void InitJoystick() {
             }
         }
     }
+    num_joysticks = 0;
+
 #ifndef NO_SDL_JOYSTICK
 #ifdef HAVE_SDL
```
royfalk: Keep only this.

engine/src/in_joystick.cpp
```diff
@@ -175,58 +179,69 @@ void InitJoystick() {
...
+                VS_LOG(important_info, (boost::format("    %1%\n") % SDL_GetJoystickNameForID(instance_id)));
+                joystick[i1] = new JoyStick(i1, instance_id);
+            } else {
+                joystick[i1] = new JoyStick(i1, 0);
+            }
+        }
+    }
+    SDL_free(joysticks);
+
 #else
```
royfalk: Remove this. We no longer have an option of running the game without SDL. We pick a library and stick to it.

engine/src/in_joystick.cpp
```diff
@@ -175,58 +179,69 @@ void InitJoystick() {
...
+        num_joysticks = MAX_JOYSTICKS;
+    }
+    VS_LOG(important_info, (boost::format("%1% joysticks were found.\n\n") % num_joysticks));
+    VS_LOG(important_info, "The names of the joysticks are:\n");
+    if (joysticks) {
+        for (int i1 = 0; i1 < MAX_JOYSTICKS; ++i1) {
+            if (i1 < num_joysticks) {
+                const SDL_JoystickID instance_id = joysticks[i1];
+                VS_LOG(important_info, (boost::format("    %1%\n") % SDL_GetJoystickNameForID(instance_id)));
+                joystick[i1] = new JoyStick(i1, instance_id);
```
royfalk, line 196: We're breaking RAII here in several ways. Why not have joystick be an std::vector?

engine/src/in_joystick.cpp
```diff
@@ -175,58 +179,69 @@ void InitJoystick() {
...
-#endif
-#endif
-        joystick[i] = new JoyStick(i);         //SDL_Init is done in main.cpp
-    }
 }
 
 void DeInitJoystick() {
-    for (int i = 0; i < MAX_JOYSTICKS; i++) {
-        delete joystick[i];
+    for (auto & i : joystick) {
```
royfalk, line 208: Also break RAII here. you have to call DeInitJoystick explicitly or leak.

engine/src/in_joystick.cpp
```diff
@@ -311,40 +329,42 @@ void JoyStick::GetMouse(float &x, float &y, float &z, int &buttons) {
...
             joy_axis[a] = 0;
         }
         x = y = z = 0;
-        joy_buttons = buttons = 0;
+        joy_buttons = 0;
+        buttons = 0L;
         return;
     } else if (mouse) {
-        GetMouse(x, y, z, buttons);
+        int btn;
```
royfalk, line 329: This is wrong. We shouldn't have mouse code in the joystick file. Place a TODO here or open an issue please.

engine/src/in_sdl.cpp
```diff
@@ -110,10 +110,10 @@ void BindDigitalHatswitchKey(int joystick, int key, int dir, KBHandler handler,
 
 void ProcessJoystick(int whichplayer) {
     float x, y, z;
-    int buttons;
+    long long buttons;
 #ifdef HAVE_SDL
 #ifndef NO_SDL_JOYSTICK
-    SDL_JoystickUpdate();     //FIXME isn't this supposed to be called already by SDL?
+    // SDL_UpdateJoysticks();     //FIXME isn't this supposed to be called already by SDL?
```
royfalk, line 116: So we update the function to SDL3 and comment it out? If it's called elsewhere, joysticks would still work.
I'm of two minds here:
1. Comment it out and you'll forget to remove the code before merging.
2. Don't comment it out and someone says the joystick don't work (unless you have a joystick) and now you have to find out where the deleted line is.

engine/src/rendertext.cpp
```diff
@@ -52,8 +52,8 @@
 
 #ifdef HAVE_SDL
 
-static SDL_mutex *_rtextSDLMutex() {
```
royfalk, line 55: I'm betting this (`rendertext.*`) and `command.*` can be deleted. I checked and command.h is referenced from main.h, which isn't referenced at all. 

Probably not a good idea to do this as part of this PR.

libraries/CMakeLists.txt
```diff
@@ -29,8 +29,8 @@
 MESSAGE(STATUS "PROCESSING DIRECTORY: ${CMAKE_CURRENT_SOURCE_DIR}")
 
 #FILE(GLOB VS_LIBRARIES LIST_DIRECTORIES true RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} *)
-SET(VS_LIBRARIES "collide2" "gfx_generic" "root_generic" "cmd" "vegadisk" "posh" 
-    "imgui" "gui")
+SET(VS_LIBRARIES "collide2" "gfx_generic" "root_generic" "cmd" "vegadisk" "posh")
```
royfalk: Only comment out gui. If imgui is giving us trouble, we definitely need to know now.

libraries/imgui/backends/imgui_impl_sdlrenderer3.h
```diff
@@ -1,20 +1,21 @@
-// dear imgui: Renderer Backend for SDL_Renderer for SDL2
-// (Requires: SDL 2.0.17+)
```
royfalk, line 2: These are all library files. Not reviewing.

libraries/vegadisk/CMakeLists.txt
```diff
@@ -105,6 +105,8 @@ IF (USE_GTEST)
 
     TARGET_LINK_LIBRARIES(
             ${TEST_NAME}
+            GLUT::GLUT
+            ${TST_LIBS}
```
royfalk, line 109: vegadisk, as a matter of policy shouldn't have GLUT, OPENAL, Vorbis and JPEG in it. 
We should move these things up.

royfalk (changes requested):
It took me 60-90 minutes or so to go over this. Even without reviewing the imgui changes, there are several discrete PRs here all mixed together.

Setting aside the cosmetic changes, of particular concern are the changes to winsys and joystick. I know pmx migrated to SDL 3 in order to fix his joystick issues, but the joystick should have followed a smaller migration to SDL3 only without anything else. Same with the winsys refactor, which rewrites large parts.

Now, I don't know if the winsys rewrite is good or bad. I do know that inside a much larger PR, these changes don't get the attention they deserve.

Having said that, I don't envy you. Taking over someone else's work and making it work is a usually thankless work. So, thank you.

kheckwrecker:
> Secondly, when navigating stations to move to different areas (shipyards, trade centres, etc.), the text is out of proper alignment - not by much, but enough to be noticeable. The area to click is still the same, but the text doesn't always show up when your mouse hovers over that location.

Just noticed that this problem was already reported in issue #1347.

kheckwrecker:
Forgot to mention, when testing this PR, the base computer at Atlantis no longer displays a text description of the mission anymore, when you click on the mission name. First time I've run into this issue.

engine/src/gldrv/winsys.cpp
stephengtuggy: The `__FUNCTION__` is going to end up becoming part of the `VS_LOG` macros. So it will be part of every log message automatically. Consider its explicit presence here only temporary.

`boost::format`'s arguments are checked, I believe. At least, CLion gives me errors if I did something wrong with them. I think the compiler does too.

engine/src/gldrv/winsys.cpp
stephengtuggy: The file and line number are also going to become part of the VS_LOG macros, and therefore part of every log message. Eventually. Hopefully soon.

engine/src/gldrv/winsys.cpp
stephengtuggy: I guess that makes sense.

This is how @pmxy had it, but maybe I will change it back.

engine/src/gldrv/winsys.cpp
stephengtuggy: Even when one already exists, from prior to git?

engine/src/gldrv/winsys.cpp
stephengtuggy: Hmm. I forget why I did that. Guess I will uncomment it again.

engine/src/gldrv/winsys.h
stephengtuggy: Okay, but how about if we delete the comment later? It was here before this PR, and deleting it will make the PR larger.

engine/src/in_joystick.cpp
stephengtuggy: Makes sense. Will do.

engine/src/in_joystick.cpp
stephengtuggy: Good point

engine/src/in_joystick.cpp
stephengtuggy: Do you have an alternate suggestion? I don't see any way to avoid this.

engine/src/in_joystick.cpp
stephengtuggy: Filed as #1475

engine/src/in_sdl.cpp
stephengtuggy: I have no idea why this is commented out. Let me take a closer look

libraries/vegadisk/CMakeLists.txt
stephengtuggy: I am not sure whether vegadisk needs GLUT functionality. I'm not sure that is part of the scope of this PR to determine. If GLUT is needed in vegadisk, then it has to be here explicitly, based on the way things are going CMake-wise.

Also, it was this way in the code I'm imitating/porting over.

For all the above reasons, I would like to leave this as is for now, with a note to come back to it later.

stephengtuggy: Also, this isn't vegadisk. It's Vega_test. (I know. Not immediately obvious from the use of the TEST_NAME variable.)

stephengtuggy:
@kheckwrecker Thanks for testing, and for reporting back. I went ahead and filed 3 bug reports based on what you reported.

That's interesting that the frame rate has only now increased for you. It did for me as soon as I started running SDL3 code.

engine/src/gldrv/winsys.h
stephengtuggy: Filed as #1480

engine/src/rendertext.cpp
stephengtuggy: Filed as #1481

stephengtuggy:
Thanks for reviewing, @royfalk . I know this was a lot of changes, and you were very thorough. Some good feedback here. I have several items to fix.

stephengtuggy:
The winsys and joystick rewrites were necessary due to the changes between SDL2 and SDL3, unfortunately. I can probably clean up the joystick code a bit more though.

kheckwrecker:
> @kheckwrecker Thanks for testing, and for reporting back. I went ahead and filed 3 bug reports based on what you reported.
> 
> That's interesting that the frame rate has only now increased for you. It did for me as soon as I started running SDL3 code.

Thanks for filing the bug reports. Hopefully, they're easy fixes.

Don't know why it happened, but I'm enjoying the higher frame rates.

engine/src/in_joystick.cpp
stephengtuggy: Oh, I think I see. DeInitJoystick should just be the destructor, shouldn't it?

engine/src/in_joystick.cpp
BenjamenMeyer: std::deque is a bit more efficient and maintains the direct access

engine/src/in_joystick.cpp
BenjamenMeyer: yes, we need to solidify some things. SDL should be one of them.

engine/src/in_joystick.cpp
royfalk: Yep.

engine/src/in_joystick.cpp
royfalk: My comment wasn't about vector vs dequeue. It was about doing

```
Joystick j(blah blah);
joysticks.push_back(j);
```

instead of

```
Joystick* j = new Joystick(blah);
joysticks[i] = j;
```

But, since you've raised this, according to Google Gemini (I know)
Feature<span jsuid="Dqvcm_t" class="uJ19be notranslate" jsaction="rcuQ6b:&amp;Dqvcm_t|npT2md" jscontroller="udAs2b" data-wiz-uids="Dqvcm_t,Dqvcm_u,Dqvcm_v" data-processed="true"><span class="vKEkVd" data-animation-atomic="" data-wiz-attrbind="class=Dqvcm_t/TKHnVd;" data-processed="true"> </span></span>
  | std::vector | std::deque
-- | -- | --
Memory Storage | Contiguous in one large block | Non-contiguous, in a series of smaller chunks/blocks
push_front/pop_front | Inefficient (linear time O(n), as elements must be shifted) | Efficient (constant time O(1))
push_back/pop_back | Efficient (amortized constant time O(1)) | Efficient (constant time O(1))
Random Access ([]) | Very fast (single pointer dereference), excellent cache performance | Slower (requires two pointer dereferences and pointer arithmetic)

As such, vector is a better choice as we initialise joysticks once and query them multiple times.

libraries/vegadisk/CMakeLists.txt
royfalk: Sorry. I missed that. However, looking at line 73 I see:
```
TARGET_LINK_LIBRARIES(vegastrike_vegadisk PRIVATE ${TST_LIBS} GLUT::GLUT)
```

so the comment is still relevant to the library, just not the PR. The original devs did not encapsulate at all. I'm not convinced the file library needs to know about 3d for example.

royfalk:
> Third, the cost of basic repair and refuel on Atlantis is now at zero credits. It's always had a cost, and this is the first time I've noticed that it has dropped to zero.

This is on purpose. It turns out that the game auto refuels the ship upon landing. With the introduction of lib_component and the mapping of all components, if something is damaged, it's also listed and has a price to fix.

```
// TODO: remove function
// We no longer do repair through basic repair.
// Kept for compatibility with python API.
int Unit::RepairCost() {
    return 0;
}
```

> Finally, for some reason, I was able to buy ship upgrades that I should not have been able to buy. For example, I was able to purchase a second radar, when I already had one installed. Interestingly enough, when I saved my game and reloaded that saved game, the second radar had disappeared from my upgrades. Also Interesting, I purchased a second Capacitor, and a second SPEC capacitor, which is fine, and both were retained when loading the saved game, which is the correct expected behaviour.

Can confirm this is a regression. 

> Forgot to mention, when testing this PR, the base computer at Atlantis no longer displays a text description of the mission anymore, when you click on the mission name. First time I've run into this issue.

This probably has something to do with recent fixes/changes in basecomputer.cpp. I'm seeing several errors:
```
Adding news
finding nonloaded quest
quest_tutorial
finding quest
quest_tutorial
nonpfact
Processing News
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
controls_factory getControl(): font_array doesn't have enough elements
```

Please open bugs for these.

kheckwrecker:
Already done - PR #1477 and PR# 1478

engine/src/gldrv/winsys.cpp
stephengtuggy: Done

engine/src/gldrv/winsys.cpp
stephengtuggy: Done

engine/src/gldrv/winsys.cpp
stephengtuggy: Done. Added the file and line number to the output here specifically, for now.

engine/src/gldrv/winsys.cpp
stephengtuggy: Changed back to err out if the specified display resolution is not found.

engine/src/gldrv/winsys.cpp
stephengtuggy: I think you're right. I should stop tracking the individual edits to the functions in their headers, from the point where we took over maintaining the source code. Will delete these

engine/src/gldrv/winsys.cpp
stephengtuggy: I think I see what you're saying about making the code more DRY in respect to the logging boilerplate. I have taken a stab at that by creating macros `VS_LOG_SDL_SUCCESS` to log successful operations, `VS_LOG_SDL_ERROR` to log failed operations (which clears the error as well), and `VS_LOG_SDL_INFO` for standalone info messages. (They have to be macros for now so that I can pick up the correct values of `__FILE__`, `__LINE__`, and `__FUNCTION__`.)

Take a look once I post the latest commit, and let me know what you think.

engine/src/gldrv/winsys.cpp
stephengtuggy: Done. Uncommented

engine/src/in_joystick.cpp
stephengtuggy: Done

engine/src/in_joystick.cpp
stephengtuggy: Done. At least, this block of code is.

engine/src/in_joystick.cpp
stephengtuggy: We could use either vector or deque. I expect we will use deque quite a bit in general, but I think I agree with @royfalk that vector makes better sense in this particular case.

engine/src/gldrv/winsys.cpp
stephengtuggy: Done

engine/src/gldrv/winsys.cpp
stephengtuggy: Done

engine/src/in_sdl.cpp
stephengtuggy: Uncommented.

libraries/CMakeLists.txt
stephengtuggy: I have uncommented imgui.

libraries/vegadisk/CMakeLists.txt
stephengtuggy: Filed for later as #1489

engine/src/gldrv/winsys.cpp
stephengtuggy: OK, posted these changes.

engine/src/in_joystick.cpp
stephengtuggy: The joystick-related changes are turning into a deeper rabbit hole than I'm comfortable with pursuing, on top of all the other changes this PR already contains. There are several other in-memory data structures that we logically need to change from arrays to vectors, deques, or something else, if we are changing the data type of `joystick`. Can we put the joystick stuff on hold for now, and come back to it in a week or two? @royfalk @BenjamenMeyer

stephengtuggy:
Uh oh. The build is failing with imgui uncommented. But it builds just fine locally when I update imgui to the latest point release, 1.92.5. Looks like I will need to commit that version. I was hoping to avoid it, because the delta is so large, but it appears I have no other choice.

stephengtuggy:
There, that's better.

libraries/vegadisk/CMakeLists.txt
BenjamenMeyer: @royfalk agreed it shouldn't. It probably does due to linker issues with building the library or something. But yes, we should get to the point where only libraries that actually need to have things like GLUT/Vorbis/etc should be adding them and linking with them, but that's probably going to require quite a bit of header restructuring and library work to get there. Once we get there we'll have known module dependencies and a hierarchy that makes sense.

engine/src/in_joystick.cpp
stephengtuggy: Filed as #1494

libraries/vegadisk/CMakeLists.txt
royfalk: I hate kicking this dead horse, but I mean something different. I think libraries should have fairly limited scope if possible. This is an opposite philosophy to the original dev team. 
For lib component, this means not knowing about graphics, 3d graphics, sound, physics, etc. 
For vegadisk it could mean not knowing about specific file formats, which would dramatically reduce the following:
```
#include "vegadisk/vsfilesystem.h"

#include <cstdio>
#include <cassert>
#include <cstdarg>
#include <exception>
#include <configuration/configuration.h>
#if defined (_WIN32) && !defined (__CYGWIN__)
#include <Shlobj.h>
#include <direct.h>
#include <cstring>
#ifndef NOMINMAX
#define NOMINMAX
#endif //tells VCC not to generate min/max macros
#include <windows.h>
#include <cstdlib>
struct dirent
{
    char d_name[1];
};
#else
#include <unistd.h>
#include <pwd.h>
#include <sys/types.h>
#include <dirent.h>
#endif
#include <sys/stat.h>
#include "root_generic/configxml.h"
#include "root_generic/vs_globals.h"
#include "src/vegastrike.h"
#include "common/common.h"
#include "root_generic/galaxy_gen.h"
#include "vegadisk/pk3.h"

#include "src/gnuhash.h"

#include "root_generic/galaxy.h"

#include <boost/filesystem.hpp>

#include "configuration/game_config.h"
#include "src/vs_exit.h"

#include <string>
#include <utility>
```

royfalk (approved):

libraries/vegadisk/CMakeLists.txt
stephengtuggy: @royfalk I agree in general: our libraries should each be as limited in scope as reasonably possible. And that means not linking to the same list of dependencies (`$TST_LIBS`) everywhere. It probably also means not linking vegadisk to GLUT.

With regard to the headers you listed, I'm not sure how many of those could actually be eliminated by removing vegadisk's knowledge of specific file formats. I'm also not sure that knowledge of specific file formats should really be out of vegadisk's scope. But I agree with the other things you said here.

Thanks for the approval. Let's get this merged, and then we can see about cleaning up the extra dependencies in a follow-up PR. Hopefully very soon.

engine/CMakeLists.txt
```diff
@@ -977,55 +957,55 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
                         SET(BOOST_VER "${Boost_DERIVED_VERSION}")
                     ENDIF (Boost_DERIVED_VERSION STREQUAL "")
-                    SET(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libjpeg-turbo8, libpng16-16t64, freeglut3, libgtk-3-0, libvorbis0a, libopenal1, libsdl2-gfx-1.0-0, libsdl2-image-2.0-0, xdg-utils, libgl1, libopengl0, ${VEGA_STRIKE_GLU_DEPENDENCY}, libboost-atomic${BOOST_VER}, libboost-chrono${BOOST_VER}, libboost-date-time${BOOST_VER}, libboost-filesystem${BOOST_VER}, libboost-log${BOOST_VER}, libboost-python${BOOST_VER}, libboost-regex${BOOST_VER}, libboost-thread${BOOST_VER}, libboost-program-options${BOOST_VER}")
+                    SET(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libjpeg-turbo8, libpng16-16t64, freeglut3, libgtk-3-0, libvorbis0a, libopenal1, xdg-utils, libgl1, libopengl0, ${VEGA_STRIKE_GLU_DEPENDENCY}, libboost-atomic${BOOST_VER}, libboost-chrono${BOOST_VER}, libboost-date-time${BOOST_VER}, libboost-filesystem${BOOST_VER}, libboost-log${BOOST_VER}, libboost-python${BOOST_VER}, libboost-regex${BOOST_VER}, libboost-thread${BOOST_VER}, libboost-program-options${BOOST_VER}")
```
BenjamenMeyer: SDL3 should be in this list

engine/CMakeLists.txt
```diff
@@ -977,55 +957,55 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
```
BenjamenMeyer: And here

engine/CMakeLists.txt
```diff
@@ -977,55 +957,55 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
```
BenjamenMeyer: and here

engine/CMakeLists.txt
```diff
@@ -1087,49 +1067,49 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
         ELSEIF (LINUX_ID STREQUAL fedora)
             IF (LINUX_VERSION_ID VERSION_EQUAL 41)
                 SET(CPACK_RPM_PACKAGE_REQUIRES "python3, boost-python3")
-                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL2, SDL2_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
```
BenjamenMeyer: and here unless we're dropping support in which case this should go away entirely

engine/CMakeLists.txt
```diff
@@ -1087,49 +1067,49 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL3, SDL3_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
             ELSE ()
                 MESSAGE (WARNING "Unsupported version of Fedora")
                 SET(CPACK_RPM_PACKAGE_REQUIRES "")
             ENDIF ()
         ELSEIF (LINUX_ID STREQUAL rhel OR LINUX_ID STREQUAL redhat)
             IF (LINUX_VERSION_ID VERSION_EQUAL 9.5)
                 SET(CPACK_RPM_PACKAGE_REQUIRES "python3, boost-python3")
-                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL2, SDL2_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
```
BenjamenMeyer: and here

engine/CMakeLists.txt
```diff
@@ -1087,49 +1067,49 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
             ELSEIF (LINUX_VERSION_ID VERSION_EQUAL 9.6)
                 SET(CPACK_RPM_PACKAGE_REQUIRES "python3, boost-python3")
-                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL2, SDL2_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
```
BenjamenMeyer: and here

engine/CMakeLists.txt
```diff
@@ -1087,49 +1067,49 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL3, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
             ELSE ()
                 MESSAGE(WARNING "This version of Red Hat is not currently supported.")
                 SET(CPACK_RPM_PACKAGE_REQUIRES "")
             ENDIF ()
         ELSEIF (LINUX_ID STREQUAL rocky)
             IF (LINUX_VERSION_ID VERSION_EQUAL 9.5)
                 SET(CPACK_RPM_PACKAGE_REQUIRES "python3, boost-python3")
-                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL2, SDL2_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
```
BenjamenMeyer: and here

engine/CMakeLists.txt
```diff
@@ -1087,49 +1067,49 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
...
             ELSEIF (LINUX_VERSION_ID VERSION_EQUAL 9.6)
                 SET(CPACK_RPM_PACKAGE_REQUIRES "python3, boost-python3")
-                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, SDL2, SDL2_image, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
+                SET(CPACK_RPM_PACKAGE_REQUIRES "${CPACK_RPM_PACKAGE_REQUIRES}, libjpeg-turbo, libpng, freeglut, gtk3, libvorbis, openal-soft, libglvnd, expat, boost-log, boost-filesystem, boost-thread, boost-regex, boost-chrono, boost-atomic, boost-program-options")
```
BenjamenMeyer: and here

libraries/vegadisk/CMakeLists.txt
BenjamenMeyer: @royfalk agreed - yes the libraries should have limited scope and only know about their specific areas; they may depend on each other in some cases so they may have to know about each other to some degree, but that's about it.

As to vegadisk - it should be the controller of the file format. Outside of it it should just appear as a data loader/saver. That was my intent with vegadisk - in part being that vegadisk would handle a compressed file and the different elements would end up being files inside it (hence the `disk` in the name).

All that said, I don't think we can hold up this PR on this discussion.

engine/CMakeLists.txt
stephengtuggy: SDL3 is not available in Ubuntu Noble on which this Linux Mint version is based. We will have to provide it ourselves somehow.

engine/CMakeLists.txt
stephengtuggy: SDL3 is not available here either.

engine/CMakeLists.txt
stephengtuggy: SDL3 is not available in Ubuntu Noble, as noted above.

engine/CMakeLists.txt
stephengtuggy: You're right. We're dropping support for Fedora 41, so this should go away entirely.

engine/CMakeLists.txt
stephengtuggy: SDL3 is not available on Red Hat/Rocky Linux 9.x either.

engine/CMakeLists.txt
stephengtuggy: Fixed.

royfalk:
> All that said, I don't think we can hold up this PR on this discussion.

I'm not. I've approved the PR last week.

evertvorster:
ImGUI messes up the game for me... 

It seems to be running in some sort of window, and paints the game in the background, darkened. 
This is not only the menu, but also in the bases and HUD. 

image

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

engine/CMakeLists.txt
BenjamenMeyer: what about the SDL2-compat module?

royfalk:
> ImGUI messes up the game for me...
> 
> It seems to be running in some sort of window, and paints the game in the background, darkened. This is not only the menu, but also in the bases and HUD.

Opened #1508 to track this.

royfalk:
I'm trying to test this PR. I ran:
`roy@horus:~/git/vegastrike_copy/Vega-Strike-Engine-Source$ cmake  .`

and got
```
...
-- Found LibArchive: /usr/lib/x86_64-linux-gnu/libarchive.so (found version "3.7.2") 
** Build Type: RelWithDebInfo
-- Configuring Unit Tests
-- Downloading GoogleTest via FetchContent_MakeAvailable
-- Found Threads: TRUE  
-- GoogleTest fetched via FetchContent_MakeAvailable
CMake Error at CMakeLists.txt:465 (FIND_PACKAGE):
  By not providing "FindSDL3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "SDL3", but
  CMake did not find one.

  Could not find a package configuration file provided by "SDL3" with any of
  the following names:

    SDL3Config.cmake
    sdl3-config.cmake

  Add the installation prefix of "SDL3" to CMAKE_PREFIX_PATH or set
  "SDL3_DIR" to a directory containing one of the above files.  If "SDL3"
  provides a separate development package or SDK, be sure it has been
  installed.


```

royfalk:
OK. I've compiled both SDL3 and SDL3_image from the source and cmake --configure works.
However, now compilation fails. The errors appear related to imgui.
```

/home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui_internal.h:769:(.text._ZN6ImPoolI10ImGuiTableE6RemoveEji[_ZN6ImPoolI10ImGuiTableE6RemoveEji]+0x4d): undefined reference to `ImGuiStorage::SetInt(unsigned int, int)'
/usr/bin/ld: ../imgui/libvegastrike_imgui.a(imgui_tables.cpp.o): in function `ImVector<ImGuiOldColumns>::reserve(int)':
/home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui.h:2235:(.text._ZN8ImVectorI15ImGuiOldColumnsE7reserveEi[_ZN8ImVectorI15ImGuiOldColumnsE7reserveEi]+0x25): undefined reference to `ImGui::MemAlloc(unsigned long)'
/usr/bin/ld: /home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui.h:2235:(.text._ZN8ImVectorI15ImGuiOldColumnsE7reserveEi[_ZN8ImVectorI15ImGuiOldColumnsE7reserveEi]+0x53): undefined reference to `ImGui::MemFree(void*)'
/usr/bin/ld: ../imgui/libvegastrike_imgui.a(imgui_tables.cpp.o): in function `ImVector<ImGuiTable>::reserve(int)':
/home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui.h:2235:(.text._ZN8ImVectorI10ImGuiTableE7reserveEi[_ZN8ImVectorI10ImGuiTableE7reserveEi]+0x23): undefined reference to `ImGui::MemAlloc(unsigned long)'
/usr/bin/ld: /home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui.h:2235:(.text._ZN8ImVectorI10ImGuiTableE7reserveEi[_ZN8ImVectorI10ImGuiTableE7reserveEi]+0x4f): undefined reference to `ImGui::MemFree(void*)'
/usr/bin/ld: ../imgui/libvegastrike_imgui.a(imgui_tables.cpp.o): in function `ImPool<ImGuiTable>::GetOrAddByKey(unsigned int)':
/home/roy/git/vegastrike_copy/Vega-Strike-Engine-Source/libraries/imgui/imgui_internal.h:764:(.text._ZN6ImPoolI10ImGuiTableE13GetOrAddByKeyEj[_ZN6ImPoolI10ImGuiTableE13GetOrAddByKeyEj]+0x12): undefined reference to `ImGuiStorage::GetIntRef(unsigned int, int)'
collect2: error: ld returned 1 exit status
gmake[2]: *** [libraries/gui/CMakeFiles/test_gui.dir/build.make:102: libraries/gui/test_gui] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:723: libraries/gui/CMakeFiles/test_gui.dir/all] Error 2
/usr/bin/ld: ../lib/libgtest.a: error adding symbols: archive has no index; run ranlib to add one
collect2: error: ld returned 1 exit status
gmake[2]: *** [engine/CMakeFiles/Vega_Strike_tests.dir/build.make:447: Vega_Strike_tests] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:840: engine/CMakeFiles/Vega_Strike_tests.dir/all] Error 2
gmake: *** [Makefile:166: all] Error 2
```

I think I'm missing something, as I was under the impression that this was a lot more polished.

engine/CMakeLists.txt
stephengtuggy: Ubuntu specifically asks us not to switch to SDL2-compat as a substitute for SDL2. I'm not sure why.

stephengtuggy: (That only applies to later versions of Ubuntu than Noble, anyway.)

engine/CMakeLists.txt
stephengtuggy: SDL2-compat likewise is not available on 9.x.

stephengtuggy:
@royfalk Sorry you're having trouble. I probably should have provided more documentation somewhere.

With the latest code, I had to figure out how to handle the fact that some Linux versions include SDL3 and SDL3_image, and others do not. With the ones that do not, we have to get the libraries from somewhere. I decided to get them using FetchContent_MakeAvailable, and control whether that is called using CMake parameters. Perhaps I would do better to make it automatic based on the Linux distro and version information?

stephengtuggy:
OK, looks like I figured out the dimmed-screen issue. @kheckwrecker @evertvorster @royfalk can you guys confirm?

stephengtuggy:
Note that if you need to, you can specify `INSTALL_SDL3=1` and `INSTALL_SDL3_IMAGE=1` as environment variables, if your distro does not include these libraries by default. This will install them using CMake FetchContent_MakeAvailable.

@BenjamenMeyer do you want to keep going this way? Or drop the distros/versions that do not include SDL3/SDL3_image? At least from our CI?

kheckwrecker:
> OK, looks like I figured out the dimmed-screen issue. @kheckwrecker @evertvorster @royfalk can you guys confirm?

Looks good to me. No flickering, no double images, and no dimmed screens. I'm getting excellent frame rates as well! Would really like to see this merged with master.

engine/CMakeLists.txt
BenjamenMeyer: > Ubuntu specifically asks us not to switch to SDL2-compat as a substitute for SDL2. I'm not sure why.

Assuming you meant `for SDL3`...

@stephengtuggy where do you see that?

evertvorster (approved):
It plays well. 
There is some window dressing to do, but gameplay wise this is good to go.

BenjamenMeyer:
> Note that if you need to, you can specify `INSTALL_SDL3=1` and `INSTALL_SDL3_IMAGE=1` as environment variables, if your distro does not include these libraries by default. This will install them using CMake FetchContent_MakeAvailable.
> 
> @BenjamenMeyer do you want to keep going this way? Or drop the distros/versions that do not include SDL3/SDL3_image? At least from our CI?

we can keep them as long as they support the SDL2 compat modules and that works.

engine/CMakeLists.txt
stephengtuggy: https://packages.debian.org/trixie/libsdl2-compat

stephengtuggy:
> > Note that if you need to, you can specify `INSTALL_SDL3=1` and `INSTALL_SDL3_IMAGE=1` as environment variables, if your distro does not include these libraries by default. This will install them using CMake FetchContent_MakeAvailable.
> > @BenjamenMeyer do you want to keep going this way? Or drop the distros/versions that do not include SDL3/SDL3_image? At least from our CI?
> 
> we can keep them as long as they support the SDL2 compat modules and that works.

I think you are misunderstanding what libSDL2-compat is. It's an implementation of libSDL2 that is a wrapper around libSDL3 underneath. As such, it is only available on platforms that have SDL3 available.

stephengtuggy:
If we drop support for platforms that do not include SDL3/SDL3_image natively, @royfalk probably won't be able to build the game anymore. He's on Linux Mint, if I understand correctly. Linux Mint does not currently have any released versions that include SDL3 or SDL3_image packages.

I've figured out how to get the game to build and run even on platforms that do not natively include SDL3/SDL3_image packages. It's working. Why not merge this thing and then figure out the platform support as a fast follow?

BenjamenMeyer:
> > > Note that if you need to, you can specify `INSTALL_SDL3=1` and `INSTALL_SDL3_IMAGE=1` as environment variables, if your distro does not include these libraries by default. This will install them using CMake FetchContent_MakeAvailable.
> > > @BenjamenMeyer do you want to keep going this way? Or drop the distros/versions that do not include SDL3/SDL3_image? At least from our CI?
> > 
> > 
> > we can keep them as long as they support the SDL2 compat modules and that works.
> 
> I think you are misunderstanding what libSDL2-compat is. It's an implementation of libSDL2 that is a wrapper around libSDL3 underneath. As such, it is only available on platforms that have SDL3 available.

yeah - you're right - I think that came from my early intro to it in one of the conversations.

Okay - getting back to the question of support...

1. It would probably be good for us to make another SDL2-based release especially if we're going to have to drop some platforms that would technically be under our support statement so that folks on those platforms have something to utilize.
2. The ImGUI + SDL3 work is showing a lot of promise and it shouldn't be delayed any longer than necessary. Granted, ImGUI is already in so we're using it with SDL2 so things will improve even with 0.10.x being just SDL2.

I'm a little in limbo between the two positions, and in some respects lean towards a `let's get 0.10.x cleaned up and released, followed by a quick turn around of releasing 0.11.x with SDL3`. But I also won't want to see this age too much.

Thoughts?

stephengtuggy:
> > > > Note that if you need to, you can specify `INSTALL_SDL3=1` and `INSTALL_SDL3_IMAGE=1` as environment variables, if your distro does not include these libraries by default. This will install them using CMake FetchContent_MakeAvailable.
> > > > @BenjamenMeyer do you want to keep going this way? Or drop the distros/versions that do not include SDL3/SDL3_image? At least from our CI?
> > > 
> > > 
> > > we can keep them as long as they support the SDL2 compat modules and that works.
> > 
> > 
> > I think you are misunderstanding what libSDL2-compat is. It's an implementation of libSDL2 that is a wrapper around libSDL3 underneath. As such, it is only available on platforms that have SDL3 available.
> 
> yeah - you're right - I think that came from my early intro to it in one of the conversations.
> 
> Okay - getting back to the question of support...
> 
>     1. It would probably be good for us to make another SDL2-based release especially if we're going to have to drop some platforms that would technically be under our support statement so that folks on those platforms have something to utilize.
> 
>     2. The ImGUI + SDL3 work is showing a lot of promise and it shouldn't be delayed any longer than necessary. Granted, ImGUI is already in so we're using it with SDL2 so things will improve even with 0.10.x being just SDL2.
> 
> 
> I'm a little in limbo between the two positions, and in some respects lean towards a `let's get 0.10.x cleaned up and released, followed by a quick turn around of releasing 0.11.x with SDL3`. But I also won't want to see this age too much.
> 
> Thoughts?

Maybe it's time to split off a 0.10.x branch (SDL2, ImGui) and start working on 0.11.x on `master` (SDL3 + ImGui).

royfalk:
I won't quote but just respond to issues raised above:
1. I'll be fine. If I'll have to, I'll build SDL3 from the source.
2. As a policy, I thought it was premature to move to SDL3. It means potentially leaving many of our users behind.
3. If we do move, there should be a way for CMake to get SDL3 as part of the build process for users.
4. How difficult would it be to support either SDL2 or SDL3 for a while? Something like:
```

FIND_PACKAGE(SDL3 REQUIRED)
    IF (SDL3_FOUND)
        LIST(APPEND VSE_TST_INCLUDES ${SDL3_INCLUDE_DIRS})
        LIST(APPEND VSE_TST_LIBS ${SDL3_LIBRARIES})
        MESSAGE(STATUS "SDL3 Found")
        SET(HAVE_SDL 1)
        SET(HAVE_SDL3 1)
        SET(SDL_WINDOWING 1)
    ELSE (SDL3_FOUND)
        FIND_PACKAGE(SDL2 REQUIRED)
        IF (SDL2_FOUND)
            LIST(APPEND VSE_TST_INCLUDES ${SDL2_INCLUDE_DIRS})
            LIST(APPEND VSE_TST_LIBS ${SDL2_LIBRARIES})
            MESSAGE(STATUS "SDL2 Found")
            SET(HAVE_SDL 1)
            SET(HAVE_SDL3 0)
            SET(SDL_WINDOWING 1)
       ELSE (SDL2_FOUND)
           MESSAGE(FATAL_ERROR "Failed to find required library SDL3/SDL2")
       ENDIF (SDL2_FOUND)
 ENDIF (SDL2_FOUND)
```

This would require some ifdefs throughout the code. As I've not worked on it, I'm not sure if this is something easy (trivial it ain't) or really hard.

stephengtuggy:
@royfalk agreed with number 3 above.

Supporting either SDL2 or SDL3 from the same codebase would be challenging, indeed. I don't recommend it.

BenjamenMeyer:
@royfalk your `#2` is my concern. I do agree that doing both SDL2 and SDL3 in the same codebase will be problematic... I doubt we can abstract it away in a meaningful way without losing too much; and IFDEF'ing it would likely make for quite a mess. We can explore if it desired.

@stephengtuggy if we split off the 0.10.x branch now that would mean we'd be maintaining both for a while and we'd have to make some of the fixes on both branches too. Do we want to have both a 0.10.x and 0.11.x release in support for a while?

Also - do we want to hash this out more in a Discussion/Issue as opposed to a PR?

stephengtuggy:
> @stephengtuggy if we split off the 0.10.x branch now that would mean we'd be maintaining both for a while and we'd have to make some of the fixes on both branches too. Do we want to have both a 0.10.x and 0.11.x release in support for a while?

@BenjamenMeyer we've been in that situation before, plenty of times. Of supporting one release while working on the next, and making many of the same fixes to both. I don't see it as much of an issue.

By the time we're ready to release 0.11.x, maybe in a year or so, hopefully the Linux distributions will have caught up a little more, and we can drop support for 0.10.x at that point. Or soon thereafter.

BenjamenMeyer:
> @BenjamenMeyer we've been in that situation before, plenty of times. Of supporting one release while working on the next, and making many of the same fixes to both. I don't see it as much of an issue.

Kind of...we haven't really kept supporting more than one branch before. We did cut 0.9.x early and took a long time to release it so we had 0.8.x (release) supported, 0.9.x in-the-works for release, and 0.10.x moving ahead. So kind of.
Just calling out that we'd have to accept it if we did.

stephengtuggy:
@BenjamenMeyer yes, that is accurate.

stephengtuggy:
This PR represents a significant step forward in the technology and performance of the game. It's been languishing for weeks, and I'm tired of waiting. I went ahead and created the `0.10.x` branch. It's a protected branch under our existing branch protection rules. This PR continues to target the `master` branch, which will now be for 0.11.x. Hopefully the above is acceptable, and we can get this out the door shortly.

Please note also that I removed all the Linux distributions and versions that do not natively include SDL3 and SDL3_image. The resulting set of Linux distros and versions is pretty short. However, all our existing supported versions of Windows and macOS continue to be supported with these changes.

engine/CMakeLists.txt
stephengtuggy: Fixed. I have removed support for all Linux versions that do not have an SDL3 package natively available.

engine/CMakeLists.txt
stephengtuggy: Fixed

engine/CMakeLists.txt
stephengtuggy: Fixed

engine/CMakeLists.txt
stephengtuggy: Fixed

engine/CMakeLists.txt
stephengtuggy: Fixed

engine/CMakeLists.txt
stephengtuggy: Fixed

engine/CMakeLists.txt
stephengtuggy: Fixed

stephengtuggy:
@BenjamenMeyer still waiting for your approval. I think I addressed your feedback in its entirety. Can you review once more?

PR #1487, opened 2025-12-16
Move text rendering to imGui

royfalk:
Text rendering will now use modern techniques with anti-aliasing and what not.

Please answer the following:

**Code Changes:**
- [ ] Have the PR Validation Tests been run? Not yet. I've focused exclusively on testing the test rendering, as that's what I've touched.

Issues:
- Mouse cursor is no longer the green arrow. May be unrelated to this PR.
- No support for font and size. Currently uses imGui default.
- No support for undocumented markup. See for example "<A Dead Man's Ship>" in the intro monologue.
- Load game menu - does not rely on the base computer code in the engine. Instead uses its own python code. Changes broke some of the functionality there. Font is default. Mouse over and selected not working.
- Text location is not perfect. It wasn't perfect before but now it's easier to see.
- Base computer (engine) still uses the old system. Not sure why.

### DIFF
--- a/engine/src/gfxlib_struct.h
+++ b/engine/src/gfxlib_struct.h
@@ -200,10 +200,10 @@ struct GFXColor {
     }
 
     // Implicit conversion FROM ImU32
-    GFXColor(ImU32 col);
+    explicit GFXColor(ImU32 col);
 
     // Implicit conversion TO ImU32
-    operator ImU32() const;
+    explicit operator ImU32() const;
 };
 
 inline GFXColor operator*(float s, const GFXColor &c) {
--- a/engine/src/cmd/base_interface.cpp
+++ b/engine/src/cmd/base_interface.cpp
@@ -457,16 +457,19 @@ void BaseInterface::Room::Draw(BaseInterface *base) {
                                 <= kYLower) {                             //check lower screenborder
                             text_pos_y = (y + fabs(text_offset_y)
                                     + text_hei);
-                        }                                   //align on top
-                        text_marker.color= GFXColor(text_color_r, text_color_g, text_color_b, links[i]->alpha);
+                        }   //align on top
+                        GFXColor temp_marker_color(text_color_r, text_color_g, text_color_b, links[i]->alpha);
+                        text_marker.color = static_cast<ImU32>(temp_marker_color);
                         text_marker.SetPos(text_pos_x, text_pos_y);
                         if (links[i]->pythonfile != "#" && text_marker.GetText().find("XXX") != 0) {
-                            GFXColor tmpbg = text_marker.background_color;
+                            GFXColor tmpbg(text_marker.background_color);
                             bool automatte = (0 == tmpbg.a);
-                            if (automatte)
-                                text_marker.background_color= GFXColor(0, 0, 0, base_text_background_alpha);
+                            if (automatte) {
+                                GFXColor temp_background_color( 0, 0, 0, base_text_background_alpha );
+                                text_marker.background_color = static_cast<ImU32>(temp_background_color);
+                            }
                             text_marker.Draw(text_marker.GetText(), 0, true, false, automatte);
-                            text_marker.background_color = tmpbg;
+                            text_marker.background_color = static_cast<ImU32>(tmpbg);
                         }
                         GFXEnable(TEXTURE0);
                     }                     //if draw_text
@@ -513,17 +516,19 @@ void BaseInterface::Room::Draw(BaseInterface *base) {
                     if (enable_markers) {
                         text_pos_y += text_hei;
                     }
-                    text_marker.color = GFXColor(1, 1, 1, 1);
+                    GFXColor white_color(1, 1, 1, 1);
+                    text_marker.color = static_cast<ImU32>(white_color);
                     text_marker.SetPos(text_pos_x, text_pos_y);
 
                     GFXDisable(TEXTURE0);
-                    GFXColor tmpbg = text_marker.background_color;
+                    GFXColor tmpbg(text_marker.background_color);
                     bool automatte = (0 == tmpbg.a);
                     if (automatte) {
-                        text_marker.background_color = GFXColor(0, 0, 0, base_text_background_alpha);
+                        GFXColor temp_background_color( 0, 0, 0, base_text_background_alpha );
+                        text_marker.background_color = static_cast<ImU32>(temp_background_color);
                     }
                     text_marker.Draw(text_marker.GetText(), 0, true, false, automatte);
-                    text_marker.background_color= tmpbg;
+                    text_marker.background_color = static_cast<ImU32>(tmpbg);
                     GFXEnable(TEXTURE0);
                 }
                 //link border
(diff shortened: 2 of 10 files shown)


evertvorster:
OK... it builds, which is a huge step forward. 
There are a couple of blockers:
1. Can't select a saved game to load. 
2. Text size does not scale to the requested size in the config.

I'll carry on testing now, but those two are pretty serious.

evertvorster:
Text in the flight UI of course also does not follow the config size. 

Text highlighting in bases is not ideal. Some of the hot spots work, others don't. On Atlantis (The only one I could get into because I was unable to load any saved games) You can cycle between locations with your right mouse button. There is no text highlighted when you hover over "Hangar"

There seems to be some description of the scene that should have been at the bottom of the screen.

Overall, this PR is moving in the right direction, but I can't approve it for merging into master when it will break game saved game loading and the base interface highlighting. 

The font size is debatable, but it is also something that master already has, and that will be going away if this PR is merged, and will then need to be fixed ASAP.

stephengtuggy:
> On play testing, the following areas need to be fixed before considering a merge into master:
> 
>     1. Font size to follow what is set in the config.
> 
>     2. Base text highlighting should be proper for each hotspot.
> 
>     3. Able to load saved games by clicking on them.

To me, item 2 above is a lower priority, since it was already broken prior to this PR. Item 1 is a medium priority, a nice-to-have. Item 3 is a show-stopper though. We definitely need to be able to load saved games, including from the main menu.

stephengtuggy:
(The font sizes will certainly need to be fixed sooner or later. I'm just not sure that it has to be in this PR.)

royfalk:
Opened a discussion as this is somewhat more complicated than it seems:
https://github.com/orgs/vegastrike/discussions/1500

royfalk:
> 1.     Font size to follow what is set in the config. 

We really can't.  The fonts are embedded in OpenGL and the sizes don't match.

> 2.     Base text highlighting should be proper for each hotspot.

As noted, this is an existing bug and unrelated to this. Need to find the source of the regression.

> 3.     Able to load saved games by clicking on them.

That one is fixed by the last commit. I won't push it because it relies on roboto.ttf and this will break all tests.

stephengtuggy:
> > 3. ```
> >    Able to load saved games by clicking on them.
> >    ```
> 
> That one is fixed by the last commit. I won't push it because it relies on roboto.ttf and this will break all tests.

Okay, what's the path forward then? @royfalk can you find another way to fix this functionality that does not break the tests? Perhaps include roboto.ttf in the test_assets folder in the engine repo or something?

stephengtuggy:
(Assuming that the absence of the Roboto font from Assets-Production is what will cause the tests to break?)

royfalk:
> Okay, what's the path forward then?

I've removed the reliance on Roboto. For now, the code uses the default imGui font at size 18, which seems OK.
I've also added support for line width background. A bit of an intuition leap. The original width value is clearly a fraction of the screen. I used `float width = fraction * configuration().graphics.resolution_x * 2;` and it's bang on. <shrug>.
Still pending, implementation of font color and startlower. This is probably why some stuff in the cockpit is overlapping.

evertvorster (approved):
This code now runs, and I can load a saved game, so from a usability point of view this is good to go... with the understanding that we'll sort out font sizes changing with config options soon, and getting rid of that overlapping fonts in the VDU.

engine/src/gfxlib_struct.cpp
```diff
@@ -36,13 +36,34 @@
...
+#include "imgui.h"
 
 #include "root_generic/options.h"
 
 #include <vector>
 
 #include "configuration/configuration.h"
 
+
+// Implicit conversion FROM ImU32
```
stephengtuggy, line 48: Please, no implicit conversions to or from other types. It gets really messy, really quick. (I think that's why the non-GCC builds are all failing.) Please make all zero-arg and one-arg constructors `explicit`. Let me know if you need help figuring out how to make the code work around that.

stephengtuggy: Note: this isn't just me saying this. It's CLion and, I believe, clangd as well.

engine/src/gfxlib_struct.h
```diff
@@ -195,6 +198,12 @@ struct GFXColor {
         return GFXColor(((r < 0) ? 0 : ((r > 1) ? 1 : r)), ((g < 0) ? 0 : ((g > 1) ? 1 : g)),
                 ((b < 0) ? 0 : ((b > 1) ? 1 : b)), ((a < 0) ? 0 : ((a > 1) ? 1 : a)));
     }
+
+    // Implicit conversion FROM ImU32
```
stephengtuggy, line 202: Same as above regarding implicit conversions

libraries/imgui/CMakeLists.txt
```diff
@@ -28,6 +28,14 @@
 
 MESSAGE(STATUS "PROCESSING DIRECTORY: ${CMAKE_CURRENT_SOURCE_DIR}")
 
+#Find SDL_image
+FIND_PACKAGE(SDL2_IMAGE REQUIRED)
```
stephengtuggy, line 32: The root CMakeLists.txt doesn't already pull this in?

royfalk:
Some additional comments:
1. The whole text code is arcane. It looks like the position is bottom left. So in imgui I need to go up one line to be where I need to be.
2. Some layout in VDU is done via new lines. Something like:
   Line 1\nLine2\n\n\nLine 5
3.In load game, I can't make the selected game background to be the line width instead of text width. If I do, it breaks other things. Not going to bother with it. I'll just add this to the tasks and fix it right.

libraries/imgui/CMakeLists.txt
stephengtuggy: Not a big deal either way. Just wondering.

engine/src/gfxlib_struct.cpp
stephengtuggy: After further thought, I'm not going to block on this. We can tackle it in a later PR if need be. Just so long as we get the CI to pass again.

engine/src/gfxlib_struct.h
stephengtuggy: Fixed. Thanks!

engine/src/gfxlib_struct.cpp
stephengtuggy: Fixed. Thanks!

libraries/imgui/CMakeLists.txt
royfalk: I thought it was to satisfy some requirement at the GUI library but no. 
I removed it and got:
```
[cmake] CMake Error at libraries/imgui/CMakeLists.txt:42 (TARGET_LINK_LIBRARIES):
[cmake]   Target "vegastrike_imgui" links to:
[cmake] 
[cmake]     SDL2_image::SDL2_image
[cmake] 
[cmake]   but the target was not found.  Possible reasons include:
```

PR #1504, opened 2025-12-28
Partial fix for bug 1488

### LINKED ISSUE #1488: Problems upgrading Dostoevsky ships
Finally got enough credits to purchase a Dostoevsky, but ran into a couple of problems.

Ships starts with 111.6 of 125 cubic meters (cm) of free space. Most upgrades only take 2 cubic meters of space.

Problem 1: If you make the mistake of purchasing Dumbfire missiles before other upgrades, your free space drops to zero, and you cannot purchase any other upgrades, Each missile is only supposed to take 0.1 cm of space, and I only got 32 missiles, so the available space should have dropped by 3.2 cm, not 111.6 cm. Also, however much free space you have left (i.e. you buy other upgrades first), the second you buy the missiles, the space drops to zero.

Problem 2: If  you try to rectify problem #2 by selling your Dumbfire missiles to get rid of them, your free space remains at zero, and you still cannot purchase any other upgrades, even though you no longer have any ammunition on your ship.

The problem also occurs with the Hyena, but it's available space is so small, it's not surprising that you run out of space. However, the issue with selling them and not recovering the space still exists with this ship.

I tried buying a Plowshare, and purchasing Dumbfires left lots of available space.

royfalk:
Ammo will not add mass or volume. This includes missiles.
This is a temporary fix for now. Ideally, we should refactor both mounts and turrets. 
However, this is a major undertaking.

Closes #1488

**Code Changes:**
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Tested bug fix only and some stuff around it.

Issues:
- Ammo should have mass and volume.
- Guns have ammo. Missiles are ammo. Not sure how the game handles the dissonance.

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -2123,9 +2123,10 @@ bool BaseComputer::isTransactionOK(const Cargo &originalItem, TransactionType tr
             have_money = item.GetPrice() * quantity <= ComponentsManager::credits;
             have_space = (playerUnit->upgrade_space.CanAddCargo(item) || item.IsWeapon());
             upgrade_already_installed = playerUnit->UpgradeAlreadyInstalled(item);
-            
+
             // Simply not allowed
             if (!UpgradeAllowed(item, playerUnit)) {
+                color_prohibited_upgrade_flag = true;
                 return false;
             }
 
@@ -3349,7 +3350,7 @@ void BaseComputer::BuyUpgradeOperation::concludeTransaction(void) {
         finish();
         return;
     }
-
+    
     int quantity = 1; // By default, we buy one unit
     if(m_newPart->mounts[0].IsMissileMount()) {
         int mount_size = playerUnit->mounts[m_selectedMount].size;
@@ -3361,8 +3362,7 @@ void BaseComputer::BuyUpgradeOperation::concludeTransaction(void) {
         // Light missile size is 64. Medium is 128. Light and medium is 192.
         // This hack produces inconsistent but plausible results.
         quantity = mount_size / missile_size;
-        std::cout << "Buying " << quantity << " missiles. (" 
-                  << missile_size << "/" << mount_size << ")\n";
+        VS_LOG(important_info, (boost::format("Buying %1% missiles (%2%/%3%)") % quantity % missile_size % mount_size));
     }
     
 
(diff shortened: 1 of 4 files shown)


royfalk:
> I'm unable to confirm whether this has resolved the issue or not, as I'm now unable to purchase any ammo or weapons at all for the Dostoevsky. Everything under ammo and weapons shows up in Ship Upgrade in purple, including the dumbfire ammo.

Well that's not good. Are you starting from a new game and buying a clean Dostoevsky?

kheckwrecker:
I started from a new game and bought a clean Dostoevsky. I also started from an older saved game and purchased a clean Dostoevsky. I was unable to purchase dumbfire missiles in either scenario.

royfalk:
Hmm. Looks like this is more complicated.
Weapons in general shouldn't count toward upgrade volume, as they are all mounted on dedicated hard points.
But, the game also has some kind of calculation for missile mounts - you can fit several light missiles in a medium mount.
Need some more tweaking.

royfalk:
Accidentally committed to task_fix_1411 and not here.
Two main issues remain:
1. Need to check we've covered all possibilities for upgrades.
2. Number of missiles upgrades is a weird function of mount sizes (really a bit field).  This hack produces inconsistent but plausible results. For example, Dostoevsky buys a light missile. light missile size is 64. Missile mount size is light (64) + medium (128) + heavy (256). Result is seven missiles.

engine/src/cmd/basecomputer.cpp
```diff
@@ -3333,46 +3341,43 @@ bool BaseComputer::BuyUpgradeOperation::checkTransaction(void) {
...
+        int mount_size = playerUnit->mounts[m_selectedMount].size;
+        const Mount mount = m_newPart->mounts[0];
+        const WeaponInfo* info = mount.type;
+        int missile_size = as_integer(info->size);
+        
+        // This is not as clear as it looks.
+        // Light missile size is 64. Medium is 128. Light and medium is 192.
+        // This hack produces inconsistent but plausible results.
+        quantity = mount_size / missile_size;
+        std::cout << "Buying " << quantity << " missiles. ("
```
stephengtuggy: Can you change this to something like


```suggestion
        VS_LOG(important_info, (boost::format("Buying %1% missiles (%2%/%3%)") % quantity % missile_size % mount_size));
```

?

royfalk: Done.

stephengtuggy: Cool. Thanks

PR #1506, opened 2025-12-29
Fix bug 1477

royfalk:
Check we don't have upgrades of this type (e.g. capacitors) installed.

Co-authored by ChatGPT - avoid code duplication in const/non-const identical methods.

**Code Changes:**
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Tested the issue only.

Issues:
- Some prohibited upgrades get red color and others purple. Not sure why.

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -2085,6 +2085,7 @@ bool BaseComputer::isTransactionOK(const Cargo &originalItem, TransactionType tr
     Unit *baseUnit = m_base.GetUnit();
     bool havemoney = true;
     bool havespace = true;
+    bool upgrade_already_installed = false;
     switch (transType) {
         case BUY_CARGO:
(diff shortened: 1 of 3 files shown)


royfalk:
This is by (new) design. The old game had support for at least three types of upgrades - value, add and multiplier and also support for multiple - for upgrades you can buy more than one of.

Capacitors, SPEC capacitors and overdrives are examples of this. These upgrades in units.json have a mult_ prefix.

I ditched everything and left only value upgrades and no more than one per type.

BenjamenMeyer:
Being able to buy more than one of something like a SPEC drive didn't make sense to me - seemed more like a bug. Two of it shouldn't be possible, nor should in really improve things - improving something should be through an upgrade ( Level 1 :arrow_forward: Level 2, etc).

BenjamenMeyer (approved):

PR #1507, opened 2025-12-29
Fix bug 1411 - purchase of cargo was crashing

### LINKED ISSUE #1411: Unable to buy armor
This has been an issue for a while. In the upgrade screen, when I click on an armor entry and click Buy, nothing happens -- at least nothing visible in the game window.

I would like to see this fixed.

royfalk:
as volume was 0 and there was a div by 0 error.

Closes #1411 

**Code Changes:**
- [ ] Have the PR Validation Tests been run? _No. Just the bug fix._

### DIFF
--- a/engine/src/resource/cargo.cpp
+++ b/engine/src/resource/cargo.cpp
@@ -47,10 +47,7 @@ static bool _parse_bool(const std::string& value) {
     return (value == "1" || value == "true");
 }
 
-// A simple utility function to prevent zero or negative mass and volume
-static double at_least_tenth(double input) {
-    return input > 0 ? input : 0.1;
-}
+static const double minimum_mass_and_volume = 0.01;
 
 // Constructors
 
@@ -62,14 +59,16 @@ Cargo::Cargo(std::string name, std::string description, int quantity, double pri
              bool component, bool installed, bool integral, bool weapon, bool passenger, 
              bool slave, double functionality) :
              name(name), description(description), quantity(quantity), price(price), 
-             category(category), mass(at_least_tenth(mass)), volume(at_least_tenth(volume)), mission(mission),
+             category(category), mass(std::max(minimum_mass_and_volume,mass)), 
+             volume(std::max(minimum_mass_and_volume,volume)), mission(mission),
              component(component), installed(installed), integral(integral) {
     this->functionality = functionality;
 }
 
-Cargo::Cargo(std::string name, std::string category, float price, int quantity, 
-          float mass, float volume): name(name), category(category),
-          price(price), quantity(quantity), mass(at_least_tenth(mass)), volume(at_least_tenth(volume)) {}
+Cargo::Cargo(std::string name, std::string category, double price, int quantity, 
+          double mass, double volume): name(name), category(category),
+          price(price), quantity(quantity), mass(std::max(minimum_mass_and_volume,mass)), 
+          volume(std::max(minimum_mass_and_volume,volume)) {}
              
 
 /*  0 name
@@ -99,8 +98,8 @@ Cargo::Cargo(std::string& cargo_text) {
         case 1: category = cargo_parts[1]; break;
         case 2: price = locale_aware_stod(cargo_parts[2]); break;
         case 3: quantity = locale_aware_stoi(cargo_parts[3]); break;
-        case 4: mass = at_least_tenth(locale_aware_stod(cargo_parts[4])); break;
-        case 5: volume = at_least_tenth(locale_aware_stod(cargo_parts[5])); break;
+        case 4: mass = std::max(locale_aware_stod(cargo_parts[4]), minimum_mass_and_volume); break;
+        case 5: volume = std::max(locale_aware_stod(cargo_parts[5]), minimum_mass_and_volume); break;
         case 6: functionality = Resource<double>(locale_aware_stod(cargo_parts[6]), 0.0, 1.0); break;
         case 7: break; // max_functionality is always 1.0. cargo_parts[6] not used.
         case 8: description = cargo_parts[8]; break;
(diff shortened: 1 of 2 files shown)


engine/src/components/components_manager.cpp
```diff
@@ -341,10 +341,12 @@ bool ComponentsManager::_Buy(CargoHold *hold, ComponentsManager *seller, Cargo *
...
     quantity = std::min(max_affordable_quantity, quantity);
 
-    // Check the maximum you can fit in your hold
-    int max_stackable_quantity = static_cast<int>(std::floor(hold->AvailableCapacity() / item->GetVolume()));
-    quantity = std::min(max_stackable_quantity, quantity);
-
+    // Check the maximum you can fit in your hold, but only if volume != 0
+    if(item->GetVolume() > 0) {
+        int max_stackable_quantity = static_cast<int>(std::floor(hold->AvailableCapacity() / item->GetVolume()));
+        quantity = std::min(max_stackable_quantity, quantity);
```
BenjamenMeyer, line 347: shouldn't this also have a `std::max` with `0` so it doesn't go negative?

royfalk: So this is a bit complex. We have a 'mult_shady_moreupgrade' upgrade, which could show up as negative volume but for now is implemented as volume=0.1. 
In reality, if we have a negative volume for actual upgrades, this will have issues elsewhere as well.
Better to fix this properly in cargo.

royfalk:
Last commit belongs in task_fix_1488. Reverting.

engine/src/components/components_manager.cpp
BenjamenMeyer: while I agree we should fix it elsewhere too, we should also protect ourselves here as well.

royfalk: OK. Fixed cargo class to prevent any option of negative or zero mass or volume.
If there's a need for some exception, we can revisit the issue.

engine/src/resource/cargo.cpp
```diff
@@ -47,6 +47,11 @@ static bool _parse_bool(const std::string& value) {
     return (value == "1" || value == "true");
 }
 
+// A simple utility function to prevent zero or negative mass and volume
+static double at_least_tenth(double input) {
+    return input > 0 ? input : 0.1;
```
BenjamenMeyer: `input > 0.1` would probably be best since `input > 0` would be true even if `input` was `0.000000000000001` - essentially the zero value for the type.

Keep in mind that floating point (float/double) do not store exactly zero - it can be `0.000000001` or `-0.9999999999999` (number of zeros/nines is dependent on the type).

BenjamenMeyer (changes requested):
I get what you're trying to do, but I don't think the code quite gets there since it doesn't do an accurate check for a zero value of a double-precision type.

kheckwrecker:
I can think of one possible exception - in the campaign, it may be possible that you get asked to transport something with zero mass for game play purposes. I'm thinking of, for example, the plot mission in Privateer where you're asked to hide the alian artifact  in a "secret hold" on your ship. As I recall, that artifact had zero mass, so that it had no effect on your ability to take on cargo runs, while you held onto the artifact.

engine/src/resource/cargo.cpp
royfalk: I am aware of this and was actually being intentional here.
if input == 0.00001, input > 0, use this value. I assume the game designers know what they're doing.
If input <= 0, even if the game designers intended this, I can't allow this and default to a small number. For volume and mass, 0.1 is a reasonable number to plug here.
Reasonable in the sense that buying/selling cargo will be reflected in the shop mass and volume.

royfalk:
> I can think of one possible exception - in the campaign, it may be possible that you get asked to transport something with zero mass for game play purposes. I'm thinking of, for example, the plot mission in Privateer where you're asked to hide the alian artifact in a "secret hold" on your ship. As I recall, that artifact had zero mass, so that it had no effect on your ability to take on cargo runs, while you held onto the artifact.

You're confusing mass and volume and the secret hold is a separate hold.
"file": "Halogen_Gases", "mass": "1.2", "volume": "25". If you fill your Llama with it, you'll have 40 units weighing 48 tons.
"file": "Gold", "mass": "19", "volume": "1". 1000 units and 19k tons.
There's no reason for the alien artifact to be 0 mass or even 0 volume.

Originally, there was one hold for upgrades, cargo and secret cargo. As a result, the game was filled with code to check what we're dealing with and where it is. Now that we've separated the holds, this is much simpler. But even then, the secret hold had a volume.

When we port this version to WC, or rather WC to this version, we'll need to check the secret hold code. I'm not sure it works properly, as I haven't tested it.

engine/src/resource/cargo.cpp
BenjamenMeyer: Please update the comment to note then since as I mentioned `0.0000000000...1` could still be zero.

BenjamenMeyer:
> > I can think of one possible exception - in the campaign, it may be possible that you get asked to transport something with zero mass for game play purposes. I'm thinking of, for example, the plot mission in Privateer where you're asked to hide the alian artifact in a "secret hold" on your ship. As I recall, that artifact had zero mass, so that it had no effect on your ability to take on cargo runs, while you held onto the artifact.
> 
> You're confusing mass and volume and the secret hold is a separate hold. "file": "Halogen_Gases", "mass": "1.2", "volume": "25". If you fill your Llama with it, you'll have 40 units weighing 48 tons. "file": "Gold", "mass": "19", "volume": "1". 1000 units and 19k tons. There's no reason for the alien artifact to be 0 mass or even 0 volume.
> 
> Originally, there was one hold for upgrades, cargo and secret cargo. As a result, the game was filled with code to check what we're dealing with and where it is. Now that we've separated the holds, this is much simpler. But even then, the secret hold had a volume.
> 
> When we port this version to WC, or rather WC to this version, we'll need to check the secret hold code. I'm not sure it works properly, as I haven't tested it.

@kheckwrecker is correct regarding the original Privateer game - they did write it as having `zero mass` and it shows up that way in the manifest for the game, which I think was to not have it impact anything performance wise of the ship and not take up cargo space for how that game functioned. IIRC it also did not impact your inventory for what your ship could carry so you could use any kind of ship regardless of how full it was with cargo to complete the mission.

That said, I also agree with @royfalk  in that it could still work here even if the mass is extremely small or it uses the secret hold to accommodate. So I don't think we're losing anything, just changing how the mechanism would work to recreate the scenario.

engine/src/resource/cargo.cpp
royfalk: As in 
`// A simple utility function to prevent (close to) zero or negative mass and volume`
?

stephengtuggy: `at_least_tenth` doesn't do what it says it does. Please, let's either rename it to something like `greater_than_zero`, or change the implementation to not allow any value less than 0.1.

royfalk: Pushed a new commit. Let's continue the discussion after it.

royfalk:
I was never happy with the naming of the function. I finally realised it's basically a max function, so I replaced it with one.
I did move the constructor parameters from float to double. The class members themselves are all double already, so it should be fine. I think it's ready for a last review.

stephengtuggy:
I like this latest version. It's elegant and clean. I'm sure it will "just work."

Issue #1027, opened 2025-02-11
Pause key does not work in Windows - locks up game


kheckwrecker:
Manual and screen says game can be paused by hitting the Pause key or the "|" key (shift+\). All that does in Windows is lock up the game. Hitting either or both do not unpause the game and leaves it locked up. I'd be interested in whether this works in Linux.

evertvorster:
Hmm... On Linux, when you hit Pause/Break... the game stops. 
About a minute later there is a "paused" graphic that pops up, and the game is then properly paused. You can move your cursor, and music is playing, but it seems the simulation is stopped. 

When you hit Pause/Break again, the simulation starts up again. Interestingly enough, while I was testing this, I was looking at the Serenity space station. When unpausing the game it fully disappears for a second or so.

kheckwrecker:
Interesting. I'll have to try waiting a bit longer in Windows and see if it behaves the same way. Regardless, it probably should not take a minute for pause to kick in.

BenjamenMeyer:
FYI - the `Pause` is really just a time manipulation in the game - it doesn't truly stop; it just changes the time to go extremely slow - something like 1 second becomes 1 billionth of a nanosecond. This ends up effecting  how long it takes for the pause/unpause to work, and it may end up being machine dependent.

@royfalk I wonder if there's a better way that we could implement the Pause...but it'll probably take a bit to figure out.

@kheckwrecker do you see anything popping up in the logs to denote that the correct key was pressed?
You should see something like:
```
FireKeyboard::TogglePause(): Pause key detected
```
that's the first thing to check.

royfalk:
> I wonder if there's a better way that we could implement the Pause...but it'll probably take a bit to figure out.

Refactor the main loop. Extract keyboard to its own thread. Pause everything else.

kheckwrecker:
> [@kheckwrecker](https://github.com/kheckwrecker) do you see anything popping up in the logs to denote that the correct key was pressed? You should see something like:
> 
> ```
> FireKeyboard::TogglePause(): Pause key detected
> ```
> 
> that's the first thing to check.

I ran the latest beta of release 0.9 to try pausing again and just wait and see what happens. The first time, about 2 minutes after pressing the Pause key, the word "Paused" appeared in the middle of the screen. The second time I tried, it took over 4 minutes before "Paused" appeared. Interestingly enough, when you hit the Pause key again, the game unpauses immediately. 

[...]

Issue #595, opened 2021-10-26
Crash When Trying to Pause (Tumbleweed)


romanyacik:
Vega Strike crashes when trying to pause on openSuSE Tumbleweed. Bug not encountered in Linux Mint

romanyacik:
Sorry for the late reply. Probably freeze is closer. What I'm experiencing seems to be more in line with the known issue that you're describing

fernando0010:
This still happens on VS 0.9.1, Ubuntu 22.04.5 LTS. Pauses freezes the game, and there's no way to unpause it but to kill the process and start again.

royfalk:
Unfortunately, this isn't a part of the code that I've looked at and won't look for a while.

PR #1516, opened 2026-01-11
Implement pause game

royfalk:
Possible fix for #1027 and #595 - needs testing.

**Code Changes:**
- [ ] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation _Only that the functionality works_

### DIFF
--- a/engine/src/cmd/ai/firekeyboard.cpp
+++ b/engine/src/cmd/ai/firekeyboard.cpp
@@ -65,8 +65,6 @@
 #include "cmd/unit_util.h"
 #include "resource/cargo.h"
 
-extern bool toggle_pause();
-
 FireKeyboard::FireKeyboard(unsigned int whichplayer, unsigned int whichjoystick) : Order(WEAPON, 0) {
     memset(savedTargets, 0, sizeof(void *) * NUMSAVEDTARGETS);
     this->autotrackingtoggle = 1;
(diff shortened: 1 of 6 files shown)


stephengtuggy:
Let's get the double-buffering fix in here please. Then I'm fine with merging it.

royfalk:
> Should note that the fix for double buffering is not in this branch, so I had to manually apply the fix in order to test this patch.

Why? Why isn't the patch in this branch. I branched from master and it should be there.
Also, why would this include a patch in a PR? If something is missing from master, we should create a separate PR to add it.

Looking at master, I see 4455e8c812f72813b27914d0f4c7d8eef3df7cef has been added already. 
I also checked gl_misc.cpp and GFXEndScene doesn't have a double buffer command. 
I think you're either applying this PR to a not-up-to-date master or some other branch.

kheckwrecker:
Apologies for the confusion. GitHub Desktop (which is what I use) is not great for applying PR to a branch, so I've just been pulling the branch this PR is in and testing it. If I can figure out how to actually apply a PR to a branch locally, I'll start doing that. I only mentioned this in the first place in case it was a problem. I'm glad that it's not, and I believe this PR is good to merge, from my perspective.

BenjamenMeyer:
> Apologies for the confusion. GitHub Desktop (which is what I use) is not great for applying PR to a branch, so I've just been pulling the branch this PR is in and testing it. If I can figure out how to actually apply a PR to a branch locally, I'll start doing that. I only mentioned this in the first place in case it was a problem. I'm glad that it's not, and I believe this PR is good to merge, from my perspective.

@kheckwrecker if you want to test a merge then create a branch from master and merge the PR branch into it:
```
$ git checkout master
$ git merge remoteOrigin/remoteBranch
```

PR #1521, opened 2026-01-13
Some minor polish to base_computer.cpp:

royfalk:
1. Remove XMLSupport - use std::to_string 
2. Remove UnsaturatedColor. Use GFXColor default parameter instead.
3. Refactor prettyPrintFloat to modern standard
4. Remove PRETTY_ADD* macros - refactor
5. Remove MODIFIES* macros - not used

Co-authored-by: ChatGPT (prettyPrintFloat)

**Code Changes:**
- [ ] Have the PR Validation Tests been run? only tested base_computer, as this does not involve actual flight code. See how to test below.

**How to test:**
1. Accept a mission. View active missions. See first mission has a prefix of 1.
2. Go to upgrade menu. Check colors are the same as before.
3. Two places:
   a. Go to ship dealer. Check ship prices are still OK.
   b. Go to cargo. Check lowest/highest prices text is OK.
4. Checked by 3.
5. Not checked. Macro not used.

**Issues:**
- Did not test in other locales (e.g.) Spain, Germany, France or Russia, where thousands separator is period and decimal point is a comma.
- Need to move the set global locale (marked with TODO) somewhere else. This is not urgent and it's very likely most of the code is going to migrate to python.

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -87,6 +87,7 @@ using VSFileSystem::SaveFile;
 #include <sys/types.h>
 #endif
 #include <sys/stat.h>
+#include <locale>
 #include "src/vega_cast_utils.h"
 
 // Can't declare in header because PyObject is problematic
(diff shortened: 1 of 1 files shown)


stephengtuggy (approved):
Nice and elegant. Looks good to me.

Issue #1529, opened 2026-01-20
Turret view/movement is buggy


royfalk:
For example, H496 you can do a 180 and clearly see a very partial view of the back of your ship.

<img width="1917" height="1120" alt="Image" src="https://github.com/user-attachments/assets/06613030-37a8-40da-908c-55afe66d8890" />

stephengtuggy:
Why is this a bug? You're seeing your ship from the perspective of the current turret, as if you were seated inside it. I regard this as the expected behavior. Unless I'm misunderstanding something.

royfalk:
So there are two bugs here:
1. You shouldn't be able to turn the turret inward toward the ship.
2. If you do, you should see the ship and **not through it**.

Generally, this is an OpenGL limitation.  At least I assume so. 
However, we should be able to identify such cases at the engine level and do something about it.

stephengtuggy:
Oh, I see what you're saying. Thank you for clarifying.

Is the turret actually parallel with the surface it's attached to, when it's at its lowest point? If so, I wouldn't necessarily consider that a bug. Although I guess an argument could be made that the turret should have a governor that automatically prevents it from pointing at parts of its own ship that stick up/out.

Regarding being able to see through the ship: That appears to be part of a larger bug I've noticed for a while now, where objects in space turn invisible as your ship approaches them. It seems to be some sort of occluder calculation error or something. Definitely needs to be fixed.

Maybe we should split this ticket into two? One for each of the sub-issues raised?

BenjamenMeyer:
yes, turrets should be range limited so they can spin in a circular motion but only move so  far towards the vessel holding it. I different kinds of turrets would have different range limitations.

A couple extremes:
1. A turret that is mostly externally mounted would have near full motion aside from a small circle at the top
2. A turret that is mostly flush mounted (think Millenium Falcon turret) would have a narrow range of motion.

Of course there are turrets in between those extremes, and that should be part of the ship design and specifications.

Issue #1528, opened 2026-01-20
Turrets only move on one axis (possible bug)


royfalk:
Examples:
- H496 has two turrets that move up and down
- Vigilance has two turrets that move left and right

It's unclear if this is a bug or expected behavior of these two ships.

stephengtuggy:
Yes, I can confirm this behavior. I think it is a bug. It's been in place for years.

PR #1530, opened 2026-01-20
Fix 1525

### LINKED ISSUE #1525: Player turrets are broken
I've been refactoring something else and needed to check something with the turrets.
I flew a Vigilance - a ship with multiple turrets.
You can switch to the first turret using [.
However, you can't fire the turret or switch back.
Looking at the code, it doesn't appear to have been tampered significantly in the last 17 years.
Last time I used functioning turrets was in Privateer in 2006.

royfalk:
closes #1525 

**Code Changes:**
- [ ] Have the PR Validation Tests been run? see tests below

**Tests conducted:**
- Play-tested with Vigilance and H496. 
- Did not the standard tests yet. 

**Issues:**
- Open #1528 and #1529 for additional issues identified.

### DIFF
--- a/engine/src/gfx/cockpit.cpp
+++ b/engine/src/gfx/cockpit.cpp
@@ -957,7 +957,6 @@ void GameCockpit::SelectProperCamera() {
 
 extern vector<int> respawnunit;
 extern vector<int> switchunit;
-extern vector<int> turretcontrol;
 
 void DoCockpitKeys() {
     CockpitKeys::Pan(KBData(), PRESS);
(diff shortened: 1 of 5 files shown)


libraries/gfx_generic/cockpit_generic.cpp
```diff
@@ -138,15 +137,16 @@ void Cockpit::Init(const char *file, bool isDisabled) {
...
+// A bad function name. 
+// This function potentially returns the current subunit, so its target can be extracted.
+// See star_system.cpp
 Unit *Cockpit::GetSaveParent() {
-    Unit *un = parentturret.GetUnit();
-    if (!un) {
-        un = parent.GetUnit();
-    }
-    return un;
+    return current_unit;
```
BenjamenMeyer, line 144: should this be:
```
if (current_unit != nullptr) {
    return current_unit;
}
return parent_unit:
```
Or do we have some kind of guarantee that the `current_unit` is always properly set?

engine/src/gfx/cockpit.cpp
```diff
@@ -1105,12 +1104,72 @@ void MapKey(const KBData &, KBSTATE k) {
...
+    }
+
+    VS_LOG(debug, (boost::format("GameCockpit::NextTurret - turrets = %1%") % number_of_turrets));
+    if(number_of_turrets == 0) {
+        return;
+    }
+
+    current_view++;
+    VS_LOG(debug, (boost::format("GameCockpit::NextTurret - current_view = %1%") % current_view));
+    if(current_view == number_of_turrets) {
```
BenjamenMeyer, line 1132: this would be better as `>=` to protect against increments that go too far

royfalk:
I'll implement the suggestions in #1541.

Issue #1512, opened 2026-01-09
Update copyright year to 2026 in all applicable game engine source files


stephengtuggy:

BenjamenMeyer:
this should hopefully be easier this year since we've standardized the statements; hopefully we can store a script that we can more easily run to do this - perhaps a GH Action that submits a PR at some point?

stephengtuggy:
@BenjamenMeyer yes, there's a script for it already.

stephengtuggy:
We do have a script for this purpose (`script/add_gpl_license.py`) but the existing version wasn't smart enough to handle all the different variations of copyright headers we have in our source files. I am attempting to make it a lot smarter this year, in order to reduce the amount of work that needs to be done manually.

stephengtuggy:
Task finished as of #1577 and #1578

PR #1577, opened 2026-04-04
Update Copyright Headers to 2026 in Remaining Files (0.10.x)

stephengtuggy:

Please answer the following:

Code Changes:

Issues:
- Finishes implementing #1512 for 0.10.x

Purpose:
- What is this pull request trying to do? Update the copyright headers in the rest of the source files

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -2,12 +2,10 @@
 # CMakeLists.txt
 #
 # Vega Strike - Space Simulation, Combat and Trading
-# Copyright (C) 2001-2025 The Vega Strike Contributors:
+# Copyright (C) 2001-2026 The Vega Strike Contributors:
 # Project creator: Daniel Horn
-# Original development team: As listed in the AUTHORS file. Specifically:
-# safemode, Anth0rx, pyramid, Nachum Barcohen, Rune Morling
+# Original development team: As listed in the AUTHORS file. Specifically: safemode, Anth0rx, pyramid, Nachum Barcohen, Rune Morling
 # Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
-#
 #
 # https://github.com/vegastrike/Vega-Strike-Engine-Source
 #
--- a/engine/objconv/blender_xmesh_export.py
+++ b/engine/objconv/blender_xmesh_export.py
@@ -9,18 +9,18 @@
 # Copyright (C) 2021 daviewales
 # Copyright (C) 2025, 2026 Stephen G. Tuggy
 #
-# This file is free software: you can redistribute it and/or modify
+# This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
-# This file is distributed in the hope that it will be useful,
+# This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
 #
 # You should have received a copy of the GNU General Public License
-# along with this file.  If not, see <https://www.gnu.org/licenses/>.
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 
 
--- a/engine/objconv/blender_xmesh_export.py
+++ b/engine/objconv/blender_xmesh_export.py
@@ -3,11 +3,10 @@
 # blender_xmesh_export.py
 #
 # Vega Strike - Space Simulation, Combat and Trading
-# Copyright (C) 2001-2025 The Vega Strike Contributors:
+# Copyright (C) 2001-2026 The Vega Strike Contributors:
 # Project creator: Daniel Horn
 # Original development team: As listed in the AUTHORS file
 # Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
-#
 #
 # https://github.com/vegastrike/Vega-Strike-Engine-Source
 #
--- a/engine/objconv/blender_xmesh_export.py
+++ b/engine/objconv/blender_xmesh_export.py
@@ -2,28 +2,25 @@
 ##
 # blender_xmesh_export.py
 #
-# Vega Strike - Space Simulation, Combat and Trading
-# Copyright (C) 2001-2026 The Vega Strike Contributors:
-# Project creator: Daniel Horn
-# Original development team: As listed in the AUTHORS file
-# Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
+# Copyright (C) dandandaman
+# Copyright (C) 2007 ace123
+# Copyright (C) 2013 cellsafemode, klaussfreire
+# Copyright (C) 2020 pyramid3d
+# Copyright (C) 2021 daviewales
+# Copyright (C) 2025, 2026 Stephen G. Tuggy
 #
-# https://github.com/vegastrike/Vega-Strike-Engine-Source
-#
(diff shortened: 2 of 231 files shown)


engine/objconv/blender_xmesh_import.py
```diff
@@ -1,5 +1,30 @@
 #!BPY
-
+##
+# blender_xmesh_import.py
+#
+# Vega Strike - Space Simulation, Combat and Trading
+# Copyright (C) 2001-2026 The Vega Strike Contributors:
+# Project creator: Daniel Horn
+# Original development team: As listed in the AUTHORS file
+# Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
```
BenjamenMeyer: This should be like the 3rd party files we imported, see L35 below

engine/tools/common.cpp
```diff
@@ -2,13 +2,11 @@
  * common.cpp
  *
  * Vega Strike - Space Simulation, Combat and Trading
- * Copyright (C) 2001-2025 The Vega Strike Contributors:
+ * Copyright (C) 2001-2026 The Vega Strike Contributors:
  * Project creator: Daniel Horn
- * Original development team: As listed in the AUTHORS file. Specifically:
- * Konstantinos Arvanitis
+ * Original development team: As listed in the AUTHORS file. Specifically: Konstantinos Arvanitis
```
BenjamenMeyer, line 7: I didnt' see `Konstantinos Arvanitis` in the AUTHORS files

engine/CMakeLists.txt
```diff
@@ -2,13 +2,11 @@
 # CMakeLists.txt
 #
 # Vega Strike - Space Simulation, Combat and Trading
-# Copyright (C) 2001-2025 The Vega Strike Contributors:
+# Copyright (C) 2001-2026 The Vega Strike Contributors:
 # Project creator: Daniel Horn
-# Original development team: As listed in the AUTHORS file. Specifically:
-# safemode, Anth0rx, pyramid, Nachum Barcohen, Rune Morling
+# Original development team: As listed in the AUTHORS file. Specifically: safemode, Anth0rx, pyramid, Nachum Barcohen, Rune Morling
```
BenjamenMeyer, line 7: I didn't see `Anth0rx` in the AUTHORS files

BenjamenMeyer (changes requested):

engine/tools/common.cpp
stephengtuggy: Then maybe he should be added. He is one of the authors of this file, according to the Git history.

engine/CMakeLists.txt
stephengtuggy: `Anth0rx` is definitely one of the authors of this file according to the Git history.

Isn't the AUTHORS file generated from two separate sources? One a static, preexisting list of authors, and the other the Git history of the repository? At least, I thought that's how it worked.

stephengtuggy: Or maybe I'm mistaken. I just checked the Git history, and I am not finding that user mentioned.

Surely this entry came from somewhere. Now the mystery is where.

engine/tools/common.cpp
stephengtuggy: This file used to be in a different directory location. It was added there to the CVS history (predating the Git history) by user @karvan , in [this commit](https://github.com/stephengtuggy/Vega-Strike-Engine-Source/commit/652b19776e284c9f746b65409f573ea85230c26e). I believe this is the Konstantinos Arvanitis referenced in the copyright header.

engine/objconv/blender_xmesh_import.py
stephengtuggy: I have definitely seen `Cub of Judah's Lion` in author lists for Vega Strike files before. I am 95% certain he was a Vega Strike contributor at one time.

This file appears to be a Blender plugin to allow Blender to import Vega Strike models for editing and conversion to other formats. It was written by the identified Vega Strike contributor. As such, it does not need the 3rd-party copyright format, AFAICT.

engine/CMakeLists.txt
stephengtuggy: Yeah, I can't find any sign of this user in the commit history. Maybe we can remove them from the authors list, unless there are any objections?

engine/tools/common.cpp
BenjamenMeyer: we should add them to the AUTHORS file

engine/CMakeLists.txt
BenjamenMeyer: we have 2 Authors file - one was generated from commit history and one was imported from Subversion and I don't know how it was generated. I would just add them to the non-git generated authors file.

engine/objconv/blender_xmesh_import.py
BenjamenMeyer: there's 3 files that make this up:
- [blender_xmesh_export.py](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/objconv/blender_xmesh_export.py)
- [blender_xmesh_import.py](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/objconv/blender_xmesh_import.py)
- [blender_xmesh_import.txt](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/objconv/blender_xmesh_import.txt)

The last one (blender_xmesh_import.txt) seems to be the defining one which includes its license and history. So I would lean towards having these 3 files honor that file and its license documentation. It's still GPLv2+ so we're still good for licensing, but I do think we need to adjust the header for the related files and copyright origin to match our 3rd party scheme.

engine/objconv/blender_xmesh_export.py
BenjamenMeyer, line 1: this file should be adjusted too - see https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1577/changes#r3052328266

engine/objconv/blender_xmesh_import.py
stephengtuggy: OK, I have updated these .py files to a generic multi-copyright format, pulling the list of authors from the actual Git history.

I see what you're saying about the .txt file. I have updated that one's `LICENSING AND DISTRIBUTION` section at the top to match blender_xmesh_import.py.

However, it looks like the .txt file only documents the import script, not the export script. Am I right?

engine/objconv/blender_xmesh_export.py
stephengtuggy: Done

engine/tools/common.cpp
stephengtuggy: Which AUTHORS file should they be added to? `doc/AUTHORS`? Or the `AUTHORS` file in the root of the repo that was auto-generated from the Git history?

engine/CMakeLists.txt
stephengtuggy: Is the non-git-generated authors file the one under the `doc/` subdirectory?

engine/objconv/blender_xmesh_import.py
BenjamenMeyer: yes it appears to be just documenting the import script; the export one though does trace to [dandandaman](https://sourceforge.net/u/dandandaman/profile/) who is more closely related to VS where Alex 'CubOfJudahsLion' Feterman is more closely related to Blender - it may have been something brought over by dandandaman.

stephengtuggy: I forgot to upload my latest changes yesterday. Should be fixed now.

engine/tools/common.cpp
stephengtuggy: Added.

engine/CMakeLists.txt
stephengtuggy: Added.

BenjamenMeyer (approved):

PR #1582, opened 2026-04-10
[Draft] C++20 - Pi Constants

stephengtuggy:

Please answer the following:

Code Changes:

Issues:

Purpose:
- What is this pull request trying to do? Begin experimenting with C++20 features. In particular, start using the templated Pi constants.


engine/objconv/asteroidgen.cpp
```diff
@@ -52,80 +50,51 @@ class Vector {
...
+        const float m = Mag();
+        i = m * std::cosf(theta);
+        j = m * std::sinf(theta);
+    }
+
+    void Pitch(const float rad) {
+        const float theta = std::atan2f(j, k) + rad;
+        const float m = Mag();
+        k = m * std::cosf(theta);
+        j = m * std::sinf(theta);
```
BenjamenMeyer, line 88: these seem to be the same methods as in the star system class. Perhaps we need another common class to provide these - not blocking for this PR but something we should probably consider going forward.

engine/src/gfx/nav/navitemstodraw.h
```diff
@@ -24,7 +24,11 @@
  * You should have received a copy of the GNU General Public License
  * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
  */
-// NO HEADER GUARD
+
+#ifndef NAVITEMSTODRAW_H
+#define NAVITEMSTODRAW_H
```
BenjamenMeyer: while I'd like us to have header guards on everything we did document a number of places we couldn't use them. Are we sure we can add it here?

libraries/root_generic/galaxy_gen.cpp
```diff
@@ -186,56 +180,26 @@ class Vector {
...
+        const float m = Mag();
+        i = m * std::cosf(theta);
+        j = m * std::sinf(theta);
+    }
+
+    void Pitch(const float rad) {
+        const float theta = std::atan2f(j, k) + rad;
+        const float m = Mag();
+        k = m * std::cosf(theta);
+        j = m * std::sinf(theta);
```
BenjamenMeyer, line 203: same as the asteroid and star system - same methods

BenjamenMeyer (commented):
I like where this is going. We'll probably identify another class level we need to do in the process too given how similar some of the code (and even was before).

engine/src/gfx/nav/navitemstodraw.h
stephengtuggy: I'm not positive. We'll see if it compiles. That's the main criterion, I figure.

Without the header guard, we might end up declaring the stuff in the header more than once, which might be an issue. So I think we should add the header guard if at all possible.

engine/objconv/asteroidgen.cpp
stephengtuggy: Yes, they are basically the same methods in all three (at least) locations in the code. We can probably merge them at some point.

libraries/root_generic/galaxy_gen.cpp
stephengtuggy: Indeed. See my reply above.

stephengtuggy:
It looks like this PR may not be as big as I was afraid of. I think I have all the Pi conversions done, and it's only around 600 lines changed.

Now for the bad news. Elsewhere in the code, our custom aligned allocator isn't compatible with C++20 as written. We'll have to refactor it significantly.

I did find a couple of C++20 aligned allocator implementations [here](https://indiegamedev.net/2022/03/27/custom-c20-memory-allocators-for-stl-containers/). One or both of them should work for our purposes, I think -- and probably speed up the game as a bonus.

stephengtuggy:
Oh, wait. I still need to make the allocator-related changes. Those might turn out to be quite substantial on their own.

I think I'll make the allocator-related changes their own PR, and try to get that one merged first, before coming back to this one.

engine/src/gfx/env_map_gent.cpp
```diff
@@ -123,7 +123,6 @@ static void Lighting( RGBColor &Col, const Vector &Norm )
 const int lmwid = 512;
 const int lmwido2 = lmwid / 2;
 const float ooWIDTHo2 = 2. / lmwid;
-const float PIoWIDTHo2 = 2 * 3.1415926535 / lmwid;
```
BenjamenMeyer, line 126: are any of these other constants actually used? If not, we could probably clean them up too; not blocking though.

engine/src/gfx/nav/navitemstodraw.h
BenjamenMeyer: if it compiles great. there's a few header files that don't have them because they get included into a few files in the middle of those files and its done for a side effect. Not a pattern I'm a fan of, but we haven't tackled fixing it in a better way yet. We marked those files with the `//NO HEADER GUARD` comment so we knew not to add it - hence my question.

engine/src/gfx/env_map_gent.cpp
stephengtuggy: Most of them don't seem to be used, but I can't guarantee that they aren't used by some Python script somewhere. Although I suppose we could check the API that gets exported/exposed to Python and get a definite answer on that.

BenjamenMeyer: They would have to be exported into the Python layer for any python script to use them.

engine/objconv/asteroidgen.cpp
```diff
@@ -27,14 +27,11 @@
...
 #include <ctime>
-#ifndef M_PI
-#define M_PI 3.14159265358979323846264338328
-#endif
-#include <float.h>
+#include <cmath>
+#include <numbers>
+#include <cstdio>
+#include <cstdlib>
 #if defined(__APPLE__) && defined (__MACH__)
```
royfalk: This is no longer needed. It was relevant when Apple had incomplete implementation of sqrtf, cosf and sinf.

engine/objconv/asteroidgen.cpp
```diff
@@ -43,6 +40,7 @@
 
 float safety_zone = 0;
 using std::vector;
+using namespace std::numbers;
```
royfalk: using std::something is not recommended.

engine/src/audio/test.cpp
```diff
@@ -45,6 +45,7 @@
 #include <string>
 //#include <limits>
 #include <cassert>
+#include <numbers>
```
royfalk: I don't think we've ever run this test.cpp. I'm not sure what I think about this? I thought about turning it into a unit test, but it can't run automatically. Someone needs to _hear_ this.

royfalk (commented):
Note to self: Reviewed up to atmosphere.cpp not including.

engine/objconv/asteroidgen.cpp
stephengtuggy: Not in header files, certainly. But what about .cpp implementation files? Is it OK there? Or not according to the style guide(s) we're following? What I read on Stack Overflow suggested that `using namespace std::something` was generally OK in an implementation file.

engine/objconv/asteroidgen.cpp
stephengtuggy: Fair enough. I will try removing it and see if it compiles on macOS still.

engine/src/audio/test.cpp
stephengtuggy: Agreed. We should probably start using this at some point. It's useful test code. But like you said, it can't really be automated.

engine/objconv/asteroidgen.cpp
BenjamenMeyer: personally, I prefer to only have `using namespace <local namespace from its own header>` as the sole entity like that.

It would probably be better if you did:
```
typedef std::numbers numbers;
...
... numbers::pi_v<float> ...
```

May be even:
```
typedef numbers::pi_v<float> PiFloat;
typedef numbers::pi_v<double> PiDouble;
```
which would be best in one of our own common math headers.

engine/src/audio/test.cpp
BenjamenMeyer: what if we saved the output to a file and then had a wave form comparison?
There's probably a way that we could automate it - though we should be careful about file sizes for that. If necessary we could use another repository to store the audio files for the tests.

stephengtuggy: I'm not sure the output would be precise enough for a wave form comparison. Also, how would we go about that?

engine/objconv/asteroidgen.cpp
stephengtuggy: I did see a comment I added in one of these files, several years ago, to the effect that entire namespaces should not be imported according to the Google Style Guide. Even in implementation files, I guess?

@BenjamenMeyer In theory, even what you posted could cause problems if someone else defined the symbols `PiFloat`, `PiDouble`, or `numbers`. What if we used a Vega-Strike-specific prefix? I.e., `VEGA_PI_FLT` and `VEGA_PI_DBL`? Those would be similar to a couple of the constant names that were already in use, only less likely to be confused with Visual Studio.

stephengtuggy: Or `kVegaPiFloat` and `kVegaPiDouble`

engine/src/audio/test.cpp
BenjamenMeyer: > how would we go about that?

*if* we could do it, we'd run it through and record it some how - may be have a version of the test that records to disk so we can validate it, and then use that as a baseline.

regarding validation:

1. if it's speech, then using speed to text tooling could process the file and validate the language is close enough
2. if it's sound (weapon fire, explosion, etc) then we could allow a fuzzy comparison
3. it might be that we setup an AI docker image to do testing with some of this too; it'd be a lot more work but could help with automation. (If we did this, I'd try to make it an option workflow that we occasionally run.)

not saying it's not going to take time to work out the details and get things working well enough to rely on it, but it could be something in the right direction.

engine/objconv/asteroidgen.cpp
stephengtuggy: Based on https://google.github.io/styleguide/cppguide.html#Constant_Names , it should probably be `kVegaPiFloat` and `kVegaPiDouble`.

engine/src/gfx/env_map_gent.cpp
stephengtuggy: Answer: Yes, these other constants are used elsewhere in this same file. So we probably can't remove them -- at this point, anyway.

engine/objconv/asteroidgen.cpp
stephengtuggy: Removed.

engine/objconv/asteroidgen.cpp
stephengtuggy: Done.

BenjamenMeyer: I'm not hung up on names, no need to :bike: :derelict_house: over it.

> to the effect that entire namespaces should not be imported according to the Google Style Guide. Even in implementation files, I guess?

Generally yes - `using namespaces` should be used minimally to prevent naming conflicts. If you import a 3rd party's namespace than you lose control of the symbols in the namespace.

engine/objconv/asteroidgen.cpp
```diff
@@ -33,6 +33,16 @@
 #include <cstdlib>
 #include "src/vs_math.h"
 
+#include "root_generic/vega_random.h"
+#include "src/vs_math.h"
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846264338328
```
BenjamenMeyer, line 40: this should get updated to `std::numbers::pi_v<float>` or `std::numbers::pi_v<double>`, probably the float version

stephengtuggy: Oh, good catch. I missed this occurrence, or else it slipped back in during a merge from `master`. This should get replaced by `kVegaPiFloat`, `kVegaPiDouble`, or both.

PR #1585, opened 2026-04-10
 Refactor random number generation to use a singleton

### LINKED ISSUE #1081: Switch random number generation to C++11 style RNG
C++11 style random number generation apparently produces higher-quality randomization. It would be good to use if we can, I think.

stephengtuggy:
…namely, `VegaRandom::Instance()`

Please answer the following:

Code Changes:

Issues Addressed:
- Closes #1081  

Known Issues Remaining:
#1609 

Purpose:
- What is this pull request trying to do? Finish the RNG conversion project


engine/src/vs_random.h
```diff
@@ -32,177 +33,145 @@
...
-
-/*
- * #define N 624
- * #define M 397
- * #define MATRIX_A 0x9908b0dfUL   // constant vector a
- * #define UPPER_MASK 0x80000000UL // most significant w-r bits
- * #define LOWER_MASK 0x7fffffffUL // least significant r bits
- */
+#ifndef VS_RANDOM_H
+#define VS_RANDOM_H
```
BenjamenMeyer, line 37: same caution about the header guard - but great if we could get this cleaned up so we can actually have a header guard

stephengtuggy:
@royfalk your Random Sanity tests uncovered some major logic errors in my new implementations of several of the randomization functions. I was multiplying a value that I thought was between 0 and 1 by `(max - min)`, then adding `min`. The value was not between 0 and 1. It was between 0 and 2^32!

It's fixed now. Or should be, anyway. Now I just create a `uniform_int_distribution` between `min` and `max`, then return a random value within that distribution/range. I have tried running the unit tests several more times, and they're all passing now.

stephengtuggy:
~This PR is still a work in progress. Lots more conversions to do! But at least the tests are all passing now.~

engine/launcher/general.cpp
```diff
@@ -237,16 +237,16 @@ char *StripExtension(char *filename) {
...
+        min = end;
+        max = start;
+    }
+    if (end == start) {
+        return start;
+    }
+    return VegaRandom::Instance().RandomInt32InRange(min, max);
 }
 
 void randcode(char *line, int length) {
```
BenjamenMeyer, line 252: @stephengtuggy seems like this should get updated too

engine/mission/include/general.cpp
```diff
@@ -247,24 +248,16 @@ void StripExtension(char *filename) {
 #ifdef _G_RANDOM
 
 int randnum(int start, int end) {
```
BenjamenMeyer, line 250: are these methods similar enough we could consolidate down to 1 implementation?

engine/setup/src/include/general.cpp
```diff
@@ -211,24 +211,16 @@ void strappend(char *dest, char *source, int length) {
 }
 
 int randnum(int start, int end) {
```
BenjamenMeyer, line 213: same question here as https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1585/changes/cdcf8d695dc635a29c28f22fafd13f55be1691c9..4cd3610f71e04b0ffa392c9874bfdf3c1241a62d#r3147853209

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -1451,9 +1451,9 @@ class FlyTo : public Orders::MoveTo {
 };
 
 static Vector randVector() {
-    return Vector((rand() / (float) RAND_MAX) * 2 - 1,
-            (rand() / (float) RAND_MAX) * 2 - 1,
-            (rand() / (float) RAND_MAX) * 2 - 1);
+    return Vector(VegaRandom::Instance().RandomFloatInRange(-1.0F, 1.0F),
+            VegaRandom::Instance().RandomFloatInRange(-1.0F, 1.0F),
+            VegaRandom::Instance().RandomFloatInRange(-1.0F, 1.0F));
```
BenjamenMeyer, line 1454: is this functionally equivalent?
```
func value():
   r = random number
   v1 = r / random max
   v2 = v1 * 2
   v3 = v2 - 1
   return v3

randomVector = Vector(value(), value(), value())
```

engine/src/main_loop.cpp
```diff
@@ -862,9 +862,9 @@ void createObjects(std::vector<std::string> &fighter0name,
             pox.j = fg->pos.j + tmp;
             pox.k = fg->pos.k + tmp;
             if (pox.i == pox.j && pox.j == pox.k && pox.k == 0) {
-                pox.i = rand() * 10000. / RAND_MAX - 5000;
-                pox.j = rand() * 10000. / RAND_MAX - 5000;
-                pox.k = rand() * 10000. / RAND_MAX - 5000;
+                pox.i = VegaRandom::Instance().RandomDoubleInRange(-5000.0, 5000.0);
+                pox.j = VegaRandom::Instance().RandomDoubleInRange(-5000.0, 5000.0);
+                pox.k = VegaRandom::Instance().RandomDoubleInRange(-5000.0, 5000.0);
```
BenjamenMeyer, line 867: same question here as https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1585/changes/cdcf8d695dc635a29c28f22fafd13f55be1691c9..4cd3610f71e04b0ffa392c9874bfdf3c1241a62d#r3147892342 - though this at least seems to account for the multiplier.

engine/launcher/general.cpp
stephengtuggy: @BenjamenMeyer `randcode` is implemented in terms of calls to `randnum`, so it didn't seem to me like it needed to get updated. The new implementation of `randnum` should suffice. Am I wrong?

engine/mission/include/general.cpp
stephengtuggy: Yes, they are very similar. In fact, I think they are identical. We could probably consolidate them.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: `rand()` returns a pseudo-random integer between 0 and `RAND_MAX`, inclusive. So `(rand() / (float) RAND_MAX)` would be a pseudo-random float between 0 and 1. Multiplied by 2, it would be a random float between 0 and 2. Subtract 1, and it ends up being a float between negative 1 and positive 1. Hence what I replaced it with. The new code should return a value in the same range as the old code, albeit perhaps with a different level of granularity.

Is the new code too long? Is that why you want to replace it with an inline (I assume) function?

engine/src/main_loop.cpp
stephengtuggy: Yes, this is a somewhat different scenario, as the old code not only divides by `RAND_MAX`, but also multiplies by 10,000. The range of the result is therefore different. The new code should return a value in the same range as the original here, as well, if I'm reading this right.

engine/src/cmd/ai/aggressive.cpp
BenjamenMeyer: thanks - just at first glance it seems like its not quite the same algorithm. Your explanation makes sense and was what I was hoping it worked as but wanted to verify.

engine/launcher/general.cpp
BenjamenMeyer: got it - you're probably right.

engine/mission/include/general.cpp
stephengtuggy: Question: the files that contain these methods each have a comment to the effect that they have been designed to act independent of the other modules (header files, really). Do we want to change that design decision?

stephengtuggy: Although I guess we've already started making that shift, as these files now rely on `VegaRandom`.

BenjamenMeyer: tldr; yes - I'm good with changing that design decision - it's one from the original code base and there's now better code practices that we should follow.

Having duplicate code does't really help with maintainability. Even if it's a `Mixin` style class (multiple inheritance, generic parentage, ???) to add the functionality it's still helpful to not duplicate it.

There's some things that are in some weird headers that don't have header guards and I think we need to move away from that kind of code inclusion design - there's better methods now to do that same kind of thing,

engine/src/vs_random.h
stephengtuggy: I have since deleted this entire file, replacing it with vega_random.h. So the point is kind of moot now.

engine/mission/include/general.cpp
stephengtuggy: Shall we consolidate these methods/files into one place in this PR, or in a subsequent one? And where should we put it? In a central location in the current Vega-Strike-Engine-Source repository, such as `/general/general.h`? Or in its own repository and library, since the launcher and perhaps other applications / code units will share it?

stephengtuggy: I've started a random-number-generation-part-2 branch already, where I'm tackling some random number calculations that are not `uniform_distribution`s, but `normal_distributions` (for example) instead. Consolidating these methods and files would probably go well in the context of that part 2 PR, seems to me.

BenjamenMeyer: @stephengtuggy works for me

BenjamenMeyer (approved):

BenjamenMeyer:
LGTM pending play testing

stephengtuggy:
> LGTM pending play testing

Play testing just completed a couple of hours ago. Mostly what I noticed was that the game was running incredibly smoothly and quickly (400+ FPS!) and that 3D objects in space looked really good.

I did also notice one issue, where after dying and then respawning, I could barely get my ship to move at all. But I think I may have already seen that with one or more previous PRs as well, so it's not a new bug. It should get documented though, and fixed at some point.

Issue #71, opened 2020-04-20
Remove Ogre related code


nabaco:
As discussed in the last community meeting, we want to heavily refactor the codebase do simplify it. This includes removing all the legacy/irrelevant code, such as the OGRE integration code which doesn't actually work.
I'm opening this issue to track the code removal. A choice of a different game engine will be tracked in a separate issue.

PR #1594, opened 2026-04-28
Visual Studio 2026 support for 0.10.x

### LINKED ISSUE #1593: Build 0.10.x under Visual Studio 2026
It is my understanding that in early May, the `windows-2025` GitHub Actions image will switch to Visual Studio 2026 as the default compiler. We'd better prepare 0.10.x to build under Visual Studio 2026, if we want to be able to release it. I am working on that right now. In fact, I think I'm nearly done.

stephengtuggy:

Please answer the following:

Code Changes:

Issues Addressed:
- Closes #1593 

Known Issues Remaining:
- #1600  
- #1597 (not a new issue)
- #1601 
- #1602 

Purpose:
- What is this pull request trying to do? Add VS2026 support to 0.10.x, and hopefully fix vcpkg caching while I'm at it


stephengtuggy:
Note: This PR includes the `ai` code cleanup changes for 0.10.x.

engine/CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ IF (NOT DISABLE_CLIENT) ##########
 ENDIF(NOT DISABLE_CLIENT) ##########
 
 IF (UNIX)
-    LIST(APPEND VSE_TST_INCLUDES ${CMAKE_PREFIX_PATH}/harfbuzz)
+    LIST(APPEND VSE_TST_INCLUDES /usr/include/harfbuzz)
```
BenjamenMeyer: this should maintain the `CMAKE_PREFIX_PATH` value

stephengtuggy: `CMAKE_PREFIX_PATH` didn't turn out to have a value at all. That's what the newly-enabled CMake warnings told me. Hence why I made this change.

stephengtuggy: (See [here](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1594/changes#diff-1712b235e02bd4a3ad866fe26447a6dc5c1b449faeea513f449d244d46d03cf5R16))

stephengtuggy: Maybe we should set `CMAKE_PREFIX_PATH` in FindLinuxDistro.cmake or something.

BenjamenMeyer: See https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html

Yeah...we're probably using it wrong - we probably should add a new FindHarbuzz.cmake file that locates the harbuzz include path and sets all the variables. That's probably the correct fix.

stephengtuggy: I just tried leaving harfbuzz out entirely. Everything seems to build fine without it -- including vegasettings, which was commented out for reasons I do not understand.

BenjamenMeyer: > including vegasettings, which was commented out for reasons I do not understand.

probably because the new app is a Python-based app where the old one was a C++ binary

BenjamenMeyer: > I just tried leaving harfbuzz out entirely. Everything seems to build fine without it

Seems like we can remove it then

stephengtuggy: > > including vegasettings, which was commented out for reasons I do not understand.
> 
> probably because the new app is a Python-based app where the old one was a C++ binary

The thing is, we haven't completely obviated the need for the old-style settings yet. Some settings are still set from vegastrike.config. Especially colors and input mappings.

The other, weirder thing is that, while the Git history for this line shows it being commented out and uncommented several times in various commits over the last several years, the main Git history view for those commits **doesn't show those changes at all!** Very strange. It's like the Git history of this file is corrupted or something.

BenjamenMeyer: > > > including vegasettings, which was commented out for reasons I do not understand.
> > 
> > 
> > probably because the new app is a Python-based app where the old one was a C++ binary
> 
> The thing is, we haven't completely obviated the need for the old-style settings yet. Some settings are still set from vegastrike.config. Especially colors and input mappings.

Last version I saw did everything and more than the old settings app. So I thought we had eliminated it for 0.10.x - if not we really need to so we can have a single method for 0.10.x for config to make config usable. @royfalk thoughts?
 
> The other, weirder thing is that, while the Git history for this line shows it being commented out and uncommented several times in various commits over the last several years, the main Git history view for those commits **doesn't show those changes at all!** Very strange. It's like the Git history of this file is corrupted or something.

Based on experience that could be *GitHub* having issues with deciding on the history view. Or are you using local tools? Probably something good to toss an AI agent at.

stephengtuggy: I was using local tools. Specifically CLion.

stephengtuggy: The new settings app does everything the old settings APP did, but not everything the old settings FILE did. Yes, we definitely need to finish migrating everything over before releasing 0.10.x.

.github/workflows/Windows-CI.yml
```diff
@@ -3,24 +3,18 @@ permissions:
   contents: read
   pull-requests: read
 
-# Controls when the action will run.
```
royfalk, line 6: Why remove these comments?
Yeah, they're obvious, to you. But someone looking into it like me or a new contributor would benefit from them.

engine/objconv/mesher/CMakeLists.txt
```diff
@@ -81,13 +83,12 @@ IF (EXPAT_FOUND)
     ENDIF()
 
     # OGRE already looked for by parent
-    # USE_SYSTEM_BOOST already set by main script
```
royfalk, line 84: Can we drop Ogre already? Also #71

From a thread on Reddit:
https://www.reddit.com/r/gamedev/comments/1imh48s/is_ogre_still_a_viable_option_for_c_3d_game/
> I found OGRE as one of the best options, but I'm afraid it's outdated.
> It was never really that popular.
> ...not because it's the best tool for the job.

engine/objconv/CMakeLists.txt
```diff
@@ -120,19 +120,22 @@ IF (WIN32)
 ENDIF()
 
 
+SET(MSH_INCLUDES "")
+SET(MSH_LIBS "")
+
 #find Expat
 FIND_PACKAGE(EXPAT REQUIRED)
 IF (EXPAT_FOUND)
```
royfalk, line 128: We should probably have a task, to remove all "XML crap" and then libexpat itself.
But for now, can we change crap to stuff or something more professional?

engine/objconv/CMakeLists.txt
```diff
@@ -120,19 +120,22 @@ IF (WIN32)
...
-    SET(MSH_INCLUDES ${MSH_INCLUDES} ${EXPAT_INCLUDE_DIRS})
-    SET(MSH_LIBS ${MSH_LIBS} ${EXPAT_LIBRARIES})
+    LIST(APPEND MSH_INCLUDES ${EXPAT_INCLUDE_DIRS})
```
royfalk, line 129: Good. I like that we're modernising CmakeLists.

engine/objconv/CMakeLists.txt
```diff
@@ -188,12 +191,11 @@ IF (WIN32)
 ENDIF()
 
 # OGRE already looked for by parent
-# USE_SYSTEM_BOOST already set by main script
-IF (OGRE_FOUND AND NOT USE_SYSTEM_BOOST)
```
royfalk, line 192: Surely we can drop Ogre here? Has anyone use objconv recently?

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -543,20 +543,20 @@ bool AggressiveAI::ProcessLogic(AIEvents::ElemAttrMap &logi, bool inter) {
     return retval;
 }
 
-Unit *GetThreat(Unit *parent, Unit *leader) {
-    Unit *th = NULL;
-    Unit *un = NULL;
+Unit *GetThreat(const Unit *parent, const Unit *leader) {
+    Unit *th = nullptr;
```
royfalk: Why not `threat` and `unit`?

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -1445,9 +1445,7 @@ class FlyTo : public Orders::MoveTo {
                 WarpToP(parent, un, true);
             } else {
                 Unit *playa = _Universe->AccessCockpit()->GetParent();
-                if (playa == NULL || playa->Target() != parent || 1) {
```
royfalk: Good catch. You gotta wonder how it got to be this way.

engine/src/cmd/ai/aggressive.cpp
```diff
@@ -1581,12 +1579,12 @@ void AggressiveAI::AfterburnerJumpTurnTowards(const Unit *target) {
     AfterburnTurnTowards(this, parent);
     const float jump_time_limit = configuration().ai.force_jump_after_time_flt;
     if (jump_time_check == 0) {
-        float dist = (target->Position() - parent->Position()).MagnitudeSquared();
+        const float dist = static_cast<float>((target->Position() - parent->Position()).MagnitudeSquared());
```
royfalk: why not `distance`?

engine/src/cmd/ai/comm_ai.cpp
```diff
@@ -86,7 +86,7 @@ bool MatchingMood(const CommunicationMessage &c, float mood, float randomrespons
 }
 
 int CommunicatingAI::selectCommunicationMessageMood(CommunicationMessage &c, float mood) {
-    Unit *targ = c.sender.GetUnit();
+    const Unit *targ = c.sender.GetUnit();
```
royfalk: Why not `target`?

royfalk (commented):
I've reviewed up to fire.cpp.

I realise I haven't been able to contribute as much these past few months. It's a combination of a regional war and me doing a one semester certification, that sucked most of my free time.

I've taken a few minutes to review what I can. I don't see any blockers. Everything I commented on is either a question or nice to have.

Thank you for picking up the slack.

.github/workflows/Windows-CI.yml
stephengtuggy: Oh, okay. I didn't realize these comments still held value. I can put them back.

engine/objconv/CMakeLists.txt
stephengtuggy: Yes, I can do that.

engine/objconv/CMakeLists.txt
stephengtuggy: No one has used the objconv tools recently, AFAIK. I have been reluctant to drop them completely because they might turn out to be useful to convert Vega Strike assets to and from more standard file formats for editing in e.g. Blender. But I guess that is more of a theoretical possibility than anything, as we don't actually install OGRE with the other dependencies on any OSes -- and indeed, it only seems to be available on a few distros.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: Good point.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: Agreed.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: Also a good point.

engine/objconv/mesher/CMakeLists.txt
stephengtuggy: Yes, we can probably drop it. See my other comment, below.

Also from that Reddit thread: https://www.reddit.com/r/gamedev/comments/1imh48s/comment/mcgg7fn/ Apparently Unigine supports double-precision calculations, which are good for space sims like ours. Interesting.

engine/objconv/CMakeLists.txt
BenjamenMeyer: @stephengtuggy we should move the tools for that kind of thing into its own repository - like our existing [modtools](https://github.com/vegastrike/Tools/tree/master/modtools); they don't necessarily need to be in the engine repo and if there are internally shared libraries then it may make sense to spin off a repo of just shared libraries for those things.

That said, our modtools repo needs some attention too.

engine/objconv/mesher/CMakeLists.txt
stephengtuggy: Done.

.github/workflows/Windows-CI.yml
stephengtuggy: Done.

engine/objconv/CMakeLists.txt
stephengtuggy: Done.

engine/objconv/CMakeLists.txt
stephengtuggy: Ogre dropped.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: Done.

engine/src/cmd/ai/aggressive.cpp
stephengtuggy: Done.

engine/src/cmd/ai/comm_ai.cpp
stephengtuggy: Done.

engine/objconv/CMakeLists.txt
stephengtuggy: Task/refactor filed as #1596

engine/objconv/CMakeLists.txt
stephengtuggy: > @stephengtuggy we should move the tools for that kind of thing into its own repository - like our existing [modtools](https://github.com/vegastrike/Tools/tree/master/modtools); they don't necessarily need to be in the engine repo and if there are internally shared libraries then it may make sense to spin off a repo of just shared libraries for those things.
> 
> That said, our modtools repo needs some attention too.

Issue filed as #1599

script/bootstrap.ps1
```diff
@@ -34,22 +34,23 @@ param(
...
 
 [Environment]::SetEnvironmentVariable('VCPKG_ROOT', "$VCPKG_PARENT_DIR\vcpkg", 'User')
 $env:VCPKG_ROOT = "$VCPKG_PARENT_DIR\vcpkg"
 [Environment]::SetEnvironmentVariable('VCPKG_ASSET_SOURCES', "$VCPKG_PARENT_DIR\vcpkgAssets", 'User')
 $env:VCPKG_ASSET_SOURCES="$VCPKG_PARENT_DIR\vcpkgAssets"
-[Environment]::SetEnvironmentVariable('X_VCPKG_ASSET_SOURCES', "clear;x-azurl,file:///$VCPKG_PARENT_DIR/vcpkgAssets,,readwrite", 'User')
-$env:X_VCPKG_ASSET_SOURCES="clear;x-azurl,file:///$VCPKG_PARENT_DIR/vcpkgAssets,,readwrite"
-[Environment]::SetEnvironmentVariable('VCPKG_BINARY_SOURCES', "clear;files,$VCPKG_PARENT_DIR/vcpkgBin,readwrite", 'User')
-$env:VCPKG_BINARY_SOURCES = "clear;files,$VCPKG_PARENT_DIR/vcpkgBin,readwrite"
+[Environment]::SetEnvironmentVariable('X_VCPKG_ASSET_SOURCES', "clear;x-azurl,file:///$VCPKG_PARENT_DIR\vcpkgAssets,,readwrite", 'User')
```
BenjamenMeyer: Windows does support the forward slash convention which prevents needing to escape the backslash.... `\v` might get translated as a vertical tab instead of the directory separator (\) and `v` in `vcpkgAssets` as intended. Similarly below.

script/bootstrap.ps1
```diff
@@ -34,22 +34,23 @@ param(
...
+$env:X_VCPKG_ASSET_SOURCES="clear;x-azurl,file:///$VCPKG_PARENT_DIR\vcpkgAssets,,readwrite"
+[Environment]::SetEnvironmentVariable('VCPKG_BINARY_SOURCES', "clear;files,$VCPKG_PARENT_DIR\vcpkgBin,readwrite", 'User')
+$env:VCPKG_BINARY_SOURCES = "clear;files,$VCPKG_PARENT_DIR\vcpkgBin,readwrite"
 [Environment]::SetEnvironmentVariable('VCPKG_BUILD_TREES', "$VCPKG_PARENT_DIR\vcpkgBuild", 'User')
 $env:VCPKG_BUILD_TREES="$VCPKG_PARENT_DIR\vcpkgBuild"
 [Environment]::SetEnvironmentVariable('VCPKG_INSTALL_OPTIONS', "--x-buildtrees-root=$VCPKG_PARENT_DIR\vcpkgBuild", 'User')
 $env:VCPKG_INSTALL_OPTIONS="--x-buildtrees-root=$VCPKG_PARENT_DIR\vcpkgBuild"
 
+git clone https://github.com/Microsoft/vcpkg.git "$env:VCPKG_ROOT"
+& "$env:VCPKG_ROOT\bootstrap-vcpkg.bat" -disableMetrics
```
BenjamenMeyer: likewise `\b` might get translated as a backspace erasing the `T` in `VCPKG_ROOT`

.github/workflows/Windows-Release.yml
```diff
@@ -74,13 +77,21 @@ jobs:
...
-          path:
-            ${{ github.workspace }}/vcpkg
-            ${{ github.workspace }}/vcpkgAssets
-            ${{ github.workspace }}/vcpkgBin
-            ${{ github.workspace }}/vcpkgBuild
+          path: |
+            ${{ github.workspace }}\vcpkgAssets
+            ${{ github.workspace }}\vcpkgBCache
+            ${{ github.workspace }}\vcpkgBin
+            ${{ github.workspace }}\vcpkgBuild
```
BenjamenMeyer: `\v` might get translated as a vertical tab instead of the directory separator (`\`) and `v` as intended. Windows does support `/` which can prevent that without having to escape the `\`.

.github/workflows/Windows-Release.yml
```diff
@@ -74,13 +77,21 @@ jobs:
...
           key: vcpkgCache-${{ github.repository_owner }}-${{ matrix.os }}-${{ matrix.triplet }}-${{ hashFiles('vcpkg.json') }}
 
+      - name: Create directories on cache miss
+        run: |
+          mkdir -p ${{ github.workspace }}\vcpkgAssets
+          mkdir -p ${{ github.workspace }}\vcpkgBCache
+          mkdir -p ${{ github.workspace }}\vcpkgBin
+          mkdir -p ${{ github.workspace }}\vcpkgBuild
```
BenjamenMeyer: same

.github/workflows/Windows-Release.yml
```diff
@@ -116,11 +129,11 @@ jobs:
...
-          path:
-            ${{ github.workspace }}/vcpkg
-            ${{ github.workspace }}/vcpkgAssets
-            ${{ github.workspace }}/vcpkgBin
-            ${{ github.workspace }}/vcpkgBuild
+          path: |
+            ${{ github.workspace }}\vcpkgAssets
+            ${{ github.workspace }}\vcpkgBCache
+            ${{ github.workspace }}\vcpkgBin
+            ${{ github.workspace }}\vcpkgBuild
```
BenjamenMeyer: same

BenjamenMeyer (commented):
We use the `/` convention even on Windows so we don't have to escape the backslash ('\').

.github/workflows/Windows-Release.yml
stephengtuggy: I am aware of the potential issue with backslashes. In practice, this appears to be working fine, as long as it's not inside quotes.

.github/workflows/Windows-Release.yml
stephengtuggy: See above.

.github/workflows/Windows-Release.yml
stephengtuggy: See above

script/bootstrap.ps1
stephengtuggy: This is a PowerShell script. No need to escape backslashes in this context.

script/bootstrap.ps1
stephengtuggy: As above. This is a PowerShell script. No need to escape backslashes in this context.

stephengtuggy:
> I've reviewed up to fire.cpp.
> 
> I realise I haven't been able to contribute as much these past few months. It's a combination of a regional war and me doing a one semester certification, that sucked most of my free time.
> 
> I've taken a few minutes to review what I can. I don't see any blockers. Everything I commented on is either a question or nice to have.
> 
> Thank you for picking up the slack.

@royfalk You're welcome. Good to have you back.

.github/workflows/Windows-Release.yml
stephengtuggy: Went ahead and switched to forward slashes in the YAML files.

.github/workflows/Windows-Release.yml
stephengtuggy: Went ahead and switched to forward slashes in the YAML files.

.github/workflows/Windows-Release.yml
stephengtuggy: Went ahead and switched to forward slashes in the YAML files.

script/bootstrap.ps1
BenjamenMeyer: yeah.... it ends up being weird from what I was just reading since some contexts may still escape it while others don't, and they changed the escape sequence initiator from `\` to  the backtick (`) instead...

still, consistency wise it's probably easier to just stick with forward slashes (`/`).

BenjamenMeyer (approved):

script/bootstrap.ps1
stephengtuggy: OK, done

Issue #1611, opened 2026-05-26
Sometimes after loading a saved game and launching into space, ship barely accelerates in any direction


stephengtuggy:
@royfalk Assigning this to you, at least temporarily, as you seem to know more than I do about this particular issue. Can you provide more details?

See also #1609

Issue #369, opened 2020-12-17
Launcher Application


BenjamenMeyer:
We presently have two applications that get run and need context (location of game assets, configuration file) to run properly.
It would probably be helpful to add a launcher program that can allow the user to launch the configuration or game from a single spot; possibly provide support for multiple games at the same time.

This application doesn't need to be complex. It should be something simple with a GUI like the following:
```
+----------------------------------------------------------------------+
| Vega Strike Launcher                                         [_] [x] |
+----------------------------------------------------------------------+
| +----------------------------------------------------------------+   |
| | [VS:UtCS Icon]  [Config Push Button] [Launch Push Button]      |   |
| +----------------------------------------------------------------+   |
+----------------------------------------------------------------------+
```

Since we're using Gtk3 for the settings application this should be another Gtk3 application.

Loki1950:
There was such a launcher at one time.

BenjamenMeyer:
Cool; will have to take a look; also see how expandable/configurable it is as it would be nice to use it as a point to kick off/configure the different games.

P-D-E:
The old launcher can still be compiled after a little change; in `general.cpp` replace line 372
```
void ShowError(char *error_msg, char *error_code, int is_fatal) {
```
with
```
void ShowError(const char *error_msg, const char *error_code, int is_fatal) {
```
and then compile with
```
g++ -D_GLIBCXX_USE_CXX11_ABI=0 -I.. saveinterface.cpp general.cpp ../src/common/common.cpp -o saveinterface `pkg-
config --cflags --libs gtk+-2.0`
```
but then it tries to run the `vegastrike` executable (not `vegastrike-engine`) and there's no current support for the data directory argument. That's coherent with the old trick of running the game from inside the data directory, in fact this launcher looks for `vegastrike.config` only in the current directory and its parent.

BenjamenMeyer:
@P-D-E thanks for the analysis. One of the things we'd like to do is make this a bigger thing that can launch multiple games, so some refactoring will probably be needed but good to know it's very close. We should probably call it `vegastrike-launcher` in line with our new naming convention for libraries and executables.

P-D-E:
@BenjamenMeyer You're welcome, in the end you're imagining something like this:

![image](https://user-images.githubusercontent.com/23652942/121691267-197f2e00-cac7-11eb-965d-ec8ef2dd4ca6.png)

This part would handle VS:UtCS and PWCU for instance;

[...]

PR #1614, opened 2026-06-13
[WIP] Refactor/Random Number Generation Part 2

stephengtuggy:

Please answer the following:

Code Changes:

Issues Partially Addressed:
- Relates to #1081 
- Relates to #369 

Known Issues Remaining:
- #1625 
- #1611 
- #1632 

I wasn't able to test buying a ship, buying upgrades, or jumping between systems due to all the issues. I covered most of the rest of the test plan though.

Purpose:
- What is this pull request trying to do? Convert some more of the random number generation code -- code that uses Normal Distributions or other distributions other than Uniform. In the process, upgrade project to C++17, and also get the old vslauncher application working.


CMakeLists.txt
```diff
@@ -115,10 +115,10 @@ PROJECT(Vega_Strike
 
 MESSAGE("== Vega Strike Version: ${VEGASTRIKE_VERSION_LONG_STR}")
 
-SET(CMAKE_CXX_STANDARD 14)
+SET(CMAKE_CXX_STANDARD 17)
 SET(CMAKE_CXX_STANDARD_REQUIRED TRUE)
 SET(CMAKE_CXX_EXTENSIONS ON)
-SET(CMAKE_C_STANDARD 11)
+SET(CMAKE_C_STANDARD 17)
```
BenjamenMeyer, line 121: we should only have to set this one place and then everything ought to inherit it - so I think we're good here with removing the other lines like you did.

libraries/common/common.cpp
BenjamenMeyer, line 1: is this file movement correct? If it's not in `engine/src` (which probably should eventually just be `src`) then it should be in `libraries/common`; but having `common` in parallel to `engine`?

libraries/common/common.h
BenjamenMeyer, line 1: same with this file

libraries/common/enum_iterator.h
BenjamenMeyer, line 1: and this one

libraries/common/general.cpp
```diff
@@ -149,7 +149,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
     current = new char[LENGTH + 1];
     length = strlen(line);
     strcpy(current, line);
-    while ((location = strstr(current, search)) > 0) {
+    while ((location = strstr(current, search))) {
```
BenjamenMeyer, line 169: we really should avoid doing assignments in the conditional. While C/C++ allows it, it's heavily frowned on. While it does lead to a line of code duplication; I tend to do favor:
```
set_conditional = to_value;
while (check_condition_of(set_conditiona)) {
    ...
    set_conditional = to_value;
}
```
hmmm... we can probably get rid of these methods now and convert to using their counterparts on std::string.... probably should do that in a different task/PR. Even if we wanted to keep the functional C interface style we could do:
```
std::string strCurrent(current);
std::string strSearchFor(search);
std::string strReplaceWith(replace);
for (size_t location = strCurrent.find(strSearchFor, location); location != std::string::npos; location = strCurrent.find(strSearchFor, location) {
    strCurrent.replace(location, strReplaceWith.length(); strReplaceWith);
}
```
or something like that...

engine/mission/include/general.cpp
```diff
@@ -158,7 +158,7 @@ char *replace(char *line, char *search, char *replace, int LENGTH) {
     }
     length = strlen(line);
     strcpy(current, line);
-    while ((location = strstr(current, search)) > 0) {
+    while ((location = strstr(current, search))) {
```
BenjamenMeyer, line 169: yeah... more duplication

libraries/common/common.cpp
stephengtuggy: I figured I would put `common` outside of any existing directories, since it is shared between `libraries` and `engine` and their subdirectories. I guess I could move it to `libraries/common` instead. Although I don't know that it will be compiled to a separate library per se.

libraries/common/general.cpp
stephengtuggy: Yeah, I don't like these either. I can fix this.

engine/mission/include/general.cpp
stephengtuggy: I'm getting to that. One thing at a time. 😄 First, I wanted to get all the code compiling, including the `launcher` and `missions` subdirectories, if possible.

libraries/common/general.cpp
BenjamenMeyer: Filed #1626 about this

BenjamenMeyer:
I'm not sure we want to move everything from `engine/src` to `engine/` in this PR. That'll be quite disruptive to other work. I'm not objecting to restructuring - I'd like to see us do it actually - but that seems out of scope for this PR.

We would also need to do it to the 0.10.x branch to make cherry-picking between the two branches easier.

libraries/common/common.cpp
stephengtuggy: I think most of the subdirectories of `engine`, apart from `src`, should eventually be moved up to the top level anyway. Since they're not really part of `vegastrike-engine`. What do you think?

stephengtuggy:
> I'm not sure we want to move everything from `engine/src` to `engine/` in this PR. That'll be quite disruptive to other work. I'm not objecting to restructuring - I'd like to see us do it actually - but that seems out of scope for this PR.
> 
> We would also need to do it to the 0.10.x branch to make cherry-picking between the two branches easier.

I meant that I wanted to move everything **except** `src/` out of the `engine/` directory. And agreed that that should not happen in this PR, but in a future one.

libraries/common/common.cpp
stephengtuggy: Should `common` be its own, proper library? Like a static library? I guess it could be, actually.

BenjamenMeyer: > I think most of the subdirectories of `engine`, apart from `src`, should eventually be moved up to the top level anyway. Since they're not really part of `vegastrike-engine`. What do you think?

eventually yes, we need to reorganize the `engine` directory to put things more top level. We just need to do it in both 0.10.x and 0.11.x branches together so their structure matches. I'm fine with `src` moving to top level and all that.

> Should `common` be its own, proper library? Like a static library? I guess it could be, actually.

that makes sense IMHO

stephengtuggy: Okay. I think I will go ahead and make `common` its own library then. I'll probably put it in the directory `libraries/common/`, and name the library and the namespace `vega_common`.

engine/mission/include/general.cpp
stephengtuggy: OK, I have combined these files into one directory location. I don't have `missions` building yet -- I'm not immediately sure how to accomplish that -- but `launcher` is building.

libraries/common/common.cpp
stephengtuggy: Should be fixed

libraries/common/common.h
stephengtuggy: Fixed

libraries/common/enum_iterator.h
stephengtuggy: Fixed

engine/launcher/CMakeLists.txt
```diff
@@ -0,0 +1,148 @@
...
+IF (CMAKE_SYSTEM_NAME STREQUAL BEOS OR CMAKE_SYSTEM_NAME STREQUAL Windows)
+    MESSAGE ("SKIPPING vslauncher - not supported on this platform at this time")
+ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
+    MESSAGE(STATUS "Configuring vslauncher for Linux")
+
+    SET(LAUNCHER_SOURCES
+            saveinterface.cpp
+        )
+
+    ADD_EXECUTABLE(vslauncher ${LAUNCHER_SOURCES})
```
BenjamenMeyer, line 39: should use our convention of `vega-launcher` (matching most of our tools) or `vegastrike-launcher` (matching the engine)?

I don't think anyone is expecting the launcher to work by the old name, so we can just ship using our new convention.

engine/launcher/saveinterface.cpp
```diff
@@ -380,10 +389,10 @@ void hello(GtkWidget *widget, gpointer data) {
                 char pwd [65535];
                 getcwd(pwd,65533);
                 pwd[65533]=pwd[65534]='\0';
-                int pid=spawnl(P_NOWAIT,"./Setup.exe",(std::string(pwd)+"/Setup.exe").c_str(),NULL);
+                int pid=spawnl(P_NOWAIT,"./vegasettings.exe",(std::string(pwd)+"/vegasettings.exe").c_str(),NULL);
```
BenjamenMeyer, line 392: this probably should point at our new application; also it should take the path as an argument for the -`d` properly too

engine/launcher/saveinterface.cpp
```diff
@@ -380,10 +389,10 @@ void hello(GtkWidget *widget, gpointer data) {
...
                 if (pid==-1) {
                     if (chdir("bin")==0) {
-                        spawnl(P_NOWAIT,"./Setup.exe",(std::string(pwd)+"/bin/Setup.exe").c_str(),NULL);
+                        spawnl(P_NOWAIT,"./vegasettings.exe",(std::string(pwd)+"/bin/vegasettings.exe").c_str(),NULL);
```
BenjamenMeyer, line 395: same here - remember the old version just did the auto-detect side; we should update to use the explicit specification instead.

libraries/common/common.cpp
BenjamenMeyer: thanks!

BenjamenMeyer (commented):
couple ideas:

1. we should abstract out the platform dependent stuff so we have a class to launch the application and different implementations for Windows and Linux/Unix/macOS so the code is cleaner instead of having headers and lots of IFDEF preprocessor blocks
2. Let's make sure we call the executables with explicit parameters (like `vegastrike-engine` requires) instead of allowing discovery (like the legacy `vegastrike` application does). This should allow us to have more control and a better operation here.
3. We should have a configuration for the launcher that can take: (a) the assets directory, (b) points to the configuration file, (c) an icon image for each game that can be launched; and then at a global level the engine and settings binaries.
4. We should update it to use the new settings application; perhaps manage a virtualenv for it as well.

If we do this right, it'll make for a well run launcher that can easily launch all our various games, prove our our command-line parameters, and make for a good user experience in the process.

I don't think we should necessarily do all of this in this one PR - the refactor (pt 1) should probably be a separate PR, as probably should be points 3 and 4. Pt 2 should probably go in here at least for using the `vegastrike-engine` call instead of using the legacy `vegastrike` call. (0.11.x is probably a good time to drop the legacy `vegastrike` engine binary, btw).

engine/launcher/saveinterface.cpp
stephengtuggy: I'll be frank: I don't like our new Kivy settings application. It's ugly; the buttons only register clicks in the text section near the middle; and important settings are hidden behind page upon page of categories. I think we need to redo the settings app yet again.

gtkmm 4 might be a good choice. It's a class-oriented, type-safe wrapper on top of GTK 4. Or maybe wxWidgets, like someone else suggested. Or maybe ImGui. Regardless, I think it needs to be written in C++ -- and use the same SDL version as the game engine -- so that the settings app uses the exact same logic to detect screens and screen sizes as the game itself does.

BenjamenMeyer: @stephengtuggy I won't disagree that it still needs help and its not necessarily as intuitive as the original tool; but the original tool doesn't work any more and it mostly does. Until we have something better we should use it.

stephengtuggy: Fair enough.

stephengtuggy: Update: Now we have an in-game settings UI! Yay!

PR #1615, opened 2026-06-13
0.10.x CI: Omit Windows 2025 VS2022 variant…

stephengtuggy:
…. No longer available. Also add Ubuntu 26.04 "resolute", Linux Mint 22.3, and Fedora 44.

Please answer the following:

Code Changes:

Issues:
N/A

Purpose:
- What is this pull request trying to do? Stop trying to build on Windows 2025 with Visual Studio 2022 for CI purposes. This combination is no longer available from GitHub Actions.


engine/CMakeLists.txt
```diff
@@ -965,30 +965,30 @@ ELSEIF (CMAKE_SYSTEM_NAME STREQUAL Linux)
                         SET(BOOST_VER "${Boost_DERIVED_VERSION}")
                     ENDIF (NOT DEFINED Boost_DERIVED_VERSION)
                     SET(CPACK_DEBIAN_PACKAGE_DEPENDS "${CPACK_DEBIAN_PACKAGE_DEPENDS}, libjpeg62-turbo, libpng16-16, freeglut3, libgtk-3-0, libvorbis0a, libopenal1, libsdl2-gfx2.0.14, libsdl2-image-2.0-0, xdg-utils, libgl1, libopengl0, ${VEGA_STRIKE_GLU_DEPENDENCY}, libboost-atomic${BOOST_VER}, libboost-chrono${BOOST_VER}, libboost-date-time${BOOST_VER}, libboost-filesystem${BOOST_VER}, libboost-log${BOOST_VER}, libboost-python${BOOST_VER}, libboost-regex${BOOST_VER}, libboost-thread${BOOST_VER}, libboost-json${BOOST_VER}, libboost-program-options${BOOST_VER}")
-                ELSEIF (LSB_LINUX_DISTRIBUTION_CODENAME STREQUAL "wilma")  # Linux Mint 22 codename "wilma"
+                ELSEIF (LSB_LINUX_DISTRIBUTION_CODENAME STREQUAL "wilma" OR LSB_LINUX_DISTRIBUTION_CODENAME STREQUAL "zara" OR LSB_LINUX_DISTRIBUTION_CODENAME STREQUAL "zena")  # Linux Mint 22 codename "wilma", Linux Mint 22.2 codename "zara", Linux Mint 22.3 codename "zena"
```
BenjamenMeyer, line 968: instead of doing all these `STREQUAL` operations; it might be good to move to the `LIST(FIND...)` operation:
```
// wherever we want to maintain the list of supported distros
SET(LINUX_MINT_DISTROS, "zara", "zena")
...
// before the checks
IS_LINUX_MINT=LIST(FIND, LINUX_MINT_DISTROS, "${LSB_LINUX_DISTRIBUTION_CODENAME}")
...
// at the check
ELSEIF(${IS_LINUX_MINT} != -1)
...
```
This would be a major refactor, but might be worth it; might make it easier to update too.

stephengtuggy: That's a good suggestion. Although these are specifically the Linux Mint 22.x versions. So the code might end up more like:

```cmake
// wherever we want to maintain the list of supported distros
SET(LINUX_MINT_22_X_DISTROS "")
LIST(LINUX_MINT_22_X_DISTROS APPEND "wilma")
LIST(LINUX_MINT_22_X_DISTROS APPEND "zara")
LIST(LINUX_MINT_22_X_DISTROS APPEND "zena")
...
// before the checks
IS_LINUX_MINT_22_X=LIST(FIND, LINUX_MINT_22_X_DISTROS, "${LSB_LINUX_DISTRIBUTION_CODENAME}")
...
// at the check
ELSEIF(${IS_LINUX_MINT_22_X} != -1)
...
```

@BenjamenMeyer what do you think?

stephengtuggy: [Looks like](https://cmake.org/cmake/help/latest/command/list.html#introduction) the list initialization could also be done this way:

```cmake
SET(LINUX_MINT_22_X_DISTROS wilma zara zena)
```

I like that a lot better.

BenjamenMeyer: looks good - not saying it needs to be in this PR but something we should consider.

BenjamenMeyer (approved):

engine/CMakeLists.txt
stephengtuggy: Filed as #1616

Issue #1618, opened 2026-06-18
Localization issues with saved games, among other things


SGD1953:
I am playing on latest master and when a game is saved it uses `.` as decimal separator in `player_fleet.json` which then crashes on loading the save. I manually patched the save and replaced all `.` with `,` (as it is in `New_Game`) which loads fine. This might be a possible localization issue.

I'm on Linux/Ubuntu with a German locale setting

Steps to reproduce:

1. Start new game
2. Save the game
3. Load the saved game
4. Crash

stephengtuggy:
Thanks for the bug report. Yes, that sounds like a localization issue, all right.

SGD1953:
I'd be happy to contribute/test, the latter is probably the better starting point.

SGD1953:
Did some more testing and its a bit non-deterministic. I managed to modify `player_fleet,json` to make it load, then I quit and then I could not load that very file anymore.

I decided to start a new game, saved it, this time it actually did load. Bought some cargo, flew to mining base, now the base is all black (but I see the ship). From previous session I knew that this is the end and the game will eventually crash. IN the console I got

```
shield_front 150,00/150,00/150,00
Reverted unit_key to player_ship_0.templateReverted unit_key to player_ship_0.blankError: WrapObject raw_object is null
Traceback (most recent call last):
  File "/home/user/vegastrike/Assets-Production/python/base_computer/upgrade_view.py", line 305, in get_upgrade_info
    text += process_line(line, unit)
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/vegastrike/Assets-Production/python/base_computer/upgrade_view.py", line 205, in process_line
    value = process_resource(key, str(unit[key]))
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/vegastrike/Assets-Production/python/base_computer/upgrade_view.py", line 149, in process_resource
    damaged = float(parts[damaged_index])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: '150,00'
Error in buildUpgradeDescription: Error: WrapObject raw_object is null
Key player_ship_0
[...]
Error: WrapObject raw_object is null
Traceback (most recent call last):
  File "/home/user/vegastrike/Assets-Production/python/base_computer/upgrade_view.py", line 299, in get_upgrade_info
[...]

PR #1629, opened 2026-06-27
Commented locale changes which break locales using comma decimal sepa…

SGD1953:
…rator

Thank you for submitting a pull request and becoming a contributor to Vega Strike: Upon the Coldest Sea.

Please answer the following:

Code Changes:

Issues:
- Fixes issue #1618

Purpose:
This is a fix for the game being broken on locales not using `.` as decimal separator

### DIFF
--- a/engine/src/cmd/basecomputer.cpp
+++ b/engine/src/cmd/basecomputer.cpp
@@ -443,12 +443,12 @@ std::string prettyPrintFloat(float value, int precision = 2) {
     // TODO: Need to place this somewhere else
     // Call this once if you want your whole program to follow system locale
 
-    std::locale::global(std::locale(""));
+    //std::locale::global(std::locale(""));
 
     std::ostringstream out;
 
     // Use the user's global locale (setlocale/std::locale::global)
-    out.imbue(std::locale(""));
+    //out.imbue(std::locale(""));
 
     // Fixed n digits after decimal + locale separators
     out << std::fixed << std::setprecision(precision) << value;
--- a/engine/src/python/unit_wrapper_class.h
+++ b/engine/src/python/unit_wrapper_class.h
@@ -35,7 +35,7 @@ using std::string;
 static const char *kNullUnitErrorMsg = "\nERROR: NULL Unit used in Python script; returning default value...";
 #define CHECKME \
     }           \
-    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, (std::string(kNullUnitErrorMsg) + " in " + __FUNCTION__).c_str()); return
+    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, (boost::format("%1%: %2%") % __FUNCTION__ % kNullUnitErrorMsg)); return
 #define WRAPPED0(type, name, def) \
     type name() {                   \
         {CHECKME def; }             \
--- a/engine/src/python/unit_wrapper_class.h
+++ b/engine/src/python/unit_wrapper_class.h
@@ -35,7 +35,7 @@ using std::string;
 static const char *kNullUnitErrorMsg = "\nERROR: NULL Unit used in Python script; returning default value...";
 #define CHECKME \
     }           \
-    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, kNullUnitErrorMsg); return
+    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, (std::string(kNullUnitErrorMsg) + " in " + __FUNCTION__).c_str()); return
 #define WRAPPED0(type, name, def) \
     type name() {                   \
         {CHECKME def; }             \
--- a/engine/src/main.cpp
+++ b/engine/src/main.cpp
@@ -248,6 +248,9 @@ int main(int argc, char *argv[]) {
     const boost::filesystem::path program_path(argv[0]);
     // const boost::filesystem::path canonical_program_path = boost::filesystem::canonical(program_path);
 
+    // set standard C locale (avoids side effects when the default locale is different)
+    std::locale::global(std::locale("C.UTF-8"));
+
     const boost::filesystem::path program_name{program_path.filename()};  //canonical_program_path.filename();
     const boost::filesystem::path program_directory_path{program_path.parent_path()};
 
--- a/engine/src/python/init.cpp
+++ b/engine/src/python/init.cpp
@@ -199,6 +199,7 @@ void Python::init() {
     PyConfig_InitPythonConfig(&config);
 
     config.isolated = 1;
+    config.coerce_c_locale = 2;
 
     // Now we can do python things about them and initialize them
     Py_Initialize();
--- a/engine/src/python/init.cpp
+++ b/engine/src/python/init.cpp
@@ -177,6 +177,7 @@ void Python::init() {
     PyPreConfig_InitPythonConfig(&py_pre_config);
 
     PyStatus status;
+    py_pre_config.coerce_c_locale = 2;
 
     status = Py_PreInitialize(&py_pre_config);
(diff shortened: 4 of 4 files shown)


engine/src/python/unit_wrapper_class.h
```diff
@@ -35,7 +35,7 @@ using std::string;
 static const char *kNullUnitErrorMsg = "\nERROR: NULL Unit used in Python script; returning default value...";
 #define CHECKME \
     }           \
-    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, kNullUnitErrorMsg); return
+    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, (std::string(kNullUnitErrorMsg) + " in " + __FUNCTION__).c_str()); return
```
SGD1953: Not technically part of the fix but adds the function name to the errors like `ERROR: NULL Unit used in Python script; returning default value... in getFullname` which should help to better pinpoint issues

stephengtuggy: Sounds good. You don't need the `.c_str()` part though

engine/src/cmd/basecomputer.cpp
```diff
@@ -440,12 +440,12 @@ std::string prettyPrintFloat(float value, int precision = 2) {
     // TODO: Need to place this somewhere else
     // Call this once if you want your whole program to follow system locale
 
-    std::locale::global(std::locale(""));
+    //std::locale::global(std::locale(""));
```
stephengtuggy, line 446: Again, I would prefer to see this changed to something like:


```suggestion
    std::locale::global(std::locale("C.UTF-8"));
```

engine/src/python/unit_wrapper_class.h
stephengtuggy: The way we've been accomplishing this elsewhere in the code is more like:


```suggestion
    Unit *me = GetUnit(); if (!me) {VS_LOG_AND_FLUSH(error, (boost::format("%1%: %2%") % __FUNCTION__ % kNullUnitErrorMsg)); return
```

stephengtuggy: But that's all temporary until we make __FUNCTION__ an automatic part of each logged message.

engine/src/cmd/basecomputer.cpp
SGD1953: That's the default though as per my understanding, why would we explicitly set the default locale for C? If we were to set it to make it explicit, should it not be at startup?

stephengtuggy: It isn't the default on my system.

Yes, we should probably set it at startup.

SGD1953: Ok, what do we do here then? I still feel that those mid-game changes are introducing hard to triage issues. The game would start for me and then depending what I did crash or run for a bit longer and crash later

stephengtuggy: I agree. Mid-game locale changes are a bad idea. Let's set the locale at startup, probably in function `main` in engine/src/main.cpp.

It's not quite as clear to me what to do about the locale for purposes of the Python code. Maybe we can set it in the `Python::init()` function in engine/src/python/init.cpp. Then hopefully that locale will take effect for the Python code for the duration of the game.

SGD1953: As part of this PR or separately?

stephengtuggy: As part of this PR, please.

stephengtuggy: Setting the locale should probably be one of the first things we do in function `main`. Maybe even the very first thing.

SGD1953: Ok, done. Lauched the game, saved, load -> I was in space. Ended game, loaded -> all good, saved again, loaded -> all good. Checked the files, all with `.` as separator.

A long way to say that it seemingly works, the "load into space" issue might be unrelated.

stephengtuggy: Sounds good.

Where were you docked when you saved your game? Were you at a temporary station (name shows up as "SS ...") by any chance?

SGD1953: Yes, I was docked at Atlantis. I quit and reloaded the game and I was properly at the base.

stephengtuggy (commented):
Looking good so far! I want to test the changes on my system before I give my final approval.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: re:Python - we should probably set it when we initialize the Python Engine. That should cover any script loaded after that.

SGD1953: Could you point me to the proper entry point please?

SGD1953:
> Looking good so far! I want to test the changes on my system before I give my final approval.

@stephengtuggy  Once your testing has finished, we should backport this change to branch 0.10.x. We should also consider having a corresponding branch in Assets-Production as it does not just contain assets.

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: @SGD1953 it should be this block: https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/python/init.cpp#L175-L205

I'm not sure exactly where in that specifically; but that should be the primary location we initialize Python from.

Python Related documentation for the methods we use to kick stuff off:
https://docs.python.org/3/c-api/interp-lifecycle.html#c.Py_Initialize
https://docs.python.org/3/c-api/interp-lifecycle.html#c.Py_InitializeFromConfig
https://docs.python.org/3/c-api/veryhigh.html#c.PyRun_SimpleString
https://docs.python.org/3/c-api/veryhigh.html#c.PyRun_SimpleFile

SGD1953: Ok, I'll add `config.coerce_c_locale = 2;` in there [which should  tell python to use what C uses](https://docs.python.org/3/c-api/init_config.html#c.PyPreConfig.coerce_c_locale). Would you conceptually agree?

BenjamenMeyer: @SGD1953 sounds good; it would be great if we can unify our C and Python spaces that way

SGD1953: @BenjamenMeyer this ended up a bit messy, but I accidentally committed with the line in the wrong place (Config instead of PreConfig). This is now corrected and does compile (for me)

stephengtuggy (approved):
Looks good to me!

stephengtuggy:
The only **potential** new issue I saw during testing was that I couldn't buy Image Recognition or Heat Seeker missiles and put them on Auto-Tracking enabled mounts. That may not be a new issue, however. And it may not even be a bug.

Overall, I think this is good to go, along with its accompanying Assets-Production PR. Let's get them merged!

SGD1953:
Missle refill is not really working for me either, I can only sell and Rebuy.

stephengtuggy:
> Missle refill is not really working for me either, I can only sell and Rebuy.

Oh, shoot. Then we may have an actual bug or two related to purchasing/refilling missiles. Better file it/them and try to get them fixed.

stephengtuggy:
> Missle refill is not really working for me either, I can only sell and Rebuy.

Filed as #1643

stephengtuggy:
> The only **potential** new issue I saw during testing was that I couldn't buy Image Recognition or Heat Seeker missiles and put them on Auto-Tracking enabled mounts. That may not be a new issue, however. And it may not even be a bug.
> 
> Overall, I think this is good to go, along with its accompanying Assets-Production PR. Let's get them merged!

Filed as #1644

PR #1639, opened 2026-07-11
Upgraded import and export plugins  to work with latest blender version

SGD1953:

Please answer the following:

Code Changes:

Purpose:
This migrated the Blender import and export plugins to the latest version. The have been gradually refined as part of https://github.com/vegastrike/Assets-Masters/issues/100

### DIFF
--- a/engine/objconv/blender_xmesh_export.py
+++ b/engine/objconv/blender_xmesh_export.py
@@ -1,5 +1,3 @@
-#!BPY
-##
 # blender_xmesh_export.py
 #
 # Copyright (C) dandandaman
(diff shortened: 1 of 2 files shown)


SGD1953:
@stephengtuggy @royfalk @BenjamenMeyer please have a look at this one, it should be relatively low-risk as the plugins simply don't work in the current state.

royfalk:
I agree this is low risk. I do have a couple of questions:
1. Did you test it? both ways on several meshes?
2. I see that xmesh is a. a proprietary file format from Amazon b. what we use. Need to check that Gemini did not implement the wrong file format.

SGD1953:
I did test this, both ways.

[This PR](https://github.com/vegastrike/Assets-Masters/pull/101/changes) imports all ship meshes (which are the most complex ones) into blender and renders HUD images which gives me a high confidence on the import. I have not validated all ships after export, but I did try Llama.

> I see that xmesh is a. a proprietary file format from Amazon b. what we use. 

Interesting

> Need to check that Gemini did not implement the wrong file format.

Not sure I understand, the code here - while implemented with AI assistance -  has been verified and tested.

engine/objconv/blender_xmesh_export.py
```diff
@@ -8,268 +6,194 @@
...
+        mat = obj.data.materials[0]
+        if mat.use_nodes:
+            for node in mat.node_tree.nodes:
+                if node.type == 'TEX_IMAGE' and node.image:
+                    if "NORMAL" in node.image.name.upper() or node.image.colorspace_settings.name == 'Non-Color':
+                        texture4 = node.image.name
+                    else:
+                        texture = node.image.name
+
+    with open(filepath, "w", encoding="utf-8") as f
```
royfalk: Missing : at the end?

engine/objconv/blender_xmesh_export.py
```diff
@@ -8,268 +6,194 @@
...
+		# Safely extract and validate scale tuple/list length
+        scale = obj.get("xmesh_scale", 1.0)
```
royfalk: Did you intend to overwrite scale variable defined above?

royfalk (commented):
There's not much I can add here, but I did ask both Claude and Gemini to review this.

**Claude**

1. scale is fetched twice, incompatibly. At the top: scale = obj.get("xmesh_scale", (1.0, 1.0, 1.0)) (vector default). Inside the with block it's silently reassigned: scale = obj.get("xmesh_scale", 1.0) and then formatted as f'scale="{scale:.6f}"'. If xmesh_scale was actually stored as a 3-vector (as the first line implies it might be), {scale:.6f} will throw TypeError since you can't format an IDPropertyArray/tuple with .6f. The first assignment is dead code — remove it or actually use a per-axis scale in the output.
2. UV-to-vertex mapping is mislabeled and effectively unused. The comment says "Map vertices to their first encountered UV coordinates," but the loop just overwrites vert_uvs[vert_idx] every time a shared vertex is visited — so it's actually last encountered, not first. More importantly, this per-vertex UV written into <Point> is redundant/wrong for any vertex shared across UV seams, since the correct per-loop UV is separately (and correctly) written per-vertex-per-polygon in the <Polygons> block. This dead/misleading computation should probably just be deleted.
3. Only one object is exported, despite the tool claiming to export "selected objects." bl_info["description"] and the docstring say "selected objects," but execute() picks a single object (active mesh, or first selected mesh) and exports only that one. If a user selects multiple meshes expecting a combined or per-object export, they'll silently get just one file with no batch/merge behavior and no warning that other selected objects were skipped.
4. Per-vertex normals average across UV/shading seams, since normals are read straight from mesh.vertices[i].normal rather than per-loop (loop.normal, requires mesh.calc_normals_split() or corner_normals in 4.x). This will visibly flatten/smooth hard edges and custom split normals on any mesh that has sharp edges or custom normals — likely not what's intended for game export.
5. No triangulation. Ngons (5+ sides) are written as <Trifan> purely based on vertex count, but the vertex loop just dumps polygon loop vertices in original winding order — it doesn't actually fan-triangulate or verify convexity/planarity. If Vega Strike's loader expects genuine triangle-fan data or triangulated geometry, non-convex or non-planar ngons will render incorrectly. Consider using a Triangulate modifier via the depsgraph or bmesh before export.
6. Texture auto-detection is fragile. It scans all nodes in the material's node tree for TEX_IMAGE, not the nodes actually connected to the shader's outputs. Unlinked/unused image nodes, multiple diffuse textures, or non-standard node graphs can cause the wrong image to be picked, and later matches silently overwrite earlier ones (last-node-wins, node order isn't guaranteed to match visual layering).
7. No exception safety around to_mesh_clear(). If writing the file throws partway through, obj_eval.to_mesh_clear() is never called, leaking the evaluated mesh for the rest of the Blender session. Wrap the write in try/finally.
8. Type assumptions on custom properties. Several values (polygonoffset, alphatest, mat_power) are format-string'd with .6f, assuming they're numeric. If a user (or a prior import step) ever stored these as strings, this throws at export time with a fairly opaque traceback. Consider explicit float(...) coercion with a friendly error, or validate types up front.

**Gemini**
1. Syntax Error: Missing Colon on with openLine: with open(filepath, "w", encoding="utf-8") as fIssue: There is a missing colon (:) at the end of the with statement. Python will throw an immediate SyntaxError and refuse to register or execute the add-on.2. Tab/Space Indentation MixtureIssue: The lines inside the with open block use tabs for indentation, whereas the rest of the script uses 4 spaces. Mixing tabs and spaces in Python 3 throws an immediate IndentationError.3. Scale Overwrite & Type CrashIssue: At the top of the function, the script safely reads scale as an expected tuple/list: scale = obj.get("xmesh_scale", (1.0, 1.0, 1.0)). However, right inside the file writing loop, it overwrites it: scale = obj.get("xmesh_scale", 1.0).The Crash: If xmesh_scale is an actual 3-element tuple (1.0, 1.0, 1.0) stored on the object, passing it to a single float format string f"{scale:.6f}" will throw a TypeError: unsupported format string passed to tuple.__format__.4. The Ngon / "Trifan" Fallback TrapIssue: When writing polygon data, the script uses a fallback string: p_type = "Tri" if ... else "Quad" if ... else "Trifan".The Gotcha: If a user forgets to triangulate their model and exports an N-gon (a face with 5 or more vertices), the script labels it <Trifan>. However, Blender's native vertex indexing for an N-gon loop is not inherently formatted as an OpenGL Triangle Fan. This will output malformed geometry that will distort or crash inside Vega Strike.

royfalk:
I would also note that we should drop xmesh and move to blender files or whatever is the industry standard natively.

SGD1953:
@royfalk thank you for the thorough review, I addressed the clear issues around the missing colon and the redefined variable. I think it worked because the original wrong definition always got overwritten, but clearly not good code.

There are many more good callouts which are probably fair, the aspiration here was not to make it perfect though but to make it work - which it does. I just deleted my Blender plugins, re-installed them (that's necessary for every change you make) and then imported a xmesh model and exported it into a different file which yielded an almost empty diff ( reflect="0.0"` vs. `reflect="0"` for Determinant). It is by no means perfect, it is a migration of the original plugins (which I assumed where built by people who understand xmesh) to the latest Blender API version.

Regarding not using xmesh, I agree with teh direction of travel here, but I feel we first need to make the engine ready to accept i.e. glTF which would help to speed things up. This needs to be a concerted effort though to get the engine ready first and then migrate the existing models (maybe the engine could support both for backward compatibility. The nice thing about glTF is that it can be loaded directly into the VRAM like DDS textures by openGL (or so I'm told). What it does not have at the moment is support for LODs and maybe other settings BFXM/XMESH provide so that we would need some sidecar JSON to complement it. This would still be an elegant solution and solve multiple issues at once:

1. Leverage an open standard
2. Improved loading of models potentially allowing us to delete a lot of legacy code
3. No more need to compile anything, the "production" format can be directly edited in Blender

BenjamenMeyer (approved):

stephengtuggy:
@SGD1953 if you could address Claude's review item number 7 (wrapping the file write in a try...finally block) and Gemini's review item number 2 (mixed tabs and spaces for indentation) I think we'll be good to go as far as I'm concerned.

SGD1953:
@stephengtuggy @royfalk implemented the suggestions, tested with an existing model. The new version produces slightly different output which however may not be a bad thing as the calculation in the past may have been a bit brute force creating sharp edges.

Tested it, it still works both ways.

engine/objconv/blender_xmesh_export.py
```diff
@@ -8,268 +6,195 @@
...
 # (at your option) any later version.
 #
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <https://www.gnu.org/licenses/>.
-#
```
stephengtuggy, line 24: Why did you remove this paragraph? This paragraph is supposed to be present in any code covered by the GPL, if I'm not mistaken.

engine/objconv/blender_xmesh_export.py
```diff
@@ -8,268 +6,195 @@
...
-
-
-"""
-Name: 'VegaStrike (selected) (.xmesh)...'
-Blender: 228
-Group: 'Export'
-Tooltip: 'Export VegaStrike (selected) (.xmesh)'
-"""
```
stephengtuggy, line 32: I think it would be good to keep this doc comment for informational purposes.

engine/objconv/blender_xmesh_export.py
```diff
@@ -8,268 +6,195 @@
...
-from __future__ import print_function
-try:
-    range = xrange
-except:
-    pass
-
-__author__	= "Dandandaman"
-__url__		= ("blender", "http://www.blender.org", "Author's homepage, http://black-paralysis.sf.net/")
-__version__	= "0.1"
```
stephengtuggy, line 41: I think we should also keep the author and URL here, though the version can be incremented.

stephengtuggy: Oh, never mind. I see that you do identify the author, further down.

engine/objconv/blender_xmesh_export.py
stephengtuggy: Looks like this is covered by the file rewrite. Never mind. Thanks!

engine/objconv/blender_xmesh_import.py
```diff
@@ -1,285 +1,400 @@
...
 # of the License, or (at your option) any later version.
 #
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, see
-# <https://www.gnu.org/licenses/>.
```
stephengtuggy, line 24: This paragraph needs to stay here as well.

stephengtuggy:
> @stephengtuggy @royfalk implemented the suggestions, tested with an existing model. The new version produces slightly different output which however may not be a bad thing as the calculation in the past may have been a bit brute force creating sharp edges.
> 
> Tested it, it still works both ways.

@SGD1953 can you elaborate further? Which model did you test with? What were the differences? If N-gons are being corrected or the like, differences might not necessarily be a bad thing. But we should be careful not to introduce incorrect changes.

engine/objconv/blender_xmesh_import.py
SGD1953: I did not really remove this, I essentially created a new file and used the standard header. I'll have a look at both file headers and will restore them.

SGD1953:
@stephengtuggy I can indeed!

In https://github.com/vegastrike/Assets-Masters/pull/101 I use the import plugin to import all `XMesh` vessel models into Blender to then render the the HUD images. That means that the import works for all models, for the export i spot checked a number of models.

If you can spare the time, I'd appreciate if someone would actually try this. You need 3 ingredients:

1. The plugins from this PR
2. The above PR with all `XMesh` models or alternatively you generate them yourself from the `BFXM` files in Production
3. Blender

This would not only help to validate this but also to spot anything in missed in terms of documentation.

engine/objconv/blender_xmesh_import.py
SGD1953: @stephengtuggy  Restored the missing header parts and some more comments

engine/objconv/blender_xmesh_export.py
stephengtuggy: Much better, thanks.

engine/objconv/blender_xmesh_import.py
stephengtuggy: Much better, thanks.

stephengtuggy (approved):
The code looks good. I will see if I can find the time to test it sometime this week. If not, I'd say it can be merged by Sunday.

royfalk:
Let's get this merged!

PR #1641, opened 2026-07-12
Fix 0.10.0 alpha installers

### LINKED ISSUE #1640: Windows Release build(s) failing
The GitHub Actions Windows 2025 images [do not include NSIS](https://github.com/actions/runner-images/issues/11754). We will need to install it ourselves as part of the installer workflow, I guess.

stephengtuggy:

Please answer the following:

Code Changes:
- [x] This is a CI/build system change only, and will require cutting an installer to test

Issues:
- Fixes #1640 for 0.10.x (hopefully)

Purpose:
- What is this pull request trying to do? Fix the Windows 2025 build failure due to NSIS not being installed on the GitHub runner image. Make other changes as necessary to fix the installer builds.

### DIFF
--- a/script/package
+++ b/script/package
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 ##
-# package
+#====================================
+# @file   : package
+# @brief  : package Vega Strike installer
+# @usage  : ./script/package --preset-name=linux-unix-makefiles-pie-enabled-glvnd-RelWithDebInfo
+# @param  : preset-name=, build-type=
+#====================================
 #
 # Vega Strike - Space Simulation, Combat and Trading
 # Copyright (C) 2001-2026 The Vega Strike Contributors:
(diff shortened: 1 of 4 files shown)


SGD1953 (approved):

script/package
```diff
@@ -184,28 +219,23 @@ function copyFiles()
...
-	make package_source
-}
+	# before running the package step, regen the CMake data so that it picks up the dependencies just generated
+	cmake --preset "$preset_name"
 
-echo "----------------------------"
-echo "--- package | 2020-12-13 ---"
-echo "----------------------------"
+	# Run the actual package step
+	cpack -V --preset "package-$preset_name"
```
BenjamenMeyer, line 226: advantage of `make package` is that it runs all the prerequisite steps before doing the build so the build system is validated before we try to package stuff. Otherwise, yes

script/package
```diff
@@ -184,28 +219,23 @@ function copyFiles()
...
 function buildPackages()
 {
-	# before running make package regen the Cmake data so that
-	# it picks up the dependencies just generated
-	cmake "${ROOT_DIR}"
-	make package
-	make package_source
-}
+	# before running the package step, regen the CMake data so that it picks up the dependencies just generated
+	cmake --preset "$preset_name"
```
stephengtuggy, line 223: This line `runs all the prerequisite steps before doing the build`. So `make package` is not needed down below.

script/package
stephengtuggy: Line 223 takes care of that. The CPack command also has the advantage of following the package preset's settings, as applicable.

BenjamenMeyer: thanks for confirmation on that; I was thinking that would likely be the case but didn't want to presume

BenjamenMeyer (approved):

PR #1660, opened 2026-07-16
Fixed font point config and migrated base computer to ImGUI

### LINKED ISSUE #671: Font size needed for ship/game
VSengine 0.8.0 after running for first time set to system resolution I found the various text in game were too small to be unreadable

SGD1953:
Code Changes:

Issues:
This re-enables the possibility to configure font_point in the config. I tested with 32 (up from 16 which was hard-coded) and it looked much better.

I now managed to find a solution for the base computer as well. In principle we should probably rewrite it from scratch using ImGUI components, this is a larger untertaking however.

possibly addresses #671

### DIFF
--- a/engine/src/cmd/base_util.cpp
+++ b/engine/src/cmd/base_util.cpp
@@ -888,12 +888,13 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
-
+//  Python accessor, not used in cpp
 float GetTextHeight(std::string text, Vector widheimult) {
     const float font_point = configuration().graphics.font_point_flt;
     return font_point * 2 / configuration().graphics.resolution_y;
 }
 
+//  Python accessor, not used in cpp
 float GetTextWidth(std::string text, Vector widheimult) {
     //Unsupported for now
     return 0;
--- a/engine/src/cmd/base_util.cpp
+++ b/engine/src/cmd/base_util.cpp
@@ -888,11 +888,10 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
+
 float GetTextHeight(std::string text, Vector widheimult) {
-    static constexpr bool force_highquality = true;
-    constexpr bool use_bit = force_highquality ||  configuration().graphics.high_quality_font;
     const float font_point = configuration().graphics.font_point_flt;
-    return use_bit ? getFontHeight() : (font_point * 2 / configuration().graphics.resolution_y);
+    return font_point * 2 / configuration().graphics.resolution_y;
 }
 
 float GetTextWidth(std::string text, Vector widheimult) {
--- a/engine/src/gui/font.cpp
+++ b/engine/src/gui/font.cpp
@@ -4,8 +4,8 @@
  * Vega Strike - Space Simulation, Combat and Trading
  * Copyright (C) 2001-2026 The Vega Strike Contributors:
  * Project creator: Daniel Horn
- * Original development team: As listed in the AUTHORS file. Specifically: Mike Byron
- * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
+ * Original development team: As listed in the AUTHORS file
+ * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy, Danny Gehl
  *
  * https://github.com/vegastrike/Vega-Strike-Engine-Source
  *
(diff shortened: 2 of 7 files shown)


libraries/gui/gui.cpp
```diff
@@ -45,7 +45,7 @@ bool gui_initialized = false;
 SDL_Window* current_window = nullptr;
 ImFont* roboto_18_font;
 
-void InitGui(SDL_Window *window, const SDL_GLContext *context) {
+void InitGui(SDL_Window *window, const SDL_GLContext *context, float fontSize) {
```
stephengtuggy: `fontSize` could be a `const`. Not essential, but nice to have.

SGD1953: I can change that, how about adding the font name as well. This is currently configurable in the config but has no effect when glut is no longer used

stephengtuggy: How about if we wait on the font name? That can be a future feature addition. @royfalk ran into some problems with the same font not being available across different OSes. So IIRC, he just went with the ImGui default font in all cases for now.

stephengtuggy: Or the default fixed-width font, or something like that.

SGD1953: @stephengtuggy made it a `const`, also managed to migrated the base computer to use the new fonts via a facade for the legacy glut font.

engine/src/gui/font.cpp
```diff
@@ -24,154 +24,132 @@
...
-    const bool tmp = configuration().graphics.high_quality_font_computer;
-    return !tmp;
+void Font::calcMetricsIfNeeded(void) const {
+    if (m_needMetrics) {
+        const_cast<Font*>(this)->calcMetrics();
+    }
 }
 
-//Calculate the metrics for this font.
-//This does the real work, and doesn't check whether it needs to be done.
```
BenjamenMeyer, line 55: we should keep function documentation

engine/src/gui/font.cpp
```diff
@@ -24,154 +24,132 @@
...
-    //space char width to the space char width we want.
-    const double eightWidth = glutStrokeWidth(GLUT_STROKE_ROMAN, '8') + m_extraCharWidth;
-    m_spaceCharFixup = eightWidth - glutStrokeWidth(GLUT_STROKE_ROMAN, SPACE_CHAR);
+    // Determine how much we need to scale the font 
+    // to reach the desired 'm_size' (0.1 = 10% of the screen height).
+    float scale_factor = m_size / configuration().graphics.font_point_dbl;
+
+    // Vertical Scaling: 
+    // We adjust by the configured GUI scale factor.
+    m_verticalScaling = scale_factor * (float)configuration().graphics.gui_font_scale_flt;
```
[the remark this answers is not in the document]
BenjamenMeyer: please cast correctly using `dynamic_cast<>`, `recast<>`, `static_cast<>`, or `const_cast<>` as appropriate. the old `(type)var` style isn't encouraged any more.

`m_verticalScaling` is a double so this should be:
```c++
m_verticalScaling = static_cast<double>(scale_factor) * static_cast<double>(configuration().graphics.gui_font_scale_flt);
```
Possibly `dynamic_cast<>`... but `static_cast<>` is probably correct. Also, check if there is `gui_font_scale_dbl` available. You don't need to necessarily add it but I know we've done some of that.

BenjamenMeyer (changes requested):
1. function documentation is essential. Documentation captures intent; code capture what it actually does. Both are essential.
2. there's a casting that needs to be fixed up
3. there's some CI/Github warnings that need to be addressed

engine/src/gui/font.cpp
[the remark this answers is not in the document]
SGD1953: Made it consistently doubles

SGD1953:
There are two issues that I noticed, the button texts for "Ship info", "Player info", "Load", "Save" and "New" shifts when you click them. It feels like those are the buttons created in cpp while the ones created from `controls.json` config work fine. ~The other issue is that modals and input fields render an excess "?" at the end, likely an unprintable character which the new font then renders as "?"~.

I played a bit more now and notices that the cockpit has some issues when you increase font too much. This is less a bug, more a consequence of the settings but some of it could probably be better handled. Like the distance to target is being pushed off-screen although there would be enough space, this could be handled by a smarter UI element setup.

With those settings it seems to run reasonably well though apart from the buttons:
```
gui_font_scale: 0.5,
font_point: 26.0,
```

engine/src/cmd/base_util.cpp
```diff
@@ -888,11 +888,10 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
+
```
royfalk: Please add comment - Only used by python.

engine/src/gui/font.cpp
```diff
@@ -4,8 +4,8 @@
  * Vega Strike - Space Simulation, Combat and Trading
  * Copyright (C) 2001-2026 The Vega Strike Contributors:
  * Project creator: Daniel Horn
- * Original development team: As listed in the AUTHORS file. Specifically: Mike Byron
```
royfalk, line 7: Looking at version 0.5 (when we took over), I see this line:
```
 * Vega Strike
 * Copyright (C) 2003 Mike Byron
```

As a rule, I prefer not to touch copyright notices that are this explicit at all.
However, as this was already done, let's keep what we have in place.

engine/src/gui/font.cpp
```diff
@@ -4,8 +4,8 @@
...
- * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
+ * Original development team: As listed in the AUTHORS file
+ * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy, Danny Gehl
```
royfalk, line 8: I assume this is you. Welcome.
@BenjamenMeyer what's our policy wrt to the current dev team line?
My feeling is that anyone who gets code into the repo should at least have his name added to that file.
Obviously, once someone is a regular, he or she should be added everywhere.

engine/src/gui/font.cpp
```diff
@@ -24,154 +24,147 @@
...
-//reference char space (it's clearly not going to be integer in the scaled space).
-static const double GLUT_WIDTH_HACK = 0.6;
-
-//The width of the space character in the outline font is too big, so we make it a special case.
-static const char SPACE_CHAR = ' ';
-
-bool useStroke() {
-    const bool tmp = configuration().graphics.high_quality_font_computer;
-    return !tmp;
+void Font::calcMetricsIfNeeded(void) const {
```
royfalk, line 35: Why move this up?

royfalk (changes requested):
My chief complaint is that you weren't bold enough in cutting through the old code.
In all honesty, for a first (?) PR in a new project, this is definitely reasonable.

This whole let's print several chars using drawChars and have it loop and call drawChar is not the way. imGUI provides us with good text printing functionality. Unless I'm missing something, this whole let's figure out the aspect ratio and the resolution is redundant. 

If it were me, I'd start at drawChars and just replace everything with imgui. There's an imgui loop somewhere that this needs to work with, so it may be a bit tricky but not too much.

Anyway, this looks better than what I would have done. I have no desire to understand the nuances of glyphs and such. 

I have one questions and two minor requests and I'm good.

libraries/gui/gui.cpp
royfalk: > How about if we wait on the font name? That can be a future feature addition. @royfalk ran into some problems with the same font not being available across different OSes. 

IIRC, it's just that we need to bundle the font with the assets and that takes a bit more admin work. Something like trying font name and if it fails using the default should be fine for now.

royfalk:
> There are two issues that I noticed, the button texts for "Ship info", "Player info", "Load", "Save" and "New" shifts when you click them. It feels like those are the buttons created in cpp while the ones created from `controls.json` config work fine. 

This was always a kludge from the beginning. There's a duplicate code in C++ (unless I removed it) doing some of what the python is doing. Working on this code is part programming, part archeology. 

I moved some code to JSON because it was repeating itself. Just FYI.

> The other issue is that modals and input fields render an excess "?" at the end, likely an unprintable character which the new font then renders as "?".

If you can't figure it out, we can handle this afterward. I don't see that as a blocker.
 
> I played a bit more now and notices that the cockpit has some issues when you increase font too much. 

This is totally reasonable. The game was designed for 4:3 and resolution of up to 1024x768. 

> With those settings it seems to run reasonably well though apart from the buttons:

Great

libraries/gui/gui.cpp
SGD1953: @royfalk I can submit a follow-up PR with it, this one seems quite complex already.

engine/src/gui/font.cpp
SGD1953: Re-added

engine/src/gui/font.cpp
SGD1953: I'll put him back, that was not by intention, I just threw the old file away entirely, implemented it from scratch and then copied a header over from another file.

engine/src/gui/font.cpp
SGD1953: You assume correctly, I'll follow your guidance here.

engine/src/cmd/base_util.cpp
SGD1953: Not sure I can follow, where do you want that comment and a bit more context would help as well. Or do you just want 
`//  Only used by python` on top of the method?

engine/src/gui/font.cpp
SGD1953: I re-implemented the entire file from scratch, hence order may have changed.

SGD1953:
> My chief complaint is that you weren't bold enough in cutting through the old code.
> In all honesty, for a first (?) PR in a new project, this is definitely reasonable.

Thank you, it is much appreciated!

>This whole let's print several chars using drawChars and have it loop and call drawChar is not the way. imGUI provides us with good text printing functionality. Unless I'm missing something, this whole let's figure out the aspect ratio and the resolution is redundant.

I understand, I started multiple times with this and then threw it in the bin after learning new details of how the current code works. First I thought, this is a GUI, we have ImGUI, I'll just build it with ImGUI (which is probably what we should do ultimately). So I looks at where to make the cut and inject the new world. Looked at `base_maker.cpp` which creates the room, but it has click handling and some logic too. Then I went further down and at every level I uncovered something new - ultimately I realised that there is a parallel UI framework with a whole bunch of logic baked in.

> If it were me, I'd start at drawChars and just replace everything with imgui. There's an imgui loop somewhere that this needs to work with, so it may be a bit tricky but not too much.

One of the things I looked into, with what I know now it could probably work somehow. The issue is, that for those components like the information boxes there is a whole custom markup language which `painttext-cpp` parses. Part of that is calculating line-breaks, there are color changes and more. In order to do this math, it uses the width and height functions of Font. Once that was clear, I felt the easiest is to draw a single char via OpenGL but using the font atlas of ImGUI. It is admittedly a bit hacky, but I learned a lot which may help to put some better structure in at a later point.

The key benefit is that we are consistently using one font across the board with adjustable size.

> Anyway, this looks better than what I would have done. I have no desire to understand the nuances of glyphs and such.

Neither do I, I understand it enough to get the job done. :) 

> I have one questions and two minor requests and I'm good.

Done :heavy_check_mark:

engine/src/cmd/base_util.cpp
royfalk: //  Only used by python

That's exactly right. 

These APIs are scattered throughout the code. Seeing a comment like this typically means I stop thinking about deleting the file for now as it requires checking 2 or 3 other repos - vegastrike assets, privateer gold and wcpu.

engine/src/gui/font.cpp
royfalk: Please follow @BenjamenMeyer guidance. I'm not certain what the policy is and he's in charge of that.

royfalk (approved):
If my two requests are answered, I'm good.

engine/src/gui/font.cpp
SGD1953: Will do once provided

BenjamenMeyer: @royfalk @SGD1953 yeah... we probably need to write it down; but yes - generally add yourself to the files you modify at the moment like you have been. @stephengtuggy generally has maintained the lines with our scripts for updating the copyrights every year.

BenjamenMeyer (approved):

SGD1953:
Closing as it is included in #1666

Issue #914, opened 2024-11-24
Fonts decrease in size as resolution increases


royfalk:
Game font size is fixed. As the resolution increased, the font decreased.

royfalk:
That's great news. Do you mean the main menu? (Campaign, Load, ...) ?

evertvorster:
Yes, the main menu seems to have a fixed spacing between text lines, rather than adjusting the spacing between lines based on the height of the font. 

However, the rest of the game has no such restriction, and the Base computer interface looks perfect, as does the HUD. 

I would say that for screen heights less than around 800, bitmap fonts are the way to go, and for higher resolutions, vector fonts are recommended, and the height of the font should be some function of the total vertical resolution to keep the text readable. 

Towards the end of vegastrike.config, you can see lines like:
```
<!-- #1280x800
  <var name="base_max_width__betternot" value="1067"/>
  <var name="base_max_height__betternot" value="800"/>
        <var name="x_resolution" value="1280"/>
        <var name="y_resolution" value="800"/>
  <var name="aspect" value="1.6"/>
  <var name="font_point" value="12"/>
#end -->
```

Smaller than 1280 x 800, all font_point values are 10. 
From 1280 x 800 and up the font_point value is gently increased with increasing vertical resolution.

evertvorster:
Hi there, @royfalk 

After a bit of a hiatus, I now have some time to do game testing again. My computer has been completely re-installed, and I have a new home directory, with no .vegastrike directory from olden times. 
To test my system, I installed vegastrike-git and vegastrike-engine-git from git master branches into my system. 

With the first game test that I tried to run on this machine, I am running into a few problems. 
It seems that the game engine makes .vegastrike in my home directory, as it should, but vegasettings crashes as soon as I try to set any variable. #1145 This of course means that I don't get a vegastrike.config file in ~/.vegastrike anymore. 

No matter, I thought, we are trying to move away from using vegastrike.config in any case. 

So, I modified /usr/share/vegastrike/config.json to start the game on my external monitor "Screen 2". It is a 4K monitor, so I set the resolution to 3840x2160. 

When launching the game, the fonts are too small to read, all over the game. When starting a new campaign, in the bases, on the HUD and in the base computer. 

So I started looking in /usr/share/vegastrike/config.json for some relief. There is a "font_point" variable that I set to 32, but still the font stayed tiny. 

[...]

Issue #195, opened 2020-07-25
In-Game Font Size Still Too Small Under Certain Circumstances


stephengtuggy:

evertvorster:
There is a fix for this that I put into Assets-Master, where the config file adjusts the text size based on screen resolution.

stephengtuggy:
@evertvorster Yes, but I think this is after your fix.

The font size is fine in windowed mode, near-4K resolution. And in full-screen 4K, the line spacing looks like the font should be about size 32. But it isn't. It's more like size 10.

evertvorster:
@stephengtuggy if the font size is about 10 in the higher resolutions, then you might have to replace your config file with the one from Assets-Production. Since the config file in your home directory is not automatically updated when a new one hits the repos, it still needs to be manually replaced.

I will of course test it all again and come back to you.

stephengtuggy:
I may have been mistaken about the circumstances under which this occurs. It might be that what makes the difference is whether you choose Bitmap or Vector fonts in vegasettings.

With vector fonts, it works. With bitmap fonts, it may not. But I'm still not sure I've figured out what the correlation is.

evertvorster:
@stephengtuggy we all agree that the fonts in game need to be bigger for higher resolutions. The font size specified in the config file works throughout the game, in that it makes the fonts bigger. 

The only place where the line spacing does not follow the font size is in the game loading screen, where you pick your save file. This is very annoying, and it is also what triggered me to look into the alternate startup, where you "wake up" in a cabin then load your save game with the game computer. 

We could conceivably make the font size a settable thing apart from the resolution, but the implementation here seems more simple. 

I have tried with vector and bitmap fonts, and the font sizes have been checked on my system to look visually the most similar to the really small resolutions that this game started out on.

BenjamenMeyer:
@evertvorster @stephengtuggy is this an issue in the engine itself? or just the assets? If the assets then we need to move the issue to Asset Masters.

stephengtuggy:
@BenjamenMeyer I'm not sure. Needs further investigation.

evertvorster:
This issue is also discussed in #671

stephengtuggy:
Hopefully this can be addressed as part of moving to Dear ImGui.

SGD1953:
@stephengtuggy can this be closed?

royfalk:
Checked GUI in main menu. 
Issue appears to be fixed.
Closing.

Issue #482, opened 2021-06-28
Adoption of Google code style and lint tool


royfalk:
At present we are not aligned to a specific style. The style currently recommended (linux kernel) is C and not C++.
I suggest we align with Google C++. I have no specific preference for it, except that it comes with a linter that supposedly identifies issues. Looking a few years into the future, it would be nice to enforce this at the PR level.

I suggest we discuss this here or in a dedicated room at gitter and then vote.

BenjamenMeyer:
I'm all for finding a linter that can enforce our coding style and would love to see that as part of the PR gates too, along with a script that can run it locally so devs can easily check it prior to submitting PRs. Let's resolve #481  and #483 first, then we can find a tool for this that matches those decisions.

That said, I've always had trouble finding one that works well for C and C++.

royfalk:
That's why I keep promoting Google's style. They wrote a linter for it. In theory, we can hack it to check for another style, but life is too short.

stephengtuggy:
@royfalk I agree with this proposal 100%.

stephengtuggy:
"Life is too short" -- yes, indeed

stephengtuggy:
If `cpplint` tries to tell us that we should be using `auto` in more places, then we can turn that warning off. I don't know that it will though

BenjamenMeyer:
@stephengtuggy about the only place that `auto` is good for is with iterators where you would otherwise have things like `x<y>::iterator` or `x<y>::const_iterator` - which can be both lengthy, and cause issues with maintenance so it's one place where it becomes very convenient; so there is a place for it,  but it's rather limited.

I've known some folks in the past that wanted to use auto for every variable (and argued from a standards perspective that it should be), which is just wrong.

Before saying more I'll read through the Google C++ guides. Right now I'm working on the ISO one. I'm a slow reader so please bear with me.

BenjamenMeyer:
@royfalk I'm assuming you mean https://google.github.io/styleguide/cppguide.html for the Google C++ Style, correct?

BenjamenMeyer:
Raw comments: https://gist.github.com/BenjamenMeyer/aebe7fa6a06fa0ae02e22b7ed0bfed12
overview:
- in general I agree with almost all of it; most of my comments are clarifying or saying we should go above what they said. So I think we can be really close to the Google Style.
- there's a few places where we deviate - for example we will be allowing exceptions since we integrate with Python and we are only C++11.

[...]

PR #1666, opened 2026-07-21
Improved imgui rendering

SGD1953:
Code Changes:
- [X] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-

This is a complete rework of the base computer rendering which now leverages ImGUI rendering entirely. This includes PR #1660 but features some more highlights

- base computer renders via Dear ImGUI, it sill leverages the custom component draw (no native buttons, menus etc.)
  - added missing ImGUI rendering context for event handlers (as they ultimately draw)
    - this is logically a bit questionable but works for now, we should probably separate logic and UI better in the future
  - Drawing routines now draw rectangles and textures via ImGUI APIs
  - **(Computer) Text now respects the normalized sizes again**, ImGUI dynamically adjusts it while rendering from the normalised font size in the UI config - this completely ignores `font_point` which I think is the superior approach, less settings and the fonts scale with your display size
    - ~Have not checked how text is handled elsewhere, but if there is a given size in norm height~ we can scale the ImGUI font
      - This impacts the main menu (loading screen) as well as other areas and makes them more readable
      - Cockpit could be updated to leverage the relative font sizes which makes the font scale with screen resolution
- Simplified text parsing for multiline elements
  - the way this is currently done is a bit wild from my perspective - the formatting allows `# n'` as newline instruction with optional line spacing, news however use `\\` and missions sometimes have `\n`, this should be consolidated from my perspective

Related issues
 -   https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/671
 -   https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/914
 -   https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/195

### DIFF
--- a/engine/src/cmd/base_util.cpp
+++ b/engine/src/cmd/base_util.cpp
@@ -888,12 +888,13 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
-
+//  Python accessor, not used in cpp
 float GetTextHeight(std::string text, Vector widheimult) {
     const float font_point = configuration().graphics.font_point_flt;
     return font_point * 2 / configuration().graphics.resolution_y;
 }
 
+//  Python accessor, not used in cpp
 float GetTextWidth(std::string text, Vector widheimult) {
     //Unsupported for now
     return 0;
--- a/engine/src/cmd/base_util.cpp
+++ b/engine/src/cmd/base_util.cpp
@@ -888,11 +888,10 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
+
 float GetTextHeight(std::string text, Vector widheimult) {
-    static constexpr bool force_highquality = true;
-    constexpr bool use_bit = force_highquality ||  configuration().graphics.high_quality_font;
     const float font_point = configuration().graphics.font_point_flt;
-    return use_bit ? getFontHeight() : (font_point * 2 / configuration().graphics.resolution_y);
+    return font_point * 2 / configuration().graphics.resolution_y;
 }
 
 float GetTextWidth(std::string text, Vector widheimult) {
--- a/engine/src/gui/font.cpp
+++ b/engine/src/gui/font.cpp
@@ -4,8 +4,8 @@
  * Vega Strike - Space Simulation, Combat and Trading
  * Copyright (C) 2001-2026 The Vega Strike Contributors:
  * Project creator: Daniel Horn
- * Original development team: As listed in the AUTHORS file. Specifically: Mike Byron
- * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
+ * Original development team: As listed in the AUTHORS file
+ * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy, Danny Gehl
  *
  * https://github.com/vegastrike/Vega-Strike-Engine-Source
  *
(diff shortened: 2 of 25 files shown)


SGD1953:
@royfalk took your advise and puled out a bigger knife to cut a bit deeper

engine/src/cmd/base_util.cpp
```diff
@@ -888,13 +888,13 @@ const Dictionary &GetEventData() {
     return _GetEventData();
 }
 
+//  Python accessor, not used in cpp
 float GetTextHeight(std::string text, Vector widheimult) {
```
BenjamenMeyer, line 892: is the `widheimult` parameter even used?

engine/src/gui/font.cpp
```diff
@@ -24,154 +24,147 @@
...
-        s->calcMetrics();
+float Font::drawChar(char c, float xOffset) const {
+    ImGuiContext* ctx = ImGui::GetCurrentContext();
+    if (!ctx) return 0.0f;
+
+    // Get the texture ID first
+    ImTextureID tex_id = ctx->IO.Fonts->TexRef.GetTexID();
+
+    // If tex_id is 0, the font atlas is not initialized or has been destroyed
+    if (!tex_id || tex_id == (ImTextureID)(intptr_t)-1) {
```
BenjamenMeyer: I'm not going to mark all the stops, but we should be using the newer casting methods.

Instead of:
```c++
(ImTextureID)(intptr_t) -1)
```
which just forces - preventing compiler warnings - it without any type safety; therefore it should be:
```c++
static_cast<ImTextureID>(static_cast<intptr_t>(-1))
```
or possibly even:
```c++
static_cast<ImTextureID>(-1)
```
which performs the type safety checks, enabling more correctly.

engine/src/gui/font.cpp
```diff
@@ -24,154 +24,147 @@
...
 
-//Check whether we need to recalc the metrics, and do it in const object.
-void Font::calcMetricsIfNeeded(void) const {
-    if (m_needMetrics) {
-        //calcmetrics is a cache.  Doesn't change the "real" state of the object.
-        Font *s = const_cast< Font * > (this);
-        s->calcMetrics();
+float Font::drawChar(char c, float xOffset) const {
+    ImGuiContext* ctx = ImGui::GetCurrentContext();
+    if (!ctx) return 0.0f;
```
BenjamenMeyer: FYI - please do:
```c++
if (!ctx) {
    return 0.0f;
}
```
instead of the 1 liner. This prevents a class of bugs when the line gets modified. We do not do implicit blocks any more; all blocks should be explicit.

engine/src/gui/guidefs.h
```diff
@@ -29,6 +29,15 @@
...
 #include "src/gfxlib.h"
 #include "gldrv/winsys.h"
+#include "imgui/imgui.h"
+
+// Macros for point conversion from VS [1-,1] coords to pixels
+#define NORM_TO_PIXEL_X(val) (((val) + 1.0f) * 0.5f * ImGui::GetIO().DisplaySize.x)
+#define NORM_TO_PIXEL_Y(val) ((1.0f - (val)) * 0.5f * ImGui::GetIO().DisplaySize.y)
+#define NORM_TO_PIXEL_W(val) ((val) * 0.5f * ImGui::GetIO().DisplaySize.x)
+#define NORM_TO_PIXEL_H(val) ((val) * 0.5f * ImGui::GetIO().DisplaySize.y)
+#define NORM_TO_PIXEL_FONT_SIZE(val) ((val) * 0.5f * ImGui::GetIO().DisplaySize.y)
```
BenjamenMeyer: it would be better to define these as C++ inline function if possible. We're trying to move away from C Preprocessor macros where we can

engine/src/gui/imguitext.cpp
```diff
@@ -0,0 +1,562 @@
...
+            fragmentStartPos = curPos;
+        } else {
+            // this consumes all chars but control chars and breakes when the line is full
+            if(input[curPos] == ' ') {
+                // we'll skip the space
+                lastWordBreakPos = curPos + 1;
+            }
+
+            // special news handling
+            if(input[curPos] == '\\') {
```
BenjamenMeyer: A `switch` case would be better:
```c++
switch (input[curPos]) {
   case `\\`:
       break;
   case `\n':
        break;
   ...
};
```
faster, more reliable and only evaluates `input[curPos]` once where the If block will evaluate it separately for every if/else-if branch.

BenjamenMeyer (changes requested):
There are some stylistic things to fix up. We're generally aiming for C++17 now; utilizing the [C++ ISO Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines), and have a few things of our own as well - like always having explicit blocks even if it's just for one line of code to prevent various kinds of bugs.

engine/src/cmd/base_util.cpp
SGD1953: I don't know to be honest, @royfalk just asked me to add the comment. I did not create this.

engine/src/gui/guidefs.h
SGD1953: Do we have a styleguide or linter settings somewhere? Happy to make those changes, but it would be great to know upfront and have some tooling to enforce the rules

engine/src/cmd/base_util.cpp
SGD1953: @BenjamenMeyer with the latest updates here, this file can basically be removed as it has been superceded by imguitext.cpp - any guidance on how to approach this?

engine/src/gui/guidefs.h
SGD1953: How about this:
```
// Namespace for coordinate conversion
namespace Coordinates {

    /// Converts normalized X coordinate [-1.0, 1.0] to display pixel space [0, DisplaySize.x]
    [[nodiscard]] inline float normToPixelX(float val) noexcept {
        return (val + 1.0f) * 0.5f * ImGui::GetIO().DisplaySize.x;
    }

    /// Converts normalized Y coordinate [-1.0, 1.0] (where +1 is top) to display pixel space [0, DisplaySize.y]
    [[nodiscard]] inline float normToPixelY(float val) noexcept {
        return (1.0f - val) * 0.5f * ImGui::GetIO().DisplaySize.y;
    }

    /// Scales a normalized width [0.0, 2.0] relative to display width
    [[nodiscard]] inline float normToPixelW(float val) noexcept {
        return val * 0.5f * ImGui::GetIO().DisplaySize.x;
    }

    /// Scales a normalized height/font size [0.0, 2.0] to pixel height
    [[nodiscard]] inline float normToPixelH(float val) noexcept {
        return val * 0.5f * ImGui::GetIO().DisplaySize.y;
    }

    /// Scales a normalized font size font size [0.0, 2.0] pixel height 
    [[nodiscard]] inline float normToPixelFontSize(float val) noexcept {
        return normToPixelH(val);
    }

} // namespace Coordinates
```

engine/src/gui/imguitext.cpp
SGD1953: You got it!

SGD1953:
@BenjamenMeyer your comments should be addressed

There is one bug which I would track separately. The Alerts for quit, save etc. don't show their text, I debugged it a bit but something quirky is going on here as the text is set but never rendered.

engine/src/gui/font.cpp
SGD1953: Done

engine/src/cmd/base_util.cpp
BenjamenMeyer: @SGD1953 please coordinate with @royfalk on that.

engine/src/gui/guidefs.h
BenjamenMeyer: > Do we have a styleguide or linter settings somewhere? Happy to make those changes, but it would be great to know upfront and have some tooling to enforce the rules

The group of us here on GitHub have discussed it and we've mostly adopted the [C++ ISO Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines); there is also https://wiki.vega-strike.org/Development:Code_style but it probably needs to be updated with our guidance. I don't think we have any linters yet, though @stephengtuggy might be able to point you in the right direction there.

Also FYI - the use of `auto` is fairly limited - namely to `for` loops and places where specifying a full iterator type (f.e `std::vector<int>::iterator`) becomes too verbose. There's a few others, but generally it's better to specify the type; use is more the exception than the rule.

Apologies on this - we really do need to update some of the documentation.

engine/src/gui/staticdisplay.h
```diff
@@ -66,11 +66,11 @@ class StaticDisplay : public Control {
 
 //Whether the text is multi-line.
     virtual bool multiLine(void) {
-        return m_paintText.widthExceeded() == PaintText::MULTI_LINE;
+        return m_paintText.lineCount() > 1;
     }
 
     virtual void setMultiLine(bool multi) {
-        m_paintText.setWidthExceeded(multi ? PaintText::MULTI_LINE : PaintText::ELLIPSIS);
+        // not implemented
```
BenjamenMeyer: this feels like a step backwards - if there is a reason its no longer needed to be implemented then a larger comment should be provided as to why.

BenjamenMeyer (changes requested):
generally most of my concerns were addressed; still keeping the block over the change from implemented to not implemented in the one method.

SGD1953:
@BenjamenMeyer well that was fun - not. Anyhow, the multline setter has now an implementation, I hope this will get us over the finish line here as this is a massive visual improvement

engine/src/cmd/base_util.cpp
SGD1953: Will do, is this a blocker for this PR?

SGD1953:
@BenjamenMeyer if we are good here, I'd highly recommend this for 0.10.x as it improves base computer readability significantly.

engine/src/gui/guidefs.h
```diff
@@ -27,8 +27,40 @@
 #ifndef VEGA_STRIKE_ENGINE_GUI_GUI_DEFS_H
 #define VEGA_STRIKE_ENGINE_GUI_GUI_DEFS_H
 
-#include "src/gfxlib.h"
+#include "gfxlib.h"
```
stephengtuggy: Please change this back to include the folder. Not a big deal, but I'm trying to keep this consistent across the project now, that `#include`s of project-internal header files include at least part of the path to the header file.

engine/src/gui/newbutton.h
```diff
@@ -28,7 +28,7 @@
 #define VEGA_STRIKE_ENGINE_GUI_BUTTON_H
 
 #include "control.h"
-#include "painttext.h"
+#include "imguitext.h"
```
stephengtuggy:
```suggestion
#include "gui/imguitext.h"
```

(see above comment about including part of the path to the file)

engine/src/gui/picker.h
```diff
@@ -28,7 +28,7 @@
 #define VEGA_STRIKE_ENGINE_GUI_PICKER_H
 
 #include "control.h"
-#include "painttext.h"
+#include "imguitext.h"
```
stephengtuggy:
```suggestion
#include "gui/imguitext.h"
```

engine/src/gui/staticdisplay.h
```diff
@@ -28,7 +28,7 @@
 #define VEGA_STRIKE_ENGINE_GUI_STATIC_DISPLAY_H
 
 #include "control.h"
-#include "painttext.h"
+#include "imguitext.h"
```
stephengtuggy:
```suggestion
#include "gui/imguitext.h"
```

engine/src/gui/imguitext.cpp
```diff
@@ -0,0 +1,546 @@
...
+ *
+ * Vega Strike is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Vega Strike.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#include "imguitext.h"
```
stephengtuggy:
```suggestion
#include "gui/imguitext.h"
```

stephengtuggy (approved):
I have a few nitpicks, but overall, this code looks good and seems to work, so I'm in favor. Nothing blocking on my side.

engine/src/gui/guidefs.h
stephengtuggy: (We should really document this somewhere.)

SGD1953:
> I have a few nitpicks, but overall, this code looks good and seems to work, so I'm in favor. Nothing blocking on my side.

Thank you, updated the includes as suggested

engine/src/cmd/base_util.cpp
stephengtuggy: @SGD1953 @BenjamenMeyer we have to keep a lot of things around for compatibility with the Python parts of the game. This is probably one of those things that should stay, at least for now. Though Roy will likely know more.

I don't think any further changes are needed here for this PR.

engine/src/gui/guidefs.h
stephengtuggy: I use the JetBrains CLion IDE, with its built-in linter along with clangd. It's pretty good. I recommend it.

SGD1953:
> There is one bug which I would track separately. The Alerts for quit, save etc. don't show their text, I debugged it a bit but something quirky is going on here as the text is set but never rendered.

Last commit fixes it and improves how the ImGUI `main_window` is handled in the event loop, I am not aware of any remaining bugs.

BenjamenMeyer:
> @BenjamenMeyer if we are good here, I'd highly recommend this for 0.10.x as it improves base computer readability significantly.

let's get this merged first; then you can port it to the 0.10.x branch; which should be a faster review since you can reference this.

engine/src/gui/guidefs.h
BenjamenMeyer: > I use the JetBrains CLion IDE, with its built-in linter along with clangd. It's pretty good. I recommend it.

it'd probably be good for us to find an independent linter that we can run in CI and easily locally.

Looks like they're using https://clang.llvm.org/extra/clang-tidy/ and then have their own for live analysis while you run the IDE, which apparently makes use of https://github.com/kde/clazy. We could probably run both under a GitHub Action.

SGD1953:
@BenjamenMeyer  Yes, please. That's the key dependency.

engine/src/gui/guidefs.h
SGD1953: How about we start with a `.clang-format` and possibly a `.clang-tidy` which we commit. All of our IDEs would be able to pick them up and flag any issues. We could start with some industry standard like Google, Microsoft or the like.

BenjamenMeyer: let's take the linter discussion over to https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/482

BenjamenMeyer (approved):

PR #1667, opened 2026-07-22
MIgrated all controls from code to controls.json enabling better them…

SGD1953:
…ing support

Code Changes:
- [X] Have the PR Validation Tests been run?

Controls in base computer are currently hard-wired in cpp which makes theming hard, this is the engine portion of a larger rework which moves all controls (excluding alerts) in the various base computer sections into a JSON in the data repository.

That resolves the 

> Hack that constructs controls in code.

I hope `dan_w` who started this 16 years ago will be proud! 

Relates to https://github.com/vegastrike/Assets-Production/pull/257


SGD1953:
@BenjamenMeyer @royfalk @stephengtuggy any thoughts on this one?

royfalk:
I hope to get to this on Sunday.

engine/src/cmd/basecomputer.cpp
```diff
@@ -531,6 +527,16 @@ GFXColor BaseComputer::getColorForGroup(std::string id) {
...
+    // Attempt loading via JSON using our controls_factory
+    if (getControls("controls.json", window())) {
+        // Successfully built GUI from controls.json!
+        return;
+    }
+
+    // ==========================================================
+    // FALLBACK: Hardcoded control construction runs if getControls returns false
+    // ==========================================================
+    VS_LOG(error, "BaseComputer: Constructing GUI using hardcoded C++ fallback.");
```
royfalk: There's a tendency to build over-resilient code. 
But, in this case, the purpose of this refactor is to move hard coded GUI code out of the engine.
If you leave it in, the next person will have to wade through the 700 or so lines that you left in the code.

When I started, basecomputer.cpp was 6000 LoC. Let's make it 0!

engine/src/cmd/controls_factory.cpp
```diff
@@ -27,66 +27,142 @@
...
 #include "gui/simplepicker.h"
 
-const std::string keys[] = {"name", "type", "rect",	"text", "textColor",
-                            "defaultTextColor", "color", "font", "id",
-                            "label", "command", "justification","thumbColor",
-                            "buttonColor", "outlineColor", "selectionColor",
-                            "highlightColor", "highlightTextColor",
-                            "textMargins", "textures", "multiline" };
+// All supported UI control property keys
+enum class ControlProp {
```
royfalk, line 49: It certainly is cleaner and longer.
My concern is that you are following in my (wrong) footsteps.
This code should move to python.

royfalk (approved):
Left two comments that try to show how I approach the engine code.
Feel free to push back.

In general, a lot of the work I've done over these 7 year has been to:
1. Reduce very large file sizes
2. Separate code to discrete concerns.
3. Lately, move anything that is not flight related and @BenjamenMeyer allows to the python side.

I'd like to expand a little on that. As a design philosophy, there a many more people who can write python and LLM is much better at it as well.

We are an open source project. Every time we move something to python, it creates an opportunity for someone less technical to change the code.

In particular, all the GUI stuff as well as economics. I'm sure there are other areas.

royfalk (changes requested):
Sorry. Please remove the extra 700 lines.
The other comment is strictly for info.

engine/src/cmd/basecomputer.cpp
SGD1953: I'm more than happy to remove it and agree in principle. It will work just fine for Vegastrike but it will likely break Privateer, that was my motivation to keep it in. We could move it into a separate file which we put into a `compatibility` folder or the like and log a deprecation warning to allow othe repos to catch up. What do you think?

engine/src/cmd/controls_factory.cpp
SGD1953: Make sense, I was thinking along the same lines as it is a bit inconsistent with the bases now. How about we merge this as it makes it consistently configurable and I create a follow-up issue to refactor this into a Python API? This is a bit more work as there is already one and I need to see how much of this is can cover. There is a lot of hidden semantics behind IDs which is hard coded in cpp.

SGD1953:
> Sorry. Please remove the extra 700 lines. The other comment is strictly for info.

@royfalk  No need to be sorry, I removed the lines. We are below 4000 lines now, way to go but progress nonetheless.

Created #1674 to track the follow-up work, from my perspective we should finalize the migration of at least all base-related screen to ImGUI and ideally align on our future UI strategy. Currently the approach I took was continuing to leverage the VS controls moving just the low-level drawing to ImGUI. That's better than what we had before, but it would be great if we could jointly define a target state.

engine/src/cmd/basecomputer.cpp
royfalk: I've broken many things in privateer and either fixed them or had someone else do it.
It's not the engine's job to do compatibility. It's the assets. 
Remove it.

Obviously, the one exception is stuff that can't be accommodated in the assets. For example, if we remove support for auto-pilot or any privateer specific code, that would be bad.

engine/src/cmd/controls_factory.cpp
royfalk: As I said below, this is strictly for info. It's fine.

royfalk (approved):
Ship it!

engine/src/cmd/basecomputer.cpp
BenjamenMeyer: As long as the assets can detect the change via the EngineAPI version I'm okay with changing things that break it but we need to get that documented so we know, it should be tracked in the changelog in https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/doc/Engine-API.md.

NOTE: We have already rev'd it for 0.10.x so we just need to record the changes.

BenjamenMeyer:
@SGD1953 please update the PR base

SGD1953:
@BenjamenMeyer rebase to `master` :heavy_check_mark: 

This will handle fonts with just the size and no weight more gracefully, thus I'll close https://github.com/vegastrike/Assets-Production/pull/252 as it is no longer needed.

SGD1953:
@BenjamenMeyer could we merge this please? This will make the cargo UI beautiful again.

Thank you

engine/src/cmd/basecomputer.cpp
stephengtuggy: > I've broken many things in privateer and either fixed them or had someone else do it. It's not the engine's job to do compatibility. It's the assets. Remove it.
> 
> Obviously, the one exception is stuff that can't be accommodated in the assets. For example, if we remove support for auto-pilot or any privateer specific code, that would be bad.

Has privateer actually been fixed for any of these compatibility breakages? Or are you just assuming that?

stephengtuggy:
@SGD1953 I think you might have misunderstood what @BenjamenMeyer was saying about updating the PR base. This PR needed to be updated with the latest changes from the `master` branch. I just took care of that.

SGD1953:
@stephengtuggy thank you! I'm pretty sure I did, maybe I did not push it.

engine/src/cmd/basecomputer.cpp
stephengtuggy: To be clear: I'm not saying we need to apply the fixes to PWCU ourselves necessarily. Nor am I saying that we need to keep the Python API rigidly the same. What I am saying is that I think we ought to coordinate with @ermo to keep PWCU updated on an ongoing basis, if possible, so that we have some assurance that our changes can be accommodated by PWCU.

And, of course, we should document all changes in the doc @BenjamenMeyer referenced.

engine/src/cmd/basecomputer.cpp
```diff
@@ -531,702 +524,14 @@ GFXColor BaseComputer::getColorForGroup(std::string id) {
...
-        ms->setTextColor(GUI_OPAQUE_WHITE());
-        ms->setTextMargins(Size(.02, .01));
-        ms->setId("Description");
-        ms->setScroller(descScroller);
-        infoGroup->addChild(ms);
-
-        infoGroup->addChild(descScroller);
-    }
+    VS_LOG(fatal, "BaseComputer: In Asset version >= 3.0 a controls.json file is required for the computer to work, please update the data directory or raise a bug for it.");
+    CockpitKeys::QuitNow();
```
stephengtuggy, line 534: There's a specific variant of `VS_LOG`, namely `VS_LOG_FLUSH_EXIT`, for logging a message and then quitting the game. I would prefer to use that here. It would go something like this:


```suggestion
    VS_LOG_FLUSH_EXIT(fatal, "BaseComputer: In Asset version >= 3.0 a controls.json file is required for the computer to work, please update the data directory or raise a bug for it.", 10);
```

stephengtuggy: Question: Should controls.json really be in the assets/data directory? Or is it the same between games supported by the engine, and should it therefore be in the Vega-Strike-Engine-Source directory? That might be a question to address separately, as I guess controls.json is already in the assets.

engine/src/cmd/controls_factory.cpp
```diff
@@ -108,210 +184,267 @@ static std::vector<double> splitAndConvert (const std::string &s, char delim) {
...
-        // Text Margin
-        if(attributes.count("textMargins")) {
-            std::vector<double> size = splitAndConvert(attributes["textMargins"], ',');
-            p->setTextMargins(Size(size.at(0), size.at(1)));
+        auto itParent = attributes.find("parent");
+        // Cache relationships requiring resolution in Pass 2
+        if (itParent != attributes.end()) {
+            VS_LOG(debug, (boost::format("Control '%1%' found parent '%2%'") % controlName % itParent->second));
+            // Child control: Pass 2 will add it to its GroupControl
+            parentLinks.emplace_back(c, itParent->second);
```
stephengtuggy, line 407: Nice!

stephengtuggy (approved):
I like this. This code appears to be both elegant and fault-tolerant. Nice work.

I had one minor nitpick and one question to be addressed separately. Nothing blocking though. LGTM!

engine/src/cmd/basecomputer.cpp
stephengtuggy: Filed as https://github.com/orgs/vegastrike/discussions/1694

stephengtuggy:
> @stephengtuggy thank you! I'm pretty sure I did, maybe I did not push it.

@SGD1953 You're welcome. It's also possible that the code on `master` simply changed again in the meantime, necessitating another update.

stephengtuggy:
@BenjamenMeyer are we good to merge this?

engine/src/cmd/basecomputer.cpp
SGD1953: Great question, I feel we could keep a copy in the engine as fallback. That would keep the code lean and in case i.e. PWGG does not have the file it would still work and not force anyone to migrate. We could emit an error encouraging adding a `controls.json`

engine/src/cmd/basecomputer.cpp
SGD1953: Once this is merged, I'm happy to create a PR for the Privateer repo which is the only other data dir I'm aware of.

BenjamenMeyer (approved):

PR #1669, opened 2026-07-25
Enhancement: AI Policy

BenjamenMeyer:

Please answer the following:

Code Changes:

Issues:
- https://github.com/orgs/vegastrike/discussions/887
- Please comment especially if something I wrote deviates from the discussion above. It was written with my best summary recollection.

Purpose:
- What is this pull request trying to do? document our AI policy
- What release is this for? N/A
- Is there a project or milestone we should apply this to? N/A


CONTRIBUTING.rst
```diff
@@ -34,6 +34,20 @@ Below are the official places to find the current developers:
 * FreeNode IRC: #vegastrike
 * Vega Strike Forums: https://forums.vega-strike.org/
 
+AI :robot:
+==========
+
+Use of AI is a contraversal topic. However, it is inevitable and now part of the world we live in. So while ideally we would not allow it that is no longer feasible as people will
```
SGD1953: This sounds a bit too fatalistic to me - we would forbid it but we can't, thus we allow it. AI has risks, but in the end it is just a tool that needs to be used appropriately.

CONTRIBUTING.rst
```diff
@@ -34,6 +34,20 @@ Below are the official places to find the current developers:
...
+then simply not declare their use of AI and then unknown risks have entered into the project. Therefore we kindly ask that anyone that wishes to contribute with AI follow the
+guidelines below.
+
+As a general rule, we follow the same policies on AI contributions as the `Linux Kernel: AI Policy`_ with the following changes:
+- Use only our accepted licenses, see `Vega Strike: Licensing`_ for details.
+- Follow our coding style
+- AI agents may not be added to the licensing statements in the files, only the human author making the contribution.
```
SGD1953, line 49: I like that, I clarifies accountability

CONTRIBUTING.rst
```diff
@@ -34,6 +34,20 @@ Below are the official places to find the current developers:
...
+- Like with human contributed changes all changes must be play tested by the author as appropriate. Do not simply use the AI to make a contribution that has never been actually appropriately tested. Lack of testing will have your PR closed.
```
SGD1953, line 50: I'd turn this around and consider all contributions human, no matter how they got generated. You submit it -you own it, no excuses.

For the testing, we could probably be a bit more specific in our test checklist on PRs. i.e. add 5-10 specific checks which the submitter has to tick and confirm they did their due diligence.

SGD1953 (changes requested):

CONTRIBUTING.rst
BenjamenMeyer: it was part of the discussion... possibly needs rewording. I'm open - this was more or less a first pass to get the discussion going.

CONTRIBUTING.rst
BenjamenMeyer: There is documentation already on what play testing means - see https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. It could be argued that what is there needs to be revisited, updated, and better defined for better clarity.

SGD1953: Yes, I'll have a look and think about it. It's quite good, but I worked a lot in the base computer lately, I did not really dock to planets, fly missions or test spec or jump drives. But if we had a longer list where you can check what you tested the reviewer can judge if there is something missing. And for the submitter, they would just have to check some boxes.

CONTRIBUTING.rst
royfalk: I missed this stuff when it was added. This could potentially litter our project with multiple licenses. 
Consider this scenario:
File A (original, GPL)
File B (new contribution, LGPL or CC - less restrictive)

Now we want to move someone's code from A to B. Without his permission, we've downgraded his contribution's license. I'm not sure if this is a good idea.

royfalk (approved):
The actual AI text looks fine to me. 
Unfortunately, I'm way behind the curve when it comes to AI.
This is the year I plan to change that.

CONTRIBUTING.rst
BenjamenMeyer: I think it was copied from previous places. Agreed we should simplify but let's do a quick audit of the code base first. I think it's nearly all GPL.

royfalk: How about:
_Please bear in mind that we are bound to accept **new** contributions under the GPLv3 or later license only._ 

Life is short and I don't want to audit all the files. Just prevent new stuff from adding new licenses. What's already here stays as is.

BenjamenMeyer: @royfalk it's a copy from https://wiki.vega-strike.org/Development:Graphics_Requirements#Licenses which covers more than just code. most of the licenses apply to the Game Assets side of things where GPL is not necessarily a good fit, such as images, music, etc. Not disagreeing that we probably need to break it down for better guidance, just giving insight into where the blob came from.

BenjamenMeyer: @royfalk addressing this in #1677 since it is not related to the AI Guidance here.

CONTRIBUTING.rst
evertvorster: My two cents:
When you pay an AI company to generate code for you, you own the code. 
This is reflected in the agreements that you sign when signing up for the service. Being the owner of the code, you can assign any copyright you want to the generated code. To be accepted into Vegastrike, that license would then be GPL-V3-or-later.

stephengtuggy: > Yes, I'll have a look and think about it. It's quite good, but I worked a lot in the base computer lately, I did not really dock to planets, fly missions or test spec or jump drives. But if we had a longer list where you can check what you tested the reviewer can judge if there is something missing. And for the submitter, they would just have to check some boxes.

@SGD1953 You didn't? That's not good. It's expected that just about every PR get play tested with the full play testing checklist, including docking to planets, accepting missions, and testing spec and jump drives, unless it's a documentation-only or build-system-only type change.

CONTRIBUTING.rst
stephengtuggy:
```suggestion
Use of AI is a controversial topic. However, it is inevitable and now part of the world we live in. So while ideally we would not allow it that is no longer feasible as people will
```

CONTRIBUTING.rst
stephengtuggy: Sometimes we get lazy and skip a test or two, but the majority of the play tests should be run every time.

stephengtuggy: If we added the contents of the play test plan to the Pull Request template, with actual check boxes next to each, would that make it clearer what the expectation is?

BenjamenMeyer: > If we added the contents of the play test plan to the Pull Request template, with actual check boxes next to each, would that make it clearer what the expectation is?

that would probably be a good step; and if someone misses someone else could check it off

stephengtuggy (approved):
Looks good to me!

CONTRIBUTING.rst
stephengtuggy: Filed task to do this as #1689

stephengtuggy:
@evertvorster @SGD1953 are you guys good with merging this as-is?

CONTRIBUTING.rst
stephengtuggy: What if we changed the wording to something like this?

"Use of AI is a controversial topic. However, it is so popular that realistically, we are unlikely to be able to avoid the subject for long. Contributors will want to use AI, so we should have guidelines for its use. Which we do. They are as follows:"

SGD1953: How about we drop this passage? It does not really add anything to the policy other than expressing a sentiment towards AI.

stephengtuggy: I'd be OK with dropping it. @BenjamenMeyer @royfalk what do you guys think?

CONTRIBUTING.rst
SGD1953: That's a great idea and makes it very clear.

CONTRIBUTING.rst
BenjamenMeyer: probably better to reword it as it sets up the section than removing it. Sentiments can be removed. Feel free to offer something up... I'll have to think about it.

SGD1953: Here is my suggestion:

> The adoption of artificial intelligence presents both significant opportunities and legitimate ethical, security, and operational concerns. Recognizing the magnitude of these challenges, we take the risks associated with AI seriously. Rather than prohibiting its use—which risks driving adoption unmonitored into the shadows—this policy establishes clear guidelines for its safe, transparent, and responsible integration. By defining explicit boundaries, we ensure that our teams can leverage AI effectively while maintaining strict oversight, quality control, and risk mitigation across all projects.

BenjamenMeyer: @SGD1953 thanks - I like that.

BenjamenMeyer: updated with @SGD1953 's paragraph.

SGD1953 (approved):

evertvorster (approved):
I'm quite happy with this.

PR #1670, opened 2026-07-25
Resolved build warnings for a less noisy build

SGD1953:
Fixed compiler warnings during the build, game still runs and cockpit looks clean


libraries/gfx_generic/mesh.h
```diff
@@ -357,7 +357,7 @@ class Mesh {
             int cloak = -1,
             float nebdist = 0); //short fix
 ///Will draw all undrawn meshes of this type
-    virtual void ProcessDrawQueue(size_t whichpass, int whichdrawqueue, bool zsort, const QVector &sortctr);
+    virtual void ProcessDrawQueue(int whichpass, int whichdrawqueue, bool zsort, const QVector &sortctr);
```
SGD1953, line 360: The other option would be to update all inheriting classes, but there are quite some of them.

BenjamenMeyer: actually...looking at the code - neither is really correct. We should upgrade this to being an enum. Most everything just passes it around - there's like may be 10 references, and then its actual interpretation is in [engine/src/gfx/mesh_gfx.cpp](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/mesh_gfx.cpp) and all its [meanings are defined there](https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/mesh_gfx.cpp#L234-L244) as well. We probably should upgrade the meanings to a shared header and just make it an enum which would give us a lot of safety checks on its usage as well.

Other locations:
- https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/technique.cpp#L388-L424
- https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/technique.h#L57

That last one would probably be best to be converted to a `std::map<enum, Pass>` in the process.
The first one in techniques is basically loading XML files, parsing the values, and then loading them into the vector that is the last one; I could see that being potentially buggy since it relies on just an incrementing number. https://github.com/vegastrike/Vega-Strike-Engine-Source/blob/master/engine/src/gfx/pass.cpp#L65-L110 would also need to get updated for using the new enum incrementally instead of the integer value.

So basics:
- at minimum the `size_t` is correct because it is unsigned vs the `int` which is signed. Since this is currently an index into a vector and fairly unchecked at that having it unsigned is a good first step.
- ideally this would all get upgraded to utilize an enumeration and move away from a vector to a map; both have constant access after initialization but overall its safer and makes things more bounded. We could also error out when the parsePass/technique loading goes beyond something we recognize.

Some more research might be needed to ensure I caught all the places this references and that its not exposed to the Python side.

BenjamenMeyer (changes requested):
If you want to get this through quickly, then the `whichpass` work needs to move to a separate PR as (a) at minimum its already correct and (b) it really needs an overhaul on its own but that's a lot bigger effort and probably requires more research to ensure its all understood well enough.

SGD1953:
This one is low priority to me, I just got annoyed by the warnings. Happy to have another look at the enum, but after the more important UI PRs are merged which fix essential parts of the game that got broken by the introduction on Dear ImGUI.

royfalk:
> If you want to get this through quickly, then the whichpass work needs to move to a separate PR as (a) at minimum its already correct and (b) it really needs an overhaul on its own but that's a lot bigger effort and probably requires more research to ensure its all understood well enough.

@BenjamenMeyer What's the downside of fixing the compiler warning, without overhauling a large and arcane code? mesh is a mess. Every time I try to touch it I get burned.

I'm fine with any contribution that takes us forward.

BenjamenMeyer:
@royfalk if the compiler warning is about using size_t then it's wrong and can be ignored.

royfalk:
> @royfalk if the compiler warning is about using size_t then it's wrong and can be ignored.

OK. But this is hiding other stuff. It's much easier to see a regression if we managed to squash all warnings.

Taking a look at `Mesh::ProcessDrawQueue`, the `size_t whichpass` parameter is used in two ways:
1. `const Pass &pass = technique->getPass(whichpass);` here `getPass(int)` converts it to `int`.
2. `ProcessShaderDrawQueue(whichpass, whichdrawqueue, zsort, sortctr); `

It looks to me like the switch to int is right. If we call `Mesh::ProcessDrawQueue` with a value of `whichpass` greater than 32bit, the results would not be desired. 

I assume it would rollover and return a smaller pass. I don't know what would happen but my guess is techniques don't really have more than a **few passes** and debating which type to use is counter-productive.

BenjamenMeyer:
@SGD1953 please update this; switch to int as @royfalk suggested and filed #1676 to update this properly

> It looks to me like the switch to int is right. If we call `Mesh::ProcessDrawQueue` with a value of `whichpass` greater than 32bit, the results would not be desired.

@royfalk we control the definition of `ProcessDrawQueue`; it's use is entirely a positive value making an unsigned value correct. Any how... filing #1676 for the enum conversation later.  Also, a negative number which `int` would allow would be bad which is probably why we moved it to `size_t` a while ago. FYI - traced this back to pre-github work - https://github.com/vegastrike/Vega-Strike-Engine-Source/commit/a82d55e7c5df25b9ff32fe29656241766c818b36 - I know its from pre-github b/c the user `dan_w` is one of the SF.net maintainers. Any how... let's fix this correctly and get rid of the warnings. I don't expect it in this PR, but let's follow-up in via #1676 .

SGD1953:
@BenjamenMeyer maybe I'm missing something, but in the current PR I did change it to `int` (as this is what I saw was used dwnstream). So are we good then with the follow-up PR and this being `int`?

stephengtuggy:
Please, let's split out the `whichpass` discussion and changes into a separate issue/discussion and PR. It's generating way too much debate for a quick-fix PR to address compile warnings.

BenjamenMeyer:
> Please, let's split out the `whichpass` discussion and changes into a separate issue/discussion and PR. It's generating way too much debate for a quick-fix PR to address compile warnings.

@SGD1953 If we can merge it then yes, leave it be.

> Please, let's split out the `whichpass` discussion and changes into a separate issue/discussion and PR. It's generating way too much debate for a quick-fix PR to address compile warnings.

@stephengtuggy @SGD1953 #1679  is for that going forward; closed #1676 as a duplicate of it - accepting the one with better detail. But yes, anything further let's push to that discussion and a separate PR.

@stephengtuggy taking this AS-IS for now since it clears some compiler warnings; otherwise agreed

stephengtuggy:
@BenjamenMeyer can you re-review this PR, please? Your requested changes are blocking merging it at the moment.

PR #1680, opened 2026-07-31
Added Cmake files and directories  to .gitignore

SGD1953:
This removes cmake artifacts from from the moified file list to avoid accidental commits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ build/
 build-*/
 cmake-build/
 cmake-build-*/
+CMakeFiles
+*.cmake
 data/
 ext/
 *.pyc


.gitignore
```diff
@@ -3,6 +3,8 @@ build/
 build-*/
 cmake-build/
 cmake-build-*/
+CMakeFiles
+*.cmake
```
BenjamenMeyer, line 7: we actually have a bunch of custom `*.cmake` files so we do not want to ignore that (see the `cmake/` directory in the root).

We can certainly ignore anything that is cached. Though I do highly recommend using an out-of-tree build:
```
$ git clone ...
$ cd Vega-Strike-Engine-Source
$ mkdir build
$ cd build
$ cmake ..
```
The above is out-of-source using a sub-directory but the `build` dir could be any where you like.

BenjamenMeyer (changes requested):

.gitignore
SGD1953: So what you are saying is that I just ran cmake in root and created directories where they should not be? Let me try a clean than, happy to close this if it is a me-issue.

SGD1953: OK, tested it, it is a me-issue, never mind. :facepalm:

PR #1681, opened 2026-08-01
Fix ship overspeed on turn

evertvorster:
Play test results:

- [x] Basic flight.
- [x] Docking to a planet.
- [x] On planet actions:
    - [x] Save the game.
    - [x] Read news.
    - [x] Buy/Sell some goods.
- [x] SPEC flight.
- [x] Save the game again.
- [x] Upgrade your new ship.
- [x] Save once more.

Ran through the beginning of Oswalds tutorial. Ship handles normally at slow speed, and negative speeds. 
Noticed that there was a bug in the code that would engage the afterburner during a turn, even when the button is not pressed. Fixed that in passing. It now works as expected. 

Out of scope for this is the fact that afterburn only gives extra thrust for almost all the ships, and do not actually provide any extra speed. This is set in Assets per ship, and we might want to review that. 

Overdrive with afterburner works correctly. 

This has been thoroughly tested. Ship handling is MUCH improved, and actually fun to play. 

Issues:
#909

Purpose:
* Clamps speed to what has been set in the FCMP while the craft is turning. 
* Its not perfect, the physics engine does seem to get in the way a little bit, but it seems much improved. 
- What release is this for?
- - 0.11
- This is actually a backportable change, it does not depend on anything else.


stephengtuggy (approved):
Looks good to me! It would be really nice to get this bug fixed.

Issue #357, opened 2020-12-06
HUD always assumes 4:3 screen ratio


stephengtuggy:
As a result, the HUD display panels are stretched out horizontally on widescreen displays. They don't look right. For example, circles are elongated, not round.

evertvorster:
Which resolution are you using? On my 16:9 monitor the circles look round...

stephengtuggy:
My monitor is 3840x2160, and I am using the 3800x2000 windowed resolution.

The main Vega Strike display is fine. It's the smaller panels around the edges that look elongated. E.g., the target view on the lower right.

Hopefully that clarifies things.

PR #1684, opened 2026-08-02
Nav screen migrated to Dear ImGUI drawing routines and mouse cursor restored

### LINKED ISSUE #1668: Hover state broken in GUI python bridge
Sprite images for "hot" state of buttons in the main menu don't work. To validate this I added debug output to `GUI.py` and the events never fire which breaks the functionality to have nice hover images for the menu.

SGD1953:
Code Changes:

This PR migrates remaining UIs to ImGUI, I'll put a list here so we have a better overview.

## Migration overview

- [X] Basecomputer - already done with #1666 
- [X] Mouse cursor restored for main menu, bases and base computer
- [X] Nav screen migrated to use ImGUI drawing
- [ ] Render bases with ImGUI
- [ ] Pause screen - need to update my bindings to get there, don't have a Pause/Break key
- [ ] Help (key overview) - for this I'd create a separate PR which renders this screen using the actual key mapping instead of a static image
- [ ] Cockpit

## Drawing order considerations

**Important** currently we use 2 rendering methods, the original direct openGL drawing approach and ImGUI which collects all drawing instructions into a drawlist and draws it at the end of the event loop. This creates issues where we mix the approaches (prior to #1666) this was primarily text. In those cases, VS direct drawing will draw something only for ImGUI to overdraw it later without it becoming actually visible to the user. This is why the nav screen was broken and why migrating to ImGUI fixes it.

## Why not bases?

Bases are in principle just sprites and text, except for the landing bay however where the 3D model of the ship is drawn. I'm open for suggestions, but we can't draw the sprite with ImGUI easily without overdrawing the ship. I'm open for suggestions here as I might be missing something. It's not a massive issue from my perspective as this concerns a few isolated places. This is the only area where this happens as far as I'm aware, cockpit, nav computer etc. could all be migrated as there the GUI overlays the rendered space scene.

## Mouse

The mouse is now consistently disabled at the beginning of the event loop and drawn with a single method removing prior duplication of code. For the mouse to work properly, I needed to comment the software mouse sensitivity functionality which is ill-placed from my perspective (and therefore causing issues) and should be removed.

```
// biModifyMouseSensitivity(x, y, false);
```
I did see that @evertvorster made a similar change in #1678 in the meanwhile, now it is double-commented. This fixes #1668 and enables wonderful things like https://github.com/vegastrike/Assets-Production/pull/263 

## Sprite rendering

To render sprites I added a new `VSSprite::DrawWithImGui` method which is not the most beautiful solution. I picked it nonetheless because I think the Sprite is where it belongs and there is a `Draw` method already which is used widely and cannot just be updated (see drawing order section). For the mouse cursor, not all the sprite functionality is required and for the lack of any other use-case I left some features unimplemented. I'd suggest we implement once those use-cases become relevant (i.e. we need to draw sprites with mipmaps or the like). For all GUI use-cases this should do.

## Misc

Simplified some code and removed unneeded imports.

### DIFF
--- a/engine/src/gfx/nav/navitemstodraw.h
+++ b/engine/src/gfx/nav/navitemstodraw.h
@@ -70,21 +70,13 @@ void NavigationSystem::DrawGrid(float &x1, float &x2, float &y1, float &y2, cons
     // Vertical grid lines
     for (int i = 1; i < 10; i++) {
         float x = x1 + i * deltax;
-        drawList->AddLine(
-            NormToPixel(x, y1),
-            NormToPixel(x, y2),
-            color, 1.0f
-        );
+        drawList->AddLine(NormToPixel(x, y1), NormToPixel(x, y2), color, 1.0f);
     }
 
     // Horizontal grid lines
     for (int i = 1; i < 10; i++) {
         float y = y1 + i * deltay;
-        drawList->AddLine(
-            NormToPixel(x1, y),
-            NormToPixel(x2, y),
-            color, 1.0f
-        );
+        drawList->AddLine(NormToPixel(x1, y), NormToPixel(x2, y), color, 1.0f);
     }
 }
 
@@ -107,17 +99,14 @@ void NavigationSystem::DrawHalfCircleTop(float x, float y, float size, const GFX
     ImDrawList* drawList = GetNavDrawList();
     ImU32 color = ToImColor(col);
 
-    const int segments = 10;
-    float radius = 0.5f * size;
-    float yOffset = y - 0.25f * size;
+    float radiusPx = Coordinates::normToPixelW(0.5f * size);
+    ImVec2 center = NormToPixel(x, y);
+    center.y -= radiusPx * 0.5f; // Offset in pixel space
 
     drawList->PathClear();
-    for (int i = 0; i <= segments; ++i) {
-        float angle = i * (static_cast<float>(M_PI) / segments);
-        float px = x + radius * std::cos(angle);
-        float py = yOffset + radius * std::sin(angle);
-        drawList->PathLineTo(NormToPixel(px, py));
-    }
+    // PathArcTo(center, radius, a_min, a_max, num_segments)
+    // 0 to PI draws the arc
+    drawList->PathArcTo(center, radiusPx, 0.0f, static_cast<float>(M_PI), 16);
     drawList->PathStroke(color, 0, NAV_LINE_WEIGHT);
 }
 
@@ -128,17 +117,13 @@ void NavigationSystem::DrawHalfCircleBottom(float x, float y, float size, const 
     ImDrawList* drawList = GetNavDrawList();
     ImU32 color = ToImColor(col);
 
-    const int segments = 10;
-    float radius = 0.5f * size;
-    float yOffset = y + 0.25f * size;
+    float radiusPx = Coordinates::normToPixelW(0.5f * size);
+    ImVec2 center = NormToPixel(x, y);
+    center.y += radiusPx * 0.5f; // Offset in pixel space
 
     drawList->PathClear();
-    for (int i = 0; i <= segments; ++i) {
-        float angle = static_cast<float>(M_PI) + i * (static_cast<float>(M_PI) / segments);
-        float px = x + radius * std::cos(angle);
-        float py = yOffset + radius * std::sin(angle);
-        drawList->PathLineTo(NormToPixel(px, py));
-    }
+    // PI to 2*PI draws the opposite arc
+    drawList->PathArcTo(center, radiusPx, static_cast<float>(M_PI), static_cast<float>(M_PI * 2.0), 16);
     drawList->PathStroke(color, 0, NAV_LINE_WEIGHT);
 }
 
@@ -149,39 +134,44 @@ void NavigationSystem::DrawPlanet(float x, float y, float size, const GFXColor &
     ImDrawList* drawList = GetNavDrawList();
     ImU32 color = ToImColor(col);
 
-    // Outer circle
-    float radius = 0.5f * size;
-    DrawCircle(x ,y, size, col);
+    ImVec2 c = NormToPixel(x, y);
+    float r = Coordinates::normToPixelW(0.5f * size);
 
-    // Planet interior lines
-    drawList->AddLine(NormToPixel(x - radius, y), NormToPixel(x, y + 0.2f * size), color, NAV_LINE_WEIGHT);
-    drawList->AddLine(NormToPixel(x, y + 0.2f * size), NormToPixel(x, y - 0.2f * size), color, NAV_LINE_WEIGHT);
-    drawList->AddLine(NormToPixel(x, y - 0.2f * size), NormToPixel(x + radius, y), color, NAV_LINE_WEIGHT);
+    // Outer circle
+    drawList->AddCircle(c, r, color, 0, NAV_LINE_WEIGHT);
+
+    // Planet interior lines (all calculated in pure pixel space relative to center)
+    drawList->AddLine(ImVec2(c.x - r, c.y), ImVec2(c.x, c.y + r * 0.4f), color, NAV_LINE_WEIGHT);
+    drawList->AddLine(ImVec2(c.x, c.y + r * 0.4f), ImVec2(c.x, c.y - r * 0.4f), color, NAV_LINE_WEIGHT);
+    drawList->AddLine(ImVec2(c.x, c.y - r * 0.4f), ImVec2(c.x + r, c.y), color, NAV_LINE_WEIGHT);
 }
 
 //**********************************
-// Draws a station icon (3x3 grid pattern)
+// Draws a station icon (Square 3x3 grid pattern)
 //**********************************
 void NavigationSystem::DrawStation(float x, float y, float size, const GFXColor &col) {
     ImDrawList* drawList = GetNavDrawList();
     ImU32 color = ToImColor(col);
 
-    float segment = size / 3.0f;
-    float startX = x - 0.5f * size;
-    float startY = y - 0.5f * size;
-    float endX = x + 0.5f * size;
-    float endY = y + 0.5f * size;
+    ImVec2 c = NormToPixel(x, y);
+    float halfSizePx = Coordinates::normToPixelW(0.5f * size);
+    float segmentPx = (halfSizePx * 2.0f) / 3.0f;
 
-    // Horizontal lines
+    float startX = c.x - halfSizePx;
+    float startY = c.y - halfSizePx;
+    float endX   = c.x + halfSizePx;
+    float endY   = c.y + halfSizePx;
+
+    // Horizontal grid lines
     for (int i = 0; i < 4; i++) {
-        float currentY = startY + i * segment;
-        drawList->AddLine(NormToPixel(startX, currentY), NormToPixel(endX, currentY), color, NAV_LINE_WEIGHT);
+        float curY = startY + i * segmentPx;
+        drawList->AddLine(ImVec2(startX, curY), ImVec2(endX, curY), color, NAV_LINE_WEIGHT);
     }
 
-    // Vertical lines
+    // Vertical grid lines
     for (int i = 0; i < 4; i++) {
-        float currentX = startX + i * segment;
-        drawList->AddLine(NormToPixel(currentX, startY), NormToPixel(currentX, endY), color, NAV_LINE_WEIGHT);
+        float curX = startX + i * segmentPx;
+        drawList->AddLine(ImVec2(curX, startY), ImVec2(curX, endY), color, NAV_LINE_WEIGHT);
     }
 }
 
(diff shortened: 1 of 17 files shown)


SGD1953:
@BenjamenMeyer @royfalk @stephengtuggy @evertvorster when you get to it, I'd appreciate your thoughts and feedback. 🙏

engine/src/gui/imgui_support.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "gui/imgui_support.h"
+#include "gfx/sprite.h"
+#include "imgui/imgui.h"
+//#include "configuration/configuration.h"
+
+
+static bool softwarePosition;
+// virtual mouse x after sensitivity adjustmemt
+static int mmx = 0;
+// virtual mouse x after sensitivity adjustmemt
```
stephengtuggy:
```suggestion
// virtual mouse y after sensitivity adjustment
```

Should we be doing sensitivity adjustment at all? The whole concept seems to have caused more trouble than it's worth.

engine/src/gui/imgui_support.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include <png.h>
+#include "gui/imgui_support.h"
+#include "gfx/sprite.h"
+#include "imgui/imgui.h"
+//#include "configuration/configuration.h"
+
+
+static bool softwarePosition;
+// virtual mouse x after sensitivity adjustmemt
```
stephengtuggy:
```suggestion
// virtual mouse x after sensitivity adjustment
```

engine/src/gfx/nav/navitemstodraw.h
```diff
@@ -25,409 +25,316 @@
...
-    GFXDraw(GFXLINESTRIP, verts);
-
-    GFXEnable(TEXTURE0);
+    drawList->PathStroke(color, 0, NAV_LINE_WEIGHT);
 }
-//**********************************
 
-//This will draw a planet icon. circle + lightning thingy
+//**********************************
+// Draws a planet icon (Circle with central axis lines)
```
stephengtuggy, line 131: Did you change the planet icon visually? It kind of looks like you did, but I can't tell for sure just by reading the code.

engine/src/gfx/nav/navitemstodraw.h
```diff
@@ -25,409 +25,316 @@
...
+    drawList->AddLine(NormToPixel(x + half, y - half), NormToPixel(x + inner, y - half), color, NAV_LINE_WEIGHT*2);
+    drawList->AddLine(NormToPixel(x + half, y - half), NormToPixel(x + half, y - inner), color, NAV_LINE_WEIGHT*2);
 }
-//**********************************
 
-//This will draw an oriented circle
+//**********************************
+// Draws a 3D projected navigation disc using smooth ImGui Ellipses
 //**********************************
 void NavigationSystem::DrawNavCircle(float x, float y, float size, float rot_x, float rot_y, const GFXColor &col) {
```
stephengtuggy, line 256: It looks like you changed what this draws visually, as well. I don't think we want that. Refactoring how something is drawn is fine. Changing the end result should be out of scope for this PR.

stephengtuggy:
OK, I reviewed this. I have some concerns. Among other things, the title of this PR doesn't match its contents. It would be more accurate to title it `Nav Computer and Mouse Pointer migrated to Dear ImGUI drawing routines`. Right?

stephengtuggy:
Also, it looks like some elements of what the nav computer draws have changed visually from what they used to be. Please do not do that in this PR if you can help it. This PR should only re-enable broken functionality and refactor how things are drawn, not what is drawn. If that makes sense.

engine/src/gfx/nav/navitemstodraw.h
SGD1953: No, I just replaced. The manual circle draws via sin/cos with `drawList->AddCircle()`

SGD1953:
Screenshot for reference:
vs_nav

SGD1953:
> OK, I reviewed this. I have some concerns. Among other things, the title of this PR doesn't match its contents. It would be more accurate to title it Nav Computer and Mouse Pointer migrated to Dear ImGUI drawing routines. Right?

In principle yes, but I have listed why the remaining UIs are not (yet) included.

> Also, it looks like some elements of what the nav computer draws have changed visually from what they used to be. 

Not intentionally, but after your review - thank you for that - I did notice one difference. `DrawCircle()` draws a circle in the pixel space while the old vertex approximation draws a circle in the normalized space which will become an ellipsis on wider screens.

Looked into this with my buddy Gemini and now it looks beautiful - thanks again for highlighting it!

vs_nav2

I did make one visual change by increasing the line width to 2px for the shapes, this is constant and can be easily edited.

stephengtuggy:
> > OK, I reviewed this. I have some concerns. Among other things, the title of this PR doesn't match its contents. It would be more accurate to title it Nav Computer and Mouse Pointer migrated to Dear ImGUI drawing routines. Right?
> 
> In principle yes, but I have listed why the remaining UIs are not (yet) included.

It's not a problem that the remaining UIs are not included yet. I just think the PR title should be changed to reflect that.

> 
> > Also, it looks like some elements of what the nav computer draws have changed visually from what they used to be.
> 
> Not intentionally, but after your review - thank you for that - I did notice one difference. `DrawCircle()` draws a circle in the pixel space while the old vertex approximation draws a circle in the normalized space which will become an ellipsis on wider screens.
> 
> Looked into this with my buddy Gemini and now it looks beautiful - thanks again for highlighting it!
> vs_nav2
> 
> I did make one visual change by increasing the line width to 2px for the shapes, this is constant and can be easily edited.

This looks great, thanks. All except for the planet icon, which appears to have an out-of-place lower half-circle in a different color. Was that there before?

Please also fix the CodeQL alerts. I believe they can be fixed by changing at least the first term in each expression to a `double`.

Other than that, I think this should be fine.

engine/src/gfx/nav/navitemstodraw.h
stephengtuggy: OK, scratch that. Applying the proper aspect ratio to what is drawn, so that circles are perfect circles and squares are perfect squares regardless of screen aspect ratio, is actually awesome. I'm very happy that you made that change. It's been needing to be done for a long time, both in the nav screen and elsewhere. See #357

engine/src/gui/imgui_support.cpp
```diff
@@ -33,9 +33,9 @@
 
 
 static bool softwarePosition;
-// virtual mouse x after sensitivity adjustmemt
+// virtual mouse x after sensitivity adjustment
 static int mmx = 0;
-// virtual mouse x after sensitivity adjustmemt
+// virtual mouse x after sensitivity adjustment
```
stephengtuggy: `adjustmemt` has been corrected to `adjustment`, but both comments still say `virtual mouse x`, even though the second one is really `virtual mouse y`.

engine/src/gfx/nav/navitemstodraw.h
stephengtuggy: OK

SGD1953:
> > > OK, I reviewed this. I have some concerns. Among other things, the title of this PR doesn't match its contents. It would be more accurate to title it Nav Computer and Mouse Pointer migrated to Dear ImGUI drawing routines. Right?
> > 
> > 
> > In principle yes, but I have listed why the remaining UIs are not (yet) included.
> 
> It's not a problem that the remaining UIs are not included yet. I just think the PR title should be changed to reflect that.

Changed

> > > Also, it looks like some elements of what the nav computer draws have changed visually from what they used to be.
> > 
> > 
> > Not intentionally, but after your review - thank you for that - I did notice one difference. `DrawCircle()` draws a circle in the pixel space while the old vertex approximation draws a circle in the normalized space which will become an ellipsis on wider screens.
> > Looked into this with my buddy Gemini and now it looks beautiful - thanks again for highlighting it!
> > vs_nav2
> > I did make one visual change by increasing the line width to 2px for the shapes, this is constant and can be easily edited.
> 
> This looks great, thanks. All except for the planet icon, which appears to have an out-of-place lower half-circle in a different color. Was that there before?

That's the your ship and that was there before. It is an odd visualization though.

> Please also fix the CodeQL alerts. I believe they can be fixed by changing at least the first term in each expression to a `double`.

Fixed.

> Other than that, I think this should be fine.

SGD1953:
This should be good now, moved migration overview to #1702

evertvorster:
Noticed you need a binding key...  You know, I can formally request you play my two PRs, and if approved they will give you that binding key. ;)

SGD1953:
Haha, I will have a look this week for certain!

engine/src/gfx/nav/navitemstodraw.h
```diff
@@ -25,409 +25,313 @@
...
+    // 1. Concentric web circles in true 3D projection
+    for (int i = 0; i < segments; ++i) {
+        float angle1 = i * (TWO_PI / segments);
+        float angle2 = (i + 1) * (TWO_PI / segments);
+
+        GFXColor ci(col.r, col.g, col.b * std::fabs(std::sin(angle1 / 2.0f)), col.a);
+        ImU32 segmentColor = ToImColor(ci);
+
+        QVector pos1(0.6f * size * std::cos(angle1), 0.6f * size * std::sin(angle1), 0.0f);
+        QVector pos2(0.6f * size * std::cos(angle2), 0.6f * size * std::sin(angle2), 0.0f);
```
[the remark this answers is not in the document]
stephengtuggy: I'm pretty sure that these CodeQL alerts are not fixed. To fix them, each of the eight occurrences of `0.6f` in this file should be changed to just `0.6`. And remove the casts to `float` on the four lines in question; they aren't necessary, and will only slow the code down.

engine/src/gui/imgui_support.cpp
stephengtuggy: Good. This is fixed. Thanks

engine/src/gui/imgui_support.cpp
royfalk: We should not. This should be an OS facility.

engine/src/gui/imgui_support.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+/** Initialises a Frame with proper settings for ImGUI and mouse */
+void StartGUIFrame(void) {
+    softwarePosition = false;
+    GFXHudMode(true);
+    // this will hide whatever is behind (the cockpit)
+    GFXClear(GFXTRUE);
+}
+
+void DrawMouseCursor(MousePointerStyle pointerStyle) {
```
royfalk, line 56: Why not just change the mouse cursor in SDL?

SDL2 / C++ or Python Bindings

SDL provides native hardware cursor creation:
C++

```
// Built-in system cursor
SDL_Cursor* cursor = SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_CROSSHAIR);
SDL_SetCursor(cursor);

// Custom surface cursor
SDL_Surface* surface = IMG_Load("cursor.png");
SDL_Cursor* customCursor = SDL_CreateColorCursor(surface, 0, 0);
SDL_SetCursor(customCursor);
```

royfalk (changes requested):
The nav screen was buggy and is still buggy.
Not sure which of these are new and which are old:
- Background too transparent. Need more opaque.
- Text in the beginning (when you press shift-m) overflows to the right (probably new). Overwriting the menu buttons on the right.
- Menu in English should be at the top or left, not right.
- Buttons should at least have borders to show they are buttons.
- There's text there to report bugs or something. See screenshot.

Other than that, the biggest issue is I'd like us to use SDL to draw the mouse.

royfalk:
nav menu

SGD1953:
@royfalk 

> The nav screen was buggy and is still buggy. Not sure which of these are new and which are old:
> 
>     * Background too transparent. Need more opaque.

This is how it was before (it was broken), no change in alpha.

>     * Text in the beginning (when you press shift-m) overflows to the right (probably new). Overwriting the menu buttons on the right.

Yes, this was most likely introduced with the TextPlane migration to ImGUI as it does currently not split lines.

>     * Menu in English should be at the top or left, not right.
> 
>     * Buttons should at least have borders to show they are buttons.
> 
>     * There's text there to report bugs or something. See screenshot.

No change here, this was the same before.

> 
> Other than that, the biggest issue is I'd like us to use SDL to draw the mouse.

Is this is a blocker? This may have some side effects to get it working with hover etc. and I don't really see what we are losing with this approach. I'm not overly passionate about either way, but I'd prefer to follow up with a more isolated mouse-only PR. Tracked as #1709 

The big issue this fixes is not making the Nav computer perfect, but to restore the proper drawing order as it is currently drawing and then overdrawn with the transparent rectangle and text. Created #1710 as a follow-up for visual improvements.

royfalk:
> Is this is a blocker?

None of it is. I'd appreciate it if you could open two issues. One for the mouse and the other for everything else. I'll approve it now.

royfalk (approved):

SGD1953:
Awesome, will do this today.

SGD1953:
@royfalk @stephengtuggy as we haven't merged this yet, I looked at the `TextPlane` and added the missing line break functionality. This fixes the overrunning text in the nav computer also impacts i.e. console which is now wrapping into its designated area again.

vs_cockpit
vs_nav3

royfalk:
> @royfalk @stephengtuggy as we haven't merged this yet, I looked at the TextPlane and added the missing line break functionality. This fixes the overrunning text in the nav computer also impacts i.e. console which is now wrapping into its designated area again.

Great. @stephengtuggy, any additional comments?

SGD1953:
If we are good for merge I will rebase on latest master again

stephengtuggy:
We worked hard to get the mouse cursor to be the hardware mouse cursor. I'd rather not take a step backwards in that regard if we can help it. Please, can you split out the mouse-related changes into a separate PR, so we can discuss them separately? Or just remove them entirely.

SGD1953:
> We worked hard to get the mouse cursor to be the hardware mouse cursor. I'd rather not take a step backwards in that regard if we can help it. Please, can you split out the mouse-related changes into a separate PR, so we can discuss them separately? Or just remove them entirely.

This work you did is by no means lost because of this PR. Could I suggest that we follow-up with a PR which migrates the cursor to SDL and removes all the software mouse code? Two reasons, while the code changes are probably not too hard, I need to re-playtest everything which is quite time consuming. The PR in its current form works and reintroduces the mouse cursor in its intended form by using the drawing mechanism of the UI framework you chose, so it is not really inconsistent.

If it is a blocker I'll have a look, but that will take a little while.

stephengtuggy:
> > We worked hard to get the mouse cursor to be the hardware mouse cursor. I'd rather not take a step backwards in that regard if we can help it. Please, can you split out the mouse-related changes into a separate PR, so we can discuss them separately? Or just remove them entirely.
> 
> This work you did is by no means lost because of this PR. Could I suggest that we follow-up with a PR which migrates the cursor to SDL and removes all the software mouse code? Two reasons, while the code changes are probably not too hard, I need to re-playtest everything which is quite time consuming. The PR in its current form works and reintroduces the mouse cursor in its intended form by using the drawing mechanism of the UI framework you chose, so it is not really inconsistent.
> 
> If it is a blocker I'll have a look, but that will take a little while.

I suppose that would be all right. @BenjamenMeyer any thoughts on this?

stephengtuggy (approved):
LGTM

SGD1953:
@evertvorster are you good with this?

evertvorster:
> @evertvorster are you good with this?

Sure buddy.. you don't need approval from **everyone**. I have not tested it, but the screenshots look cool. As soon as you have approval from one person you can merge. :) Unless I am mistaken?

SGD1953:
> > @evertvorster are you good with this?
> 
> Sure buddy.. you don't need approval from **everyone**. 

Yes, but since it broke master last time it felt appropriate to at least give you the chance to have a look.

> I have not tested it, but the screenshots look cool.

Glad you like them, most importantly the cross-system navigation becomes usable again and is better readable on larger screens.

> As soon as you have approval from one person you can merge. :) Unless I am mistaken?

Yes, but I cannot merge as I lack the permission, but if you are ok I'd ask @BenjamenMeyer to merge it

PR #1688, opened 2026-08-03
Fix prohibited upgrades doubling

### LINKED ISSUE #1630: Savegame file keeps growing
I played for a while I noticed how my computer is under heavy load when saving an loading. I looked into the file and it seems that we keep appending to `Prohibited_Upgrades` which leads to a (truncated) list like this (without the linebreaks, they are just there for readability):

```
mult_shady_moreupgrade:1
mult_shady_morethrust:1
mult_shady_moreshields:1
mult_shady_moregunrecharge:1
mult_shady_moreshieldrecharge:1
mult_shady_moreturning:1
upgrades/Overdrive:1
upgrades/Capacitors:1
upgrades/Reactors:1
upgrades/Armor:1
upgrades/Sensors:1
upgrades/Shield_Systems:1
upgrades/Passenger_Quarters:1
upgrades/Repair_Systems:1
upgrades/ECM_Systems:1
mult_shady_moreupgrade:1
mult_shady_morethrust:1
mult_shady_moreshields:1
mult_shady_moregunrecharge:1
mult_shady_moreshieldrecharge:1
mult_shady_moreturning:1
upgrades/Overdrive:1
upgrades/Capacitors:1
upgrades/Reactors:1
upgrades/Armor:1
upgrades/Sensors:1
upgrades/Shield_Systems:1
upgrades/Passenger_Quarters:1
upgrades/Repair_Systems:1
upgrades/ECM_Systems:1
mult_shady_moreupgrade:1
mult_shady_morethrust:1
mult_shady_moreshields:1
mult_shady_moregunrecharge:1
mult_shady_moreshieldrecharge:1
mult_shady_moreturning:1
upgrades/Overdrive:1
upgrades/Capacitors:1
upgrades/Reactors:1
upgrades/Armor:1
upgrades/Sensors:1
upgrades/Shield_Systems:1
upgrades/Passenger_Quarters:1
upgrades/Repair_Systems:1
upgrades/ECM_Systems:1
mult_shady_moreupgrade:1
mult_shady_morethrust:1
mult_shady_moreshields:1
mult_shady_moregunrecharge:1
mult_shady_moreshieldrecharge:1
mult_shady_moreturning:1
upgrades/Overdrive:1
upgrades/Capacitors:1
upgrades/Reactors:1
upgrades/Armor:1
upgrades/Sensors:1
upgrades/Shield_Systems:1
upgrades/Passenger_Quarters:1
upgrades/Repair_Systems:1
upgrades/ECM_Systems:1
mult_shady_moreupgrade:1
mult_shady_morethrust:1
mult_shady_moreshields:1
mult_shady_moregunrecharge:1
mult_shady_moreshieldrecharge:1
mult_shady_moreturning:1
```

My latest save is 700MB

### LINKED ISSUE #1687: Saved files are growing ridiculously large.
One of my testing save files is over 2.3Gb in size. 

On a little investigation it turns out that I have some prohibited upgrades, and every time the game is saved, the prohibited upgrades section is written twice. 

This causes logarithmic file size growth...

evertvorster:

Please answer the following:

Code Changes:
This does not affect gameplay, only stopping writing double entries for prohibited upgrades. When a saved game is written.
It is achieved by reading only unique upgrades and discarding the duplicates. Then, when saving, it only saves one copy. 

Issues:
- -Fixes #1687
- - Fixes #1630 

Purpose:
- - Fix a bug where save files get really large
- What release is this for?
- - Git master
N/A

### DIFF
--- a/engine/src/components/components_manager.cpp
+++ b/engine/src/components/components_manager.cpp
@@ -44,6 +44,10 @@ Resource<double> ComponentsManager::credits = Resource<double>(0.0, 0.0);
 void ComponentsManager::Load(std::string unit_key) {
     mass = base_mass = UnitCSVFactory::GetVariable(unit_key, "Mass", 0.0);
 
+    // Clear any previously-loaded prohibited upgrades so repeated Load() calls
+    // (e.g. the two calls made from Unit::LoadRow) do not accumulate duplicates.
+    prohibited_upgrades.clear();
+
     // Consumer
     std::string prohibited_upgrades_string = UnitCSVFactory::GetVariable(unit_key, "Prohibited_Upgrades", std::string());
 
--- a/engine/src/components/components_manager.cpp
+++ b/engine/src/components/components_manager.cpp
@@ -62,17 +62,26 @@ void ComponentsManager::Load(std::string unit_key) {
         std::vector<std::string> parts;
         boost::split(parts, upgrade, boost::is_any_of(":"));
         if (parts.size() == 1) {
-            const std::string& category = parts[0];
-            prohibited_upgrades.emplace_back(category, 0);
+            AddProhibitedUpgrade(parts[0], 0);
         } else if (parts.size() == 2) {
-            const std::string& category = parts[0];
-            const int limit = locale_aware_stoi(parts[1]);
-            //const std::pair<const std::string, const int> pair(category, limit);
-            prohibited_upgrades.emplace_back(category, limit);
+            AddProhibitedUpgrade(parts[0], locale_aware_stoi(parts[1]));
         } else {
             VS_LOG(error, (boost::format("%1%: Invalid format in prohibited upgrades string: %2%") % __FUNCTION__ % upgrade));
         }
     }
+}
+
+// The pre-existing doubling bug (Load() called twice per unit construction)
+// could have left many identical duplicate groups in a save's string. Adding
+// is done deduplicated so a bloated save is collapsed to one copy and repaired
+// on the next save.
+void ComponentsManager::AddProhibitedUpgrade(const std::string& category, int limit) {
+    for (const auto& existing : prohibited_upgrades) {
+        if (existing.first == category && existing.second == limit) {
+            return;
+        }
+    }
+    prohibited_upgrades.emplace_back(category, limit);
 }
 
 void ComponentsManager::Serialize(std::map<std::string, std::string>& unit) const {
--- a/engine/src/components/components_manager.h
+++ b/engine/src/components/components_manager.h
@@ -68,6 +68,9 @@ class ComponentsManager {
 
     std::vector<std::pair<const std::string, const int>> prohibited_upgrades;
 
+    // Deduplicated add — repairs the doubling-bug ballooned save lists.
+    void AddProhibitedUpgrade(const std::string& category, int limit);
+
     friend class CargoHold;
     friend class Movable;
 
--- a/libraries/cmd/unit_csv.cpp
+++ b/libraries/cmd/unit_csv.cpp
@@ -586,7 +586,17 @@ void Unit::LoadRow(std::string unit_identifier, string modification, bool saved_
     string tmpstr;
     csvRow = unit_identifier;
 
-    this->Load(unit_identifier);
+    // Load the named unit's components. For the LEGACY save-game format the
+    // resolved unit_key below is remapped to "player_ship", so this first
+    // load seeds the named ship's row before the remap; the Load(unit_key)
+    // call near the end then applies the saved per-ship state. For the NEW
+    // save-game format unit_key stays unit_identifier, making this first load
+    // redundant with the later one.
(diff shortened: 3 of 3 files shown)


stephengtuggy (commented):
The code looks good. This needs to be play-tested using the standard PR Validation Test plan, however. That test plan includes loading and saving, along with some additional features that we should test pretty nearly every time we make a functional change to the code.

Also, the `master` branch equals the 0.11.x release. At least for now. I have associated this PR with the 0.11.x project accordingly.

This is a good change that definitely needs to make it into the code. Thanks for the contribution, @evertvorster .

Question: Should we also update the code that calls Load twice, so that it doesn't do that anymore?

stephengtuggy:
We will probably want to port this back to 0.10.x as well.

stephengtuggy:
One more question. Should the prohibited upgrades be stored in a `std::set` or similar to optimize looking for duplicates? Looping through all the existing items in the list is not very efficient.

This might be a micro-optimization that doesn't really matter with the number of elements we're working with here. But I thought I would at least bring it up.

evertvorster:
> One more question. Should the prohibited upgrades be stored in a `std::set` or similar to optimize looking for duplicates? Looping through all the existing items in the list is not very efficient.
> 
> This might be a micro-optimization that doesn't really matter with the number of elements we're working with here. But I thought I would at least bring it up.

Yes, I briefly considered it. However, this was the simpler option code-wise. All this little routine does is fix save files that already have been accumulating duplicates in prohibited upgrades. 
Its used once per corrupted save file, and then never again. The real fix is not to produce duplicates in the first place, which is the main purpose of this PR.  

In other words, on a saved game with no duplicates, there is no benefit in efficiency looking for duplicates.

evertvorster:
> The code looks good. This needs to be play-tested using the standard PR Validation Test plan, however. That test plan includes loading and saving, along with some additional features that we should test pretty nearly every time we make a functional change to the code.
> 
> Also, the `master` branch equals the 0.11.x release. At least for now. I have associated this PR with the 0.11.x project accordingly.
> 
> This is a good change that definitely needs to make it into the code. Thanks for the contribution, @evertvorster .
> 
> Question: Should we also update the code that calls Load twice, so that it doesn't do that anymore?

Thanks for taking a look!
To answer your question, I don't know why we are calling load multiple times. There might be a good reason. However, we don't want stuff accumulating, so I thought it simpler to just clear before loading again. 

This PR fixes the issue, and I'm actually focusing on thoroughly play testing another branch that I'm working on. This bug showed itself while play-testing that. 
Its not related to the code changes I made in that branch, but now the fix is rolled into that branch, so it is getting a good workout right now.

SGD1953:
Love it, fixes the bug I reported! :+1: 

> Question: Should we also update the code that calls Load twice, so that it doesn't do that anymore?
Definitely worth looking into, but as this won't hurt I'm for merging.

SGD1953 (approved):

evertvorster:
> Love it, fixes the bug I reported! 👍
> 
> > Question: Should we also update the code that calls Load twice, so that it doesn't do that anymore?
> > Definitely worth looking into, but as this won't hurt I'm for merging.

This bug was getting in the way of my testing, and the fix is the answer to the question: What is the smallest change we can make that resolves the issue?

I did not know _why_ it was called twice, there might be a perfectly good reason. Without fully investigating this there is no way to be sure. What I did know for a fact is that you only need one list of prohibited upgrades.

evertvorster:
> The code looks good. This needs to be play-tested using the standard PR Validation Test plan, however. That test plan includes loading and saving, along with some additional features that we should test pretty nearly every time we make a functional change to the code.
> 
> Also, the `master` branch equals the 0.11.x release. At least for now. I have associated this PR with the 0.11.x project accordingly.
> 
> This is a good change that definitely needs to make it into the code. Thanks for the contribution, @evertvorster .
> 
> Question: Should we also update the code that calls Load twice, so that it doesn't do that anymore?

Would it be too terrible to merge this now, and open a issue to figure why this code is being called twice? 
I have added #1696 as a reminder to come back to this.

evertvorster:
OK... I should have done this earlier, was an easy fix. 
The double call was due to us supporting both old and new style saved games. The last commit here gates the calls depending on which type of saved game is loaded. 

Fixes #1696

royfalk:
> The double call was due to us supporting both old and new style saved games. The last commit here gates the calls depending on which type of saved game is loaded.

I wrote this code and I'm not convinced this is the case. I think I just made a mistake and called it twice.

I was unhappy with the solution proposed here so I made my own in #1701.
- It's shorter and cleaner.
- It removes the extra call. This is key. It was never intended to be called twice.
- It implements a map (not set) instead of vector. We need a pair and map gives us that.

I'm not wild about writing a second fix to a bug and potentially making the work done by you (Evert) redundant, either. And so I leave it in your hands to decide what to do, or rather which to merge.

evertvorster:
Superceded by Roy's more correct fix

PR #1692, opened 2026-08-04
Add files .clang-format and .clang-tidy

### LINKED ISSUE #1695: Add .clang-format and .clang-tidy to repository
> @SGD1953 wrote: `How about we start with a .clang-format and possibly a .clang-tidy which we commit. All of our IDEs would be able to pick them up and flag any issues. We could start with some industry standard like Google, Microsoft or the like.` Sounds like a plan! 

 _Originally posted by @stephengtuggy in [#482](https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/482#issuecomment-5174481665)_

stephengtuggy:

Please answer the following:

Code Changes:

Issues Addressed:
- Closes #1695 

Known Issues Remaining:
- Nothing new that I'm aware of

Purpose:
- What is this pull request trying to do? Add .clang-format and .clang-tidy files to the repository, thus implementing #1695 
- What release is this for? 0.11.x and following, I guess
- Is there a project or milestone we should apply this to? 0.11.x, I guess


SGD1953:
What about

```
clang-diagnostic-double-promotion,bugprone-misplaced-widening-cast
```

for `.clang-tidy`? This should check for overflows checked by CodeQL like 

> CodeQL / Multiplication result converted to larger typeMultiplication result may overflow 'float' before it is converted to 'double'.

royfalk (approved):
OK. It looks reasonable based on the explanation of Gemini.

One thing that jumped out to me was `modernize-*`, which advocates for modern C++ standards (like C++11 up to C++20/C++23). It flags outdated patterns and suggests modern replacements like nullptr, auto, override, and range-based for loops.

This is just a linter, so we can ignore it, but we're currently in C++14 and have self limited ourselves in use of auto to for loops (?).

SGD1953:
I think we can add

```
ExtraArgs:
  - '-std=c++14'
```

which modernize should respect.

.clang-tidy
```diff
@@ -0,0 +1,24 @@
...
+  - mpi-buffer-deref
+  - mpi-type-mismatch
+  - openmp-use-default-none
+  - abseil-*
+  - misc-*
+  - -misc-multiple-inheritance
+  - bugprone-*
+  - clang-analyzer-*
+  - llvm-*
+  - modernize-*
```
BenjamenMeyer, line 22: per @royfalk 's comment about `modernize-` - from what I can tell we need to add the following config:
```
CheckOptions:
  - key: modernize-use-auto.MinTypeNameLength
    value: '999'
```
Google is saying that will block it for short types while leaving it in place for the iterators and other things we care about it. Still might need tweaking but it's probably a start.

Per https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-auto.html#cmdoption-arg-MinTypeNameLength it defaults to `5` so we should definitely raise it up. 999 might be too long; but it should probably be pretty high.

BenjamenMeyer (changes requested):
we should configure modern-use-auto to have a more reasonable mintypenamelength value per our coding style than its default of `5`.

SGD1953:
I copied the files into my workspace, I think we are good to go here as a baseline, we can always refine when we feel we need to. As a minimum things will be consistent.

.clang-tidy
stephengtuggy: @BenjamenMeyer OK, applied your suggestion.

stephengtuggy:
> What about
> 
> ```
> clang-diagnostic-double-promotion,bugprone-misplaced-widening-cast
> ```
> 
> for `.clang-tidy`? This should check for overflows checked by CodeQL like
> 
> > CodeQL / Multiplication result converted to larger typeMultiplication result may overflow 'float' before it is converted to 'double'.

Those are good suggestions. I believe `bugprone-misplaced-widening-cast` is already included as part of `bugprone-*` on line 19 or so. As for `clang-diagnostic-double-promotion`, I have added it to the file.

stephengtuggy:
> we should configure modern-use-auto to have a more reasonable mintypenamelength value per our coding style than its default of `5`.

Done.

stephengtuggy:
> I think we can add
> 
> ```
> ExtraArgs:
>   - '-std=c++14'
> ```
> 
> which modernize should respect.

Added.

BenjamenMeyer (approved):

Issue #1698, opened 2026-08-05
Capacitor drains and never refills on stock Llama


evertvorster:
# Bug: Shield maintenance drains the capacitor even at full shields

**Status:** ROOT-CAUSED + FIXED on branch `fix-shield-maintenance-drain` (commit `7e1414e9f`, off master `36206c257`), 2026-08-04. Play-test pending (Evert). Evert will file the issue and PR.
**Files:** `engine/src/components/shield.cpp` (Regenerate), config.json `components.shield.*`
**Reported by:** Evert (2026-08-04) — capacitor drains in flight; Llama no longer refills after firing.

## Symptom

- The ship's **main capacitor drains continuously while flying**, even with no weapons firing.
- The **Llama used to refill its capacitor after firing** (pre-2025) — now it doesn't.
- Save evidence (HeadingOut_2, player_fleet.json): the legit ships sit near full (Llama 197/200 = 98.5%, Dostoevsky 98.5/100) while the upgraded Vendetta is at **213.5/1500 = 14%** and keeps draining.
- Any ship with a large shield upgrade drains fast: Vendetta (quadshield10 = 4×1000 shield) goes 1500→0 in ~18s of flight.

## What used to happen (pre-Feb 2025) — the OLD model

Before commit `ca3df7f9c` (Feb 18 2025, PR #1016 "Implement hull, armor and shield components"), shield energy use lived in `Energetic` (engine/src/cmd/energetic.cpp):

**`ExpendEnergyToRechargeShields()`** (old energetic.cpp:173):
```cpp
double maximum_charge = std::min(max_shield_value - current_shield_value, regeneration);
// At FULL shields: max - current = 0  →  maximum_charge = 0  →  NO energy spent
...
double energy_required_to_charge = actual_charge * VSDPercent() * simulation_atom_var;
ExpendEnergy(energy_required_to_charge);
```

**`MaintainShields()`** (old energetic.cpp:157):
```cpp
const float shield_maintenance = unit->shield->GetRegeneration() * VSDPercent()
        / configuration()->physics_config.shield_energy_capacitance  // 0.2
        * shield_facets * configuration()->physics_config.shield_maintenance_charge  // 0.25
        * simulation_atom_var;
```

Key properties of the OLD model:
1. **Full shields → zero energy cost** (charge = min(deficit, regen) = 0).
[...]

PR #1699, opened 2026-08-05
Fix shield maintenance draining the capacitor at full shields and lasers getting ammo.

evertvorster:

Please answer the following:
The author of the changes should run the game and do the actions described below. Starting a new game, and completing these steps in this order should take around 10-15 minutes.

- [x] Basic flight.
- [x] Docking to a planet.
- [x] On planet actions:
    - [x] Save the game.
    - [x] Read news.
    - [x] Buy/Sell some goods.
- [x] SPEC flight.
- [x] Save the game again.
- [x] Upgrade your new ship.
- [x] Save once more.

Code Changes:
Added in https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1681
and
https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1688
To ease play testing. Review those first and they'll disappear from the changes here, if approved.

Issues:
- #1698 
- #1700

Purpose:
- Fix two simple bugs that affect game play. Full documentation on the bugs are linked above.
- What release is this for?
- 0.11
- 0.11


engine/src/cmd/basecomputer.cpp
```diff
@@ -2648,8 +2648,12 @@ void BaseComputer::BuyUpgradeOperation::concludeTransaction(void) {
 
     if(playerUnit->BuyUpgrade(baseUnit, &m_part, quantity)) {
         double percent = 0;
-        // Modify upgrade quantity according to cargo quantity
-        m_newPart->mounts[0].ammo = quantity;
+        // Only missiles carry ammo. Non-missile weapons (beams, bolts) are
```
royfalk, line 2651: This part is fine.

engine/src/components/shield.cpp
```diff
@@ -340,34 +340,35 @@ void Shield::Regenerate(const bool player_ship) {
...
-    const double shield_maintenance_cost = TotalMaxLayerValue() * configuration().components.shield.maintenance_factor_dbl;
-    SetConsumption(shield_maintenance_cost);
-    const double actual_maintenance_percent = Consume();
-    if(Percent() > actual_maintenance_percent) {
-        Decrease();
-        return;
-    }
-
-    // Manually throttle shield strength
+    // Manually throttle shield strength (must run before the fully-charged
```
royfalk, line 343: As noted in the review of the issue, this hard codes a value in the engine that can be set in the assets.

engine/src/components/shield.cpp
```diff
@@ -340,34 +340,35 @@ void Shield::Regenerate(const bool player_ship) {
...
-    double regen = actual_regeneration_percent * regeneration.AdjustedValue() * simulation_atom_var;
+    /* Shield Regeneration - energy-limited by what the reactor can supply.
+    * The energy cost scales with the shield deficit being rebuilt, so a
+    * heavily damaged shield demands proportionally more power. If the
+    * reactor can't supply the full cost, Consume() returns the fraction it
+    * could actually provide and regen slows by that same fraction - e.g.
+    * half the energy -> half the recharge rate -> twice as long to fill
+    * (matches the pre-2025 behavior).
+    */
+    const double shield_recharge_cost = (TotalMaxLayerValue() - TotalLayerValue()) * configuration().components.shield.maintenance_factor_dbl;
```
royfalk, line 368: Also, this changes the behavior of the shield in a way I don't agree with.

royfalk (changes requested):
Bit of  a mixed review. 
First bug fix is good. 
The second is IMO not an engine bug.

Don't be shy about making changes to the assets. You have strong opinions and actually play the game.

engine/src/components/shield.cpp
evertvorster: Ah. I do have a very strong opinion here. Vega Strike 0.5 played well. Really well... well enough to have it still hack at it 20 years later. One part of that playing well were shields that behaved well.

You may not agree with how shields worked in VegaStrike, but they are a core part of the game as they were, and having them work in a different way is a bug. As it stands now, if you take a stock llama off of a new campaign, the shields drain and never recover. You would have noticed this if you fought anybody during play testing. 

As shields are implemented now breaks the game totally.

This fix reverts back the shields to mostly  how the game was designed, and adds a little dynamicism that "feels" really well when you play. 

Now, we can have a chat about how shields should work, and see if there is a way to reach a common ground, but for now, this fix makes the game playable again, and IMHO improves is a little bit.

engine/src/cmd/basecomputer.cpp
evertvorster: Thanks, that fixes the ammo on lasers bug.

evertvorster: Thank you for reviewing this.

engine/src/components/shield.cpp
evertvorster: We have over 1400 settable options, what's one more? 

However, if we merge this fix as is, then I can open an issue to create a setting and put a default value in it. That has to happen first before I can have the engine support having an extra setting. 

So, lets open an issue to return to shields and work through how you think shields should work, and then we'll discuss how we can get that into the game that makes playing the game still fun.

engine/src/components/shield.cpp
stephengtuggy: Agreed that the existing behavior of the shields is completely broken and needs to be fixed.

Can't we just remove the ` - TotalLayerValue()` clause on this line, and merge the rest? The rest of @evertvorster 's changes here look good to me.

engine/src/components/shield.cpp
stephengtuggy: @royfalk I don't understand. What value is being hard-coded?

royfalk: > We have over 1400 settable options, what's one more?

1401? But seriously, _components.shield.maintenance_factor_ is one of these 1400 settings.  It already exists.

> @royfalk I don't understand. What value is being hard-coded?

When you delete this code:
```
// Shield Maintenance
    // TODO: lib_damage restore efficiency by replacing with shield->efficiency
    //const double efficiency = 1;

    const double shield_maintenance_cost = TotalMaxLayerValue() * configuration().components.shield.maintenance_factor_dbl;
    SetConsumption(shield_maintenance_cost);
    const double actual_maintenance_percent = Consume();
    if(Percent() > actual_maintenance_percent) {
        Decrease();
        return;
    }
```
you are effectively hard coding _components.shield.maintenance_factor_ to 0.

engine/src/components/shield.cpp
royfalk: > Ah. I do have a very strong opinion here. Vega Strike 0.5 played well. Really well... well enough to have it still hack at it 20 years later. One part of that playing well were shields that behaved well.

I agree. I just think that tweaking a number is a simpler solution.

> Can't we just remove the  - TotalLayerValue() clause on this line, and merge the rest?

You can but that would change the code from what was in 0.5 as the AI itself documents:
> The energy cost scales with the shield deficit being rebuilt, so a heavily damaged shield demands proportionally more power.

The old code was:
```cpp
double maximum_charge = std::min(max_shield_value - current_shield_value, regeneration);
// At FULL shields: max - current = 0  →  maximum_charge = 0  →  NO energy spent
...
double energy_required_to_charge = actual_charge * VSDPercent() * simulation_atom_var;
```

This is also true in the my code. A fully charged shield needs no regeneration cost, only maintenance.

evertvorster:
Superceded by other PRs.

engine/src/components/shield.cpp
evertvorster: Yes, you are right, and I was wrong. 
I'm right now play testing a PR against Assets that sets this setting. 
The other shield thing I'm clamouring for I'm putting into a separate PR all by itself, and I have approved your laser ammo fix PR, so effectively this PR has no point, and I closed it.

PR #1713, opened 2026-08-08
Energy-limited shield recharge: scale recharge cost by shield deficit

evertvorster:

Please answer the following:

Code Changes:
  - [x] Demonstrate that the issue is fixed or the feature exercised
    - [x] Specific testing where a clear, present, demonstrable test can be provided; OR
    - [x] Play test (see below) that demonstrates the feature/issue; OR
    - [x] Build test when things directly impact how the build functions or some part of the build is generated
  - [x] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - In the event that a user has a shield that requires for some or the other reason more power than the reactor can give, maybe it got damaged or the user made a stupid mistake in buying it.  The current code disallows the shield to work at all, but it should work in a degraded fashion. IMHO. 

Purpose:
- - This fix would allow the user to still charge up the shield, but at a much lower rate, which is determined by the shortfall in power delivery. 

Ie, if the shield requires 100Mj/s to charge at its optimum rate, but the reactor can only deliver 90Mj/s, then the shield will charge at 90% of its normal speed. If the shortfall is much larger, then the shields will charge much, much slower, and be practically useless in a fight. 

Think of it a little like charging your Tesla at the wall outlet. Its slow, and a pain, but it might just give you a little hope. :)

- What release is this for?
- 0.11

### DIFF
--- a/engine/src/components/shield.cpp
+++ b/engine/src/components/shield.cpp
@@ -364,7 +364,10 @@ void Shield::Regenerate(const bool player_ship) {
         return;
     }
 
-    const double shield_regeneration_cost = regeneration.AdjustedValue() * configuration().components.shield.regeneration_factor_dbl;
+    // Energy-limited recharge: the cost scales with the deficit being rebuilt, so a
+    // shield too big for the reactor charges at a reduced rate (slower) rather than
+    // at full speed. (Consume() returns the fraction the reactor could actually supply.)
+    const double shield_regeneration_cost = (TotalMaxLayerValue() - TotalLayerValue()) * configuration().components.shield.regeneration_factor_dbl;
     SetConsumption(shield_regeneration_cost);
     const double actual_regeneration_percent = Consume();
     double regen = actual_regeneration_percent * regeneration.AdjustedValue() * simulation_atom_var;


royfalk:
```cpp
const double shield_regeneration_cost = (TotalMaxLayerValue() - TotalLayerValue()) * 
   configuration().components.shield.regeneration_factor_dbl;
```

Let's take some simple numbers.
TotalMaxLayerValue = 100
regeneration_factor_dbl = 1

| TotalLayerValue | Cost |
| --- | --- |
| 0 | 100 |
| 10 | 90 |
| 20 | 80 |
| 30 | 70 |
| 40 | 60 |
| 50 | 50 |
| 60 | 40 |
| 70 | 30 |
| 80 | 20 |
| 90 | 10 |
| 100 | 0 |

That means that if I get hit in combat with an overpowered shield and my shield needs to recharge, the more I'm hit the less chance I have to actually charge it.

Did you mean?
```cpp
const double shield_regeneration_cost =  TotalLayerValue() * 
   configuration().components.shield.regeneration_factor_dbl;
```

> Ie, if the shield requires 100Mj/s to charge at its optimum rate, but the reactor can only deliver 90Mj/s, then the shield will charge at 90% of its normal speed.

That actually already happens below:
```cpp
const double actual_regeneration_percent = Consume(); // let's say you get 90%
double regen = actual_regeneration_percent * regeneration.AdjustedValue() * simulation_atom_var; // You'd get 90% of regen here.
```

evertvorster:
OK, on double-checking you are right. I'll close this PR. 

On a side note, shields are supposed to cost more when they are down, its good game design. It also fits with shields not using much power when fully charged. We also could consider a constant charge rate, where shields recharge at the same rate until its full, when it drops to maintenance rate. 

The game punishes you for letting your shields get down to nothing... this is also the spot where you usually die. 
It builds suspense, and gives you the feeling of danger... It makes it feel more real. 

This might be something we might consider in the future for a difficulty setting in the game. We could go from brutal to realistic to easy, and stuff like this we could then adjust based on the setting. I like my games brutal and difficult, and others might like something else. Something to think about.

royfalk:
> On a side note, shields are supposed to cost more when they are down, its good game design. It also fits with shields not using much power when fully charged. 

Open a discussion on it. If you get buy-in from others, it should be trivial to add as an option to the current implementation. I personally like my game easy.

> We also could consider a constant charge rate, where shields recharge at the same rate until its full, when it drops to maintenance rate.

I think that's how we do it now. As long as you have the power and your shields are not at max, the regen rate is linear.

stephengtuggy:
There is a difficulty setting in the game. It isn't working at the moment. I don't think it's being set, and its setting may not be honored everywhere it should be, either.

stephengtuggy:
> > We also could consider a constant charge rate, where shields recharge at the same rate until its full, when it drops to maintenance rate.
> 
> I think that's how we do it now. As long as you have the power and your shields are not at max, the regen rate is linear.

This is how I would like to see it work. It is also how it has worked in the past, AFAIK.

evertvorster:
Done via a different angle. Closing

Issue #1707, opened 2026-08-07
Dear ImGui fails to initialize when run in VM without GPU passthrough


stephengtuggy:
I have further confirmed that once those two problems are solved, the game gets a little further, but still fails to start. This time the problem is Dear ImGui looking for an OpenGL 3 backend, and only finding an OpenGL 1.1 implementation, due to my Windows VM not having working GPU passthrough, hence falling back to plain old GDI/OpenGL 1.1 software emulation.

What are we going to do about this last issue, you guys? It might really hinder our release testing efforts if we can't test the game in virtual machines without GPU passthrough.

_Originally posted by @stephengtuggy in https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1683#issuecomment-5212557318_

PR #1720, opened 2026-08-15
Fix crash due to incorrect ImGUI frame nesting

### LINKED ISSUE #1717: Crash when running in debug mode
Game crashes with 
```
Launching bases for Special/Empty
Adding news
conditioning
nonzeroing
[{"Afterburner_Accel":"17000.00/17000.00/17000.00","Afterburner_Speed_Governor":"120.00/120.00/125.0
Parsing 0
Create old format player ship Llama.begin with index 0, faction 25 and flight group Shlimazel
vegastrike-engine: /home/roy/git/vegastrike/Vega-Strike-Engine-Source/libraries/imgui/imgui.cpp:10778: void ImGui::ErrorCheckNewFrameSanityChecks(): Assertion `(g.FrameCount == 0 || g.FrameCountEnded == g.FrameCount) && "Forgot to call Render() or EndFrame() at the end of the previous frame?"' failed.
```

This only happens in Debug because Release strips out the assert.

**Core Dump**
```
#8  0x0000650faf43ca9c in ImGui::NewFrame ()
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/libraries/imgui/imgui.cpp:5410
#9  0x0000650faf0a0474 in bootstrap_draw (message="Loading black.png", 
    newSplashScreen=0x0)
    at /home/roy/git/vegastrike/Vega-Strike-Engine-Source/engine/src/main.cpp:481
#10 0x0000650faed22157 in Texture::Load (this=0x650fe4cdbd50, 
    FileName=0x650fe43f5b28 "black.png", stage=0, mipmap=MIPMAP, 
    target=TEXTURE2D, imagetarget=TEXTURE_2D, force_load=0 '\000', 
    maxdimension=65536, detailtexture=0 '\000', nocache=0 '\000', 
    address_mode=DEFAULT_ADDRESS_MODE, main=0x0)

```

SGD1953:
This fixes #1717, it is not the most elegant solution but it implements the desired behavior.

Please answer the following:

Code Changes:
- [X] Have the PR Validation Tests been run?
  - [X] Demonstrate that the issue is fixed or the feature exercised
    - [X] Specific testing where a clear, present, demonstrable test can be provided - no more crash

Issues:
- #1717

Purpose:
- this properly ends and creates imGUI frame preventing incorrect nesting
- currently the game is creating a new frame while one is already open and draws it which then crashes when on draw as this is not supposed to happen.


engine/src/main.cpp
```diff
@@ -466,6 +467,10 @@ void bootstrap_draw(const std::string &message, Animation *newSplashScreen) {
         }
     }
 
+    // Automatically detect if ImGui is currently between NewFrame() and Render()
+    ImGuiContext* g = ImGui::GetCurrentContext();
+    const bool owns_frame = (g == nullptr || !g->WithinFrameScope);
```
royfalk, line 475: This is actually quite elegant.

engine/src/main.cpp
```diff
@@ -475,12 +480,14 @@ void bootstrap_draw(const std::string &message, Animation *newSplashScreen) {
...
-    ImGui_ImplOpenGL3_NewFrame();
-    ImGui_ImplSDL3_NewFrame();
-    ImGui::NewFrame();
-    // End ImGui Init
-
+    // create new frame if none is active
+    if(owns_frame) {
+        // ImGui Init
+        ImGui_ImplOpenGL3_NewFrame();
+        ImGui_ImplSDL3_NewFrame();
```
royfalk, line 491: Why are we using **both** backends?

royfalk (approved):
LGTM.

engine/src/main.cpp
SGD1953: No idea, kept that as I assumed the original author had a reason.

BenjamenMeyer: can we please test to see if we can only use one? that might help solve #1707.

engine/src/main.cpp
```diff
@@ -495,14 +502,23 @@ void bootstrap_draw(const std::string &message, Animation *newSplashScreen) {
...
     ImGui_ImplOpenGL3_RenderDrawData(ImGui::GetDrawData());
     SDL_Window* current_window = SDL_GL_GetCurrentWindow();
     SDL_GL_SwapWindow(current_window);
     // End ImGui
 
+    // open a new frame if there was an active one
+    if(!owns_frame) {
+        // ImGui Init
+        ImGui_ImplOpenGL3_NewFrame();
+        ImGui_ImplSDL3_NewFrame();
```
BenjamenMeyer, line 522: same here

engine/src/main.cpp
SGD1953: Which one? SDL3?

SGD1953: Commented each of the backend -> game crashes. Hence I'd leave it for now, if we feel this needs to be looked into we can file a follow-up PR.

BenjamenMeyer: given this is on master/0.11.x we probably should be using only the SDL3 one here, thus removing the OpenGL one.

SGD1953: As said, that leads to a crash

stephengtuggy: We should settle on one backend eventually, probably the SDL3 one. But both backends are used across multiple locations in the code currently. They all need to be updated together.

If we can get this done in this PR, great. But I don't think it's a blocker for this pull request.

stephengtuggy: If we settle on the SDL3 backend, that might have the advantage of allowing testing on virtual machines without 3D graphics acceleration again. (See #1707 )

stephengtuggy: Filed as #1732

SGD1953:
@stephengtuggy @BenjamenMeyer are we good to merge here?

engine/src/main.cpp
```diff
@@ -86,6 +86,7 @@
 #include "src/vs_exit.h"
 
 #include "imgui/imgui.h"
+#include "imgui/imgui_internal.h"
```
stephengtuggy, line 89: Why are we including an "internal" header file? That seems less than ideal.

stephengtuggy: What are we using from it?

SGD1953: I agree, line 472 is the key element here to figure out if there is a frame already and unfortunately it is only accessibly via internals

stephengtuggy: Got it. OK. It is what it is then.

stephengtuggy (approved):
LGTM

PR #1016, opened 2025-01-30
Implement hull, armor and shield components

royfalk:
- Delete Health class. Very similar to Resource.
- Move functions to components - e.g. Percent, Get,
- Move shield functionality like AdjustPower from DamageableLayer to Shield class
- Disable enhance shields from collision. This should probably be handled by a python script.
- Delete many functions that are now unused.
- Delete x4 sparkling missile
- Delete planet.cpp fawg weird hull value
- Delete incorrect GetHullPercent by planet. For now it can be damaged and destroyed.
- Move some python API from macros to actual code.
- Fix issue in energy_container, where Capacitor and FTL Capacitor were not saved.
- Add two shield factors in configuration. Currently not read from file.

This PR requires a matching Assets PR - https://github.com/vegastrike/Assets-Production/pull/155.

Code Changes:
- [x] Have the PR Validation Tests been run? See https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation. Also, tests of the actual components. 

Issues:
- Damage capability of missiles and guns in ship_view.py and upgrade_view.py seems high. There's an open issue somewhere to that effect already.
- Reload after death - progeny had no image in VDU and when docked.
- Landed on a rocky planet after reload - it had no features. Couldn't take off either. I think this area of the game is quite buggy.
- Fixes #1011 
- Fixes #1008 
- Fixes #1007 (probably)
- Fixes #1006 
- Fixes #1019 
- Possibly fixes #1025. Game now saves (again) a bunch of stuff including unit scale.
- Convert Radar_Range to int before saving - prevent exponent format for upgrade view.
- Fix issue in armor and shield, where Resource was initiated without max value. Use proper string constructor instead.
- Fix issue in armor, hull and shield, where PercentOperational did not use actual libdamage damage.
- Fix issue in EnergyContainer applied modifier in the wrong direction. Also, use Serialize to save all three values.
- Fix issue in Reactor, where Serialize was not used.
- Game crashes with the following messages:
```
  fatal force
  fatal error in camera
```
   Recorded in #1013 . Will keep an eye on it.

PR #1726, opened 2026-08-19
Fix shield maintenance draining the capacitor at full shields

evertvorster:
Regression from ca3df7f9c (Feb 2025, PR #1016): Shield::Regenerate drained max_shield x maintenance_factor from the primary capacitor every atom, even at full shield health. For ships whose shield max exceeds their reactor output (e.g. Mule 18600 shields vs 80/s reactor), that left the weapon capacitor empty -- so neither the shields nor the weapons could function. AI ships with such shields never fired (Armed::Fire saw energy 0), and capital-ship shields disappeared.

Why tuning maintenance_factor can't fix this for large vessels:
  maintenance = max_shield x factor, but reactor output is a fixed per-ship
  value. For a ship to refill its capacitor you need maintenance < reactor,
  i.e. factor < reactor/max_shield. A Mule (18600 shields, 80/s reactor) needs
  factor < 0.0043; a capital ship (40000 shields, 115/s reactor) needs
  factor < 0.0029. No single global factor works for all ships -- the drain
  must not scale with max shield at all.

Fix (per devs' linear-recharge preference): move the fully-charged gate to the top of Regenerate, before any energy consumption. Full shields now cost 0 capacitor energy; the linear regeneration below is unchanged. This restores the pre-2025 behavior where shields only spend energy while actually recharging.

TL;DR The game was **designed** around full shields not costing anything, and any value that would allow a capship to keep its shields are effectively 0. So, lets not beat around the bush here and make it 0. 

Or, you know we spend the next few years re-balancing the game.

### DIFF
--- a/engine/src/components/shield.cpp
+++ b/engine/src/components/shield.cpp
@@ -340,6 +340,18 @@ void Shield::Regenerate(const bool player_ship) {
     * Finally, you adjust whatever used it to the value in question
     */
 
+    // Fully charged shields need no energy - return before any consumption.
+    // (Regression fix: the maintenance drain below used to run even at full
+    // shields, draining the primary capacitor (which the weapons also use) at
+    // max_shield x maintenance_factor per second. For ships whose shield max
+    // exceeds their reactor output - e.g. a Mule (18600 shields) or any
+    // capital ship - that left the capacitor empty, so neither the shields nor
+    // the weapons could ever function. See the 'capacitor never refills after
+    // firing' / 'AI ships never fire' reports.)
+    if (TotalLayerValue() == TotalMaxLayerValue()) {
+        return;
+    }
+
     // Shield Maintenance
     // TODO: lib_damage restore efficiency by replacing with shield->efficiency
     //const double efficiency = 1;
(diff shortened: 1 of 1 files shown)


stephengtuggy (approved):
Isn't there supposed to be a shield maintenance factor that is different from the shield recharge factor? I.e., shields at full health aren't supposed to take zero energy, but they are supposed to draw a lot less than while recharging.

Thank you for going with the linear refill rate. I appreciate that.

Ultimately, I think we need to go back to something very similar to the 0.5 formula, perhaps excluding the unit-of-measure weirdness and the `if` block that makes no difference to the end result. I understand that this formula is complicated, though, and not the easiest to work with. I plan to take a stab at making this change myself.

Your PR will work for now though, @evertvorster , as a short-term fix. I agree that it's urgent that we get the shields and capacitors charging again.

evertvorster:
> Isn't there supposed to be a shield maintenance factor that is different from the shield recharge factor? I.e., shields at full health aren't supposed to take zero energy, but they are supposed to draw a lot less than while recharging.
> 
> Thank you for going with the linear refill rate. I appreciate that.
> 
> Ultimately, I think we need to go back to something very similar to the 0.5 formula, perhaps excluding the unit-of-measure weirdness and the `if` block that makes no difference to the end result. I understand that this formula is complicated, though, and not the easiest to work with. I plan to take a stab at making this change myself.
> 
> Your PR will work for now though, @evertvorster , as a short-term fix. I agree that it's urgent that we get the shields and capacitors charging again.

There is, but it does not scale well with larger shields. Capships lost their shields with just about any value above 0. 

We can always come back and revisit this one. 

I just don't see the point, though. We'd have to carefully balance all the shield sizes in assets, then play-test just about every ship, many, many PRs... all to get back to where we are today.

Issue #1661, opened 2026-07-18
Consistent font handling with ease of use


SGD1953:
### Objective

This is an umbrella enhancement issue to collect the work that has been going into migrating fonts from the legacy glut stroke and bitmap fonts to a more modern way of handling them via ImGUI (based on teh amazing work by @royfalk in #1487 ). Ideally we can use the opportunity to simplify the configuration as well.

### Fixes / Restoring functionality

- [x] font_point config has no impact - #1666 
- [x] base computer (cargo, missions, news, load/save screens) still leverage legacy fonts - #1666 
- [x] controls_factory getControl(): font_array doesn't have enough elements errors - https://github.com/vegastrike/Assets-Production/pull/252
- [ ] reestablish font type configuration in config (currently hard-coded to roboto)

### Ease of use / Configuration

There are a quite some configurations around fonts, not all of them make sense after the rework.

```
"graphics":
        "font": "helvetica18",
        "font_antialias": true,
        "font_point": 32.0,
        "font_width_hack": 0.925,
        "gui_font_scale":0.55,
        "high_quality_font": false,
        "high_quality_font_computer": true,
        "high_quality_sprites": false,
        "bases":
            "font": "helvetica18",
```

- [ ] Remove `high_quality_*` settings, in the new world we can control font size via the point setting which could be adjusted in game (i.e. for larger headings)
- [ ] Remove font_width_hack - this is no longer needed from my perspective, but still used in a few spots
- [ ] font_width_hack - questionable as well, still leveraged in #1660 but it sort of neutralizes itself with font_width_hack, I feel that this should just be a constant (0.5)
- [ ] font_antialias - this is legacy handling too for the bitmap fonts, in principle controlling font AA does make sense

### Related issues

- #671 
- #914
- #195

PR #1729, opened 2026-08-20
Added config.json font config option and added RobotoMono as default font

SGD1953:
Please answer the following:

Code Changes:
  - [X] Demonstrate that the issue is fixed or the feature exercised
    - [X] Play test (see below) that demonstrates the feature/issue; OR
  - [X] Play Tests
    - [X] From the Main Menu, start a new Campaign.
    - [X] Basic flight.
    - [X] Docking to a planet.
    - [ ] On planet actions:
        - [ ] Save the game.
        - [X] Read news.
        - [ ] Buy/Sell some goods.
    - [ ] SPEC flight.
    - [ ] Save the game again.
    - [ ] Upgrade your new ship.
    - [ ] Save once more.

Issues:
- #1661 - this adds support for TTF and OTF fonts
- Includes RobotoMono font and the corresponding license as default for backward compatibility, this is used when the configured font in the assets directory cannot be found
- regular font lookup happens in the data dir's `fonts` directory
- the touched files contain some unrelated linter fixes as well (i.e. `F` added or capitalized for floats)
- various versions of a font are loaded including italic and bold - currently the game only utilizes bold but this allow for a richer experience moving forward
- if we agree that we stick to this font, a fast follow adjusting the font sizes in the assets repo would make some of the screens like news more readable (font is currently a bit too tiny for my taste)
- all font sizes are now a function of the window height and scale with resolution
  - we may want to consider having a scale factor here (we could use `gui_font_scale`)
  - there are limits however which can make certain text invisible or run over if you scale too much
  - scale factor is better than a fixed size as not all texts are the same size, this allows for larger headings, smaler footnotes etc.
- a separate PR in Assets will introduce the agreed Saira font using the engine capabilities introduced here

A few screenshots (using Saira)
grafik
grafik
grafik
grafik
grafik

Screenshots with default (RobotoMono)
image
image
image
image

Feedback welcome!


stephengtuggy:
OK, that hangar view doesn't look right. The ship is squished horizontally. Not the right aspect ratio at all.

And why Zekton? I thought we had converged on Saira as the best default.

We need to think some more about how the font sizes are determined. I'm not convinced that we have a good approach figured out yet.

stephengtuggy:
See https://github.com/orgs/vegastrike/discussions/1722 for a few more details. Although most of the discussion took place in Gitter/Matrix, unfortunately, making it difficult to capture.

SGD1953:
> OK, that hangar view doesn't look right. The ship is squished horizontally. Not the right aspect ratio at all.

That's unrelated and due to the aspect set to my screen aspect. For the screenshots I have changed the game resolution to make the images a bit smaller.

> And why Zekton? I thought we had converged on Saira as the best default.

I have updated the PR, it is Saira.

> We need to think some more about how the font sizes are determined. I'm not convinced that we have a good approach figured out yet.

I think the approach chosen is actually quite consistent. The base computer uses a relative font size in norm space, this PR replicates this approach to the rest of the game. We should probably make this a config, but as config is currently being worked on I'd prefer to do this as a fast-follow.

stephengtuggy:
> > OK, that hangar view doesn't look right. The ship is squished horizontally. Not the right aspect ratio at all.
> 
> That's unrelated and due to the aspect set to my screen aspect. For the screenshots I have changed the game resolution to make the images a bit smaller.
> 
> > And why Zekton? I thought we had converged on Saira as the best default.
> 
> I have updated the PR, it is Saira.
> 
> > We need to think some more about how the font sizes are determined. I'm not convinced that we have a good approach figured out yet.
> 
> I think the approach chosen is actually quite consistent. The base computer uses a relative font size in norm space, this PR replicates this approach to the rest of the game. We should probably make this a config, but as config is currently being worked on I'd prefer to do this as a fast-follow.

OK, sounds good. One idea is to use a relative font size, as we are doing now, but with a minimum size setting to ensure that the font doesn't get too small even in small windows/resolutions. But we can probably tackle that in a future PR.

engine/src/gldrv/winsys.cpp
```diff
@@ -559,7 +559,16 @@ static bool setup_sdl_video_mode() {
     get_screen_measurements();
 
     // Initialize imgui
-    InitGui(window, &context, configuration().graphics.font_point_dbl);
+    namespace fs = boost::filesystem;
```
stephengtuggy, line 563: Nice!

engine/src/gui/imguitext.cpp
```diff
@@ -446,7 +441,9 @@ FormattedLayout ImGuiText::parseText(const std::string& input, const float width
             }
             // Process format
             curPos++; // Move past '#'
-            if (curPos >= endPos) break;
+            if (curPos >= endPos) {
+                break;
+            }
```
stephengtuggy, line 446: Good catch!

engine/src/gui/imguitext.h
```diff
@@ -92,7 +92,7 @@ class ImGuiText {
     }    
     // Draws the text
     void draw(int firstLineToDraw=0);
-    ImVec2 getTextWidth(const std::string text, const float fontSize);
+    ImVec2 getTextWidth(std::string text, Font fontSize);
```
stephengtuggy: I wonder if this should be:


```suggestion
    ImVec2 getTextWidth(std::string text, const Font& fontSize);
```

so that the Font object doesn't have to be copied every time this method gets called. Just a minor suggestion.

engine/CMakeLists.txt
```diff
@@ -801,6 +801,11 @@ ELSE (WIN32)
     INSTALL(PROGRAMS "${Vega_Strike_BINARY_DIR}/vegastrike-engine" RENAME vegastrike DESTINATION bin)
 ENDIF (WIN32)
 
+# Copy fonts
+FILE(
+    COPY "fonts"
+    DESTINATION ${CMAKE_BINARY_DIR}
+)
```
stephengtuggy, line 808: Thanks for adding this in here. This is something that we should not forget whenever we add files or directories to the game.

libraries/gui/gui.cpp
```diff
@@ -60,8 +62,53 @@ void InitGui(SDL_Window *window, const SDL_GLContext *context, const float fontS
...
     io.Fonts->Clear();
     ImFontConfig cfg;
-    cfg.SizePixels = fontSize;
-    io.FontDefault = io.Fonts->AddFontDefault(&cfg);
+    cfg.OversampleH = 3; // 3x horizontal oversampling for smooth edges
+    cfg.OversampleV = 2; // 2x vertical oversampling
+    cfg.PixelSnapH = true; // Align glyphs to whole pixels to prevent blurriness
+    // cfg.GlyphOffset.y = -1.5F;
+
+    // Default font-size is 0.1 norm height
```
stephengtuggy: Hmm. This comment doesn't seem to match the code below it. Maybe it does, but if so, that isn't obvious. Can you expand a little more on the reason for multiplying by `0.045F` and `0.5F`?

stephengtuggy:
Where is the part that actually sets Saira as the default font? I don't see that anywhere.

Which might not be a bad thing. It just doesn't match what the PR says it does.

stephengtuggy:
Also: Someone is going to need to do some more play testing on this, covering more of the test cases, before we merge it.

engine/src/gui/imguitext.h
SGD1953: That's fair, I'll change that.

libraries/gui/gui.cpp
SGD1953: It does not indeed. I started with 0.1F and it is WAY too large. Then I was looking at the basecomputer and what's there and picked a value which I felt makes sense. This should move to configuration, we could (mis-)use font_point for it but instead of points we use a norm space size (like the 0.045F). It is then multiplied by 0.5 to make get a value in the interval [0,1] which is then multiplied by the window height. This is how the Font class does it an thus it is consistent.

SGD1953:
@stephengtuggy there will be a PR for the assets dir which sets the font to `fonts/Saira-Regular.ttf`. I will link it once I have it, it is not a hard dependency though.

I can do more play testing, but I don't want to play test 10 times. So let's get this in mergable state first and then I'll do more testing. Pretty much all areas which have text are in the screenshots though except for the bar.

libraries/gui/gui.cpp
stephengtuggy: OK. That sounds like a perfectly reasonable implementation. All that remains, then, is to update or remove the comment.

stephengtuggy:
> @stephengtuggy there will be a PR for the assets dir which sets the font to `fonts/Saira-Regular.ttf`. I will link it once I have it, it is not a hard dependency though.

So it sounds like this PR doesn't actually set a default font. Which might be fine; we just need to update the PR title and description to match. Consistency is key.

> I can do more play testing, but I don't want to play test 10 times. So let's get this in mergable state first and then I'll do more testing. Pretty much all areas which have text are in the screenshots though except for the bar.

I know it's a lot of work, but in general, any PR that modifies the engine code should be run through all or most of the play test plan **before** merging it. That way, we make sure to keep the `master` branch clean, build-able, and playable at all times, as much as possible.

SGD1953:
> I know it's a lot of work, but in general, any PR that modifies the engine code should be run through all or most of the play test plan before merging it. 

We are perfectly aligned here, what I said was that I want to get all approvals first and make necessary adjustments before I do the full test run. Otherwise I'd have to do it twice.

stephengtuggy:
> > I know it's a lot of work, but in general, any PR that modifies the engine code should be run through all or most of the play test plan before merging it.
> 
> We are perfectly aligned here, what I said was that I want to get all approvals first and make necessary adjustments before I do the full test run. Otherwise I'd have to do it twice.

Oh, I see. I'm sorry. I misunderstood what you were saying. I get it now.

SGD1953:
I'll probably move the font to assets, the config is there anyways and the game won't run without assets dir, therefore the notion of a default font makes limited sense except we want to include Roboto for backward compatibility (if helvetica or roboto is set -> use roboto).

SGD1953:
With the feedback from the PR and some more reflection on how to best introduced it, the following revised approach:

- this PR adds the possibility to configure a font in the assets dir but does not yet introduce Saira, instead it will fallback to RobotoMono in case the configured font is not found (which is the case for "roboto" or "helvetica12") which creates a better backward-compatibility
- this has no data dir dependencies and current configs will just work, only with the new font which scales nicely and comes with a proper bold version

evertvorster:
On my config, the font rendered far too small. I play at 1368×768 with font_point = 17.0, but the text came out tiny. My read on why: InitGui bakes the font at defaultNormTextSize × 0.5 × pixelHeight and never uses the font_point value it's passed (fontSize2) — so the configured font-size setting has no effect on the rendered height. That's the sizing concern Stephen raised, and it matches.

I'm working on a fix that actually respects the font_point parameter — bake at font_point (with the resolution-relative value as fallback), and set style.FontSizeBase to the bake so the rendered height follows. I'll fold that in on my side.

On sequencing: I'd recommend merging #1729 as the font-rendering foundation, and I'll rebase our unify-text-coords work (TextPlane→ImGuiText migration + font_point-driven sizing) on top of it. #1729's real font-metric rendering (removing the scaleFactor FIXME, using Font::getFont()/CalcTextSizeA) is the cleaner approach than the workaround I would otherwise use. My PR work will keep the good bits (the compat Draw() for migrated TextPlane call sites + font_point sizing) and adapt to #1729's API. I'll resolve the imguitext.cpp/gui.cpp conflicts on the rebase.

So: please go ahead with the merge — I will incorporate properly respecting the font_point height parameter on our overlay.

SGD1953:
> On my config, the font rendered far too small. I play at 1368×768 with font_point = 17.0, but the text came out tiny. My read on why: InitGui bakes the font at defaultNormTextSize × 0.5 × pixelHeight and never uses the font_point value it's passed (fontSize2) — so the configured font-size setting has no effect on the rendered height. That's the sizing concern Stephen raised, and it matches.
> 
> I'm working on a fix that actually respects the font_point parameter — bake at font_point (with the resolution-relative value as fallback), and set style.FontSizeBase to the bake so the rendered height follows. I'll fold that in on my side.

Yes, but that is on purpose. The game essentially defined font sizes in norm space which we should use to calculate the point sizes instead of using a fixed `font_point` value. The issues currently is that TextPlane ignores those settings (`myFontMetrics.j`) and instead brute forces a single size. Therefore I was thinking we could use a `font_scale_factor` which allows to increase the various sizes by a percentage, this could come with guardrails so that text always runs nicely and the game remains playable.

engine/CMakeLists.txt
```diff
@@ -801,6 +801,11 @@ ELSE (WIN32)
     INSTALL(PROGRAMS "${Vega_Strike_BINARY_DIR}/vegastrike-engine" RENAME vegastrike DESTINATION bin)
 ENDIF (WIN32)
 
+# Copy fonts
+FILE(
+    COPY "fonts"
+    DESTINATION ${CMAKE_BINARY_DIR}
```
BenjamenMeyer, line 807: this should probably go into the shared data directory - e.g /usr/share/vegastrike - during installation:
```
SET(VS_FONT_DIR="${CMAKE_INSTALL_DATADIR}/vegastrike/fonts"
INSTALL(FILES <fontfile> DESTINATION ${VS_FONT_DIR})
```
You'll want to use `FILES(GLOB ...)` to find the font files and store them into a CMake Variable, and then use the `INSTALL` function to them on the target system.

BenjamenMeyer (changes requested):
Generally looks good; but we need to ensure the fonts get installed to the right location.

The `CMAKE_BINARY_DIR` is simply where things are built to.
They should get installed into a directory under the Data directory (`CMAKE_INSTALL_DATADIR` - `%ProgramFiles%` on Windows, `/usr/share/vegastrike` on MacOS/Linux). I provided a little of the CMake text to use, but it's certainly not complete.

engine/fonts/OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2015 The Roboto Mono Project Authors (https://github.com/googlefonts/robotomono)
+
```
royfalk, line 2: I don't agree we should have any default fonts, other than the actual default font, in the engine.

engine/src/gldrv/winsys.cpp
```diff
@@ -559,7 +560,16 @@ static bool setup_sdl_video_mode() {
...
+    fs::path fontPath = fs::path(VSFileSystem::datadir) / "fonts" / configuration().graphics.font;
```
royfalk, line 564: should be font_path as per our conventions.

engine/src/gui/imguitext.cpp
```diff
@@ -412,7 +407,7 @@ void ImGuiText::parseFormat(std::string input, size_t startPos, //Location of be
     *resultPos = curPos;
 }
 
-FormattedLayout ImGuiText::parseText(const std::string& input, const float widthInPixels) {
+auto ImGuiText::parseText(const std::string& input, const float widthInPixels) -> FormattedLayout {
```
royfalk, line 410: Why would you move from 
`FormattedLayout ImGuiText::parseText(const std::string& input, const float widthInPixels) {`

to this? IOW, why use arrow return type?

engine/src/gui/imguitext.cpp
```diff
@@ -502,7 +494,7 @@ FormattedLayout ImGuiText::parseText(const std::string& input, const float width
                     if(!m_multiLine) {
                         currentFragmentText += ELLIPSIS_STRING;
                     }
-                    if(currentLine.width + getTextWidth(currentFragmentText,  m_fontStack.back().size()).x > widthInPixels) {
+                    if(currentLine.width + getTextWidth(currentFragmentText,  m_fontStack.back()).x > widthInPixels) {
```
royfalk, line 497: I got to here.

royfalk (commented):
Spent some time reviewing this, but didn't get a chance to really dig in or finish before time ran out.

engine/src/gui/imguitext.cpp
SGD1953: That's what the linter suggested

engine/fonts/OFL.txt
SGD1953: Noted, please add your thoughts to the fonts discussion. There are a couple of advantages of including a font:
- it's ttf and not as pixelated
- it's a set of fonts which include bold and italic fonts (no hacks required)
- Roboto was one of the original options and therefore it should keep the original experience for the mods like Privateer

royfalk: Not sure where the font discussion is, but this is a different issue.
I'm saying that as an engine designer/developer, I'm fine with the `default font` being pixelated, not support bold/italic, etc.
If you as an asset designer/developer want a font, include it there.

I said it before and I'll say it again. We have a tendency to write defensive code that can handle missing assets. We shouldn't. It makes the code longer, dirtier and harder to understand without payoff.

evertvorster:
Hmm... just looking at this again. I have an open PR (Still in draft) that would make a lot of this PR moot. It updated imgui, and uses ttf fonts natively without baking. It also uses the roboto font included with imgui as a default, and looks for ttf fonts in assets which it then populates in a list in the settings app.

SGD1953:
@evertvorster go for it, I'll close this PR

engine/fonts/OFL.txt
SGD1953: Discussion is here:  https://github.com/orgs/vegastrike/discussions/1739

PR #1735, opened 2026-08-22
Implement an in-game settings screen, accessed by Alt+C. (Can be re-bound)

evertvorster:

Please answer the following:

The author of the changes should run the game and do the actions described below. Starting a new game, and completing these steps in this order should take around 10-15 minutes.

- [x] Basic flight.
- [x] Docking to a planet.
- [x] On planet actions:
    - [x] Save the game.
    - [x] Read news.
    - [x] Buy/Sell some goods.
- [x] SPEC flight.
- [x] Save the game again.
- [x] Upgrade your new ship.
- [x] Save once more.

## Areas of interest

[](https://github.com/vegastrike/Vega-Strike-Engine-Source/wiki/Pull-Request-Validation#areas-of-interest)

While completing those steps, the tester should look for oddities, especially in the areas that he changed. In general the following aspects should be looked at:

- Graphics (textures, light, geometry, etc)
- Audio (noise, volume, synchronisation, etc)
- Physics (momentum, gravity, collision, etc)

---

# Config Screen Play Test — in-game config utility (Alt+C)

**Branch:** `vs-settings-ng-json` (Vega-Strike-Engine-Source)
**Companion assets:** Assets-Production with the split config files (config.json / bindings.json / engine.json)

> Copy of the basic play test above, expanded with a **config-screen-specific** test list.
> Every feature implemented in the config utility needs a test. Open the config screen with **Alt+C**.

## Entry point & navigation

- [x] Alt+C opens the config screen from the main menu.
- [x] Alt+C opens the config screen while in-flight (world paused / non-blocking).
- [x] Alt+C opens the config screen from a base computer / docked.
- [x] Alt+C closes the config screen and returns to the game.
- [x] Mouse cursor is visible and clickable inside the config screen for all frames.
- [x] No click-through: clicking a config control does not fire a game action behind it.
- [x] Config screen does not segfault on repeated open/close cycles.

## Display frame — resolution, aspect, fullscreen, font

- [x] Resolution list populates from the detected display modes.
- [x] Selecting a resolution **applies live** (windowed and fullscreen) without restart.
- [x] Changing resolution toggles between windowed and fullscreen correctly.
- [x] Fullscreen toggle applies live.
- [ ] Monitor selector (if present) works on a multi-display setup.
- [x] Screen aspect selector updates the base viewport letterbox.
- [x] Base aspect persists: bases render letterboxed to the chosen aspect (default 16:10).
- [ ] Base-aspect letterbox does NOT shift mouse coordinates / hit boxes in bases.
- [ ] Font type selector present and selectable.
- [x] Font size / Text Height applies **live** and resizes text (not just blurs it).
- [x] Ideal font height suggestion scales with horizontal resolution (roughly width/80).
- [x] Font size persists across restart.

## Flight Control device selector — hot-swapping input devices

- [x] Flight control selector lists Keyboard / Mouse / Joystick.
- [x] Selecting the active device enables its settings; the non-active device's settings are greyed out / disabled.
- [x] **Hot-swap** to Keyboard: keyboard flight controls work immediately (no restart).
- [x] **Hot-swap** to Mouse: mouse flight works immediately — GetMouse sensitivity, warp recenter, cursor.
- [x] **Hot-swap** to Joystick: joystick flight works immediately.
- [x] Switching device after Save persists to the next launch.
- [x] No autofire / axis-leakage bug recurs when switching devices (the hotplug/attach fix holds).

## Mouse settings

- [x] Mouse flight mode (glide / warp) selector works.
- [x] Camera-cursor toggle.
- [x] Warp zone slider.
- [x] Mouse exponent slider.
- [x] Mouse deadband.
- [x] Mouse y-inversion derived from the preset (glide/warp invert y) and persisted to `input.mouse.inverse_y`.
- [x] Crosshair cursor visible in the HUD for mouse flight.

## Joystick settings

- [x] Per-axis live binding (assign a physical axis to an in-game axis).
- [x] Joystick hotplug: plug in a joystick after launch and it attaches without re-bind (no autofire).
- [x] Auto Deadband toggle/slider.
- [ ] Force-feedback device selector (if a FFB joystick is available).
- [ ] Joystick device selector.

## Bindings dialog

- [x] Bindings grouped by category.
- [x] Bindings shown per device (keyboard / mouse / joystick / hat).
- [x] **Real capture:** pressing a key/button/hat captures the binding live.
- [x] Add a new binding.
- [x] Delete an existing binding.
- [x] Conflict detection: binding an already-used key warns / blocks.
- [x] Bindings persist across restart.
- [x] Joystick axis unbind persists (a deliberately-unbound axis stays unbound).

## Presets frame

- [x] Preset dropdowns populate from `engine.json -> presets` (Computer, Physics, Geometry, Textures, Shaders, Sound, MusicAndVolume, Text, FactionTextures, Autodocker, Censorship).
- [x] Selecting a preset applies its resolved variables to `Configuration`.
- [x] Preset selection persists to config.json (survives restart).
- [x] Preset dropdowns render above the bottom button bar (not overlapping).
- [x] Presets frame auto-scales with the window.

## Save / Close (write-back)

- [x] Live-apply input bindings/axes/device on Save (re-runs bindKeys() without restart).
- [x] Save applies changes and **stays open** (does not close).
- [x] Close discards unsaved changes and closes.
- [x] Close is red when the screen has unsaved (dirty) changes.
- [x] Save writes a sparse overlay to `~/.vegastrike/config.json` (only overridden keys).
- [x] Save writes bindings to `~/.vegastrike/bindings.json`.
- [x] Unknown keys in config.json are preserved (sparse overlay does not clobber them).

## Persistence across restart

- [x] Every config change survives a full quit + relaunch (Load a saved game, reopen config screen).
- [x] Resolution / fullscreen persists.
- [x] Base aspect persists.
- [x] Font size persists.
- [x] Flight-control device persists.
- [x] Mouse + joystick settings persist.
- [x] Preset selections persist.
- [x] Input bindings persist.
- [x] homedir config.json wins over datadir config.json (user override respected).

## Config-overlay lifecycle (open/close regression)

Bugs fixed on this branch: stuck input states, stuck continuous-press sound, and cursor visibility per control device.
- [x] Open the config overlay, then close it — **no key/button state sticks** (no continuous fire, no stuck mouse button).
- [x] After closing the config overlay, **no continuous-press sound** (e.g. weapon fire/afterburner) keeps playing.
- [x] Flight Control: **Joystick** — config overlay shows **no mouse cursor** (joystick mode never shows a cursor; `mouse_cursor` forced false when switching from mouse back to joystick).
- [x] Flight Control: **Mouse** — config overlay shows the mouse cursor (needed for clicking).
- [x] Repeated Alt+C open/close does not leak input state across cycles.

## Config overlay clobber protection (homedir merge)

- [x] The config overlay **merges** dirty paths into the existing `~/.vegastrike/config.json` — it does **not** replace/overwrite the whole file.
- [x] Existing user-set keys (unrelated to the config screen) are preserved after a Save.
- [x] `bindings.json` overlay likewise merges rather than clobbers.

## Regression / broader integration

- [x] Base computer (ship view / upgrade view) loads without the `KeyError: 'constants'` / `'components'` errors.
- [x] HUD, radar, navscreen render correctly after config changes.
- [x] Audio volume / mute toggles (if surfaced) apply.
- [x] Colors correct (theme.json) — no white border around base screens.
- [ ] No `vegastrike.config` XML reads at runtime.
- [x] Game still loads a save, respawns after death, and exits without segfault after config changes.
Continue where you left off.

Issues:
- - Game has no in-game settings mechanism

Purpose:
- - Replace the non-working settings app that used to be with the game with an in-game settings screen.
- What release is this for?
- - 0.11
- - 0.11


evertvorster:
There are still a few open bugs on this:

Bindings list is not a sparse overlay in the user directory — ~/.vegastrike/bindings.json writes the full bindings list, not just dirty/overridden entries. Should be a sparse overlay like config.json so user additions/overrides are preserved and merged on load (may currently clobber).

Shaders don't hot-apply — restart required — Changing the Shaders preset doesn't take effect live; needs a restart. (Likely needs a shader/program re-init hook.)

Physics presets don't hot-apply — restart required — Same as shaders; keeps old physics values until restart even though saved.

Flight mode keyboard: joystick z axis (and non x/y axes) still active — When flight control = keyboard, x/y are correctly dead but the z axis still responds. Non-x/y axes aren't inhibited when switching away from joystick. <- **Fixed**

Rendered cursor does not hot-apply when switching control device — Changing flight-control device doesn't change the rendered cursor until restart; saved/persisted but the SDL cursor isn't re-applied live.

I'll get to them in short order, but this is already quite usable, and I can file issues for these,

engine/src/cmd/base_interface.cpp
```diff
@@ -146,8 +146,37 @@ std::vector<unsigned int> base_keyboard_queue;
 
 #define mymin(a, b) ( ( (a) < (b) ) ? (a) : (b) )
 
+// Letterboxed base viewport rect (in configured-resolution space), centered, from
+// graphics.bases.max_width/height. Returns false if no letterbox applies (the base
+// should fill the whole screen). This is the base-aspect viewport the imgui bases
+// render into; the GL viewport maps it onto native pixels.
+static bool base_viewport_rect(int &x, int &y, int &w, int &h) {
```
SGD1953, line 153: This is great, thanks for restoring this!

engine/src/cmd/base_interface.cpp
```diff
@@ -734,15 +763,23 @@ void base_main_loop() {
     }
 
     // ImGui Init
+    ImGui_ApplyPendingFontSize();   // apply a pending font-size change before laying out
```
SGD1953, line 766: I'd prefer we don't mess with fonts here, there are a lot of other PRs this will conflict with. I'd rather accept that setting font point has no impact.

engine/src/configuration/tests/json_config_model_tests.cpp
```diff
@@ -0,0 +1,163 @@
...
+using boost::json::value;
+using std::vector;
+using std::string;
+
+// A small model with the shape of config.json: nested sections + scalars.
+static JsonConfigModel make_model(const value& assets, const value& user = value(nullptr)) {
+    return JsonConfigModel(assets, user.is_null() ? nullptr : &user);
+}
+
+TEST(JsonConfigModel, MergeUserOverride) {
```
SGD1953, line 45: This file could use some documentation, it is not quite obvious what each of those tests is doing. It is great we have unit tests though!

engine/src/config_xml.cpp
```diff
@@ -436,6 +460,7 @@ CommandMap initGlobalCommandMap() {
     commandMap["SheltonKey"] = FlyByKeyboard::SheltonKey;
     commandMap["MatchSpeedKey"] = FlyByKeyboard::MatchSpeedKey;
     commandMap["PauseKey"] = FireKeyboard::TogglePause;
+    commandMap["ConfigKey"] = FireKeyboard::ToggleConfigScreen;
```
SGD1953, line 476: Why not replacing `Esc` with this? It effectively useless anyways once you start modifying the bindings.

evertvorster:
> Pulled this and asset master, launch the game, launch, hit `Alt-C` -> nothing happens.

That was added into Assets master fairly recently. You have to test against the absolute latest Assets master, please.

engine/src/in_kb.cpp
```diff
@@ -62,6 +63,45 @@ static bool kbHasBinding(int key, int modifiers) {
     return keyBindings[map_key].function != defaultHandler.function;
 }
 
+// Global (always-active) actions fire in ANY context (in-flight, docked, nav, text),
```
SGD1953, line 67: I like that!

engine/src/in_kb.cpp
```diff
@@ -62,6 +63,45 @@ static bool kbHasBinding(int key, int modifiers) {
...
+// not just the HUD. Add handlers here as more global actions are needed.
+static const KBHandler kGlobalActionHandlers[] = {
```
SGD1953: Is an array the best way to store those? A set or a map simplifies he lookups like below.

engine/src/universe.cpp
```diff
@@ -861,5 +871,32 @@ void Universe::TogglePause() {
     paused = !paused;
 }
 
+void Universe::ToggleOptionsActive() {
```
SGD1953, line 874: Could we consider moving this into a config class to avoid growing universe further and further?

SGD1953 (changes requested):

engine/src/cmd/base_interface.cpp
evertvorster: The game has always had a font size setting, and arguably this is something that the user wants to set. The game has to deal with it gracefully. It is a feature I want to keep. 

The main reason I want to keep it is that I play at 4K or 2.7K, and the default font size is too small, and the calculated fonts sizes might not be to my liking. 

Now, we both have work ongoing in this regard, and hopefully those PRs will go in first and this setting will work without breaking other things. Hang tight here, and we'll revisit this one when the others land. I suspect this PR will take some time to merge, its a monster.

engine/src/config_xml.cpp
evertvorster: You need to have SOMETHING bound to get into the settings screen. You can re-bind it to whatever you want, but esc is already taken by default.

engine/src/universe.cpp
evertvorster: We could. But we'll have to open an issue for it. Hopefully this is the last addition to universe.

engine/src/in_kb.cpp
evertvorster: This is the first, and hopefully only binding that we want to work *everywhere* in the game. It seems that the complexity of doing it another way is just not worth it.

engine/src/configuration/tests/json_config_model_tests.cpp
evertvorster: If you can point me to the documentation for the other unit tests, I'll gladly add what each of these tests do to that. :)

engine/src/config_xml.cpp
SGD1953: Maybe I did not articulate my thought very well, I meant we should remove the current key mapping screen an instead show the config screen.

engine/src/configuration/tests/json_config_model_tests.cpp
stephengtuggy: AFAIK, none of the unit tests have any documentation apart from the code itself. The `TEST(..., ...)` clause should serve as at least basic documentation.

If we want to document things further, perhaps we could add a comment above each test case.

engine/src/config_xml.cpp
stephengtuggy: Disagree. You have to bring up the show keybindings screen before `Q` will work to quit the game from flight. We can't get rid of it, at least not very easily.

I think Evert's code here is probably correct.

engine/src/cmd/base_interface.cpp
stephengtuggy: Like I said on one of your font-size-related PRs, @SGD1953 , I don't think we have the font size issue quite figured out yet. For now, let's continue allowing users to set it, please, like @evertvorster has it here. We can discuss this matter further in a GitHub discussion, and make any further changes once we have reached a consensus.

stephengtuggy: GitHub Discussion created as https://github.com/orgs/vegastrike/discussions/1739

engine/src/cmd/base_interface.cpp
evertvorster: Don't be so quick to thank me yet. Setting this to anything but 16:9 or 16:10 causes horrible issues. The proper fix for those are in a different PR, and will un-do a lot of the work you are doing now. 
I do apologize, but you can't say I didn't warn you.

engine/src/configuration/tests/json_config_model_tests.cpp
evertvorster: My point exactly. If we had a place to document unit tests, I would be happy to contribute to it.

engine/src/config_xml.cpp
evertvorster: For this PR, I don't want to broaden the scope too much. However, you are free to re-bind esc to opening the menu in your game. I'll even add a quit button to the config screen, and a show keybindings button that brings up the original screen that was shown with hitting esc. 

But this is a feature, and one that changes the game for other users. In this area, we have to be extra careful. 
So, good idea, but wrong place and time.

stephengtuggy: The show keybindings screen was what I meant. That's existing functionality. I agree that we should not broaden the scope of this PR too much.

engine/src/configuration/tests/json_config_model_tests.cpp
```diff
@@ -0,0 +1,163 @@
+/*
+ * json_config_model_tests.cpp
+ *
+ * Vega Strike - Space Simulation, Combat and Trading
+ * Copyright (C) 2001-2026 The Vega Strike Contributors:
+ * Project creator: Daniel Horn
+ * Original development team: As listed in the AUTHORS file
+ * Current development team: Roy Falk, Benjamen R. Meyer, Stephen G. Tuggy
```
stephengtuggy, line 8: @evertvorster you can feel free to add yourself to this line in the files you add or change.

engine/src/configuration/tests/json_config_model_tests.cpp
stephengtuggy, line 1: This is a pretty solid set of unit tests. Nice job.

engine/src/in_joystick.cpp
```diff
@@ -187,18 +188,29 @@ void InitJoystick() {
...
+        joystick[i1]->nr_of_axes = 0;
+        joystick[i1]->nr_of_buttons = 0;
+        joystick[i1]->nr_of_hats = 0;
+        if (joysticks && i1 < num_joysticks) {
+            // A real device is present at startup: attach it to this slot.
+            joystick[i1]->Attach(joysticks[i1]);
         }
     }
-    SDL_free(joysticks);
+    if (joysticks) SDL_free(joysticks);
```
stephengtuggy: Part of our coding style is to put braces around all `if` blocks, `for` loops, `while` loops, etc., no matter how short. Not a blocker though.

engine/src/gui/config_screen.cpp
```diff
@@ -0,0 +1,1754 @@
...
+    base_aspect_text = aspect_opts[sel_base_aspect];
+}
+
+// ---------------------------------------------------------------------------
+// Configuration read/write (the vs-05 model_set_var/get_var, mapped to the
+// engine's Configuration struct).
+// ---------------------------------------------------------------------------
+
+static vega_config::Configuration &cfg() {
+    return const_cast<vega_config::Configuration &>(configuration());
```
stephengtuggy: I think it's time we changed the signature of `configuration()` (and the object inside it) to not `const`. Then we won't have to keep casting away the `const`ness over and over. After all, the settings can now be modified within the game itself.

Doesn't have to happen in this PR though.

engine/src/in_kb.cpp
stephengtuggy: I agree with @SGD1953 here. A map would be preferable. Not going to block on it though.

engine/CMakeLists.txt
```diff
@@ -248,6 +249,7 @@ SET(LIBGUI_SOURCES
     src/gui/font.cpp
     src/gui/glut_support.cpp
     src/gui/imgui_support.cpp
+    src/gui/config_screen.cpp
```
stephengtuggy, line 252: Optional: Add `config_screen.h` as well. It's probably better practice to include the header files, though not strictly required. And the existing code didn't do this very much.

stephengtuggy (approved):
@evertvorster this looks really good! Yes, it's a bit on the larger side, but no larger than some of the PRs I have submitted. And almost all of the changes look very necessary.

Looks good to me!

evertvorster:
Just added the quit button, and centered the bottom row buttons. 
Also fixed warp mouse mode. Tuning it is now well understood, so if it does not do exactly what you want it to do, file an issue and copy me in on it.

royfalk:
Got here too late. Will open an issue for it.

PR #1736, opened 2026-08-23
Return proper values for base width and height

SGD1953:
Return the proper values to allow maintaining aspect ratio in bases

The original comment in the file was misleading, the main reason the resolutions matched was that they were mostly 4:3 aspect ratio resolutions. With wider screens this is no longer true however and the base width properties allow to not have stretched base screens.

Personally I'd go a step further and would calculate the following properties:

1. Aspect
2. base height (normally screen height but there could be screens with as aspect `< 4/3`
3. base width

This would simplify the configuration of the game.

### DIFF
--- a/libraries/root_generic/universe_util_generic.cpp
+++ b/libraries/root_generic/universe_util_generic.cpp
@@ -177,14 +177,13 @@ namespace UniverseUtil {
                 return std::to_string(configuration().graphics.resolution_y);
             }
             if (option == "base_max_width") {
-                // The GUI uses base_max_width as its screen width; in the old
-                // XML it always matched the render resolution (x_resolution),
-                // and bases.max_width (1440x1080) is the art's virtual size,
-                // not what the GUI should lay out at.
-                return std::to_string(configuration().graphics.resolution_x);
+                // The GUI uses base_max_width as its screen width; this can be
+                // a lower value than the actual max width to preserve the aspect ratio
+                // of the original game assets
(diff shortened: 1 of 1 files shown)


evertvorster:
Please hold off on this one. The base aspect ratio is directly addressed in two of my PRs.
In the config screen PR, base resolutions are defined and passed to the engine, so you can read them directly, and get the aspect ratio (not that it matters)
In the PR I'm working on now, I'm unifying the coordinate systems used for all the bases and HUD, so development can be easier. 

Sorry, should have chatted with you earlier so we can unify our efforts. 
Can I add you to the reviewers for the config screen PR?

SGD1953:
Sure, we can wait for the config changes to be merged. Not fixing the width/height issue here leads to incorrect line breaks in for example load screen and intro monologue when the font size is larger. I'll be looking at your PR in any case and we need to absolutely restore the coordinate system handling. The only thing I'd be careful with is the font_point config, in principle this should not be needed as most parts of the game define norm font sizes which also allow text to be smaller/larger in specific areas. The new TextPlane implementation however brute-forces a single font size which is incorrect.

evertvorster:
Not a worry! Drawing things in VS at the moment is a mess, so I am reworking that entire system. I fixed the line breaks in passing, but more importantly there will be only one coordinate system moving forward, and so we don't need to continually fix things when we use other fonts, etc. 

You'll have something to play with there shortly too.

SGD1953:
Closing as #1737 tackles this more holistically

PR #1737, opened 2026-08-23
Unify text rendering on ImGuiText; drive font size from font_point 

evertvorster:
This PR addresses a structural issue in the game code. As such, it is massive and ugly, and for that I do apologize. 
I want to build a UI for the settings app, and found a multitude of ways to do it, all conflicting and I did not want to roll my own. 

The multitude of different coordinate systems keeps on tripping us up, and causes issues when we make changes in one part of the game, something somewhere else, seemingly unrelated, breaks. 

Since we have moved over to using imGUI for everything, we might as well standardize on it, and in part this is what this PR achieves. 

There is a document in the docs folder of the engine that lays out this one unified layout, how it works, and what to avoid. Of course, to unify I had to get rid of the alternatives, and work through all instances of it, hence the size and scope of this PR. 

The biggest offender to the standardized UI implementation was the base computer. It completely ignored font settings provided by the engine, and did its own thing, causing inconsistencies in how just about everything is handled. Hence, most of that is now gone, and it still works fine, but takes queues from the engine now. 

**This PR also updates imGUI to the latest released version**, mainly to take advantage of its native ttf handling. There has been some work on getting some ttf fonts in the game, and this opens the door for doing it in a consistent and easy to implement way. It uses the roboto font that is supplied with imGUI, but should be able to read fonts from Assets with minimal modification. 

I'm not against sweeping changes, so on your first review of the code, please look at the structure and scope, and I will implement suggestions like moving things around and maybe splitting into separate files the largest of the files if there is a clear advantage in doing so. 

For colors, I used the format imGUI supports natively, and that of course meant going through both engine and assets to change it over. So, this PR depends on https://github.com/vegastrike/Assets-Production/pull/280
That one depends on this one, so please make sure you have both installed before testing. 

**Again, I profusely apologize if this PR is affecting any open PR.**

Issues:
- - There were multiple conflicting coords systems in the game
- - Font height was not respected in all areas of the game.

Purpose:
- - Unify coords systems
- - Consistent Font handling
- - Ttf fonts natively supported
- What release is this for?
- - 0.11
- - 0.11


doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
+# Vega Strike — Text, Text Boxes & Coordinate Systems (Developer Standard)
```
SGD1953, line 1: I love that we are documenting this!

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+- `#b<width>` bold / stroke width.
+
+**Color notations** — the parser accepts **both**, so authors use whichever fits:
+
+- `#RRGGBB` — industry-standard hex (the primary authoring form; artists know it).
+  `#000000` (black) is a **reset to the default color**, not literal black.
+- `#cR:G:B#` / `#cR:G:B:A#` — float 0..1 channels, for colors that need alpha or
+  fractional channels.
+
+> `#c` and `#b` are reserved prefixes for the float-color / stroke codes, so a hex
```
SGD1953, line 75: This is the reason why I have not merged TextPlane and ImguiText earlier, it's not ideal that you cannot use certain colors.

We could probably let the parser do a look ahead and figure out it is a hex code or not.

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+> color must not begin with `b` or `c` (e.g. write `#bb0000`, not `#b0000`).
+
+### `TextPlane` (legacy — migrating)
```
SGD1953, line 78: Love it!

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+`TextPlane` (`engine/src/gfx/hud.cpp`) is the base/HUD text box. It is being migrated onto
+`ImGuiText`. **New code must use `ImGuiText`.** Its separate `#c`-only parser will be removed.
+
+### `PaintText` (dead)
```
SGD1953, line 82: We should delete it, does not have to happen in this PR but once deleted there is no confusion.

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+- **Glyph size from `font_point`** (`configuration().graphics.font_point_flt`). Author a
+  *relative scale* on `font_point` (default 1.0); the engine renders it as a pixel glyph height
+  (`font_point * scale`). Do **not** hardcode a per-control font size, and do **not** do
+  resolution-relative math — the engine's standard font metrics handle resolution.
+- **Line spacing from the RENDERED glyph height.** `ImGuiText` measures each line with ImGui's
+  rendered metrics (`CalcTextSizeA` / `dimensions.y`), never `font.size()` or a literal — so a
+  future config font-height change propagates automatically.
+- The base computer sets font from `controls.json`'s per-control `font`, which is now a *relative
+  scale* (`font_point * scale`), not an override.
+- The engine's `Font` class is a thin `{size, strokeWeight}` carrier feeding `ImGuiText`. It does
```
SGD1953, line 160: Nice, wanted to do that as well.

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Target:** ONE interpretation (canonical top-left `(-1,+1)`). Where an element genuinely needs a
+different origin (e.g. sprite center-point), expose it as an explicit `origin` flag on a single
+method — not three divergent helpers.
+
+---
+
+## 3. Font size & line spacing
+
+- **Glyph size from `font_point`** (`configuration().graphics.font_point_flt`). Author a
```
SGD1953, line 151: As said earlier, I'd prefer a relative text size with a configurable scale factor - `font.size() * scale_factor`

doc/UI-Drawing-Standard.md
```diff
@@ -0,0 +1,215 @@
...
+
+`type`, `text`, `label`, `command`, `texture`, `rect`, `justification`, `font`, `textMargins`,
+`multiline`, `color`, `textColor`, `downColor`, `downTextColor`, `highlightColor`,
+`highlightTextColor`, `borderColor`, `endBorderColor`, `buttonColor`, `outlineColor`,
+`selectionColor`, `shadowWidth`, `parent`, `scroller`, `cycleTime`, `id`.
+
+Control types: `button`, `staticDisplay`, `picker`, `scroller`, `groupControl`, `staticImageDisplay`,
+`textInputDisplay`.
+
+If bases are unified onto the `Control`/`ImGuiText` world, they inherit this theming for free — the
```
SGD1953, line 191: 🎉

engine/src/cmd/base_interface.cpp
```diff
@@ -1251,6 +1299,9 @@ BaseInterface::BaseInterface(const char *basefile, Unit *base, Unit *un) :
     CurrentBase = this;
     CallComp = false;
     lastmouseindex = 0;
+    // The fixer/campaign streaming text (dialogue typed word by word) renders at
+    // 1.4x; curtext and other base-room text stay at the normal size.
+    othtext.setTextScale(1.4f);
```
SGD1953, line 1304: Let's make this configurable or as a minimum let's add a TODO to clean it up later

engine/src/cmd/base_interface.cpp
```diff
@@ -1261,12 +1312,12 @@ BaseInterface::BaseInterface(const char *basefile, Unit *base, Unit *un) :
     curlinkindex = 0;
     this->baseun = base;
     float x, y;
-    curtext.GetCharSize(x, y);
-    curtext.SetCharSize(x * 2, y * 2);
-    curtext.SetSize(1 - .01, -2);
-    othtext.GetCharSize(x, y);
-    othtext.SetCharSize(x * 2, y * 2);
-    othtext.SetSize(1 - .01, -.75);
+    curtext.getCharSize(x, y);
```
SGD1953, line 1317: Does this still make sense? Should we not just set set a size? Above we set a `1.4f` scale and here we are doubling the font size. Also, the x value is essentially pointless.

engine/src/cmd/base_interface.cpp
```diff
@@ -1577,6 +1628,19 @@ void BaseInterface::Draw() {
     GFXColor(0, 0, 0, 0);
     SetupViewport();
     StartGUIFrame();
+    // Base room text (curtext = room default/link text, othtext = talk text) lays out
+    // against the base resolution, matching the rest of the base (BaseText, link labels,
+    // hitboxes).
+    const float base_max_w = static_cast<float>(configuration().graphics.bases.max_width);
```
SGD1953, line 1636: We should move this to a method so we can apply the base resolution and offset to textboxes consistently without duplicating this code.

engine/src/cmd/basecomputer.cpp
```diff
@@ -4410,7 +3693,7 @@ void BaseComputer::LoadSaveQuitConfirm::init(void) {
     text->setTextColor(GFXColor(.7, 1, .4));
     text->setMultiLine(true);
     text->setColor(GUI_CLEAR);
-    text->setFont(Font(.07, 1.25));
+    text->setFont(Font(configuration().graphics.font_point_flt * 0.79f, 1.25));
```
SGD1953, line 3696: This should move to the `controls.json`, but can be done in a follow up PR.

engine/src/gfx/nav/drawgalaxy.cpp
```diff
@@ -80,37 +80,37 @@ void DrawNodeDescription(string text,
...
     const float background_alpha = configuration().graphics.hud.text_background_alpha_flt;
     int length = text.size();
     float offset = (float(length) * 0.005);
     if (ignore_occupied_areas) {
-        displayname.SetPos((x_ - offset), y_);
-        displayname.SetText(text);
-        displayname.SetCharSize(size_x, size_y);
+        displayname.setPos((x_ - offset), y_);
+        displayname.setText(text);
+        displayname.setCharSize(size_x, size_y);
```
SGD1953, line 91: Not going to comment every occurrence, but `setCharSize()` should be revisited.

engine/src/gldrv/winsys.cpp
```diff
@@ -561,8 +562,15 @@ static bool setup_sdl_video_mode() {
...
     get_screen_measurements();
 
-    // Initialize imgui
-    InitGui(window, &context, configuration().graphics.font_point_dbl);
+    // Initialize imgui with the configured font. graphics.font is "Roboto" (the
+    // embedded engine font) or a .ttf filename in the data fonts/ directory.
+    std::string font_file;
+    const std::string &font_name = configuration().graphics.font;
+    if (!font_name.empty() && font_name != "Roboto") {
+        font_file = VSFileSystem::datadir + "/fonts/" + font_name;
```
SGD1953, line 582: String is not ideal here, you can borrow the logic from https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1729/changes#diff-9a84003d1a9d89427bf7662dd4293507b69079fcf2e0e567239b8338f31877d4R564 which is more robust.

engine/src/vsdraw/imguitext.h
```diff
@@ -71,15 +93,33 @@ struct FormattedLayout : public std::vector<Line> {
...
+    // (othtext) sets this to 2.0 so it renders larger; other text stays at 1.0.
+    void setTextScale(float s) { m_textScale = s; }
+    float textScale() const { return m_textScale; }
+    // Set the text color from a packed ImU32 (TextPlane-compatible interface).
+    void setColorU32(ImU32 c) { m_colorU32 = c; m_color = GFXColor(c); }
+    // Background rectangle behind the text (TextPlane `background_color`).
+    void setBackgroundColor(ImU32 c) { m_backgroundColor = c; }
+    ImU32 backgroundColor() const { return m_backgroundColor; }
+    // Whether to suppress the background rectangle (TextPlane `automatte`).
+    void setAutomatte(bool a) { m_automatte = a; }
```
SGD1953, line 126: This name is very not obvious, please change to something like `setTransparent()` and `isTransparent()`/`transparent()` while we are touching this

engine/src/vsdraw/imguitext.h
```diff
@@ -92,14 +132,56 @@ class ImGuiText {
...
     Font m_font;
     std::string m_text;
     Justification m_justification = LEFT_JUSTIFY;
     bool m_multiLine = false;
     GFXColor m_color;
+    ImU32 m_colorU32 = IM_COL32(255, 255, 255, 255);
+    float m_textScale = 1.0f;    // Text render scale (1.0 = normal).
+    ImU32 m_backgroundColor = IM_COL32(0, 0, 0, 0);
+    bool m_automatte = false;
+    float m_charW = 0.06f;
```
SGD1953, line 185: This redundant (`m_charH`) with `Font.size`, we should consolidate. I don't think `m_charW` is ever used.

SGD1953 (changes requested):
Did a first pass, really love the direction as it is very much in line with what I had envisioned. I also like that you documented the style guide for text and already migrated the text in the game to use the new style guide.

I disagree with `font_point` being the starting point of calculations and would prefer relative sizes with a scale factor (could be applied centrally in `ImguiText` constructor.

The are some minor inconsistencies like the size being in Font and ImguiText which should not be too hard to address.

Will try to play-test over the weekend.

engine/src/vsdraw/imguitext.cpp
```diff
@@ -89,7 +103,10 @@ void ImGuiText::draw(int firstLineToDraw) {
         }
         
         for (const auto& frag : line) {
-            float pixelFontSize = Coordinates::normToPixelFontSize(frag.font.size());
+            // Font::size() IS the pixel glyph height (font_point-relative). Round to a
```
SGD1953, line 106: It would be better if Font::size() would be norm size and not pixel size and we convert here

doc/UI-Drawing-Standard.md
CheetahPixie: personally, I would not ever reserve prefixes that could potentially collide with something else. I'd either name them something else, or straight up use some symbol other than #, if possible, for the prefixes. as to colors, if regex could narrow it down to A-F a-f 0-9, then, say, #c_pref would immediately disqualify, and resolve the collission issue outright, both due to the underscore, and the two letters (p and r) that the regex would exclude. of course, changing the # symbol would also just not ever allow this to be an issue.

engine/src/vsdraw/imguitext.h
CheetahPixie: what about alpha? or is this not an alpha thing?

doc/UI-Drawing-Standard.md
CheetahPixie: for system dpi, there should be one scale factor, and then calculate actual font size from dpi as set in config.json. that is:

system scale factor * (font size in config.json * in-game scale relative to the size specified in config.json)

that in game scale bit is cruical for several screens, where text in one block might differ from another, and we need to have the scale of that other text be larger by some specific percentage than the base size.

the base size here, say, 16, should *always* be the smallest text seen in game, and every relative size multiplied from that size. that way, you don't get possibly unreadable text on small font sizes, and will always have a correct scale ratio between individual elements. and if, for example, we have a scale factor of 1.25 for an element, then, at size 16, the scaled font will be 20. at size 8, that will become 10. or, in hidpi territory, we might have a system scale of 200%, wherein we have the equation

200% * (16 * 1.25) = 40pt

which, given that scale, plus the base size, plus the element's own scale, would result in the correct and expected 40 point font size, instead of some other silly value.

if we did this by a fixed amount, say, 4, 16 would become 20, as we want, but 32 would become 36, and 8 would become 12, both of which are completely wrong. this is exactly why fractions are what we want.

PR #1752, opened 2026-08-29
Imgui font backend

evertvorster:
This PR is an effort to reduce the text coords PR size, which currently has about 3600 lines added and removed. 
The imgui backend needs updating to accommodate ttf fonts, and is rolled into that one.

This PR updates the backend seperately, and adds in the roboto ttf that comes with imgui.  (And reduces the other PR's added lines by 1200. )

Please answer the following:

Code Changes:
  - [x] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - Old imgui did not support native TTF.

Purpose:
- - Upgrade imgui to latest release, and switch to native ttf, using included roboto font.
- What release is this for?
- - 0.11
- - 0.11


libraries/gui/gui.cpp
```diff
@@ -40,17 +40,26 @@
 #include "backends/imgui_impl_opengl3.h"
 #include "backends/imgui_impl_sdlrenderer3.h"
 
+// TrueType font (Roboto-Medium) embedded as a compressed base85 array so the engine
```
royfalk, line 43: As I noted in @SGD1953 's PR, I don't think we should bundle **any** font with the engine. If the assets don't provide one, the game should run with the default font.

This has the benefit of keeping the engine font cleaner and it's highly unlikely any game using our engine will not use an appropriate font.

libraries/gui/gui.cpp
```diff
@@ -40,17 +40,26 @@
...
 SDL_Window* current_window = nullptr;
-ImFont* roboto_18_font;
 
 // Font size requested by a settings change, applied on the next frame. Stored as a
 // float so we can defer the atlas rebuild to a safe point (start of a frame, before
 // any text is laid out), avoiding mid-frame texture destruction.
 static float s_pending_font_size = 0.0f;
 
-void InitGui(SDL_Window *window, const SDL_GLContext *context, const float fontSize) {
+// Font file requested by a settings change, applied on the next frame. Empty means
```
royfalk, line 57: This is an excellent example of overkill. 
If a player changes the font size, we can do this in the next run and save whatever is happening here.

royfalk (commented):
Two initial comments.

libraries/gui/gui.cpp
evertvorster: This is the font that is distributed with imgui. Its available by default, and is the basis for the fallback font in any case. 
We need it to get through the next few hurdles. We can always drop it later, but this is the engine default font...  It just changed from bitmap to ttf.

libraries/gui/gui.cpp
evertvorster: This is for font hot-apply in the settings screen. In fact, a almost all things hot apply in the settings screen, and this is definitely needed.

libraries/gui/gui.cpp
royfalk: This is from _roboto.h_:

```cpp
static const char RobotoMediumFont_compressed_data_base85[144680+1] =
    "7])#######hb'0,'/###I),##bw9hLbu2@$v(+h<CiTgY#3_Y#)u(###9'e=N9o6U.0bY#9D)##G)te=1?`b&6K)##u3k34;vA0F[Px;/3''##Tj&##-V:'ISHUPSCX%##[/k--'/Y-G"
    "nvS2t0))##:I0R3&7#=BejUS%ns1_A,$j--r2/-DtuF'[dvp;#9TM:#%Z$^DUj<d6jDH&#dg[0#'m@aDGigjs*GCgLspIiLh$tLD8-s8&;.#wL)mGoL<x$KDLG?a.)RX*MgXL=#+cunE"
    "5im/klnM5#h,0%#ano:Fur6D%/Xm--nGq92>4^=BAE-pB6r[Y#_arI0/4HiF#a$AFDpN-Qctq--AcFVCYfwrs2o[f1s<Mt-tnRfLV_6##oN.gW;he<6ba?iL)B$##?a,e([_a/1t-?@<"
```

I'm going to assume this is the actual font, embedded into a header file. If Roboto were truly embedded by default, we wouldn't need this.

libraries/gui/gui.cpp
royfalk: I know. I'm saying hot apply is cool but provides marginal improvement for most features. Fonts are usually not modified by users and so providing hot apply is not really useful.

Let's see what others think.

evertvorster: Hmm, later in the very large PR I'm reducing into smaller more digestible bits, the settings screen actually scans the Assets directory for fonts, and adds them to the font picker. You can pick the font, and for some reason the font that is in assets now has a much smaller glyph size, and the user might want to set the font size to something nice in the settings app. To see what you are doing, and what the effect will be, we need hot apply. 

Actually, hot apply has been a requested feature from Benjamin, and every setting is hot apply now, apart from shaders/techniques which do some magic on init that I have not been able to figure out. That is the only exception to hot apply.

libraries/gui/gui.cpp
evertvorster: Let me double-check on this.

evertvorster: OK, after some double-checking on my side. 

We're not introducing a new font dependency — we're changing the engine's default font from one bundled bitmap font to a bundled TrueType font. The old default was ImGui's built-in ProggyClean.ttf, a tiny bitmap font that's compiled into the imgui library. With the move to native TTF handling, that default became obsolete, so we switched to Roboto — the TTF ImGui itself ships and uses as its default.

Embedding it in source isn't our invention; it's exactly how ImGui ships its own default font. ImGui provides binary_to_compressed_c, which bakes a TTF into a C array (ProggyClean is embedded in imgui_draw.cpp the same way). We used that tool's base85 form and put it in a header (roboto_font.h) so gui.cpp includes it — same pattern, ~145 KB, no runtime file dependency, and a guaranteed consistent fallback.

libraries/gui/gui.cpp
stephengtuggy: My recommendation is to keep hot apply as long as the effort has already been made to implement it.

libraries/gui/gui.cpp
stephengtuggy: Wait. Is ImGui's default font ProggyClean or Roboto? Do older versions of ImGui use ProggyClean as the default, while newer versions use Roboto?

I don't have a problem introducing a new default font if that's what we decide to do. As long as we're doing that, I think I would prefer that the font be included as a .ttf file, rather than encoded in a C/C++ header. .TTF is the proper format for a TrueType font.

stephengtuggy (approved):
You know what? This is fine the way you have it, Evert. Approving.

Issue #1742, opened 2026-08-24
Autopilot adjusts ship direction in a juddery, "two steps forward and one step back" fashion


stephengtuggy:
As stated in the title. This is another extremely annoying autopilot behavior that we need to fix. It makes no sense why the autopilot would need to do this.

BenjamenMeyer:
IIRC, from @Loki1950 this is due to some of the gravity well avoidance logic

evertvorster:
Yes.. It does not do it when there is nothing to crash into, like at the jump points. 
I have been looking at this, but right now I have a few bigger fish to fry. I just kick out of autopilot when I arrive at the destination. 

Actually, while I wait for some of my PRs to merge, let me dig into this a little. If it makes you feel any better, Vs-05 version already did this, so its not something new.

stephengtuggy:
Yes, the autopilot did this a little ever since I first played the game. But it's much worse than it used to be.

evertvorster:
OK, I solved the autopilot issues, but to get the smooth SPEC ramp-down, I had to change the SPEC model from gravity based to distance based. 

It plays really well, but it also changes the way SPEC feels. (More sharp run-up, but consistently smooth bleed down to normal speed. 

https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1754

The PR still needs some dressing up, wiring through the tunables and eliminating the unused ones, which will be a separate Assets PR. But, before I do that, I want some opinions on whether the changes here are acceptable.

stephengtuggy:
@evertvorster is this now fixed as of #1754 ?

evertvorster:
I would say so. 
When you first engage autopilot, it will just coast until it is a few km clear of objects, then it will engage ftl. If you are in a heavily congested area, it might look like the autopilot swings wildly, but its actually looking for clear space so it can speed up. Once it finds some clear space, it heads to the destination. So, its not necessarily picking the shortest route, it's picking the fastest one.

stephengtuggy:
> I would say so. When you first engage autopilot, it will just coast until it is a few km clear of objects, then it will engage ftl. If you are in a heavily congested area, it might look like the autopilot swings wildly, but its actually looking for clear space so it can speed up. Once it finds some clear space, it heads to the destination. So, its not necessarily picking the shortest route, it's picking the fastest one.

[...]

Issue #1741, opened 2026-08-24
Autopilot takes way too long to start heading toward destination


stephengtuggy:
As of 0.10.x, the autopilot is buggier than ever. Now when you start anywhere near a planet, station, or jump; target another planet, station, or jump; and turn on the autopilot to start traveling there, the autopilot consistently steers away from the target, for no reason that I can determine. It continues to head in a different direction for several minutes -- far longer than it should. Finally, it will eventually start heading towards the target.

This behavior is extremely annoying, and it never used to happen in older versions. We need to get this fixed as a matter of priority.

evertvorster:
If its any consolation, 0.5 also has this, but to a lesser extent. I think its the autopilot thinking its going to crash into something, and steering away. I'll look at it closer once the dust settles a bit. However, I did notice that the overspeed fix in master greatly helps here.

stephengtuggy:
Is this fixed as of #1754 ?

kheckwrecker:
I always assumed what was happening was that autopilot was taking your ship directly away from the planet in a straight line, to escape the gravity well as expeditiously as possible, before moving back onto  the targeted course. In other words, if your target is on a line from your launch point that runs parallel to the planet, you'll spend longer in the gravity well if it tries to go on that line right away. So the autopilot addresses that by accelerating directly away before paralleling the planet at a safe (no gravity well) distance. As far as I know, it's always been like this.

This has been less of a problem in more recent versions due to autodocking occuring further out. It's an interesting quirk of the game that you launch from a planet/base at the same distance at which you docked. So the first time you launch on a new game, it's slow becuase of the gravity well. Subsequent launches will not have this annoyance because of how far out you docked (and then launch at same distance), which I have found to usually be outside the gravity well.

So, I'm not sure this is a new issue, but it is annoying.

stephengtuggy:
> Is this fixed as of [#1754](https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1754) ?

I can answer my own question. This is fixed in #1754 , at least in 0.11.x. The autopilot still sometimes takes a bit to head in the direction of the target, but it seems like there are good reasons for that now when it happens.

stephengtuggy:
Thanks very much for addressing this, @evertvorster

stephengtuggy:
[...]

Issue #1655, opened 2026-07-13
Ship speed updates far slower, causing jerky motion and SPEC movement


CheetahPixie:
Instead of the expected constant updates, the ship's speed instead updates only about every second. This is independent of the simulation atom, and applies only to the ship's own acceleration.

Turning a ship that already has velocity reveals that the speed value itself does constantly update, but for some reason, its acceleration happens in what can best be described as stair steps.

This was encountered on 0.10.0-alpha3, under Kubuntu Resolute (26.04).

CheetahPixie:
This also seems to affect target selection, which looks to update in sync with the speed itself.

CheetahPixie:
It also seems NPC ships are affected. They jerk around similarly.

CheetahPixie:
This is not caused by the locale issues. That is, an LC_ALL does not fix this.

stephengtuggy:
@CheetahPixie can you test this again with the latest code on the master branch? I think #1754 may have fixed it.

Issue #1255, opened 2025-08-12
Fighter Barracks goes into SPEC flight and flies away


royfalk:
It's quite clear because:
1. Shields went down
2. It was slowly moving away from me even though I was flying at full speed.

kheckwrecker:
I experienced this issue this evening as well, in the latest master. In addition, I noted my own speed was dropping, even though I was at full throttle. Eventually, it stopped dropping (but it's painful to wait for this to stop as it takes quite a while), and was able to eventually get back to docking range.

evertvorster:
I have seen weirdness on vessel speed in git master too. 
I was assuming that it was because I was too close to Atlantis, and after a while I was able to get up to speed again. There is some weight issue in the save file that we need to sort, but this space station going into SPEC sounds awfully similar to what I was calling a positioning uncertainty for a very long time now. 

I run into it in two separate ways. 
1. When I SPEC to a space installation (I have yet to see this happen with a planet) sometimes as you come out of SPEC you can see the station in front of you for one frame, and then suddenly you are behind it by 100km, as if you have warped through it, or hear me out on this one, it changed position. 

2. When launching from a space station, I sometimes see it "Jitter" around in space, along one axis. Everything else is stationary, but it can look like this thing is entering SPEC. 

That is reported in #866 

I have a hunch that it is related to: #594

PR #1754, opened 2026-08-30
FTL space compression

evertvorster:
FTL is space compression. (Its how we break the hard speed limit of light through a vacuum) 

How much space can be compressed depends now only on what's in it — the nearest object sets the radius of the compressible sphere. The old per-object interdiction power-law (CalculateNearestWarpUnit + default_interdiction/warp_region*/warp_curve_degree) is removed.

The warp multiplier is now a simple linear ramp of the significant distance to the nearest object:
```
multiplier = 1 (no SPEC)                 within warp_min_range (3 km)
multiplier = max_mult · (nearest − warp_min_range)
             / (max_effective_velocity − warp_min_range)   above it
```

No SPEC within 3 km — it's space compression and needs empty space, so inside weapons range you just fly normally (this also means combat isn't at warp speed).

Linear from 3 km to max_effective_velocity (≈29 M km) — full speed only when the whole column of space you'd compress in the next second is clear.

The nearest object counts bases, stations, and ships (added to the gravitational set), not just planets.

The autopilot (AutoLongHaul) steers with a vector sum to keep SPEC fast: repulsion from the nearest objects inside a collapsing bubble pushes it into clear space, a steady destination pull brings it home, and the destination itself never repels. The bubble is a full SPEC sphere at range and collapses to nothing by arrival.

Four new tunables (physics.*): warp_min_range, warp_clearance_repel, warp_clearance_attract, warp_clearance_range_mult. ~13 old interdiction/warp configs are retired. warp.max_effective_velocity/warp_multiplier_max are reused unchanged as the range/scale anchors.

Please answer the following:

Code Changes:
  - [ ] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1742#issuecomment-5457806400
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1741
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1655
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1255
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1012
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/881
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/854
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/205
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/343

Purpose:
- - Fix SPEC and autopilot
- What release is this for?
- - 0.11
- - 0.11


engine/src/cmd/ai/navigation.cpp
```diff
@@ -49,6 +52,27 @@ using namespace Orders;
 
 constexpr float M_PI_FLT = M_PI;
 
+// Collects up to a bounded number of units (other than ourselves) in range for the
+// SPEC clear-space steering -- used as the action for UnitWithinRangeLocator over
```
royfalk: The engine does not recognize SPEC or warp as proper terms. It uses the neutral ftl.

engine/src/cmd/ai/navigation.cpp
```diff
@@ -528,14 +552,6 @@ void AutoLongHaul::SetParent(Unit *parent1) {
...
 
-QVector AutoLongHaul::NewDestination(const QVector &curnewdestination, double magnitude) {
-    return curnewdestination;
-}
-
-static float mymax(float a, float b) {
-    return a > b ? a : b;
-}
-
 static float mymin(float a, float b) {
```
royfalk: Please delete as well and replace with std::min.

royfalk (commented):
Initial review. Did not go deep but it seems like the right direction.

engine/src/cmd/ai/navigation.cpp
evertvorster: Done

engine/src/cmd/ai/navigation.cpp
evertvorster: Done

engine/src/cmd/ai/navigation.cpp
```diff
@@ -623,88 +641,125 @@ void AutoLongHaul::Execute() {
...
+        const float attract = configuration().physics.warp_clearance_attract_flt;
+        // The bubble stays at a full ftl sphere while the target is outside the
+        // gather range (warp_clearance_range_mult x the bubble). Once the target
+        // enters that range we collapse the bubble, so that it is already gone by the
+        // time the target reaches the 1x bubble itself -- on arrival we simply get to
+        // the target instead of dodging ships there.
+        double bubble = max_compression_range;
+        if (destinationdistance < gather_range) {
+            const double ratio = (destinationdistance - max_compression_range)
+                    / (gather_range - max_compression_range);
```
BenjamenMeyer: we should use another set of `()` around this since it's multi-line and has the division in it:
```c++
const double ratio = (
    (destinationdistance - max_compression_range) /
    (gather_range - max_compression_range)
);

that would make it a bit cleaner - no opinion on whether the `/` should be at the end of the line or beginning of it beyond what our coding style says.

BenjamenMeyer (commented):
generally LGTM

engine/src/cmd/ai/navigation.cpp
evertvorster: I would like to see a copy of that coding style at some point, so my agent can code more in line with what you expect.

evertvorster: Done

BenjamenMeyer: > I would like to see a copy of that coding style at some point, so my agent can code more in line with what you expect.

@evertvorster best I can point you to at the immediate is https://github.com/vegastrike/editor-configs. We do need to do better with coding style documentation.

BenjamenMeyer (approved):
haven't playtested, but LGTM

stephengtuggy (approved):
Haven't tested, but LGTM. I especially like that we no longer drop out of SPEC when turning. That was a really dumb thing that the previous autopilot implementation did.

evertvorster:
Do we also want the NPC to have the same autopilot? If so, I can add this in a separate PR.

stephengtuggy:
> Do we also want the NPC to have the same autopilot? If so, I can add this in a separate PR.

Probably.

PR #1756, opened 2026-08-31
Ttf picker and fixes

evertvorster:
As it says on the tin, this PR leverages the newly gained native ttf abilities of the engine. 
It scans a specific directory in Assets for ttf fonts, and makes them available for configuration.

Removed old classic bitmap modes and fonts the engine no longer supports.
Ensures that if you are silly enough to pick a ridiculously large font size, you can still click the Apply button to undo your horrors. 

Please answer the following:

Code Changes:
  - [ ] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - Settings has no ability to change fonts

Purpose:
- - Scan Assets for fonts, and make them available for configuration.
- What release is this for?
- - 0.11
- - 0.11


engine/src/gui/config_screen.cpp
```diff
@@ -80,32 +80,56 @@ static bool bind_capturing = false;
...
+    if (fs::exists(fonts_dir) && fs::is_directory(fonts_dir)) {
+        std::vector<std::string> ttf;
+        for (fs::directory_iterator it(fonts_dir), end; it != end; ++it) {
+            if (it->path().extension() == ".ttf") {
+                ttf.push_back(it->path().filename().string());
+            }
+        }
+        std::sort(ttf.begin(), ttf.end());
+        for (const auto &f : ttf) {
+            s_font_list.push_back(f);
```
royfalk, line 111: What's the difference between s_font_list and s_font_values?

engine/src/gui/config_screen.cpp
```diff
@@ -449,37 +457,22 @@ void draw_display_frame() {
             if (ImGui::MenuItem(aspect_opts[i])) { sel_screen_aspect = (int)i; refresh_screen_aspect_text(); dirty = true; }
         ImGui::EndPopup();
     }
-    // Font type picker.
+    // Font picker: Roboto (engine default) + .ttf fonts from the data fonts/ dir.
+    if (s_font_list.empty()) {
```
royfalk, line 461: By enabling a font picker for the user to specify the font, you are taking away the option to support more than one font in the game.

royfalk (commented):
Two comments. One I think is pretty substantial.

engine/src/gui/config_screen.cpp
stephengtuggy: > By enabling a font picker for the user to specify the font, you are taking away the option to support more than one font in the game.

Does the game have that ability today? I don't think it does.

evertvorster: I don't follow either. The game only ever supported one font at a time, and never had a way to switch fonts. 
This font picker just makes picking which font the game runs with easier. 
So, it adds functionality that was never there, and takes nothing away. 

Moreover, if you wanted to use a specific font for something in-game, there is nothing preventing you from adding that font to assets and referring to it directly. The font picker just picks the default font at this point.

engine/src/gui/config_screen.cpp
evertvorster: s_fonts_list vs s_fonts_values
They're parallel arrays, kept in sync, but they serve different roles:

Purpose	Used for:
1. s_fonts_list	Display — the labels shown in the dropdown	ImGui::Button(s_font_list[sel_font]), ImGui::MenuItem(s_font_list[i])
2. s_fonts_values	Persistence — the graphics.font config value for each entry	select_font_from_value(g.font) reverse-lookup, and graphics.font = s_font_values[sel_font] on Save

So s_fonts_list is what the user sees, and s_fonts_values is what gets written to the config when that entry is selected. select_font_from_value() scans s_fonts_values to map a persisted graphics.font string back to the matching dropdown index.

The reason they look identical: right now each display label is the font name (Roboto, then the .ttf filenames), so both lists contain the same strings. The separation is intentional so a display label can differ from the persisted value — e.g. a friendlier menu name ("Saira (Default)") mapping to the actual filename (Saira-Regular.ttf), or showing "Roboto" while storing the engine-default sentinel. That's the whole reason the two exist rather than one.

So: functionally identical today, but list is the UI label and values is the persisted value — a display↔value mapping kept separate on purpose. 

Eventually we want to make the config screen more user-friendly, and then this will become genuinely useful.

engine/src/gui/config_screen.cpp
```diff
@@ -80,32 +80,56 @@ static bool bind_capturing = false;
...
+    }
+}
+
+// Set sel_font from a persisted graphics.font value ("Roboto" or a .ttf filename).
+static void select_font_from_value(const std::string &value) {
+    if (s_font_values.empty()) {
+        refresh_font_list();
+    }
+    sel_font = 0;
+    for (size_t i = 0; i < s_font_values.size(); ++i) {
```
stephengtuggy, line 126: This is not the most elegant code for finding a match. You could use `std::find`. Better yet, you could use an `std::map` or similar.

I'm not going to block on it though. With the number of entries we're dealing with, lookup performance doesn't matter much.

stephengtuggy (approved):
Looks good to me. @royfalk can you take a look at Evert's and my responses when you get a chance?

Issue #1773, opened 2026-09-05
Base computer font not adjusting with font settings


evertvorster:
When adjusting font height settings, the base computer does not adjust its font sizes.

stephengtuggy:
@evertvorster this is fixed in the above two PRs, right?

evertvorster:
> [@evertvorster](https://github.com/evertvorster) this is fixed in the above two PRs, right?

Correct

PR #1774, opened 2026-09-05
Fixes base computer internal representation of font sizes. 

evertvorster:

**THIS PR HAS AN ACCOMPANYING ASSETS PR**
https://github.com/vegastrike/Assets-Production/pull/287

Please answer the following:

Code Changes:
  - [x] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1773

Purpose:
- - Bring the base computer font size handling in line with the rest of the game. 
- - Allow for the user to set the font size in the config, and it to be applied in the base computer.
- - Revert a scaling factor that was applied in 2011 that made all the font sizes appear wrongly sized. 
- What release is this for?
- - 0.11
- - 0.11


stephengtuggy:
@evertvorster looks like engine/src/gui/painttext.cpp needs to have some conflicts resolved.

evertvorster:
> @evertvorster looks like engine/src/gui/painttext.cpp needs to have some conflicts resolved.

Yeah, we actually deleted it in this branch. :)

engine/src/gui/font.h
```diff
@@ -43,13 +43,16 @@ static const float BOLD_STROKE = 1.5;
 class Font {
 public:
 //Font size.  Vertical distance in identity space.
+    //
+    // Removed the legacy `* 0.5` (added 2011, commit e3b743d16a, as a scoped
```
royfalk: Comments like this should not appear in the code. Instead, after you file a PR, you can review yourself and add them.

Once we remove something from the code, there really is no need to have a record of it anymore.

engine/src/gui/font.h
```diff
@@ -43,13 +43,16 @@ static const float BOLD_STROKE = 1.5;
...
+    // "stroke font size" fix) so size() returns the honest m_size. The font
```
royfalk: Reword this comment. It should be something like:
This function returns the font size as a percentage of the screen height.
It should then move to above the definition of m_size and not the function which is a simple getter.

engine/src/gui/imguitext.cpp
```diff
@@ -228,17 +230,18 @@ int ImGuiText::visibleLineCountStartingWith(int lineNumber, float vertInterval)
...
+ImVec2 ImGuiText::getTextWidth(const std::string text) {
+    // Glyph height comes from the user-set Text Height (font_point), matching what
+    // draw() renders at, so layout (wrap width, line height, centering, scroll
+    // counts) agrees with what is actually drawn.
+    const float pixelFontSize = configuration().graphics.font_point_flt;
+    ImFont* font = ImGui::GetFont();
+    if (font && font->IsLoaded()) {
+        return font->CalcTextSizeA(pixelFontSize, FLT_MAX, -1.0f, text.c_str());
+    }
+    // No font bound yet (measuring outside a render window): fall back to the
```
royfalk, line 242: This is why I hate the default font. We now have these small code snippets everywhere.

engine/src/gui/modaldialog.cpp
```diff
@@ -86,7 +86,7 @@ static void CreateControlsForAlertWindow(Window *window) {
     title->setJustification(CENTER_JUSTIFY);
     title->setTextColor(GUI_OPAQUE_WHITE());
     title->setColor(GUI_CLEAR);
-    title->setFont(Font(.07, BOLD_STROKE));
+    title->setFont(Font(.045, BOLD_STROKE));
```
royfalk, line 89: These should really be enum somewhere.
Something like:
```cpp
enum FontHeight {
   extra_small = 0.04,
   small = 0.045,
   medium = 0.05,
   large = 0.055,
   ...
}
```

royfalk (commented):
A few minor comments.
I assume this change is primarily because we changed how the font heights are calculated in another PR?

engine/src/gui/modaldialog.cpp
evertvorster: Hmm, could be. 
However, the relative font height picked here is for the author of the screen to decide. I agree with you, in basecomputer.cpp this would make a lot of sense.

engine/src/gui/font.h
evertvorster: Done

engine/src/gui/font.h
evertvorster: Done

engine/src/gui/imguitext.cpp
evertvorster: There is definitely room for improvement. :) Can we tackle this in a separate issue, though?

engine/src/gui/modaldialog.cpp
evertvorster: You are right. 
However, the level where this should be implemented is unclear. 

Should it be per interface, or global? 

The intent in basecomputer is to draw buttons text and info text and of course some of these will be larger than others... but I'm not sure if it would be better to have one definitive button_text height defined, or leave it up to the author of the screen to pick how large a certain piece of text should be as a default. (Remember that the user can still select a larger font, which will propagate throughout the game)

engine/src/gui/imguitext.cpp
royfalk: Yes. This is just me venting and trying to convince other people to drop the whole default font and just trust the assets.

engine/src/gui/modaldialog.cpp
royfalk: I think you misunderstood me. If we implement this enum, the code would be something like:
```cpp
title->setFont(Font(FontSize.small, BOLD_STROKE));
```

This makes it much clearer to anyone reading the code.

evertvorster: OK, but enums hold integers, these are fractions. 

Since we are still battling out the coords system, maybe we switch to a coord of 100 x 100, and then text sizes can be represented as whole numbers, ie: 2 = 2% of screen height?

The point here is, we should not block this PR on this discussion, as we are going to overhaul this area anyways, with a targeted PR. The purpose of THIS PR is to bring the base computer in line with the rest of the game.

engine/src/gui/imguitext.cpp
evertvorster: We've had a default font for a long time, ever since we started porting over to imgui. 
imGui is not using the configured helvetica, as it cannot load bitmapped fonts. 

So, we were using the fallback included Roboto font... A default, NOT in Assets. 

I think my point is made.

However, I still think that there is room for improvement, but for defensive coding, we need a fallback, and we can design it into the system  better so there does not have to be all these little snippets of code hanging around. 

You might agree that there is quite a bit of housekeeping to do on this code base.

engine/src/gui/modaldialog.cpp
royfalk: Yeah, that's fine.

royfalk (approved):
Go ahead and ship it.

PR #1784, opened 2026-09-07
Fix ship turning inertia not tracking cargo mass

### LINKED ISSUE #1783: turning rate not updating live with cargo weight .
So, here's the situation:

You load a saved game where your ship is empty. Then you load up on heavy cargo. 
If you just take off, your ship turns like it is empty. 
If you saved your game, exited VS and then reloaded your game, your ship now turns slowly, like it is actually trying to turn more weight. 

Same with getting rid of heavy cargo. 
If you load a saved game with heavy cargo, and sold the cargo then continued flying, your ship will continue to feel sluggish when turning until you have saved and loaded the save file.

evertvorster:
The ship's moment of inertia (Momentofinertia) was captured once at load (LoadRow: Momentofinertia = GetMass()) and never updated when cargo or upgrades changed the ship's mass during a session. Turning/handling divides by GetMoment(), so if the unit was created/loaded while carrying a heavy load, the ship stayed sluggish (reduced turn responsiveness) even after selling all the cargo, until a reload rebuilt the unit at the lighter mass.

Same bug family as #1268/#1273 (mass handling moved into ComponentsManager by 1fbd2a5ea). The codebase already treats GetMoment() as the ship's mass (e.g. navigation.cpp uses GetMoment() as mass for braking/deceleration), so it should track the live mass, not a load-time snapshot.

Fix: make Movable::GetMoment() virtual and override it on Unit to return the current GetMass() (base + cargo + upgrades), keeping angular inertia in sync with what the ship is actually carrying.

Please answer the following:

Code Changes:
  - [ ] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [ ] Save the game again.
    - [ ] Upgrade your new ship.
    - [ ] Save once more.

Issues:
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1783

Purpose:
- - Fix a bug with ship handling. 
- What release is this for?
- - 0.11
- - 0.11


evertvorster:
A normal play test won't catch this one. You need to be loading up heavy cargo and saving and exiting and generally hustling to notice this bug. 

In any case, I'll do the normal play testing shortly. The fix is fairly simple, though, and unlikely to touch any other part of the game.

royfalk (commented):
We should not be using `GetMoment` at all. 

If you look at the definition of `mass` and `base_mass` in _components_manager.h_, they replace the old `mass` and `moment` variables.

Mass is updates to include both cargo and upgrades.

evertvorster:
> We should not be using `GetMoment` at all.
> 
> If you look at the definition of `mass` and `base_mass` in _components_manager.h_, they replace the old `mass` and `moment` variables.
> 
> Mass is updates to include both cargo and upgrades.

Reverted my initial fix, and implemented your intention. This of course touched a few more files....

engine/src/python/python_unit_wrap.h
```diff
@@ -192,8 +192,6 @@ voidWRAPPED1( SetAngularVelocity, Vector, v
 
 #if (WHICH_UNIT_EXPORT_NUMBER == 1) || (WHICH_UNIT_EXPORT_NUMBER == 2)
 #else // (WHICH_UNIT_EXPORT_NUMBER == 1) || (WHICH_UNIT_EXPORT_NUMBER == 2)
-WRAPPED0(float, GetMoment,
```
royfalk, line 195: Be very careful removing python API unless you check, at a minimum, it's not called by VS.

libraries/cmd/movable.cpp
```diff
@@ -323,8 +323,9 @@ Vector Movable::ResolveForces(const Transformation &trans, const Matrix &transma
     if (NetTorque.i || NetTorque.j || NetTorque.k) {
         temp1 += InvTransformNormal(transmat, NetTorque);
     }
-    if (GetMoment()) {
-        temp1 = temp1 / GetMoment();
+    const double angular_mass = unit->GetMass();
```
royfalk: I wonder if it wouldn't be cleaner to just do:
```cpp
const float angular_mass = unit->GetMass();
```

royfalk (approved):
Overall looks fine. 
Two minor comments. 
It would be a good idea for a second person to review this.
It seems a minor thing but it's touching pretty critical stuff.
Bugs introduced here are a pain to figure out compared to say _base_computer_.

libraries/cmd/movable.cpp
evertvorster: Yes, thanks for the tip.

engine/src/python/python_unit_wrap.h
evertvorster: Yes, it was a bit of a pain to track down all the call sites to properly remove GetMoment, as I believe your intent was?
I'll wait for a second pair of eyes to look over this PR before merging.

stephengtuggy: I believe we also need to check the contents of the python*.zip files in Assets-Production. At least the Python 3.x one. And for the sake of completeness, we should probably check the Assets-Master repo as well.

stephengtuggy: https://github.com/vegastrike/Assets-Production/blob/master/py3_modules.7z , to be precise.

evertvorster: **Checked for `GetMoment` callers in shipped Python**

`GetMoment` is the only symbol this PR removes from the Python API
(`WRAPPED0(float, GetMoment, 0)` in `engine/src/python/python_unit_wrap.h`), and it is gone from
the engine entirely. What I checked for callers:

| Where | Result |
|---|---|
| Assets-Production: the contents of `py3_modules.7z` and `py27_modules.7z` | no references. Extracted and grepped **all** files (2263 and 2182), so `.pyc` attribute names would have matched too. |
| Assets-Production: every tracked text file — missions, XML, `.py`, JSON | no references. Mission files embed Python, so an inline script calling it would show up here. |
| Assets-Masters | no references in its 593 text files, and it ships no `py3_modules.7z` (or any other Python archive). |
| Assets-Production `0.5.4` | ships a *different* archive from master: only CPython `modules/builtin/*.pyc` stubs, no game code, no references. |
| Installed game data, and my own campaign/save data | no references. |

The two module bundles are also byte-identical on `origin/master` and my playground branch
(both `763f668a…`), so there is no unmerged-assets angle.

Roy's caution — be very careful removing Python API without checking it is not called by VS — is
what this covers, and his `unit->GetMass()` suggestion is in (`144d379ee`).

stephengtuggy: Well, I ran that check, and didn't find anything, so it looks like we're OK. We'll just need to make sure to document this change to the appropriate version of the engine API.

evertvorster: So, with this commit, GetMoment is officially removed from VS

stephengtuggy (approved):
LGTM!

PR #1791, opened 2026-09-14
Fix fleet transfer price

### LINKED ISSUE #1790: Fleet ship shows no Transfer button unless you can afford to buy the ship outright
Flying a Llama with a Plowshare in the fleet (no jump drive on the Plowshare: it
could not follow a jump). Jumped out of the system, saved, jumped back in to
collect it. In the ship dealer the Plowshare appears in "My Fleet", but there is
no Transfer button at all, and it cannot be flown to — it is simply "elsewhere".

Cause: isTransactionOK() gates BUY_SHIP on the item's price. For a ship you
already own that is the purchase price (master_part_list.csv: Plowshare.stock =
180000), not the transport cost (PlayerShip::transfer_price = 1000 in-system).
With 1082 credits the check fails, hideCommitControls() runs, and no button is
drawn — even though the transfer itself is affordable.

Fix: test transfer_price for an owned ship (item.index > 0). Also clamp the jump
count in UpdateTransportPricesForOwnedShips(), where an empty jump path made it
-1 and priced the transfer below zero.

Expected: if you can afford the in-system transport cost, the ship comes to you,
regardless of where it is parked or whether that base still exists.

evertvorster:
Fleet ships showed no Transfer button unless you could afford to buy the ship
outright.

BUY_SHIP is one transaction covering two different operations. isTransactionOK()
gated it on the item's price alone, but buyShip() dispatches on item->index:

  index > 0  "already own ship" -> charges PlayerShip::transfer_price
                                  (0 same base, 1000 in-system, 25000/jump)
  index == 0 a real purchase    -> charges item->GetPrice()

Those prices are wildly different — a Plowshare.stock is 180000 to buy and 1000
to move in-system. Applying the purchase price to an owned ship meant that with
less than 180000 credits the check failed, hideCommitControls() ran, and the
button was not drawn at all. A ship parked at a base that had since flown away
therefore looked unreachable even though moving it was affordable.

The gate now mirrors buyShip()'s own dispatch, so the two agree on affordability
in both branches. As a side effect the transfer branch's existing
`transfer_price > credits` check is reachable again — it was dead code, which is
also why refusing a transfer failed silently: it always returned from the gate,
never from there. (item->index > 0 is the codebase's existing owned-marker; for
buying, item->GetPrice() is still what's tested.)

Also fixes a related pricing bug in the same area:
UpdateTransportPricesForOwnedShips() computed the jump count as
jumps_vector.size() - 1, which is -1 when getJumpPath() finds no route. A
negative jump count priced the transfer below zero — affordable-looking, and it
would have added credits instead of spending them.

Not addressed here, worth separate issues: a refused transfer still fails
silently, and a PlayerShip::base naming a base that has gone away is never
repaired.

Please answer the following:

Code Changes:
  - [ ] Play Tests
    - [ ] From the Main Menu, start a new Campaign.
    - [ ] Basic flight.
    - [ ] Docking to a planet.
    - [ ] On planet actions:
        - [ ] Save the game.
        - [ ] Read news.
        - [ ] Buy/Sell some goods.
    - [ ] SPEC flight.
    - [ ] Save the game again.
    - [ ] Upgrade your new ship.
    - [ ] Save once more.
- [x] Simple Bugfix

Issues:
- - #1790 

Purpose:
- - Fix a small bug
- What release is this for?
- - 0.11
- - 0.11


royfalk (approved):
This is one of the reasons the old code is so difficult.
The original devs should not have recycled buy for transferring ships.
It should be separate, making the code not just clearer but also more robust.

I never encountered the bug because I always play with a lot of money in such scenarios, since I need it to buy ships for testing.

evertvorster:
Yeah, the old coders certainly knew a thing or two about code re-use. Sometimes to a fault, as is evidenced here. Thanks for the review.

PR #1801, opened 2026-09-18
Frame rate limiter in config, and also optionally hide fps counter.

evertvorster:
The game runs incredibly fast, drawing many frames you simply will not see. 
This is wasteful, and makes gaming laptops noisy as they run flat out for no reason. 

Implemented a few popular frame rate mechanisms. 

vsync: just syncronizes to the monitor. 
Half monitor is simply half the refresh rate the monitor supports, so my laptop supports 240fps, and this would make it an effective 120fps. 

Then there is a drop-down with the refresh rates your monitor actually supports. 

This makes the game run way cooler and quieter, and you still have the unbounded option when doing optimizations. 
Lastly, FPS counter should not be a part of the game, and is disabled by default, but can be enabled for aformentioned optimizations.

Please answer the following:

Code Changes:
  - [ ] Play Tests
    - [ ] From the Main Menu, start a new Campaign.
    - [ ] Basic flight.
    - [ ] Docking to a planet.
    - [ ] On planet actions:
        - [ ] Save the game.
        - [ ] Read news.
        - [ ] Buy/Sell some goods.
    - [ ] SPEC flight.
    - [ ] Save the game again.
    - [ ] Upgrade your new ship.
    - [ ] Save once more.

Issues:

Purpose:
- What release is this for?


stephengtuggy:
Yes please! I am very interested in the frame rate limiter feature. I was going to request that you submit a Pull Request for it, @evertvorster , but you beat me to it.

Just one change I would like to request. The FPS display in the upper right of the HUD has been part of the game for years. Can we keep it in by default, while allowing it to be hidden if desired?

engine/src/gldrv/winsys.cpp
```diff
@@ -191,6 +191,68 @@ void winsys_swap_buffers() {
...
+
+void winsys_apply_frame_limit() {
+    const auto &g = configuration().graphics;
+
+    int interval = 1; // on (monitor)
+    if (g.vsync == "off") {
+        interval = 0;
+    } else if (g.vsync == "adaptive") {
+        interval = -1;
+    }
```
stephengtuggy: Maybe this could be a `class enum`, with three possible values?

engine/src/gldrv/winsys.cpp
```diff
@@ -191,6 +191,68 @@ void winsys_swap_buffers() {
...
+int winsys_monitor_refresh() {
+    SDL_Window *window = SDL_GL_GetCurrentWindow();
+    if (!window) {
+        return 0;
+    }
+    const SDL_DisplayMode *mode = SDL_GetCurrentDisplayMode(SDL_GetDisplayForWindow(window));
+    if (!mode) {
+        return 0;
+    }
+    return static_cast<int>(mode->refresh_rate + 0.5f);
```
stephengtuggy: The `0.5f` is for rounding purposes, isn't it? That wasn't immediately obvious to me as I read it.

I was going to suggest using something like `std::rint`, but the error handling for that is much too complicated. Maybe `std::nearbyint`? But you would still have to add

```c++
#include <cfenv>
#include <cmath>
#pragma STDC FENV_ACCESS ON
```

to the top of the file. And even then, I'm not sure it's guaranteed to work across all compilers.

engine/src/configuration/configuration.h
```diff
@@ -1724,6 +1724,12 @@ namespace vega_config {
         double fov_dbl = 37.5;
         float fov_flt = 37.5;
         bool framerate_changes_shader = false;
+        // Frame-rate limiting: vsync mode, and a software cap when it is below what
+        // the GPU would otherwise produce.
+        std::string vsync = "on";                    // "off" | "on" | "adaptive"
+        std::string frame_limit_mode = "unlimited";   // "unlimited" | "half" | "fixed"
+        int max_framerate = 0;                        // fixed cap, frames per second
+        bool show_fps = false;
```
stephengtuggy: I'd like to ask that this be `true` by default. It is most likely what longtime players will expect.

engine/src/gldrv/winsys.cpp
stephengtuggy: Something like:

```c++
class enum VegaVSyncMode {
    kOff = 0,
    kOnMonitor = 1,
    kAdaptive = -1,
}
```

stephengtuggy: And parse the JSON config setting into this enum internally.

engine/src/gfx/cockpit.cpp
```diff
@@ -2009,6 +2009,15 @@ void GameCockpit::Draw() {
         }
     }
 
+    // Optional FPS readout, drawn over everything. The shipped cockpits have no FPS
```
stephengtuggy: Huh? They don't?

engine/src/gldrv/winsys.cpp
stephengtuggy: https://en.cppreference.com/cpp/numeric/math/nearbyint

engine/src/configuration/configuration.h
evertvorster: done

stephengtuggy (approved):
LGTM (Looks Good To Me)! Let's ship it.

PR #1803, opened 2026-09-19
shield: drop the max-scaled upkeep and charge the rated recharge rate

evertvorster:
This PR returns shields generation to how they used to generate. 
Sorry, but the new model is just broken. 

In my bug report, the displayed value for power draw in shield info is much, much less than what the generator can generate, and so the shields should never drain.

Shield::Regenerate charged max_shield * maintenance_factor every frame while the shields were below full.
For a shield pool bigger than the reactor can supply, that out-drew the reactor and drained the primary capacitor; and it stalls just short of full and never reached the free-at-full gate.

Removed that upkeep block. The shield now draws its rated recharge rate (Shield_Recharge, from the unit data) while it is below full, scaled by its efficiency: a damaged shield recharges more slowly and draws more energy for the same charge, but still reaches full - and a full shield costs nothing.

This restores the pre-#1016 behaviour of charging on the regeneration rate rather than the shield pool, and implements the old TODO to scale by shield efficiency. maintenance_factor and regeneration_factor are no longer read.

Please answer the following:

Code Changes:
  - [x] Play Tests
    - [x] Basic flight.
    - [x] Docking to a planet.
    - [x] On planet actions:
        - [x] Save the game.
        - [x] Read news.
        - [x] Buy/Sell some goods.
    - [x] SPEC flight.
    - [x] Save the game again.
    - [x] Upgrade your new ship.
    - [x] Save once more.

Issues:
- - https://github.com/vegastrike/Vega-Strike-Engine-Source/issues/1804

Purpose:
- - Fix shields for everyone
- What release is this for?
- - 0.11
- - 0.11


engine/src/components/shield.cpp
```diff
@@ -352,29 +352,26 @@ void Shield::Regenerate(const bool player_ship) {
...
     }
 
-    // Manually throttle shield strength
-    if(Percent() > max_power) {
-        Decrease();
+    // Shield Regeneration. The shield draws its rated recharge rate from the
+    // capacitor (Shield_Recharge, mj/s), scaled by its efficiency: a damaged
+    // shield recharges more slowly and draws more energy for the same charge,
+    // but still reaches full (see the gate above).
+    const double shield_efficiency = PercentOperational();
```
royfalk: Unlike every other component, shields can be damaged in two ways:
1. Damage to the **shield generated**
2. Damage to the **shield generator**
It's important to note this difference.

PercentOperational() is not a shield function. It's a component function. It deals with damage to the **generator**.

Take a look at `void Shield::Damage()`. It should be clear enough.
We randomly damage each shield facet (front, rear, ...) and the `regeneration` figure. We calculate `operational` as an average of them.

engine/src/components/shield.cpp
```diff
@@ -352,29 +352,26 @@ void Shield::Regenerate(const bool player_ship) {
...
+    if (shield_efficiency <= 0.0) {
```
royfalk: Obviously. If the shield generator doesn't work, no need to do anything.
But I think you wanted to calculate damage to the **shield generated**, no?

engine/src/components/shield.cpp
```diff
@@ -352,29 +352,26 @@ void Shield::Regenerate(const bool player_ship) {
...
         return;
     }
 
-    // Shield Regeneration
-    const double shield_regeneration_cost = regeneration.AdjustedValue() * configuration().components.shield.regeneration_factor_dbl;
+    const double shield_regeneration_cost = regeneration.MaxValue() / shield_efficiency;
     SetConsumption(shield_regeneration_cost);
     const double actual_regeneration_percent = Consume();
-    double regen = actual_regeneration_percent * regeneration.AdjustedValue() * simulation_atom_var;
+    double regen = actual_regeneration_percent * regeneration.MaxValue() * shield_efficiency
```
royfalk: By switching from `regeneration.AdjustedValue` to `regeneration.MaxValue,` you've removed the ability to damage the shield generator in that aspect.

engine/src/components/shield.cpp
```diff
@@ -352,29 +352,26 @@ void Shield::Regenerate(const bool player_ship) {
...
+    // shield recharges more slowly and draws more energy for the same charge,
+    // but still reaches full (see the gate above).
+    const double shield_efficiency = PercentOperational();
+    if (shield_efficiency <= 0.0) {
         return;
     }
 
-    // Shield Regeneration
-    const double shield_regeneration_cost = regeneration.AdjustedValue() * configuration().components.shield.regeneration_factor_dbl;
+    const double shield_regeneration_cost = regeneration.MaxValue() / shield_efficiency;
```
royfalk: Same as below. We could argue that a damaged shield generator still draws the same power. But I'm not sure this was intentional.

royfalk (changes requested):
> Removed that upkeep block. ...
> This restores the pre-https://github.com/vegastrike/Vega-Strike-Engine-Source/pull/1016 behaviour of charging on the regeneration rate rather than the shield pool

I'm not sure what shield pool means in this context, but the old behavior had shield maintenance costs.
The old formula was:
```cpp
energy -= shield.recharge * VSD
        / ( 100 * (shield.efficiency ? shield.efficiency : 1) ) 
        / shieldenergycap * shield.number * shield_maintenance_cost
        * simulation_atom_var * ( (apply_difficulty_shields) ? g_game.difficulty : 1 );
```

I don't care enough to block this change, so it'll be up to someone else to approve. 

-------------------
Edit: Nope. I suspect the AI missed a few things this time. Blocking until fixed.

engine/src/components/shield.cpp
evertvorster: Its plenty clear, and good catch. The way it was implemented here diluted the damage to the generator itself. A generator that is 50% damaged should work 50% as good as an undamaged unit. 

Now, what that degradation is might be a whole topic of discussion. Here I picked that it would take more energy and charge the shield slower, but still eventually build a shield with 100% strength. But, we could pick other metrics, like disabling facets, or total shield strength... the list is endless. But, for this PR, just having some degradation should suffice, right?

engine/src/components/shield.cpp
evertvorster: Hmm, are you saying a damaged shield generator should not be able to generate the same strength of shield as an undamaged unit?

Unfortunately, this would cause a constant power draw as the shields can never get to 100%?

There are other knock-on effects as well, which would make this PR balloon out of control.

evertvorster:
Thanks for taking a look @royfalk. I implemented some changes to make it more faithful to how it used to work. Shields do need a discussion, but for now this PR allows me to fly my Vendetta as I used to, and also some of the larger and heavier ships.

engine/src/components/shield.cpp
evertvorster: Good catch. Fixed

engine/src/components/shield.cpp
evertvorster: Same same

engine/src/components/shield.cpp
royfalk: I'm not sure why you're implementing this functionality as if there isn't already an implementation.
The current one has the following:
1. Regeneration - damage reduces regeneration
2. Facets - damage reduces facet max strength.

engine/src/components/shield.cpp
royfalk: > Hmm, are you saying a damaged shield generator should not be able to generate the same strength of shield as an undamaged unit?

Yes.

> Unfortunately, this would cause a constant power draw as the shields can never get to 100%?

Not as I wrote it. If you use `AdjustedValue` instead of `MaxValue`, it wouldn't either.

> There are other knock-on effects as well, which would make this PR balloon out of control.

Why? I don't agree with what you're doing, but doing it right should be a few lines.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
         return;
     }
 
-    // Manually throttle shield strength
-    if(Percent() > max_power) {
-        Decrease();
+    // Shield Regeneration. The generator's own efficiency sets the recharge rate and the
+    // power drawn: a damaged generator recharges more slowly and costs more, but the
+    // shields still reach full (see the gate above).
+    const double shield_efficiency = (regeneration.MaxValue() > 0.0)
```
royfalk: I thought you wanted damaged shields to regenerate slower.
Here you are taking regeneration, which is a `Resource<double>` and trying to implement `Percent`, which is already implemented in order to calculate the adjusted regeneration value, which is also implemented.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
+            ? regeneration.AdjustedValue() / regeneration.MaxValue() : 1.0;
+    if (shield_efficiency <= 0.0) {
```
royfalk: And Resource supports these checks implicitly, so no need if you use regeneration.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
-    SetConsumption(shield_regeneration_cost);
+    // Upkeep while below full, plus the cost of the charge being rebuilt (capped at the
+    // regeneration rate); both scale with the generator's efficiency.
+    const double vsd_percent = configuration().components.fuel.vsd_mj_yield_dbl / 100.0;
+    const double shield_maintenance_cost = regeneration.MaxValue() * vsd_percent
+            / shield_efficiency
+            / configuration().physics.shield_energy_capacitance_dbl
+            * static_cast<double>(number_of_facets)
+            * configuration().physics.shield_maintenance_charge_dbl;
+    const double shield_deficit = TotalMaxLayerValue() - TotalLayerValue();
```
royfalk: Simply call `Percent`. It is the DamageableLayer method that does this.
```cpp
double DamageableLayer::Percent() const {
    if(number_of_facets == 0) {
        return 0;
    }

    double percents = 0;
    for(const Resource<double>& facet:facets){
        percents += facet.Percent();
    }

    return percents/number_of_facets;
}
```

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
+    if (shield_efficiency <= 0.0) {
         return;
     }
 
-    // Shield Regeneration
-    const double shield_regeneration_cost = regeneration.AdjustedValue() * configuration().components.shield.regeneration_factor_dbl;
-    SetConsumption(shield_regeneration_cost);
+    // Upkeep while below full, plus the cost of the charge being rebuilt (capped at the
+    // regeneration rate); both scale with the generator's efficiency.
+    const double vsd_percent = configuration().components.fuel.vsd_mj_yield_dbl / 100.0;
```
royfalk, line 361: So we're bringing the arcane back into the code?
`vsd_mj_yield_dbl` is Vegastrike specific. 
I'm not even sure what the acronym stands for.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
+    const double shield_maintenance_cost = regeneration.MaxValue() * vsd_percent
```
royfalk, line 362: You're mixing old code with new. 
Shield maintenance should be energy to cost to maintain the current shields. You're calculating it using `regeneration`, which implies the cost to regenerate.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
+            / shield_efficiency
```
royfalk, line 363: Also regeneration and not maintenance.

engine/src/components/shield.cpp
```diff
@@ -352,29 +353,36 @@ void Shield::Regenerate(const bool player_ship) {
...
+            / configuration().physics.shield_energy_capacitance_dbl
+            * static_cast<double>(number_of_facets)
+            * configuration().physics.shield_maintenance_charge_dbl;
+    const double shield_deficit = TotalMaxLayerValue() - TotalLayerValue();
+    const double maximum_charge = std::min(shield_deficit, regeneration.AdjustedValue());
```
royfalk, line 368: These two are unrelated. One is in shield strength and the other is regeneration. Why do a minimum function of them?

royfalk (commented):
I don't know where you're going here or whether DeepSeek has gone off the rails. 

If you want to set maintenance at 0, just delete the following line:
```cpp
SetConsumption(shield_maintenance_cost);
const double actual_maintenance_percent = Consume();
```

If you want regeneration to be quicker the stronger the shields are:
```cpp
double shield_efficiency = std::min(Percent(), 0.25);
double regen = actual_regeneration_percent * regeneration.AdjustedValue() * simulation_atom_var * shield_efficiency;
```

Note that I added the minimum 0.25 so if you lose shields, they'll actually regenerate. Obviously, you can change that to a settings variable. 

---------------------------------
If you just want to play with Vendetta or you think the current shield costs are wrong (they probably are), figure out the right values for `maintenance_factor_dbl` and `regeneration_factor_dbl`.

evertvorster:
> If you just want to play with Vendetta or you think the current shield costs are wrong (they probably are), figure out the right values for maintenance_factor_dbl and regeneration_factor_dbl.


I think you are missing the point here.

This is a bugfix for code that worked as designed until #1016 refactored it. Shields used to draw their rated recharge rate; #1016 replaced that with max_shield × maintenance_factor, and that is what drains the capacitor.

There is no single factor on the shield pool that reproduces the old behaviour. Drawing pool × f needs f = 0.0035 at tier 6 and 0.0043 at tier 1 — a 23% spread across the fifteen buyable shields, because the strength-per-recharge ratio isn't constant. The old model is one constant: 0.324 × the shield's rated Shield_Recharge, identical at every tier. That rated rate is the advertised figure on the ship info screen, next to the reactor's recharge — the number a player compares against their reactor when choosing a shield. Replacing per-type authored data with a pool formula is what broke this.

The constants you called arcane are the shipped tuning values that produced that behaviour; re-deriving them in a new shape improves nothing, so #1803 keeps them.

I did take your API points: efficiency is now regeneration.Percent(), and the hand-rolled Percent is gone. The full-charge gate now runs on the damaged maximum, so a damaged generator does cap shield strength — and reaching that ceiling is free, which is what stops the constant draw you were worried about.

What remains is empirical, and it's in the table below:
| Shield | Per facet x4 | Pool | Advertised recharge (MJ/s) | Old model (MJ/s) | old / adv | Current code @mf=0.05 (MJ/s) | / adv | Current code @mf=0.01 (MJ/s) |
|---|---|---|---|---|---|---|---|---|
| quadshield01 | 75 x4 | 300 | 4 | 1.3 | 0.324 | 15.2 | 3.80 | 3.2 |
| quadshield02 | 150 x4 | 600 | 8 | 2.6 | 0.324 | 30.4 | 3.80 | 6.4 |
| quadshield03 | 300 x4 | 1200 | 14 | 4.5 | 0.324 | 60.7 | 4.34 | 12.7 |
| quadshield04 | 400 x4 | 1600 | 18 | 5.8 | 0.324 | 80.9 | 4.49 | 16.9 |
| quadshield05 | 500 x4 | 2000 | 22 | 7.1 | 0.324 | 101.1 | 4.60 | 21.1 |
| quadshield06 | 600 x4 | 2400 | 26 | 8.4 | 0.324 | 121.3 | 4.67 | 25.3 |
| quadshield07 | 700 x4 | 2800 | 32 | 10.4 | 0.324 | 141.6 | 4.42 | 29.6 |
| quadshield08 | 800 x4 | 3200 | 36 | 11.7 | 0.324 | 161.8 | 4.49 | 33.8 |
| quadshield09 | 900 x4 | 3600 | 40 | 13.0 | 0.324 | 182.0 | 4.55 | 38.0 |
| quadshield10 | 1000 x4 | 4000 | 44 | 14.3 | 0.324 | 202.2 | 4.60 | 42.2 |
| quadshield11 | 1200 x4 | 4800 | 54 | 17.5 | 0.324 | 242.7 | 4.49 | 50.7 |
| quadshield12 | 1400 x4 | 5600 | 62 | 20.1 | 0.324 | 283.1 | 4.57 | 59.1 |
| quadshield13 | 1600 x4 | 6400 | 72 | 23.3 | 0.324 | 323.6 | 4.49 | 67.6 |
| quadshield14 | 1800 x4 | 7200 | 80 | 25.9 | 0.324 | 364.0 | 4.55 | 76.0 |
| quadshield15 | 2000 x4 | 8000 | 90 | 29.2 | 0.324 | 404.5 | 4.49 | 84.5 |

Old model    = regeneration x 0.054 / 0.2 x 4 x 0.25 (upkeep) + min(deficit, regeneration) x 0.054 (rebuild)
Current code = pool x maintenance_factor + regeneration x regeneration_factor
old / adv    = 0.054 x (4 x 0.25 / 0.2 + 1) = 0.324, constant for every shield

| Shield | Pool factor needed to reproduce the OLD behaviour | Pool factor needed to match the ADVERTISED rate |
|---|---|---|
| quadshield01 | 0.00432 | 0.01333 |
| quadshield02 | 0.00432 | 0.01333 |
| quadshield03 | 0.00378 | 0.01167 |
| quadshield04 | 0.00364 | 0.01125 |
| quadshield05 | 0.00356 | 0.01100 |
| quadshield06 | 0.00351 | 0.01083 |
| quadshield07 | 0.00370 | 0.01143 |
| quadshield08 | 0.00364 | 0.01125 |
| quadshield09 | 0.00360 | 0.01111 |
| quadshield10 | 0.00356 | 0.01100 |
| quadshield11 | 0.00364 | 0.01125 |
| quadshield12 | 0.00359 | 0.01107 |
| quadshield13 | 0.00364 | 0.01125 |
| quadshield14 | 0.00360 | 0.01111 |
| quadshield15 | 0.00364 | 0.01125 |

Old behaviour: 0.00351 (tier 6) to 0.00432 (tier 1) - a 23% spread, so no single value works.
Advertised:     0.01083 (tier 6) to 0.01333 (tier 1) - also a 23% spread; even 0.01 is only right on average,
                and it would draw roughly 3x the old model.

engine/src/components/shield.cpp
evertvorster: The current implementation is also the cause of the bug report with this PR.

engine/src/components/shield.cpp
evertvorster: Fixed

engine/src/components/shield.cpp
evertvorster: Done

royfalk:
The formula in #1016 for shield cost was:
```cpp
( (shield.recharge*VSD/100*simulation_atom_var*shield.number/shieldenergycap)
                                        > energy ) ? (energy*shieldenergycap*100/VSD
                                                      /shield.number) : shield.recharge*simulation_atom_var
```

Which means (unless I made a mistake plugging in the numbers):
```
If recharge * 0.054 * 0.01 * (2 or 4) / 0.2 > energy
then
cost = energy * 0.2 * 100 / 5.4 / 4
else
cost = recharge * simulation
```

Now I don't get what or why they did this and what this is supposed to be.
But I'm guessing we can simplify this to
cost = recharge * 0.01
Which is basically nothing.

I have not yet gotten around to looking at your calculations. I'm getting on a plane today, so I may not get to it until the 30th.

royfalk:
> Shield 	Per facet x4 	Pool 	Advertised recharge (MJ/s) 	Old model (MJ/s) 	old / adv 	Current code @mf=0.05 (MJ/s) 	/ adv 	Current code @mf=0.01 (MJ/s)

What is the **advertised** recharge? I assume it's the recharge. But then you have old model, but we didn't change the regeneration figures.

Can you explain the columns in your table please.

evertvorster:
This should be a more clear version of the table. :)

| Shield | Pool | Rated recharge (MJ/s) | OLD draw (MJ/s) | old ÷ rated | CURRENT @mf=0.05 | ÷ rated | CURRENT @mf=0.01 |
|---|---|---|---|---|---|---|---|
| quadshield01 | 300 | 4 | 1.3 | 0.324 | 15.2 | 3.80 | 3.2 |
| quadshield02 | 600 | 8 | 2.6 | 0.324 | 30.4 | 3.80 | 6.4 |
| quadshield03 | 1200 | 14 | 4.5 | 0.324 | 60.7 | 4.34 | 12.7 |
| quadshield04 | 1600 | 18 | 5.8 | 0.324 | 80.9 | 4.49 | 16.9 |
| quadshield05 | 2000 | 22 | 7.1 | 0.324 | 101.1 | 4.60 | 21.1 |
| quadshield06 | 2400 | 26 | 8.4 | 0.324 | 121.3 | 4.67 | 25.3 |
| quadshield07 | 2800 | 32 | 10.4 | 0.324 | 141.6 | 4.42 | 29.6 |
| quadshield08 | 3200 | 36 | 11.7 | 0.324 | 161.8 | 4.49 | 33.8 |
| quadshield09 | 3600 | 40 | 13.0 | 0.324 | 182.0 | 4.55 | 38.0 |
| quadshield10 | 4000 | 44 | 14.3 | 0.324 | 202.2 | 4.60 | 42.2 |
| quadshield11 | 4800 | 54 | 17.5 | 0.324 | 242.7 | 4.49 | 50.7 |
| quadshield12 | 5600 | 62 | 20.1 | 0.324 | 283.1 | 4.57 | 59.1 |
| quadshield13 | 6400 | 72 | 23.3 | 0.324 | 323.6 | 4.49 | 67.6 |
| quadshield14 | 7200 | 80 | 25.9 | 0.324 | 364.0 | 4.55 | 76.0 |
| quadshield15 | 8000 | 90 | 29.2 | 0.324 | 404.5 | 4.49 | 84.5 |

Columns, in order:
  Pool              - per facet x 4, from units.csv
  Rated recharge    - the Shield_Recharge column. UNCHANGED between the two models:
                      this is the number the ship info screen prints as "Shield recharge: N MJ/s".
  OLD draw          - energy drawn from the capacitor per second under the pre-#1016 code
  CURRENT draw      - energy drawn per second under the current code, at the given maintenance_factor
  old / rated       - the OLD draw as a multiple of the rated recharge: 0.324 for every shield.

So the recharge figures are identical in both models - that is the point. What changed is what the
DRAW is calculated from: the rated recharge (old) versus the shield pool (current).

evertvorster:
Fair question — the columns were ambiguous, so here they are spelt out. The short version: the
recharge delivered is the same in both models. What changed is what the energy draw is calculated
from.

| Shield | Pool | Rated recharge (/facet MJ/s) | Recharge delivered (4x rated, pts/s) | OLD draw (MJ/s) | old / rated | CURRENT draw (MJ/s) | current / rated | current / old |
|---|---|---|---|---|---|---|---|---|
| quadshield01 | 300 | 4 | 16 | 1.3 | 0.324 | 15.2 | 3.80 | 11.7 |
| quadshield02 | 600 | 8 | 32 | 2.6 | 0.324 | 30.4 | 3.80 | 11.7 |
| quadshield03 | 1200 | 14 | 56 | 4.5 | 0.324 | 60.7 | 4.34 | 13.4 |
| quadshield04 | 1600 | 18 | 72 | 5.8 | 0.324 | 80.9 | 4.49 | 13.9 |
| quadshield05 | 2000 | 22 | 88 | 7.1 | 0.324 | 101.1 | 4.60 | 14.2 |
| quadshield06 | 2400 | 26 | 104 | 8.4 | 0.324 | 121.3 | 4.67 | 14.4 |
| quadshield07 | 2800 | 32 | 128 | 10.4 | 0.324 | 141.6 | 4.42 | 13.7 |
| quadshield08 | 3200 | 36 | 144 | 11.7 | 0.324 | 161.8 | 4.49 | 13.9 |
| quadshield09 | 3600 | 40 | 160 | 13.0 | 0.324 | 182.0 | 4.55 | 14.0 |
| quadshield10 | 4000 | 44 | 176 | 14.3 | 0.324 | 202.2 | 4.60 | 14.2 |
| quadshield11 | 4800 | 54 | 216 | 17.5 | 0.324 | 242.7 | 4.49 | 13.9 |
| quadshield12 | 5600 | 62 | 248 | 20.1 | 0.324 | 283.1 | 4.57 | 14.1 |
| quadshield13 | 6400 | 72 | 288 | 23.3 | 0.324 | 323.6 | 4.49 | 13.9 |
| quadshield14 | 7200 | 80 | 320 | 25.9 | 0.324 | 364.0 | 4.55 | 14.0 |
| quadshield15 | 8000 | 90 | 360 | 29.2 | 0.324 | 404.5 | 4.49 | 13.9 |

**What the columns are:**

- **Pool** — per-facet shield x 4 facets, straight from `units.csv`.
- **Rated recharge** — the `Shield_Recharge` column. **Unchanged between the two models.** It is a *per-facet* figure: the regeneration loop adds it to every facet, so a quad shield restores 4x this per second. The ship info screen prints this single number as "Shield recharge: N MJ/s", which is why it reads like a power draw.
- **Recharge delivered** — rated x 4, the shield points per second actually put back. **Identical in both models** — same rated value, same per-facet delivery.
- **OLD draw** — energy per second drawn from the capacitor by the pre-#1016 code.
- **CURRENT draw** — energy per second drawn by the current code, at `maintenance_factor` 0.05.
- **old / rated** — 0.324 for every shield.
- **current / rated** — 3.8 to 4.7.
- **current / old** — about 12-14x.

**The two formulas, for reference:**

- OLD: `rated x 0.054/0.2 x 4 x 0.25` (upkeep) + `min(deficit, rated) x 0.054` (rebuild) = `rated x 0.324`
- CURRENT: `POOL x maintenance_factor` + `rated x regeneration_factor`

So to your point: you're right that the regeneration figures didn't change, and that is the point. The recharge behaviour is the same; only the price of it changed. The old draw was computed from the rated recharge — one constant, 0.324x, for every shield. The current draw is computed from the shield **pool**, a number the player never sees and never chose, so it scales with shield size rather than recharge, and lands 3.8-4.7x the rated value and about 14x the old draw.

(That is also why no value of `maintenance_factor` fixes it: reproducing the old behaviour as `pool x f` needs f = 0.00351 at tier 6 and f = 0.00432 at tier 1 — a 23% spread, because shield strength per unit of recharge isn't constant. 0.324 x the rated recharge is exact at every tier.)

